PR #911, opened 2025-09-02
Add Semaphore over ECDSA

jadnohra:
# Add Semaphore ECDSA Circuit Implementation

This PR implements a Semaphore anonymous group membership proof system using ECDSA key derivation on secp256k1.

## Circuit Hierarchy

SemaphoreProofECDSA
├─ IN: message (Vec) - Public message being signed
├─ IN: scope (Vec) - Public scope/context for nullifier
├─ IN: merkle_root ([u8; 32]) - Public Merkle tree root
├─ OUT: nullifier ([u8; 32]) - Public nullifier (prevents double-spending)
├─ AUX: secret_scalar ([u8; 32]) - Private ECDSA key (witness)
├─ AUX: merkle_siblings (Vec<[u8; 32]>) - Merkle proof path (witness)
├─ AUX: leaf_index (usize) - Position in tree (witness)
└─ Internal Components:
├─ Secp256k1ScalarMult: secret_scalar → (pubkey_x, pubkey_y)
├─ Keccak256: (pubkey_x || pubkey_y) → commitment
├─ Keccak256: (scope || secret_scalar) → nullifier
└─ MerkleProof: (commitment, siblings, index) → root_verification

## Example Usage

The `semaphore_ecdsa` example demonstrates the circuit with configurable parameters:

### Circuit Parameters

- `--tree-height <N>` (default: 2) - Merkle tree height, determines max group size (2^N members)
- `--message-len-bytes <N>` (default: 32) - Maximum message length in bytes
- `--scope-len-bytes <N>` (default: 24) - Maximum scope length in bytes

### Instance Parameters

- `--group-size <N>` (default: 4) - Number of group members to create
- `--prover-index <N>` (default: 1) - Index of member generating the proof (0-based)
- `--message <STRING>` (default: "I vote YES on proposal #42") - Message to include in proof
- `--scope <STRING>` (default: "dao_vote_2024_q1") - Scope for nullifier generation

### Example Commands

```
# Basic usage with defaults
cargo run --example semaphore_ecdsa --release

# Large group with custom message
cargo run --example semaphore_ecdsa --release -- \
  --tree-height 16 \
  --group-size 100 \
  --prover-index 42 \
  --message "Approve budget proposal" \
  --scope "dao_2024_q1_vote"

# Benchmark mode (runs circuit N times)
cargo run --example semaphore_ecdsa --release -- --bench 10
```

Scaling

Circuit complexity scales with tree height but is independent of message size:

| Tree Height | Total Constraints | Scaling Factor |
| --- | --- | --- |
| 1 | ~15,000 | Baseline |
| 2 | ~18,000 | \+2,931/level |
| 4 | ~24,000 | \+2,931/level |
| 8 | ~36,000 | \+2,931/level |

- Tree height adds ~2,931 AND constraints per level (Merkle proof verification)
- Scope size has minimal effect (only affects Keccak padding at boundary crossings)
- Message size has zero constraint impact (public input only, not processed)
- MUL constraints remain constant (~12,000 from secp256k1 operations)


crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+impl IdentityCommitmentECDSA {
+	fn new(builder: &CircuitBuilder, curve: &Secp256k1) -> Self {
+		let secret_scalar: [Wire; SCALAR_LIMBS] = std::array::from_fn(|_| builder.add_witness());
+
+		let scalar_biguint = BigUint {
+			limbs: secret_scalar.to_vec(),
+		};
+
+		let generator = Secp256k1Affine::generator(builder);
+		let public_key = scalar_mul_naive(builder, curve, SCALAR_BITS, &scalar_biguint, generator);
```
onesk, line 99: Use `scalar_mul` - it leverages `secp256k1` endomorphism and is thus twice more efficient

crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+		let siblings: Vec<[Wire; KECCAK_DIGEST_LIMBS]> = (0..tree_height)
+			.map(|_| std::array::from_fn(|_| builder.add_witness()))
+			.collect();
+
+		let mut hashers = Vec::new();
+		let mut current = leaf;
+		let mut index = leaf_index;
+
+		for sibling in &siblings {
+			let is_even = builder.bnot(builder.band(index, builder.add_constant_64(1)));
```
onesk, line 188: nit: instead of threading `index` value through the circuit, you could've just shifted the bit in question to the MSB

crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+					.unwrap();
+
+			// Create witness wires for Keccak message
+			let message_wires: Vec<Wire> = (0..PUBKEY_MESSAGE_BYTES / BYTES_PER_WIRE)
+				.map(|_| builder.add_witness())
+				.collect();
+
+			// Constrain message wires to match multiplexer output
+			for i in 0..SCALAR_LIMBS {
+				builder.assert_eq(format!("merkle_message_left[{}]", i), message_wires[i], left[i]);
```
onesk, line 208: why constrain instead of just using `left` and `right` wires as-is?

crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+
+			if wire_start_byte >= scope_len_bytes {
+				builder.assert_eq("scope_wire_padding_zero", wire, zero);
+			} else if wire_end_byte > scope_len_bytes {
+				for byte_offset in 0..BYTES_PER_WIRE {
+					let global_byte_idx = wire_start_byte + byte_offset;
+					if global_byte_idx >= scope_len_bytes {
+						let byte_val = builder.shr(wire, (byte_offset * BITS_PER_BYTE) as u32);
+						let byte_masked = builder.band(byte_val, builder.add_constant_64(0xFF));
+						builder.assert_eq("scope_padding_zero", byte_masked, zero);
```
onesk, line 318: nit: this could've been a single AND constraint, with one of `0xFF`..`0xFFFFFF_FFFFFFFFFF` masks. See how Keccak / Blake2s / etc. handle this. But given that this is done only once per nullifier that doesn't have much effect on circuit size.

crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+			builder.assert_eq(
+				format!("nullifier_secret_equals_identity[{}]", i),
+				secret_scalar_wires[i],
+				identity_secret_scalar[i],
+			);
+		}
+
+		// Create witness wires for Keccak message
+		let total_message_wires = scope_wires + SCALAR_LIMBS;
+		let nullifier_message_wires: Vec<Wire> = (0..total_message_wires)
```
onesk, line 339: nit: this impl assumes that scope is always padded to the nearest multiple of 64 bits in size, whereas reference impl can work with scopes of arbitrary length

crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+			let wire_start_byte = wire_idx * BYTES_PER_WIRE;
+			let wire_end_byte = wire_start_byte + BYTES_PER_WIRE;
+
+			if wire_start_byte >= message_len_bytes {
+				builder.assert_eq("message_wire_padding_zero", wire, zero);
+			} else if wire_end_byte > message_len_bytes {
+				for byte_offset in 0..BYTES_PER_WIRE {
+					let global_byte_idx = wire_start_byte + byte_offset;
+					if global_byte_idx >= message_len_bytes {
+						let byte_val = builder.shr(wire, (byte_offset * BITS_PER_BYTE) as u32);
```
onesk, line 432: nit: may be useful to reuse this logic between scope & message computations

crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+		builder.assert_eq_v("merkle_root_check", merkle_proof.root, merkle_root);
+
+		let nullifier_gen = NullifierGeneratorECDSA::new(
+			builder,
+			scope_len_bytes,
+			identity_commitment.secret_scalar,
+		);
+
+		Self {
+			message,
```
onesk, line 456: How exactly is message used? It does not feed anywhere? Is there any attempt to replicate "dummy square" functionality of the original Semaphore?

jadnohra: Discussed with Ben and Jim. The way this works in binius64 is to 'observe' the message in the transcript. So in our case, we should not even have the message in the circuit/witness. I will remove it.

crates/frontend/src/circuits/semaphore_ecdsa/circuit.rs
```diff
@@ -0,0 +1,687 @@
...
+}
+
+impl IdentityCommitmentECDSA {
+	fn new(builder: &CircuitBuilder, curve: &Secp256k1) -> Self {
+		let secret_scalar: [Wire; SCALAR_LIMBS] = std::array::from_fn(|_| builder.add_witness());
+
+		let scalar_biguint = BigUint {
+			limbs: secret_scalar.to_vec(),
+		};
+
```
paulcadman, line 97: Should `scalar_biguint` be constrained to be non-zero? There is an assert! in `compute_public_key_coords`  in populate_witness.

PR #916, opened 2025-09-02
Add --repeat to examples

fkondej:


fkondej:
Closing - will use part of this logic in a separate PR.

PR #1107, opened 2025-09-09
Implement optimized SIMD version of build_g_triplet

GraDKh:
### TL;DR

Optimize the shift protocol's phase 1 prover by leveraging packed field operations for better performance. On my laptop with AVX2 support this reduces the `build_g_triplets` time by 40% and for the `hashsign` the overall proving time is reduced by 10%.

This comes at the cost of restricting `P::WIDTH` to not exceed 8, but in real life our packed 128b fields never have bigger width than 4

### What changed?

- Modified `prove_phase_1` and `prove` function signatures to separate the `PackedField` trait bound from the function declaration and move it to the `where` clause
- Added `WithUnderlier<Underlier: UnderlierWithBitOps>` constraint to the packed field type
- Removed the static `BITS_MAP` lookup table and replaced it with a more efficient approach using packed field operations
- Optimized the `build_g_triplet` function to process multiple bits at once using packed fields
- Changed the accumulator size to be dependent on the packed field width
- Created a packed masks map to efficiently handle bit operations
- Updated the `build_multilinear_triplet` function to work with packed fields directly

### How to test?

- Run the existing test suite to ensure correctness is maintained
- Benchmark the shift protocol performance before and after the changes
- Verify that the optimized implementation produces the same results as the original one

### Why make this change?

This optimization significantly improves the performance of the shift protocol's phase 1 by:
1. Reducing memory usage by processing multiple bits at once using packed fields
2. Eliminating the need for the static lookup table which improves cache efficiency
3. Taking advantage of SIMD-like operations through the packed field implementation
4. Reducing the number of operations needed by processing bits in batches

The changes maintain the same functionality while making better use of the underlying hardware capabilities.

### DIFF
--- a/prover/prover/src/protocols/shift/phase_1.rs
+++ b/prover/prover/src/protocols/shift/phase_1.rs
@@ -50,7 +50,7 @@ const LOG_LEN: usize = LOG_WORD_SIZE_BITS + LOG_WORD_SIZE_BITS;
 /// Proves the first phase of the shift reduction.
 /// Computes the g and h multilinears and performs the sumcheck.
 #[instrument(skip_all, name = "prover_phase_1")]
-pub fn prove_phase_1<F, P: PackedField<Scalar = F>, C: Challenger>(
+pub fn prove_phase_1<F, P, C: Challenger>(
 	key_collection: &KeyCollection,
 	words: &[Word],
 	bitand_data: &PreparedOperatorData<F>,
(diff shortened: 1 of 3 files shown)


GraDKh:
* **#1107** <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius64/1107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/IrreducibleOSS/binius64/1107?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

## How to use the Graphite Merge Queue

Add the label <kbd>merge-ready</kbd> to this PR to add it to the merge queue.

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/IrreducibleOSS?ref=merge-queue-instructions-in-stack-comment&prId=1107).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

Please do not merge from GitHub as this will restart CI on PRs being processed by the merge queue.

</sub>

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

josephjohnston (approved):

PR #1125, opened 2025-09-10
remove "data structures" tutorial

benediamond:
this is actually an anti-pattern.

in order to "concatenate" two wire-bundles of fixed / known lengths, we should simply concatenate them as arrays. "concatenation" in this trivial sense is free / a no-op at the level of circuits, since we simply "move the wires together".

what this example instead did is define a new bundle of wires, of length equal to the sum of the first two, and then constrain equality wire-by-wire. this is useless, and a waste of constraints.

this has been removed from the docs; removing it here too.

concatenation of _variable-length_ inputs is a much more important, and hard problem. at some point it would be good to get that into a tutorial.


jadnohra:
Do you have a good idea for the simplest indicative example?

jadnohra (approved):

PR #1140, opened 2025-09-11
Gruen 3.2 memory savings

onesk:
This PR plays with the `multilinear::evaluate` trick of representing the eq indicator as an outer product of two smaller eq inds, and exploiting that in mlechecks. The primary target of this optimization is `SelectorMlecheck` in MUL reduction, which has to instantiate 64x equality indicators, but others benefit as well.

Pros:
 * The eq ind related memory consumption can be brought from `O(n)` to `O(sqrt(n))`
* Smaller chunk sizes pave the way to Four Russians style optimizations on binary mlechecks, similar to the ones AND reduction already employs

Cons:
 * This may be in the way of ZK sumcheck efforts (not knowledgeable enough on this)
 * This is a complication to the code (albeit small and limited scope)
* Sumcheck has to do additional `O(n/sqrt(n))` large field muls (though this is more than offset by the savings of not having to fully expand eq inds).

Apart from memory savings, benchmarking shows modest 2-5% perf improvement in single-threaded x86 MUL throughtput as well.

### DIFF
--- a/prover/prover/src/protocols/sumcheck/selector_mle.rs
+++ b/prover/prover/src/protocols/sumcheck/selector_mle.rs
@@ -2,8 +2,6 @@
 
 #![allow(dead_code)]
 
-use std::cmp::max;
-
 use binius_field::{Field, PackedField};
 use binius_math::{FieldBuffer, multilinear::fold::fold_highest_var_inplace};
 use binius_utils::{bitwise::Bitwise, rayon::prelude::*};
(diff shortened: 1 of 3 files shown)


jimpo:
Looking today

prover/prover/src/protocols/sumcheck/selector_mle.rs
```diff
@@ -174,7 +183,7 @@ where
 		// This prover has multiple evaluation points and cannot implement MleCheckProver.
 		let (prime_coeffs, round_coeffs) = izip!(&self.gruen32s, sums, packed_prime_evals)
 			.map(|(gruen32, &sum, packed_prime_evals)| {
-				gruen32.interpolate2(sum, packed_prime_evals.sum_scalars(self.n_vars()))
+				gruen32.interpolate2(sum, packed_prime_evals.sum_scalars(self.n_vars() - 1))
```
jimpo, line 186: Was this a bug? Seems unrelated

onesk: Yep, it was a bug; it's actually strange this does not reproduce with other sumchecks, probably because it's accidentally correct there. The `n_vars - 1` is the correct value here.

onesk:
Agree with the reasoning, changed this to only affect `SelectorMlecheckProver`.

PR #1142, opened 2025-09-11
[examples] prove_load

pepyakin:
A new subcommand that loads a constraint system and parameters from the file
system instead of generating them.


jimpo:
Having this function integrated into the ExampleCLI is good though!

PR #1156, opened 2025-09-15
[frontend] breakdown shifted value indices

pepyakin:
Now the circuit stats breaks down the number of distinct shifted value indices
and unshifted value indices.

```
Gates & Instructions
├─ Number of gates: 27,625
├─ Number of evaluation instructions: 27,625
├─ Distinct value indices: 41,877
│  ├─ Distinct shifted value indices: 34,604
│  └─ Distinct unshifted value indices: 7,273
```

### DIFF
--- a/prover/examples/snapshots/blake2b.snap
+++ b/prover/examples/snapshots/blake2b.snap
@@ -3,7 +3,9 @@ blake2b circuit
 Gates & Instructions
 ├─ Number of gates: 1,363
 ├─ Number of evaluation instructions: 1,363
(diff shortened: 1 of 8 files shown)


jimpo (approved):

jimpo:
> This runs post-fusion. Not sure what you mean though: change relative to that?

I just meant that it gets output next to the gate information, which I was confused by.

pepyakin:
yup, should be fixed in the next PR in the stack.

PR #1248, opened 2025-09-26
[ci] sccache cache-buster value

jimpo:
Compilation is getting SIGKILLED, perhaps with illegal instructions.
This should help debug.


jimpo:
Downgrading Rust version to 1.89.0 solved the problem instead.

PR #1413, opened 2026-03-05
[transcript] Resolve TODO expects in TranscriptWriter with documented invariants

wstran:
Replace three `expect("TODO: propagate error")` calls in `TranscriptWriter` with proper expect messages documenting why serialization is infallible
 
Add doc comments to `TranscriptWriter` struct and its `write`, `write_slice`, and `write_scalar_iter` methods with `# Panics` sections per the style guide
 
Transcript buffers are always growable (`BytesMut`), so `SerializeBytes::serialize` cannot fail due to insufficient space — same reasoning as the existing `expect("challenger has infinite buffer")` pattern at line 126


jimpo (approved):

wstran:
Re-triggered CI — the previous `rust-checks-x86_64` failure on this PR is a flaky GitHub Actions runner issue (SIGILL across ~100 tests, not a code issue). The job uses `-Ctarget-cpu=native` so the binary's instruction set depends on whichever runner CPU GitHub assigns; the all-doc-comment changes in this PR cannot cause SIGILL. The portable variant, arm64, wasm, and all circuit snapshots passed. Will report once the rerun completes.

wstran:
All checks now green. @jimpo this should be ready to merge whenever you have a moment — thanks!

jimpo (approved):

PR #1472, opened 2026-05-05
[ci] Differentiate Rust cache key by CPU feature set

### LINKED ISSUE #1471: [ci] rust-checks-x86_64 fails with SIGILL on AMD runners due to Intel-built cache reuse
## Summary

`rust-checks-x86_64` is currently failing on `main` (and on every PR rebased onto it) with `SIGILL` in the `binius-circuits::rs256::tests::*` group. The portable variant (`rust-checks-x86_64-portable`) and `rust-checks-arm64` pass. The root cause is a CPU/cache mismatch, not a code regression.

## Evidence

GitHub-hosted `ubuntu-24.04` runners are drawn from a heterogeneous pool:

| CI run on `main`              | Runner CPU                    | AVX-512F | GFNI | Conclusion |
|-------------------------------|-------------------------------|----------|------|------------|
| April 28 (last passing)       | Intel Xeon Platinum 8370C     | yes      | yes  | success    |
| April 29 (115108d), May 5 (7c9bed9) | AMD EPYC 7763           | no       | no   | failure (SIGILL) |

Failing tests (consistent across runs):
- `binius-circuits::rs256::tests::test_real_rsa_signature_verification_with_message`
- `binius-circuits::rs256::tests::test_real_rsa_signature_verification_with_wrong_message`
- `binius-circuits::rs256::tests::test_real_rsa_signature_with_invalid_prefix`

Sample failure log: https://github.com/binius-zk/binius64/actions/runs/25363027923/job/74366872130

## Mechanism

The `rust-checks-x86_64` job builds with `RUSTFLAGS=-Ctarget-cpu=native`, then caches `target/` via `actions-rust-lang/setup-rust-toolchain@v1` (which wraps `Swatinem/rust-cache`). The default cache key includes `Cargo.lock` and toolchain version but not CPU feature information.

When a job runs on an Intel Xeon, the resulting `.rlib` artifacts contain AVX-512/GFNI-encoded code (used by gates in `binius-field` such as `packed_ghash_512`, `simd::avx512`, `gfni`). When a subsequent job is scheduled on an AMD EPYC and restores the same cache, those artifacts are linked into the test binary and execute illegal instructions at runtime — hence `SIGILL`.

`x86_64-portable` is unaffected because its build emits no CPU-specific code. `arm64` is unaffected because GitHub's `ubuntu-24.04-arm` pool is homogeneous.

## Proposed fix

Namespace the cache key by CPU feature class so heterogeneous runners do not share artifacts. PR #XXXX hashes `/proc/cpuinfo` flags into `setup-rust-toolchain`'s `cache-key`. Cache reuse is preserved within a feature class; only cross-class restores get separate caches.

Alternative considered: drop `-Ctarget-cpu=native` for CI and use `x86-64-v3` instead. Rejected because that would bypass the AVX-512/GFNI/VAES code paths that `binius-field` exposes via `cfg(target_feature = ...)` gates, leaving them untested in CI.

wstran:
Fixes #1471.

## Problem

`rust-checks-x86_64` recurringly fails on `main` with `SIGILL` in `binius-circuits::rs256::tests::*`. Root cause is documented in #1471: GitHub's `ubuntu-24.04` runner pool mixes Intel Xeon (with AVX-512/GFNI) and AMD EPYC (without). With `RUSTFLAGS=-Ctarget-cpu=native` and `Swatinem/rust-cache` keyed only by `Cargo.lock` + toolchain, a build cached on an Intel runner is reused on an AMD runner and crashes at runtime.

## Fix

Hash `/proc/cpuinfo` flags into `setup-rust-toolchain`'s `cache-key`. Same-class runners keep sharing cache; cross-class runs get separate caches. Falls back to `cpu-portable` on systems without `/proc/cpuinfo`.

## Why this approach

- Keeps the existing `-Ctarget-cpu=native` strategy: AVX-512, GFNI, VAES, VPCLMULQDQ code paths in `binius-field` are still exercised when CI lands on Intel runners.
- Surgical: 15 added lines in one job, no other CI changes.
- Self-correcting: even if GitHub adds a third CPU SKU later, it just gets its own cache bucket — no further CI work needed.

## Alternatives considered

- **Drop `-Ctarget-cpu=native`, use `-Ctarget-cpu=x86-64-v3`**: bypasses AVX-512/GFNI/VAES `cfg(target_feature=...)` gates in `binius-field`, leaving them untested.
- **Disable cache for the native job**: significantly slower CI for every run, not just the cross-class restores.
- **Pin to one runner SKU**: GitHub does not expose this knob for `ubuntu-24.04`.

## Test plan

- [x] YAML validates (`ruby -ryaml ...`)
- [x] Bash snippet exercised locally (Linux path produces 16-hex digest from `flags`; non-Linux path falls back to `cpu-portable`)
- [ ] CI on this PR runs across Intel and AMD runners, both pass without SIGILL

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,13 +69,14 @@ jobs:
       # GitHub-hosted x86_64 runners include both Intel (with AVX-512/GFNI) and
       # AMD (without) SKUs. With `-Ctarget-cpu=native`, build artifacts cached
       # on one SKU produce illegal instructions when restored on another.
-      # Hash the CPU flag set into the cache key so heterogeneous runners do
-      # not share artifacts.
+      # Hash the CPU feature set into the cache key so heterogeneous runners do
+      # not share artifacts. Skipped for portable builds (CPU-agnostic output).
       - name: Compute CPU feature cache key
+        if: matrix.arch.rustflags != ''
         id: cpu
         run: |
           if [ -r /proc/cpuinfo ]; then
-            HASH=$(grep -m1 '^flags' /proc/cpuinfo | sha256sum | cut -c1-16)
+            HASH=$(grep -m1 -E '^(flags|Features)[[:space:]]*:' /proc/cpuinfo | sha256sum | cut -c1-16)
           else
             HASH=portable
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,10 +66,25 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v6
+      # GitHub-hosted x86_64 runners include both Intel (with AVX-512/GFNI) and
+      # AMD (without) SKUs. With `-Ctarget-cpu=native`, build artifacts cached
+      # on one SKU produce illegal instructions when restored on another.
+      # Hash the CPU flag set into the cache key so heterogeneous runners do
+      # not share artifacts.
+      - name: Compute CPU feature cache key
+        id: cpu
+        run: |
+          if [ -r /proc/cpuinfo ]; then
+            HASH=$(grep -m1 '^flags' /proc/cpuinfo | sha256sum | cut -c1-16)
+          else
+            HASH=portable
+          fi
+          echo "key=cpu-${HASH}" >> "$GITHUB_OUTPUT"
       - name: Setup Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           components: clippy
+          cache-key: ${{ steps.cpu.outputs.key }}
       - name: Install cargo-nextest
         uses: taiki-e/install-action@v2
         with:


jimpo:
Amazing, thank you! I literally just wrote up a ticket for this before I saw your PR. :-)

jimpo (changes requested):
Claude reviewed and pointed out two minor issues:

  Minor issues

  1. The CPU-hash step runs on all three matrix entries, not just the two that set -Ctarget-cpu=native. For x86_64-portable this means the cache also splits by SKU, even though portable builds are CPU-agnostic — a small avoidable cache-miss cost. Easy fix would be to make the step conditional on matrix.arch.rustflags != '', or attach cache-key only for native arches. Not worth blocking on, but cleaner.
  2. arm64 isn't handled explicitly. ARM Linux /proc/cpuinfo uses Features:, not flags:. grep -m1 '^flags' returns empty on ubuntu-24.04-arm, so sha256sum hashes the empty string — yielding the same key (cpu-e3b0c4429...) for every arm64 runner. That happens to be the correct behavior (arm64 pool is homogeneous, so all runs should share cache), but it works by accident. A small robustness improvement would be to grep both ^flags and ^Features, or just hash the whole file and rely on identical SKUs producing identical content.

wstran:
CI validates the fix. The `rust-checks-x86_64` job in this PR's CI run landed on the same AMD EPYC 7763 SKU that was producing SIGILL on `main` (https://github.com/binius-zk/binius64/actions/runs/25365263389/job/74374384791) and passed cleanly:

```
cache-key: cpu-9e64157087e39a63
CPU: AMD EPYC 7763 64-Core Processor
Detected: aes, avx, avx2, pclmulqdq, sha, vaes, vpclmulqdq
```

The new cache key forced a fresh build on AMD instead of restoring AVX-512/GFNI artifacts from a prior Intel run, so no illegal-instruction crashes on rs256 tests.

wstran:
Addressed both items in 5fafc96:

1. **Skip CPU hash on portable**: gated the new step on `matrix.arch.rustflags != ''`. The portable row reverts to the default cache key — same SKU classes share cache again.
2. **arm64 Features**: regex is now `^(flags|Features)[[:space:]]*:`, so ubuntu-24.04-arm produces a real digest instead of hashing an empty match.

Verified locally: YAML still parses; `grep` against both an x86-style `flags:` line and an arm64-style `Features:` line returns the expected line in each case.

wstran:
CI on 5fafc96 confirms both refinements work as intended (run 25367273138):

| Job | Observed cache-key | Behavior |
|---|---|---|
| `rust-checks-x86_64-portable` | _(no `cpu-` suffix; step skipped)_ | Default cache key restored — no per-SKU split |
| `rust-checks-x86_64` | `cpu-9e64157087e39a63` (AMD EPYC 7763) | Real hash; got a same-class cache **hit** from the earlier run |
| `rust-checks-arm64` | `cpu-67943d1370fc80af` | Real hash from `Features:` line, distinct from the empty-input digest `e3b0c4429...` |

All checks green.

jimpo (approved):

PR #1454, opened 2026-04-16
Delayed reduction for GF(2^128) sumchecks via widening multiply

quangvdao:
## Summary

Introduces a `WideningMul` trait for `GF(2^128)` packed fields that separates the multiply step (leaves an unreduced "wide" product) from the reduce step, and threads it through the prover-side sumcheck and inner-product hot paths. Accumulating in wide form lets us amortize the per-word reduction across all products in an accumulator, which is the main win of this patch on sumcheck workloads over `GF(2^128)`.

## Reference

This is an application of the "delayed reduction" technique analyzed in _Speeding Up Sum-Check Proving (Extended Version)_, Dao, DeStefano, Bagad, Domb, Thaler — Cryptology ePrint Archive, Paper 2026/587: <https://eprint.iacr.org/2026/587> (see also the implementation in [Jolt](https://github.com/a16z/jolt)). Our setting is `GF(2^128)` (CLMUL-based), which was mentioned in an earlier version of the paper. The present PR is a from-scratch application to Binius's prover pipeline; it is not a port of Jolt's code.

## Headline e2e results

Apple M4 Max, 16 cores, AArch64; release + `RUSTFLAGS="-C target-cpu=native"`; baseline upstream `main` at `8ff2b6a`, branch tip `d4e82cb`; interleaved A/B pairs via `scripts/tracing-ab/run_ab.py`. Full per-span breakdowns and caveats are in the "End-to-end benchmark results" section below.

| workload                          | main    | branch  | Δ (median) | note                  |
|:----------------------------------|--------:|--------:|:----------:|:----------------------|
| Ethsign / ecrecover, n=1          | 204.8 ms | 192.7 ms | **-5.9%** | clean, low variance (sd ~11%) |
| Keccak 256 KiB                    | 155.0 ms | 143.1 ms | **-7.7%** | median, but sd is high on this run |

## What's new

- `WideningMul` trait with `widening_mul` / `reduce_wide`, and a wide accumulator type `WideRoundEvals2` for round-evaluation collection.
- Two unreduced product representations:
  - `WideGhashProduct` (schoolbook, 4 independent CLMULs + 2 reduction CLMULs), used on AArch64 and portable backends.
  - `WideKaratsubaGhashProduct` (3-mul Karatsuba form + 2 reduction CLMULs), used on x86_64 with `pclmulqdq` / `vpclmulqdq`.
[...]

PR #1506, opened 2026-06-10
Add WideMul trait for deferred-reduction GF(2^128) multiply

jimpo:
## Summary

Isolates the `binius-field` half of #1454 per [jimpo's review](https://github.com/binius-zk/binius64/pull/1454#pullrequestreview-4357159068). Introduces `WideMul`, which separates the multiply step (an unreduced "wide" product) from the reduction step so that GF(2^128) inner products can amortize the per-word reduction across an entire accumulator — the basis for the delayed-reduction sumcheck optimization. Prover-side adoption of the trait follows in a separate PR.

Original work by Quang Dao (@quangvdao commit author preserved).

## Changes vs. #1454, addressing review

- **`WideMul` is a parent trait of `PackedField`** (not a propagated bound). Every packed field implements it: the trivial impl (eager multiply, identity reduce) for all fields except the CLMUL-accelerated GF(2^128) packings. Non-GHASH packed types get the trivial impl via a `define_packed_binary_fields!` macro dispatch that skips `BinaryField128bGhash`; the blanket `PackedField` impls gain a `Self: WideMul` bound.
- **Renames** per review: `WideningMul`→`WideMul`, `type Wide`→`Output`, `widening_mul`→`wide_mul`, `reduce_wide`→`reduce`.
- **Narrowed associated-type bound**: `Output: Default + Add + AddAssign + Sub + SubAssign` (added `Sub`/`SubAssign` to the wide-product type); `Debug + Send + Sync + 'static` live on the `PackedField` supertrait bound.
- **Benchmark moved out of `arith-bench`** (which shouldn't depend on `binius-field`) into `crates/field/benches/ghash_widening.rs`. The `inner_product_wide` bench, which needs `binius-math`, is deferred to the prover-side PR.
- **Dropped `WideKaratsubaGhashProduct`.** Benchmarked schoolbook vs. Karatsuba on x86_64 (pclmulqdq + vpclmulqdq): Karatsuba was parity-to-~10% slower across M128/M256, so x86 now uses the schoolbook `WideGhashProduct` like AArch64. This removes the second wide-product type, the `xor_halves` CLMUL helper, and the per-arch `cfg_if` branching. (Measured on a noisy virtualized host; happy to revisit with bare-metal numbers.)

## Correctness / verification

- proptest correctness + linearity across all packed widths, plus an accumulation stress test (`crates/field/src/packed_ghash.rs`, `arch/shared/ghash.rs`).
- `cargo test` green for `binius-field`, `binius-ip-prover`, `binius-spartan-prover`, `binius-prover`; workspace `cargo build` green; `binius-field` builds for `wasm32-wasip1`.
- `cargo fmt` clean; `cargo clippy` clean for `binius-field`.

### DIFF
--- a/crates/field/benches/ghash_widening.rs
+++ b/crates/field/benches/ghash_widening.rs
@@ -1,0 +1,56 @@
+// Copyright 2026 The Binius Developers
+
+//! Benchmark widening (deferred-reduction) `GF(2^128)` multiplication against full multiplication.
+//!
+//! Measures the inner-product pattern `sum_i a_i * b_i` two ways: a full multiply per term (which
+//! reduces every product), and a widening multiply that accumulates unreduced products and reduces
+//! once at the end.
+
+use std::hint::black_box;
+
+use binius_field::{PackedField, WideMul, arch::OptimalPackedB128};
+use criterion::{
+	BenchmarkGroup, Criterion, Throughput, criterion_group, criterion_main, measurement::WallTime,
+};
+
+fn bench_at_n<P: PackedField>(group: &mut BenchmarkGroup<'_, WallTime>, label: &str, n: usize) {
+	let mut rng = rand::rng();
+	let a_vals: Vec<P> = (0..n).map(|_| P::random(&mut rng)).collect();
+	let b_vals: Vec<P> = (0..n).map(|_| P::random(&mut rng)).collect();
+
+	group.throughput(Throughput::Elements((n * P::WIDTH) as u64));
+
+	group.bench_function(format!("full_mul/{label}/n={n}"), |b| {
+		b.iter(|| {
+			let mut acc = P::default();
+			for i in 0..n {
+				acc += black_box(a_vals[i]) * black_box(b_vals[i]);
+			}
+			black_box(acc)
+		})
+	});
+
+	group.bench_function(format!("wide_mul/{label}/n={n}"), |b| {
+		b.iter(|| {
+			let mut acc = <P as WideMul>::Output::default();
+			for i in 0..n {
+				acc += P::wide_mul(black_box(a_vals[i]), black_box(b_vals[i]));
+			}
+			black_box(P::reduce(acc))
+		})
+	});
+}
+
+fn bench_ghash_widening(c: &mut Criterion) {
+	let mut group = c.benchmark_group("ghash_widening");
+
+	let label = format!("OptimalPacked_{}x128b", OptimalPackedB128::WIDTH);
+	for &n in &[16, 256, 4096] {
+		bench_at_n::<OptimalPackedB128>(&mut group, &label, n);
+	}
+
+	group.finish();
+}
+
+criterion_group!(benches, bench_ghash_widening);
+criterion_main!(benches);
--- a/crates/field/src/arch/aarch64/packed_ghash_128.rs
+++ b/crates/field/src/arch/aarch64/packed_ghash_128.rs
@@ -1,4 +1,5 @@
 // Copyright 2023-2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 // Copyright (c) 2019-2023 RustCrypto Developers
 
 //! ARMv8 `PMULL`-accelerated implementation of GHASH.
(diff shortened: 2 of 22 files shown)


crates/field/benches/ghash_widening.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright 2025 Irreducible Inc.
```
jimpo: This new copyright header is wrong. See CONTRIBUTING.md for guidance.

crates/field/src/arch/aarch64/packed_ghash_128.rs
```diff
@@ -82,6 +82,21 @@ impl TaggedSquare<GhashStrategy> for PackedBinaryGhash1x128b {
 	}
 }
 
+// Implement WideMul
+impl crate::arithmetic_traits::WideMul for PackedBinaryGhash1x128b {
```
jimpo: import WideMul and drop the qualification path

crates/field/src/arch/aarch64/packed_ghash_128.rs
```diff
@@ -82,6 +82,21 @@ impl TaggedSquare<GhashStrategy> for PackedBinaryGhash1x128b {
...
+	type Output = crate::arch::shared::ghash::WideGhashProduct<M128>;
```
jimpo: Same here, but use partial qualification. It should read `ghash::WideGhashProduct`

crates/field/src/arch/x86_64/packed_ghash_512.rs
```diff
@@ -129,6 +129,30 @@ cfg_if! {
 	}
 }
 
+// Implement WideMul
+cfg_if! {
+	if #[cfg(all(target_feature = "vpclmulqdq", target_feature = "avx512f"))] {
+		impl crate::arithmetic_traits::WideMul for PackedBinaryGhash4x128b {
```
jimpo: Again, use `crate::arithmetic_traits::WideMul` and `crate::arch::shared::ghash`.

crates/field/src/arithmetic_traits.rs
```diff
@@ -1,11 +1,71 @@
...
+				wide
+			}
+		}
+	};
+}
+
+pub(crate) use impl_trivial_wide_mul;
+
+/// Every scalar field has a trivial widening multiply: multiply eagerly, reduce to the identity.
+impl<F: crate::Field> WideMul for F {
```
jimpo: We probably don't want this blanket impl and instead have `Field` impl structs implement `WideMul` themselves. That way the Ghash field can have a specialized `WideMul`.

jimpo:
Thanks for the review — addressed all five comments (force-pushed):

- **Copyright** — new bench file now `// Copyright 2026 The Binius Developers`; added the Binius Developers line to the modified files per CONTRIBUTING.md.
- **Qualification (×3)** — the arch GHASH impls now `use` the `WideMul` trait and the `ghash` module, so they read `impl WideMul …` / `ghash::WideGhashProduct`. On x86 these imports are `#[cfg]`-gated to the CLMUL feature that gates the impls, so a no-CLMUL x86 build doesn't trip `unused_imports` under `-D warnings`.
- **Blanket `impl<F: Field> WideMul for F`** — removed. Each scalar field now implements `WideMul` itself (`BinaryField1b`, `AESTowerField8b`, `BinaryField128bGhash`), and the GHASH scalar's impl is its own (trivial for now) so it can be specialized later.

**One design decision to flag** on that last point: removing the blanket made generic `F: Field` code lose its `WideMul` impl, which reintroduced exactly the bound-propagation into `binius-math` you objected to in #1454. To avoid that I made `WideMul` a **parent trait of `Field`** as well as `PackedField`, so every field carries it without propagating bounds. Two things worth your call:
1. `FieldOps` is the single shared base of both `Field` and `PackedField`, so one `FieldOps: WideMul` could replace listing it on both — I kept it on `Field`+`PackedField` to match your "add to PackedField's parents" wording, but happy to switch to the DRY single home.
2. The GHASH **scalar** is left trivial-but-specializable; if you wanted the real deferred-reduction impl on the scalar now, that's a follow-up.

Verification: `cargo test` green for field (211) / math / ip-prover / spartan / prover; workspace build; `wasm32-wasip1`; `aarch64` check; clippy clean across CLMUL-on/off x86 configs; fmt clean.

jimpo:
Per your follow-up: `BinaryField128bGhash` now has a real deferred-reduction `WideMul` (no longer trivial). It reuses the width-1 packed type — `Output = <PackedBinaryGhash1x128b as WideMul>::Output`, so it's an unreduced `WideGhashProduct` on CLMUL backends (accumulate many, reduce once) and falls back to the trivial multiply on portable/wasm. Added proptests for scalar correctness and deferred accumulation. Green across native / CLMUL-off / wasm32 / aarch64.

PR #1508, opened 2026-06-10
Adopt WideMul deferred reduction in inner products and MLE-checks

jimpo:
Follow-up to #1506 (the `WideMul` trait, now merged): the "use the trait throughout" half requested in the [original review](https://github.com/binius-zk/binius64/pull/1454#pullrequestreview-4357159068) — prover-side adoption of deferred GF(2^128) reduction.

Because `WideMul` is a parent trait of `Field`/`PackedField`, adoption needs **no trait-bound propagation** — call sites just accumulate `wide_mul` products and `reduce` once. The diff is ~half the size of the original prototype for the same coverage, and touches no ZK-wrapper/channel/config plumbing.

## binius-math — inner products use widening
- `inner_product_wide_par` / `inner_product_wide_buffers` / `inner_product_wide_packed`, with a widening-vs-standard proptest.
- FRI `fold_interleaved` (the hot interleaved-fold inner product) now uses `inner_product_wide_buffers`.

## binius-ip-prover — MleCheck audit
`WideRoundEvals2` is an unreduced degree-2 round-eval accumulator. I audited every MleCheck/Sumcheck prover and applied widening where there is a GF(2¹²⁸) multiply-then-sum hot loop:

| Prover | Applied? | Note |
|---|---|---|
| `BivariateProductSumcheckProver` | ✅ | per-point products → wide, reduce once |
| `BivariateProductMultiMlecheckProver` | ✅ | wide accumulators persist across the whole fold |
| `QuadraticMleCheckProver` | ✅ | `C(..)·eq` widened |
| `BatchQuadraticMleCheckProver` | ✅ | wide M-claim arrays (`array::from_fn`, since `Output: !Copy`) |
| `SelectorMlecheckProver` / `RerandMlecheckProver` | ✅ | per-chunk wide accumulation |
| `MultilinearEvalProver` | ✅ | the round eval *is* an inner product `⟨M(·,1), eq⟩` |
| `BivariateProductMlecheckProver` | — | delegates to the multi-mle prover |
| frac-add / padded / batch / zk-mlecheck decorators | — | no GF(2¹²⁸) accumulation of their own |

## Notes / decisions
- **`Output` is not `Copy`** (the narrowed bound from #1506). I kept that bound and wrote the accumulators Copy-free (`array::from_fn` + in-place `AddAssign`) rather than re-widening the trait. Happy to add `Copy` to the bound instead if you'd prefer simpler prover code.
- **Small-instance size gate not ported.** The prototype gated `BivariateProductSumcheckProver` (`PAR_THRESHOLD_LOG_HALF`) because the wide accumulator + Rayon overhead regressed very short buffers. Here wide is applied unconditionally; small-instance tuning is left as follow-up.
- **Perf** not benchmarked here — the sandbox is too noisy for reliable sumcheck deltas (the prototype made the same caveat). Recommend an A/B on controlled hardware; correctness is covered by the existing prover round-trip tests (wide output == eager output).

## Verification
`cargo test` green: math / ip-prover / iop-prover / prover / spartan-prover. clippy + fmt clean. `binius-math` builds for `wasm32-wasip1`.

### DIFF
--- a/crates/ip-prover/src/sumcheck/bivariate_product_multi_mle.rs
+++ b/crates/ip-prover/src/sumcheck/bivariate_product_multi_mle.rs
@@ -1,14 +1,15 @@
 // Copyright 2023-2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
 use std::cmp::max;
 
-use binius_field::{Field, PackedField};
+use binius_field::{Field, PackedField, WideMul};
 use binius_ip::sumcheck::RoundCoeffs;
 use binius_math::{FieldBuffer, multilinear::fold::fold_highest_var_inplace};
 use binius_utils::rayon::prelude::*;
 use itertools::{Itertools, izip};
 
-use super::{common::SumcheckProver, error::Error, gruen32::Gruen32, round_evals::RoundEvals2};
+use super::{common::SumcheckProver, error::Error, gruen32::Gruen32, round_evals::WideRoundEvals2};
 use crate::sumcheck::common::MleCheckProver;
 
 /// Multiple claim version of `BivariateProductMlecheckProver` that can prove mlechecks
@@ -101,8 +102,9 @@ where
 		let packed_prime_evals = (0..1 << (self.n_vars() - 1 - chunk_vars))
 			.into_par_iter()
 			.fold(
-				|| vec![RoundEvals2::default(); sums.len()],
-				|mut packed_prime_evals: Vec<RoundEvals2<P>>, chunk_index| {
+				|| vec![WideRoundEvals2::default(); sums.len()],
+				|mut packed_prime_evals: Vec<WideRoundEvals2<<P as WideMul>::Output>>,
+				 chunk_index| {
 					let eq_chunk = self.gruen32.eq_expansion().chunk(chunk_vars, chunk_index);
 
 					for (round_evals, (evals_a, evals_b)) in
@@ -126,8 +128,11 @@ where
 							let evals_a_inf_i = evals_a_0_i + evals_a_1_i;
 							let evals_b_inf_i = evals_b_0_i + evals_b_1_i;
 
-							round_evals.y_1 += eq_i * evals_a_1_i * evals_b_1_i;
-							round_evals.y_inf += eq_i * evals_a_inf_i * evals_b_inf_i;
+							// The final multiply (by `evals_b_*_i`) of each product is accumulated
+							// in unreduced (wide) form; the wide accumulators persist across the
+							// whole fold and are reduced a single time at the end.
+							round_evals.y_1 += P::wide_mul(eq_i * evals_a_1_i, evals_b_1_i);
+							round_evals.y_inf += P::wide_mul(eq_i * evals_a_inf_i, evals_b_inf_i);
 						}
 					}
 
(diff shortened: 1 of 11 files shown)


crates/math/src/inner_product.rs
```diff
@@ -98,13 +99,106 @@ where
 		.sum()
 }
 
+/// Parallel inner product using widening (unreduced) multiplication with deferred reduction.
+///
+/// Accumulates `wide_mul(a_i, b_i)` in wide form, then reduces once at the end. For packed
+/// `GF(2^128)` fields this amortizes the reduction cost across all products, which is the main
+/// win of this routine over [`inner_product_par`].
+#[inline]
+pub fn inner_product_wide_par<F, P, DataA, DataB>(
```
jimpo: Why make new inner_product_wide_* methods instead of simply modifying implementations of the existing methods?

crates/ip-prover/src/sumcheck/bivariate_product_multi_mle.rs
```diff
@@ -101,8 +102,15 @@ where
 		let packed_prime_evals = (0..1 << (self.n_vars() - 1 - chunk_vars))
 			.into_par_iter()
 			.fold(
-				|| vec![RoundEvals2::default(); sums.len()],
-				|mut packed_prime_evals: Vec<RoundEvals2<P>>, chunk_index| {
+				// `WideMul::Output` is not `Clone`, so the accumulator vec is built without the
```
jimpo: Just add Clone to the WideMul::Output bounds in the trait.

crates/math/src/inner_product.rs
jimpo: Done — folded the widening into the existing functions. `inner_product_par` and `inner_product_packed` (and `inner_product_buffers` via the latter) now accumulate in unreduced form and reduce once; the three `inner_product_wide_*` functions are removed. Non-`GF(2^128)` fields use the trivial widening multiply, so the result is identical to the old product-then-sum. FRI `fold_interleaved` now picks up the deferred reduction transparently — its call site is unchanged, so that file is back to pristine. Replaced the wide-vs-standard test with an inner-product-vs-naive-scalar correctness test.

crates/ip-prover/src/sumcheck/bivariate_product_multi_mle.rs
jimpo: Done — added `Clone` to `WideMul::Output` in the trait, and reverted this accumulator back to `vec![WideRoundEvals2::default(); sums.len()]`. (`BatchQuadraticMleCheckProver` keeps `array::from_fn`, since fixed-size array repeat `[x; M]` needs `Copy` rather than `Clone` — `from_fn` is the idiomatic Clone-free form there and isn't contorted.)

crates/math/src/inner_product.rs
```diff
@@ -90,12 +94,11 @@ where
...
-	iter::zip(a, b)
-		.map(|(a_i, b_i)| a_i * b_i)
-		.sum::<P>()
-		.into_iter()
-		.take(1 << log_n)
-		.sum()
+	// See `inner_product_par` for why the products are accumulated in unreduced (wide) form.
+	let wide_sum = iter::zip(a, b)
+		.map(|(a_i, b_i)| P::wide_mul(a_i, b_i))
+		.fold(<P as WideMul>::Output::default(), |acc, w| acc + w);
```
jimpo: Add Sum as a bound on WideMul::Output type so we can use it here

jimpo: Done — added `Sum` to the `WideMul::Output` bound (with a `Sum` impl for `WideGhashProduct`; the trivial `Output` types get it via `FieldOps: Sum`). Both `inner_product_par` (rayon `.sum()`) and `inner_product_packed` (`iter::sum()`) now use it instead of `fold(default, +)`.

PR #1513, opened 2026-06-11
feat(frontend): expose eval-form bytecode + value-vec accessors

pauldelucia:
Three read-only accessors, no behavioral change:

- `Circuit::eval_form() -> &EvalForm`
- `EvalForm::bytecode() -> &[u8]`
- `WitnessFiller::value_vec()` / `value_vec_mut() -> &/&mut ValueVec`

### Why
We build a fresh per-shape Binius64 circuit for almost every Ethereum block — proving the block's receipts trie is fully indexed — so `CircuitBuilder::build`'s constraint lowering dominates our prover time (~25 s at our circuit sizes). The lowering is needed only to drive `populate_wire_witness`; the witness fill itself is the `EvalForm` bytecode, which is a flat register program over value-vec indices and is composable per gadget.

Exposing the bytecode lets an external evaluator extract per-gadget eval programs once and replay them per block — skipping the lowering for repeated circuit shapes while staying byte-identical to the frontend. These accessors are the minimal surface needed; happy to adjust naming/visibility to your preference (e.g. behind a `cfg` or a dedicated trait) if you'd rather not widen the public API.

### DIFF
--- a/crates/circuits/src/keccak/mod.rs
+++ b/crates/circuits/src/keccak/mod.rs
@@ -30,6 +30,13 @@ pub struct Keccak256 {
 }
 
 impl Keccak256 {
+	/// The padded-message witness wires (keccak padding applied; filled by
+	/// [`Self::populate_message`]). Exposed for external evaluators that fill the
+	/// witness without the frontend filler (mirrors the public `message`/`digest`).
+	pub fn padded_message(&self) -> &[Wire] {
+		&self.padded_message
+	}
+
 	/// Create a new keccak circuit using the circuit builder
 	///
 	/// # Arguments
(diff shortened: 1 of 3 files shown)


crates/circuits/src/keccak/mod.rs
```diff
@@ -30,6 +30,11 @@ pub struct Keccak256 {
 }
 
 impl Keccak256 {
+	/// Returns the padded-message witness wires (filled by [`Self::populate_message`]).
+	pub fn padded_message(&self) -> &[Wire] {
```
jimpo, line 34: FYI, this `Keccak256` gadget struct will soon turn into a simple gadget function that doesn't store the message. (eg. an interface more like `binius_circuits::keccak::fixed_length::keccak256`).

Happy to accept this for now, but it'll break soon.

jimpo (approved):

jimpo:
Interesting. I'm happy to merge these accessors, as they are very unobtrusive, but I'll note that building a fresh circuit per-statement is not really a supported/recommended use case.

PR #1514, opened 2026-06-11
feat(hash): single-block SHA-256 ParallelDigest specialization (BINIUS-75)

jimpo:
## Summary

Implements `ParallelSha256Digest`, a `ParallelDigest` that specializes leaf hashing for the common Merkle-tree case where every leaf fits — together with SHA-256 padding — in a single 64-byte block. Such a leaf is hashed with **one `compress256` call** from the SHA-256 IV, skipping the per-leaf `update`/`finalize` bookkeeping the generic `ParallelDigestAdapter` performs.

Closes BINIUS-75.

### What changed
- **`ParallelDigest::digest_with_const_leaves`** (new trait method): a specialized entry point for leaves built from exactly `n_items_per_input` items of a `FixedSizeSerializeBytes` type, so each leaf has a compile-time-derivable length. Default impl forwards to `digest`.
- **`ParallelSha256Digest`**: specializes the above. The padding suffix (`0x80` terminator, zero fill, big-endian bit length) is precomputed once per Rayon worker; each leaf only overwrites the message prefix, then `compress256` runs from the IV and the big-endian state is emitted as the digest. Leaves longer than 55 bytes fall back to `ParallelDigestAdapter`.
- A head-to-head benchmark (`sha256_const_leaves`) for the evaluation case.

## Evaluation gate

> *"This specialized version must be 10% faster than the unspecialized version for parallel digest with 2 B128 elements per leaf."*

Measured with `RUSTFLAGS="-C target-cpu=native"` (prod flags), 1 MiB of B128, 2 elements (32 bytes) per leaf:

| Path | Time | Throughput |
|---|---|---|
| unspecialized (`ParallelDigestAdapter`) | 2.1285 ms | 470 MiB/s |
| **specialized** (`ParallelSha256Digest`) | **1.0161 ms** | **984 MiB/s** |

**~52% faster** (2.1× throughput), CIs non-overlapping by a wide margin — well past the 10% gate.

Reproduce:
```bash
RUSTFLAGS="-C target-cpu=native" cargo bench -p binius-hash --bench sha256 -- sha256_const_leaves
```

## Notes
- Correctness is verified against `<Sha256 as Digest>::digest` for leaf sizes 16/32/48 bytes (single-block) and 64 bytes (fallback) in `crates/hash/src/sha256.rs`.
- This is the simple form (plain overwrite into a reused, pre-padded block template); the issue's suggested XOR / custom-`BufMut` trick was held in reserve since it wasn't needed to clear the gate.
- **Not yet wired into `Sha256HashSuite::ParLeafHash`** — kept to experiment scope (measure the win). Happy to follow up with adoption into the Merkle leaf-hashing path if you want it.

### DIFF
--- a/crates/hash/benches/sha256.rs
+++ b/crates/hash/benches/sha256.rs
@@ -128,7 +128,7 @@ fn bench_const_leaves(c: &mut Criterion) {
 	group.bench_function("specialized", |b| {
 		b.iter(|| {
 			let out = &mut digests.spare_capacity_mut()[..n_leaves];
-			specialized.digest_with_const_leaves(
+			specialized.digest_with_const_len(
 				BATCH_SIZE,
 				black_box(elements.as_slice())
 					.par_chunks(BATCH_SIZE)
--- a/crates/hash/benches/sha256.rs
+++ b/crates/hash/benches/sha256.rs
@@ -3,7 +3,7 @@
 use std::hint::black_box;
 
 use binius_field::{BinaryField128bGhash as B128, Random};
-use binius_hash::{ParallelDigest, ParallelDigestAdapter, StdDigest};
+use binius_hash::{ParallelDigest, ParallelDigestAdapter, ParallelSha256Digest, StdDigest};
 use binius_utils::rayon::{prelude::*, slice::ParallelSlice};
 use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
 use digest::{Digest, Output};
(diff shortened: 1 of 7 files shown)


jimpo (commented):
Great results!

In this PR, add two more commits that complete the speedup:

- Use `ParallelSha256Digest` in the SHA-256 HashSuite
- Call digest_with_const_leaves in hash_leaves for the binary Merkle tree

jimpo:
Done — added in three commits:

1. **`feat(hash): use ParallelSha256Digest in the SHA-256 HashSuite`** — `Sha256HashSuite::ParLeafHash` now points at `ParallelSha256Digest`.
2. **`feat: hash Merkle leaves with digest_with_const_leaves`** — threads the per-leaf element count (`n_items_per_input`) from `commit` / `commit_field_buffer` down through `MerkleTreeProver::commit_iterated` → `binary_merkle_tree::build_from_iterator` → `hash_leaves`, which now calls `digest_with_const_len` instead of `digest`. Salted leaves pass `n_items_per_input + salt_len` (the exact element count each leaf yields). No new trait bounds — `Field: FixedSizeSerializeBytes` already holds. The signature change to `commit_iterated`/`build_from_iterator` is contained; there are no other callers in the workspace.
3. **`refactor(hash): rename digest_with_const_leaves to digest_with_const_len`** — per your follow-up.

Verified in **debug** (so the count `debug_assert` and the verifier round-trip both fire): binius-hash (49) + binius-iop-prover (31) green, including the hiding/salted open+verify tests, FRI `test_commit_prove_verify_*` (exercises both `to_par_scalar_big_chunks` and `small_chunks`), and the BaseFold ZK round-trips. Full `--workspace` build, clippy (`--tests --benches`), and `cargo doc --document-private-items -D warnings` all clean.

One thing worth flagging on the motivating ZK case: with hiding commitments the salt enlarges the leaf, so the single-block fast path engages only while `(batch_size + salt_len) * 16 ≤ 55` bytes — i.e. ≤ 3 B128 elements per leaf. For `log_batch_size = 1` (2 elements) that's `salt_len ≤ 1` (48 bytes → fast path); `salt_len = 2` lands at 64 bytes and correctly falls back to the generic path (no speedup, but identical output). Let me know if the production ZK salt length would push it into the fallback.

PR #1517, opened 2026-06-11
Optimize BatchBrakedownFolder::fold (BINIUS-78)

benji-potamus:
## Summary

Optimizes `BatchBrakedownFolder::fold` ([BINIUS-78](https://linear.app/jimpo/issue/BINIUS-78/optimize-batchbrakedownfolderfold)) by fusing the per-oracle codeword folding with the cross-oracle accumulation, eliminating wasted allocations and redundant multiplications.

Previously `fold` folded each oracle's interleaved codeword into its own buffer (via `ProxTestFolder::fold` → `fold_interleaved`), then in a separate pass scaled by the outer-challenge tensor value and accumulated into the combined codeword. This had three inefficiencies:

1. **Discarded temporary buffers** — each per-oracle folded codeword was allocated only to be summed in and dropped.
2. **Double scaling** — every input value was scaled once by the inner-challenge tensor (inside the inner product) and again by the outer-challenge tensor value.
3. **Redundant lifting muls** — when a codeword is lifted (`log_lift > 0`), the scaling multiplication was recomputed for each of the `2^log_lift` duplicated copies.

## Changes

For each oracle, zipped with its outer-challenge tensor value `scalar`:

- **Pre-scale the inner tensor** by `scalar`, so each folded entry comes out already scaled by the outer-challenge value. By linearity (`⟨chunk, tensor⟩·scalar = ⟨chunk, tensor·scalar⟩`) this folds the two scalings into one pass over the `2^log_batch_size`-element tensor instead of one mul per output entry.
- **`log_lift == 0`:** fold and accumulate directly into the combined codeword on the fly, with no temporary buffer.
- **`log_lift > 0`:** fold into a temporary buffer, then accumulate each folded entry into its `2^log_lift` contiguous copies — the scaling mul is not repeated per copy.

The contiguous duplication (`lifted[j] = folded[j >> log_lift]`) is preserved exactly, matching the verifier's index translation in `crates/iop/src/fri/batch.rs` (`index >> log_lift`).

This makes `ProxTestFolder::fold` and the private `fold_interleaved` helper unused, so both are removed (their fold logic is now inlined into the batch path).

## Testing

No behavior change. The existing FRI prove→verify roundtrip tests pin the exact semantics across both code paths and pass:

- `test_commit_prove_verify_lifted_multi_oracle` — mixed-size oracles, `log_lift ∈ {2, 0, 4}` (exercises lifting).
- `test_commit_prove_verify_batched_multi_oracle` — equal-size oracles, `log_lift == 0`.

`cargo test -p binius-iop-prover` (31 tests, incl. basefold) and `cargo clippy -p binius-iop-prover --tests --benches -- -D warnings` both pass.

### DIFF
--- a/crates/iop-prover/src/fri/fold.rs
+++ b/crates/iop-prover/src/fri/fold.rs
@@ -1,4 +1,5 @@
 // Copyright 2024-2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
 use std::{iter, mem};
 
@@ -360,43 +361,6 @@ where
 	}
 }
 
-/// Fold the interleaved codeword into a single codeword using the given challenges.
-///
-/// ## Arguments
-///
-/// * `codeword` - an interleaved codeword.
-/// * `challenges` - the folding challenges. The length must be at least `log_batch_size`.
-/// * `log_len` - the binary logarithm of the code length.
-/// * `log_batch_size` - the binary logarithm of the interleaved code batch size.
-///
-/// See [DP24], Def. 3.6 and Lemma 3.9 for more details.
-///
-/// [DP24]: <https://eprint.iacr.org/2024/504>
-#[instrument(skip_all, level = "debug")]
-fn fold_interleaved<F, P>(
-	codeword: FieldSlice<P>,
-	challenges: &[F],
-	log_len: usize,
-	log_batch_size: usize,
-) -> FieldBuffer<F>
-where
-	F: Field,
-	P: PackedField<Scalar = F>,
-{
-	assert_eq!(codeword.log_len(), log_len + log_batch_size);
-	assert_eq!(challenges.len(), log_batch_size);
-
-	let tensor = eq_ind_partial_eval(challenges);
-
-	// For each chunk of size `2^chunk_size` in the interleaved codeword, fold it with the folding
-	// challenges.
-	let values = codeword
-		.chunks_par(log_batch_size)
-		.map(|chunk| inner_product_buffers(&chunk, &tensor))
-		.collect::<Vec<_>>();
-	FieldBuffer::new(log_len, values.into_boxed_slice())
-}
-
 /// FRI-fold the codeword using the given challenges.
 ///
 /// ## Arguments
@@ -451,33 +415,6 @@ where
 {
 	pub fn log_folded_len(&self) -> usize {
 		self.codeword.log_len() - self.log_batch_size
-	}
-
-	pub fn fold(
-		self,
-		merkle_prover: &'a MTProver,
-		challenges: &[F],
-	) -> (FieldBuffer<F>, BrakedownOracleProver<'a, P, MTProver>) {
-		assert_eq!(challenges.len(), self.log_batch_size); // precondition
-
-		let ProxTestFolder {
-			log_batch_size,
-			log_lift,
-			codeword,
-			merkle_committed,
-		} = self;
-
-		let log_len = codeword.log_len() - log_batch_size;
-		let folded_codeword =
-			fold_interleaved(codeword.to_ref(), challenges, log_len, log_batch_size);
-		let oracle = BrakedownOracleProver::new(
-			codeword,
-			merkle_committed,
-			merkle_prover,
-			log_batch_size,
-			log_lift,
-		);
-		(folded_codeword, oracle)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


benji-potamus:
## Performance impact

Measured by proving the `bip32 --zk` example (the change has no criterion benchmark) and reading the `tracing-forest` profiling tree. Baseline = `main`, branch = this PR, each built `--release` with `RUSTFLAGS="-C target-cpu=native"` in its own workspace. **6 iterations each, interleaved** (base, branch, base, …) to average out machine drift, on a 19-core x86_64 sandbox.

The optimized code is `BatchBrakedownFolder::fold`, which the tree reports as the **`FRI Initial Fold`** span (`log_len=24, arity=3`). That span also contains an unchanged Merkle-commit child, so "fold only" below subtracts that child to isolate the folding work.

| Metric | Baseline (median) | Branch (median) | Δ median | Δ min (least noisy) |
|---|---|---|---|---|
| **`FRI Initial Fold` — fold only** | 28.2 ms | 14.3 ms | **−49%** | 26.1 → 12.3 ms (**−53%**) |
| `FRI Initial Fold` — span total (incl. Merkle commit) | 37.2 ms | 20.8 ms | −44% | 30.8 → 16.6 ms (−46%) |
| `Prove` (end-to-end) | 940 ms | 925 ms | −1.6% (within noise) | — |

Raw fold-only samples (ms): baseline `[26.1, 27.9, 27.9, 28.4, 32.5, 32.7]`, branch `[12.3, 13.2, 14.1, 14.4, 18.7, 22.9]`.

**The fold itself is ~2× faster** — consistent with folding the outer-challenge scalar into the inner tensor (one pass over the small tensor instead of a mul per output entry) and dropping the per-copy scaling mul during codeword lifting.

End-to-end prove time doesn't move measurably: the initial fold is only ~2–3% of total proving here, so the ~14 ms saved sits inside run-to-run noise at the top level. The win is real and local; it'll matter more where the first-round fold is a larger share (more / larger batched oracles).

crates/iop-prover/src/fri/fold.rs
```diff
@@ -537,29 +474,66 @@ where
...
+					.par_iter_mut()
+					.zip(codeword.chunks_par(log_batch_size))
+					.for_each(|(acc, chunk)| *acc += inner_product_buffers(&chunk, &tensor));
+			} else {
+				// Lifting: fold into a temporary buffer, then accumulate each folded entry into its
+				// `2^log_lift` contiguous copies. The scaling mul is not repeated per copy.
+				let folded = codeword
+					.chunks_par(log_batch_size)
+					.map(|chunk| inner_product_buffers(&chunk, &tensor))
+					.collect::<Vec<_>>();
```
jimpo: Seems we can avoid this collect and just accumulate on the fly for the log_lift != 0 case as well.

jimpo (changes requested):

crates/iop-prover/src/fri/fold.rs
```diff
@@ -537,29 +474,66 @@ where
...
+				let combined = combined_codeword.as_mut();
+				for (f, &value) in folded.iter().enumerate() {
+					let base = f << log_lift;
+					for acc in &mut combined[base..base + (1 << log_lift)] {
```
jimpo: Better to do a `combined.par_chunks_mut(1 << log_lift)` and zip with that.

crates/iop-prover/src/fri/fold.rs
benji-potamus: Done — dropped the `collect` and the `log_lift == 0` / `> 0` split entirely. Both cases are now one `par_chunks_mut(1 << log_lift)` accumulating on the fly (chunk size is `1` when there's no lifting), so nothing is allocated in either case.

crates/iop-prover/src/fri/fold.rs
benji-potamus: Done. Now:
```rust
combined_codeword
    .as_mut()
    .par_chunks_mut(1 << log_lift)
    .zip(codeword.chunks_par(log_batch_size))
    .for_each(|(copies, chunk)| {
        let value = inner_product_buffers(&chunk, &tensor);
        for acc in copies {
            *acc += value;
        }
    });
```
Each folded entry's inner product is computed once and added to its `2^log_lift` contiguous copies in parallel, with no temp buffer. Pushed as `23b8179`; FRI tests (incl. `test_commit_prove_verify_lifted_multi_oracle`) and crate clippy green.

jimpo (approved):

PR #1518, opened 2026-06-11
Consolidate secp256k1 EC multiplication onto GLV-Straus MSM (BINIUS-76)

benji-potamus:
## Summary

Makes **GLV-Straus (`msm_strauss_endo`) the single secp256k1 elliptic-curve multiplication routine** and removes everything it subsumes. Rebased onto the now-merged #1516.

## Changes

- `ecdsa::ecrecover` and `ecdsa::bitcoin::verify`: `u1·G + u2·P` via `msm_strauss_endo(MSM_WINDOW, [u1, u2], [G, P])` instead of `shamirs_trick_endomorphism`.
- `ecdsa::scalar_mul::scalar_mul`: reimplemented as a single-point `msm_strauss_endo(MSM_WINDOW, [scalar], [point])`.
- `bitcoin::p2pkh`: `private_key·G` now uses `scalar_mul` (→ MSM) instead of `scalar_mul_naive`.
- Removed `shamirs_trick_endomorphism`, `shamirs_trick_naive`, `scalar_mul_naive`, **and the plain (non-GLV) `msm_strauss`** — GLV-Straus is always at least as good, so it's the only variant kept. `scalar_mul.rs` drops ~440 lines.
- Added `MSM_WINDOW = 4` (the empirically-optimal Straus window); `ec_msm` example now always uses `msm_strauss_endo` (dropped the `-e` toggle; default `-w 4`).
- Re-blessed snapshots:
  - `bitcoin_p2pkh`: 380,532 → **168,874 committed (−56%)** — GLV-Straus is far cheaper than the old naive double-and-add.
  - `ethsign`: 216,721 → 236,920 (+9.3%, the GLV-Straus-vs-endo-Shamir delta at n=2).
  - `ec_msm`: re-blessed (example default is now GLV-Straus, w=4).

## Why accept the ECDSA n=2 regression

Endo-Shamir is ~6–9% smaller than GLV-Straus only at n=2–3; GLV-Straus wins everywhere else and scales linearly. Keeping separate routines purely for the 2-point case isn't worth the extra implementations. (For the single-scalar `p2pkh` path it's a large *win*.)

## Testing

- `cargo test -p binius-circuits --lib ecdsa` / `… p2pkh` pass, including `test_ecdsa_recover_test_vector`, `test_bitcoin_ecdsa_test_vector`, `test_p2pkh_circuit_privkey_external`, `test_scalar_mul_with_endomorphism`, and the `msm_strauss_endo` window-1/2/3 k256 tests.
- crate-scoped clippy + fmt clean; `cargo doc --document-private-items` has no broken links; `ethsign`/`bitcoin_p2pkh`/`ec_msm` snapshots verified.

### DIFF
--- a/crates/examples/src/circuits/ec_msm.rs
+++ b/crates/examples/src/circuits/ec_msm.rs
@@ -4,7 +4,7 @@ use std::iter;
 use anyhow::Result;
 use binius_circuits::{
 	bignum::{BigUint, assert_eq},
-	ecdsa::scalar_mul::{msm_strauss, msm_strauss_endo},
+	ecdsa::scalar_mul::msm_strauss_endo,
 	secp256k1::{N_LIMBS, Secp256k1, Secp256k1Affine},
 };
 use binius_core::word::Word;
@@ -19,8 +19,8 @@ use rand::prelude::*;
 use crate::ExampleCircuit;
 
 /// Example circuit that computes a multi-scalar multiplication `Σ_i scalars[i] · points[i]` over
-/// secp256k1 with a statically-known number of points, using the fixed-window (Straus) algorithm
-/// (see [`msm_strauss`]).
+/// secp256k1 with a statically-known number of points, using the GLV fixed-window (Straus)
+/// algorithm (see [`msm_strauss_endo`]).
 ///
 /// The points and scalars are public inputs (the points are constrained to lie on the curve), and
 /// the claimed result is a public input that the circuit checks against the in-circuit MSM.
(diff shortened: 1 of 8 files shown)


crates/examples/src/circuits/ec_msm.rs
jimpo: Remove this. Since `msm_strauss_endo` is always faster, it's the only version we should keep.

jimpo (changes requested):

crates/examples/src/circuits/ec_msm.rs
benji-potamus: Done. Removed the plain `msm_strauss` entirely (function + its window-1/2/3 tests), so `msm_strauss_endo` is the only Straus variant. The `ec_msm` example now always uses it (dropped the `-e` flag and the plain branch; default window is 4). `strauss_accumulate`/`build_strauss_table` stay (used by the endo path); `msm_strauss_endo`'s doc is now self-contained. fmt + clippy + `cargo doc` (no broken links) clean; `ec_msm` snapshot re-blessed.

Also rebased this PR onto the now-merged #1516.

jimpo (approved):

PR #1519, opened 2026-06-11
Prove a truncated BIP32 tree (BINIUS-79)

benji-potamus:
Implements the truncated-BIP32 circuit from [BINIUS-79](https://linear.app/jimpo/issue/BINIUS-79/update-circuit-for-truncated-bip32-tree), stacked on the BINIUS-60 `bip32-demo-cli` branch (hence the base; diff is just this change).

Instead of running a full BIP32 derivation in-circuit, this proves a pubkey hash derived from a **non-hardened-only** chain rooted at an xprivkey that is *either* the seed KDF master *or* a single hardened derivation from a supplied parent extended private key. The expensive hardened levels of a path are folded into the private `parent_xprivkey` witness offline, so the in-circuit path loop carries no hardened-vs-normal branching.

### Circuit statement (`binius-bip32-sig`)
- **Public input:** `pk_hash` — a SHA-256 digest.
- **Private:** `seed` (64 B), `parent_xprivkey` (64 B = `private_key ‖ chain_code`), `hardened_index` (31-bit; bit 63 carries the `use_seed` flag), non-hardened `path`, `path_length`.
- If `use_seed`, start from the seed master key; otherwise take one hardened step from `parent_xprivkey`. Then derive along the non-hardened `path`, SHA-256 the compressed public key, and assert it equals `pk_hash`.

### Commits
1. **Expose BIP32 per-step derivation primitives** — refactors the `bip32` module in `binius-examples` to expose `master_from_seed`, `extended_key_parts` (parse a 64-byte `private_key ‖ chain_code` into scalar + chain code), `hardened_child`, `non_hardened_child`, and the branch-selecting `child` (hardened vs. non-hardened sharing a single HMAC-SHA512). The hardened step uses the parent *scalar* directly — no scalar multiplication. The generic `bip32_derive_compressed` loop is rewritten to call `child`; its behavior and the standalone example are unchanged.
2. **Prove a truncated BIP32 tree in bip32-sig** — composes those primitives into the truncated derivation (`truncated_derive_compressed`), the circuit statement (`TruncatedBip32`), a `bitcoin`-crate reference oracle, and the CLI rewiring. `split_derivation` maps an arbitrary path onto the inputs by splitting at the path's **last** hardened level (any prefix mix is derived offline into `parent_xprivkey`). `MAX_DEPTH` is the non-hardened suffix depth (2); address-binding, proof-file format, and verify flow are unchanged.

The reusable per-step gadgets live in the library (`binius-examples`); the application-specific truncated composition lives in the consumer (`binius-bip32-sig`).

### Tests
Circuit tests for the seed and hardened-parent branches, `path_length = 0`, full depth, short-path padding, and hardened-path-index rejection; plus end-to-end checks that `split_derivation` reproduces the directly-derived public key for both a standard BIP44 path and a mixed path (`m/0/1'/2`, hardened after non-hardened). `cargo test` and crate-scoped `clippy` are green on `binius-bip32-sig` and `binius-examples`.

### DIFF
--- a/crates/bip32-sig/src/address.rs
+++ b/crates/bip32-sig/src/address.rs
@@ -18,7 +18,9 @@ use bitcoin::{
 use ripemd::{Digest as _, Ripemd160};
 use sha2::Sha256;
 
-/// Bit 31 of a derivation index marks a hardened child (matches `Bip32Example`).
+use crate::derive::DerivationInputs;
+
+/// Bit 31 of a derivation index marks a hardened child.
 pub const HARDENED_BIT: u32 = 0x8000_0000;
 
 /// All address types the demo supports. Each maps to its standard BIP purpose.
@@ -70,7 +72,7 @@ impl AddressType {
 }
 
 /// A BIP44-style path `m/purpose'/0'/account'/change/index` as raw indices (hardened bit set on the
-/// first three levels). Depth 5, matching the circuit's `max_depth`.
+/// first three levels). Depth 5: three hardened levels and a two-level non-hardened suffix.
 pub fn bip44_path(ty: AddressType, account: u32, change: u32, index: u32) -> Vec<u32> {
 	vec![
 		ty.purpose() | HARDENED_BIT,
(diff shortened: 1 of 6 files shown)


crates/bip32-sig/src/address.rs
```diff
@@ -144,6 +146,80 @@ pub fn derive_compressed_pubkey(seed: &[u8; 64], path: &[u32]) -> Result<Compres
...
+/// Map a full BIP32 path onto the truncated circuit's inputs.
+///
+/// The circuit reaches the start of its non-hardened chain by either the seed master key or a
+/// single hardened step from a parent extended private key. So a path of zero-or-more hardened
+/// levels followed by only non-hardened levels (the standard BIP44 shape) is split as: all-but-the-
+/// last hardened level into `parent_xprivkey` (derived offline), the last hardened level into the
+/// in-circuit hardened step, and the non-hardened suffix into `path`. A path with no hardened
+/// levels derives straight from the seed master key.
+///
+/// The non-hardened suffix must fit `max_depth`, and hardened levels may not follow a non-hardened
```
jimpo: That's an arbitrary limitation. Be robust to the case when a there are hardened levels following non-hardened levels.

crates/examples/src/circuits/bip32.rs
```diff
@@ -123,6 +115,122 @@ pub fn bip32_derive_compressed(
...
+///   Entries beyond `path_length` are still derived but never selected.
+/// * `path_length` - the actual non-hardened depth (`<= max_depth`) selecting which level to output
+///
+/// # Returns
+/// The 33-byte compressed public key as nine four-byte big-endian words (see [`compress_pubkey`]).
+///
+/// # Assumptions
+/// As with [`bip32_derive_compressed`], the negligible-probability BIP32 invalid-key cases are not
+/// constrained; the gadget assumes a valid witness.
+pub fn truncated_bip32_derive_compressed(
```
jimpo: This is too specific of an interface to expose from the crate for the other crate's application. This crate could just expose `master_from_seed`, `hardened_child`, and a new `non_hardened_child`, and a general `child` (which does both hardened and non-hardened and selects between them, but sharing the HMAC-SHA512 circuitry).

jimpo (changes requested):

crates/bip32-sig/src/address.rs
benji-potamus: Fixed in e4df667. `split_derivation` now splits at the path's **last** hardened level: everything before it (any hardened/non-hardened mix) is derived offline into `parent_xprivkey`, that last hardened level is the in-circuit hardened step, and the suffix after it — necessarily non-hardened — becomes `path`. So a path like `m/0/1'/2` is now handled. Added a `matches_mixed_hardened_path` test that checks the split + circuit reproduces the directly-derived pubkey for exactly that shape.

crates/examples/src/circuits/bip32.rs
benji-potamus: Done in e4df667 (commit 1). The `bip32` module now exposes the per-step primitives instead: `master_from_seed`, `extended_key_parts` (parses a 64-byte `private_key || chain_code` into scalar + chain code), `hardened_child`, `non_hardened_child`, and the branch-selecting `child` — which selects hardened vs. non-hardened on bit 31 sharing a single HMAC-SHA512, and is exactly what the generic `bip32_derive_compressed` loop now calls.

`truncated_bip32_derive_compressed` is removed; the truncated composition now lives in `binius-bip32-sig` (`truncated_derive_compressed` in `derive.rs`), built from those primitives. That also resolves the placement seam I'd flagged in the PR description — the reusable gadgets are in the library, the application-specific composition is in the consumer.

jimpo (approved):

PR #1520, opened 2026-06-11
Instrument ZK setup sub-steps for timing attribution (BINIUS-61)

benji-potamus:
## Summary

Phase 0 of [BINIUS-61](https://linear.app/jimpo/issue/BINIUS-61) ("Make ZK prover/verifier setup reusable"): instrument the ZK setup path so we can attribute where the ~setup time actually goes before committing to the later phases.

- Adds `tracing` spans around the three `ZKProver::setup` sub-steps that previously had none:
  - **Build key collection** (`build_key_collection`)
  - **Build / Compile ZK wrapper circuit** (symbolic-exec → `compile`)
  - **Precompute NTT domain** (`generate_from_subspace` + BaseFold prover compiler)
- Fixes a latent bug in the existing `ZKVerifier::setup` spans: they used `let _ = span.entered()`, where the `_` pattern **drops the guard immediately**, so the span closed before the wrapped work ran and recorded ~0 ns. Bound to a named guard (`let _guard`) so the span covers the block. The new prover spans use the same pattern.

No behavior change — instrumentation only.

## Why

Running the instrumented depth-5 BIP32 circuit (`bip32 prove --zk`, `--release` + `target-cpu=native`, 19 threads) shows **`build_key_collection` dominates ZK setup (~53%, ~620 ms)**, while the symbolic-exec (~27 ms total) and NTT precompute (~3.5 ms) are nearly free. That measurement re-orders the issue's later phases (full write-up posted on the Linear issue). The spans are the prerequisite for that measurement and have standalone value for setup-time attribution, so they ship as their own small PR.

### DIFF
--- a/crates/prover/src/zk_config.rs
+++ b/crates/prover/src/zk_config.rs
@@ -60,7 +60,10 @@ where
 	pub fn setup(zk_verifier: ZKVerifier<H>) -> Result<Self, Error> {
 		// Build the inner IOPProver.
 		let inner_iop_verifier = zk_verifier.inner_iop_verifier().clone();
-		let key_collection = build_key_collection(inner_iop_verifier.constraint_system());
+		let key_collection = {
+			let _guard = tracing::debug_span!("Build key collection").entered();
+			build_key_collection(inner_iop_verifier.constraint_system())
+		};
 		let inner_iop_prover = IOPProver::new(inner_iop_verifier.clone(), key_collection);
 
 		// Re-derive the outer constraint system and layout via symbolic execution.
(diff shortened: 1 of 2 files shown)


crates/prover/src/zk_config.rs
```diff
@@ -58,24 +58,35 @@ where
 {
 	/// Constructs a ZK prover from a [`ZKVerifier`].
 	pub fn setup(zk_verifier: ZKVerifier<H>) -> Result<Self, Error> {
+		let _setup_guard = tracing::debug_span!("Setup ZK prover").entered();
```
jimpo: I've found it better to avoid putting the top level span within the function, preferring to put it around the calls. Often the caller has some context that is useful to put into the log message (regarding what that specific call's purpose is). So keep the subspans below, but remove this one and if needed add a tracing span around the call to `setup`.

crates/prover/src/zk_config.rs
```diff
@@ -91,16 +102,19 @@ where
...
-		let log_num_shares = binius_utils::rayon::current_num_threads().ilog2() as usize;
-		let ntt = NeighborsLastMultiThread::new(domain_context, log_num_shares);
-		let merkle_prover = BinaryMerkleTreeProver::<_, H>::new();
-		let basefold_compiler = BaseFoldZKProverCompiler::from_verifier_compiler(
-			zk_verifier.basefold_compiler(),
-			ntt,
-			merkle_prover,
-		);
+		let basefold_compiler = {
+			let _guard = tracing::debug_span!("Precompute NTT domain").entered();
```
jimpo, line 105: This span message doesn't match everything below. Keep this span message, but it should only be around the `GenericPreExpanded::generate_from_subspace` call.

jimpo (changes requested):

crates/prover/src/zk_config.rs
benji-potamus: Removed the in-function `Setup ZK prover` span; the sub-spans below stay. Didn't add a call-site span — the one caller, `setup_zk` in `crates/examples/src/lib.rs`, already brackets both `ZKVerifier::setup` + `ZKProver::setup` in `info_span!("ZK setup")`, so the sub-spans nest under that with the `log_inv_rate` context. (Left `ZKVerifier::setup`'s pre-existing top-level `Setup ZK verifier` span alone since it's outside this PR's scope — happy to give it the same treatment in a follow-up if you'd like the symmetry.)

crates/prover/src/zk_config.rs
benji-potamus: Narrowed `Precompute NTT domain` to wrap only the `GenericPreExpanded::generate_from_subspace` call; the NTT/merkle/compiler construction is now outside the span.

jimpo (approved):

PR #1521, opened 2026-06-12
Check that Rust docs build without warnings (BINIUS-82)

benji-potamus:
## Summary

Adds a local prek hook that checks the Rust documentation builds cleanly, and brings the CI `rust-doc` job up to the same standard.

Resolves [BINIUS-82](https://linear.app/jimpo/issue/BINIUS-82/prek-hook-for-checking-rust-docs).

### Changes

- **`rustdoc` pre-push hook** (`tools/check-rustdoc.sh`, registered in `prek.toml`): runs the same `cargo doc --document-private-items --no-deps` as the `rust-doc` CI job, with `RUSTDOCFLAGS="-D warnings"`. Registered at the `pre-push` stage with `pass_filenames = false`, mirroring the existing `clippy` hook (doc build is too slow for `pre-commit`).
- **CI `rust-doc` job**: set the same `RUSTDOCFLAGS="-D warnings"`. The job previously relied only on `actions-rust-lang/setup-rust-toolchain`'s default `RUSTFLAGS=-D warnings`, which **is not passed to rustdoc** — so rustdoc-only lints (broken intra-doc links, missing crate-level docs) only warned and never failed CI. Hook and CI now enforce the same standard.

### Notes for review

- **Why touch CI too?** Without it the local hook would be silently *stricter* than CI, producing pre-push failures that CI wouldn't flag. The one-line CI change keeps them in parity. Happy to split this out if you'd rather scope the PR to just the hook.
- **No copyright header on `check-rustdoc.sh`** — matches the sibling `tools/clippy-changed-packages.sh` (the copyright check only scans `.rs` files). Easy to add if you'd prefer.

## Test plan

- `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items --no-deps` builds cleanly on this branch (all 19 crates), confirming `main`'s docs already satisfy the stricter check — the hook/CI change adds no new failures today.
- `prek validate-config prek.toml` → all configs valid.
- `bash tools/check-rustdoc.sh` exits 0.

### DIFF
--- a/prek.toml
+++ b/prek.toml
@@ -31,6 +31,18 @@ hooks = [
     entry = "bash tools/clippy-changed-packages.sh",
     pass_filenames = true,
     stages = ["pre-push"]
+  },
+  {
+    id = "rustdoc",
+    name = "Check Rust docs build",
+    types = ["rust"],
+    language = "system",
+    # RUSTDOCFLAGS="-D warnings" promotes rustdoc lints (broken intra-doc links,
+    # missing crate-level docs) to errors, matching the `rust-doc` CI job.
+    entry = "cargo doc --document-private-items --no-deps",
+    env = { RUSTDOCFLAGS = "-D warnings" },
+    pass_filenames = false,
+    stages = ["pre-push"]
   }
 ]
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,10 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@v1
       - name: Build documentation (including private items for CI checks)
         run: cargo doc --document-private-items --no-deps
+        env:
+          # Promote rustdoc lints (broken intra-doc links, missing crate-level
+          # docs) to errors, matching the `rustdoc` pre-push hook in prek.toml.
+          RUSTDOCFLAGS: "-D warnings"
 
   circuits-snapshot:
     name: circuits-snapshot (${{ matrix.example }})


prek.toml
```diff
@@ -31,6 +31,15 @@ hooks = [
     entry = "bash tools/clippy-changed-packages.sh",
     pass_filenames = true,
     stages = ["pre-push"]
+  },
+  {
+    id = "rustdoc",
+    name = "Check Rust docs build",
+    types = ["rust"],
+    language = "system",
+    entry = "bash tools/check-rustdoc.sh",
```
jimpo: prek has an `env` key for local hooks: https://prek.j178.dev/reference/configuration/?h=local#args

Seems like we could use that and inline the command into entry, instead of a separate script

jimpo (changes requested):

prek.toml
benji-potamus: Done — dropped `tools/check-rustdoc.sh` and inlined the command into `entry` with `RUSTDOCFLAGS` set via prek's `env` key:

```toml
entry = "cargo doc --document-private-items --no-deps",
env = { RUSTDOCFLAGS = "-D warnings" },
```

Verified `prek validate-config` passes and that the `env` value actually reaches the command (ran a throwaway hook that echoed `$RUSTDOCFLAGS` → `-D warnings`). This also moots the copyright-header note, since there's no longer a new file. Pushed as `2d7914a`.

jimpo (approved):

PR #1522, opened 2026-06-12
Make ZKProver/ZKVerifier serializable for cross-process reuse (BINIUS-81)

benji-potamus:
## Summary

Phase 2 of [BINIUS-61](https://linear.app/jimpo/issue/BINIUS-61) (parent), tracked as [BINIUS-81](https://linear.app/jimpo/issue/BINIUS-81). The Phase 0 measurement showed `build_key_collection` is the dominant ZK-setup cost (~53% / ~620 ms on the depth-5 BIP32 circuit). It's witness-independent, so it can be built once, serialized, and reloaded in a fresh process to skip the rebuild — the high-leverage cold-start win.

Implemented per review feedback as `SerializeBytes`/`DeserializeBytes` **on the structs** rather than exposing `KeyCollection` through a `setup_with_key_collection` signature:

1. **`Implement SerializeBytes/DeserializeBytes for ZKVerifier`** — `ZKVerifier` retains `log_inv_rate` and serializes as its seed `(constraint_system, log_inv_rate)`; `deserialize` reconstructs the derived state (outer CS, oracle specs, BaseFold compiler) by routing through `setup`.
2. **`Add SerializeBytes/DeserializeBytes for ZKProver; privatize key-collection injection`** — `ZKProver` serializes as `(constraint_system, log_inv_rate, key_collection)`; `deserialize` reuses the deserialized `KeyCollection` (the ~620 ms part) and recomputes the cheaper derived state (symbolic-exec / outer circuit ~27 ms, NTT domain ~3.5 ms). `setup_with_key_collection` is now **private**, used only by `setup` and `deserialize`, so the public surface is `setup` / `serialize` / `deserialize` (no `KeyCollection` leak). Setup/validation failures in `deserialize` map to `SerializationError::InvalidConstruction`.

So a process with a serialized `ZKProver` calls `ZKProver::deserialize(bytes)` instead of `ZKProver::setup(verifier)` and skips the dominant cost. The blob is self-contained (embeds the CS); NTT sharding is thread-count-dependent but internal to proving, so a blob is portable across machines.

## Validation

`cargo test -p binius-prover --test prove_verify` — new `test_zk_prove_verify_serialized` round-trips both structs through serialize/deserialize, then proves with the reloaded prover and verifies with the reloaded verifier (accept, not byte-equality — ZK proofs are randomized).

## Scope

CLI plumbing is out of scope here (under this design the CLI would load a serialized `ZKProver`, not a `KeyCollection`); the real consumer is the `binius-bip32-sig` CLI, gated on BINIUS-60. Phase 1 (de-dup symbolic-exec, ~27 ms) and Phase 3 (serialize NTT, ~3.5 ms) remain dropped.

### DIFF
--- a/crates/prover/src/zk_config.rs
+++ b/crates/prover/src/zk_config.rs
@@ -6,7 +6,10 @@
 //! Spartan-based zero-knowledge wrapper. The prover counterpart to
 //! [`binius_verifier::zk_config::ZKVerifier`].
 
-use binius_core::{constraint_system::ValueVec, word::Word};
+use binius_core::{
+	constraint_system::{ConstraintSystem, ValueVec},
+	word::Word,
+};
 use binius_field::{
 	BinaryField128bGhash as B128, PackedExtension, PackedField, UnderlierWithBitOps, WithUnderlier,
 };
@@ -19,13 +22,16 @@ use binius_spartan_verifier::{
 	constraint_system::ConstraintSystemPadded, wrapper::IronSpartanBuilderChannel,
 };
 use binius_transcript::{ProverTranscript, fiat_shamir::Challenger};
-use binius_utils::SerializeBytes;
+use binius_utils::{DeserializeBytes, SerializeBytes, serialization::SerializationError};
 use binius_verifier::{IOPVerifier, zk_config::ZKVerifier};
+use bytes::{Buf, BufMut};
 use digest::Output;
 use rand::CryptoRng;
 
 use crate::{
-	IOPProver, merkle_tree::prover::BinaryMerkleTreeProver, protocols::shift::build_key_collection,
+	IOPProver,
+	merkle_tree::prover::BinaryMerkleTreeProver,
+	protocols::shift::{KeyCollection, build_key_collection},
 };
 
 type ProverNTT<F> = NeighborsLastMultiThread<GenericPreExpanded<F>>;
@@ -58,12 +64,22 @@ where
 {
 	/// Constructs a ZK prover from a [`ZKVerifier`].
 	pub fn setup(zk_verifier: ZKVerifier<H>) -> Result<Self, Error> {
-		// Build the inner IOPProver.
-		let inner_iop_verifier = zk_verifier.inner_iop_verifier().clone();
 		let key_collection = {
 			let _guard = tracing::debug_span!("Build key collection").entered();
-			build_key_collection(inner_iop_verifier.constraint_system())
+			build_key_collection(zk_verifier.inner_iop_verifier().constraint_system())
 		};
+		Self::setup_with_key_collection(zk_verifier, key_collection)
+	}
+
+	/// Constructs a ZK prover from a verifier and a prebuilt [`KeyCollection`], skipping the
+	/// dominant key-collection build. Private: external callers use [`Self::setup`], or
+	/// [`DeserializeBytes::deserialize`] to reuse a serialized prover.
+	fn setup_with_key_collection(
+		zk_verifier: ZKVerifier<H>,
+		key_collection: KeyCollection,
+	) -> Result<Self, Error> {
+		// Build the inner IOPProver.
+		let inner_iop_verifier = zk_verifier.inner_iop_verifier().clone();
 		let inner_iop_prover = IOPProver::new(inner_iop_verifier.clone(), key_collection);
 
 		// Re-derive the outer constraint system and layout via symbolic execution.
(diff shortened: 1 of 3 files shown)


jimpo (commented):
The method signature `setup_with_key_collection` leaks too much information about internals (the KeyCollection struct).

What if instead we made `ZKVerifier` and `ZKProver` both implement `SerializeBytes` and `DeserializeBytes` themselves? The deserialize could have additional logic for recomputing the cheaper stuff (like NTT domains, outer circuit, etc.), but the interface to it would be easy to use. Then, if there's a serialized ZKProver available, instead of calling `ZKProver::setup`, one could call `ZKProver::deserialize`.

benji-potamus:
Good call — leaking `KeyCollection` through the public API is the wrong side of ease-of-use. Taking your direction: implement `SerializeBytes`/`DeserializeBytes` on `ZKVerifier` and `ZKProver`, and make `setup_with_key_collection` **private** (used only by `deserialize`), so the public surface is just `setup` / `serialize` / `deserialize`.

Plan:
- **Seed written:** `ZKVerifier` ↔ `(constraint_system, log_inv_rate)`; `ZKProver` ↔ `(constraint_system, log_inv_rate, key_collection)`. `deserialize` recomputes the cheap derived state — symbolic-exec / outer circuit (~27 ms) and NTT domain (~3.5 ms) — by routing through the existing `setup` path, and injects the deserialized `KeyCollection` (the ~620 ms part). Both structs gain a `log_inv_rate` field; setup/validation failures inside `deserialize` map to `SerializationError::InvalidConstruction`.
- **Validation:** a `serialize → deserialize → prove → verify` unit test (replacing the KeyCollection round-trip test). NTT sharding is thread-count-dependent but that's internal to proving, not proof validity, so a blob is portable across machines.

Two calls I'd like your read on:
1. **Self-contained blob?** Planning `ZKProver::deserialize(bytes)` to embed the CS so it needs nothing else (matches "call `deserialize` instead of `setup`"). Wrinkle: the bip32-sig CLI already persists the CS separately and needs `cs.value_vec_layout` to rebuild the witness, so a self-contained `ZKProver` blob duplicates the CS on disk. I think that's fine (the CLI can stop saving a standalone CS and read the layout off the deserialized prover), but if you'd rather `deserialize` take the CS as a parameter to avoid the dup, say so.
2. **CLI scope:** I'll **drop** the examples `load-prove --zk` rework from this PR — under this design the CLI would load a serialized `ZKProver`, not a `KeyCollection`, and the unit test already exercises the path. Keeping this PR to the prover/verifier-crate API; the real consumer is the bip32-sig CLI (gated on BINIUS-60).

Implementing the core now and will push + re-request review; will adjust (1)/(2) if you prefer otherwise.

benji-potamus:
Pushed the redesign. `ZKVerifier` and `ZKProver` now implement `SerializeBytes`/`DeserializeBytes`; `setup_with_key_collection` is private; the examples-CLI rework is dropped. Went with the self-contained blob (embeds CS) and the test as the validation path, per my notes above — say the word if you'd prefer `deserialize` to take the CS as a param instead. CI re-running.

crates/prover/src/zk_config.rs
```diff
@@ -45,6 +51,9 @@ where
 	outer_iop_prover: binius_spartan_prover::IOPProver<B128>,
 	outer_layout: WitnessLayout<B128>,
 	basefold_compiler: BaseFoldZKProverCompiler<P, ProverNTT<B128>, ProverMerkleProver<B128, H>>,
+	/// Retained so the prover can be reconstructed from its seed via [`SerializeBytes`] /
+	/// [`DeserializeBytes`].
+	log_inv_rate: usize,
```
jimpo: Can't this be read off of the `basefold_compiler` somehow instead of storing as a denormalized field?

crates/verifier/src/zk_config.rs
```diff
@@ -46,6 +50,9 @@ pub struct ZKVerifier<H: HashSuite> {
 	inner_iop_verifier: IOPVerifier,
 	outer_iop_verifier: IronSpartanIOPVerifier<B128>,
 	basefold_compiler: BaseFoldZKVerifierCompiler<B128, BinaryMerkleTreeScheme<B128, H>>,
+	/// Retained so the verifier can be reconstructed from its constraint system via
+	/// [`SerializeBytes`] / [`DeserializeBytes`].
+	log_inv_rate: usize,
```
jimpo: Can't this also be read off the `basefold_compiler`?

crates/prover/src/zk_config.rs
benji-potamus: Good call — dropped the field. `log_inv_rate` is now read off the compiler: `self.basefold_compiler.fri_params().rs_code().log_inv_rate()` (`rs_code()` and `log_inv_rate()` are both public getset getters). No denormalized state.

crates/verifier/src/zk_config.rs
benji-potamus: Done — same fix here: dropped the field and the `log_inv_rate()` accessor now returns `self.basefold_compiler.fri_params().rs_code().log_inv_rate()`.

jimpo (approved):

PR #1523, opened 2026-06-12
Track compile-time len_range on ByteVec to simplify concat constraints (BINIUS-47)

benji-potamus:
Closes BINIUS-47.

Adds a compile-time `len_range: Range<usize>` to `ByteVec` recording the known bound on `len_bytes`, and uses it in `concat` to drop the dynamic slice/compare machinery when offsets and lengths are statically known.

## Headline result

`hashsign` (hash-based signatures — `circuit_tweaked_keccak` concatenates many constant-length terms) drops from **1,189,986 → 838,500 AND constraints** (−29.5%). More importantly it crosses the **2²¹ → 2²⁰ bucket boundary**, so the power-of-two-padded committed size roughly halves. See the `hashsign.snap` diff for the full before/after `CircuitStat`. No other snapshot changes (`zklogin` is bit-identical — see below).

## Design

`len_range` records the compile-time bound on `len_bytes` (`start <= len_bytes <= end`, end inclusive to mirror capacity). Invariant: `len_range.end <= capacity`, asserted in the constructors.

The range is only ever established where it is genuinely enforced:
- `new` / `new_inout` / `new_witness` default to the full `0..capacity` (behavior-preserving).
- `new_const_len` pins `len_bytes` to a compile-time constant wire ⇒ `len_range = N..N`, structurally enforced.
- `truncate` / `slice_const_range` carry their constant length.
- `new_with_len_range` (used by `concat`) records an externally-enforced range — `concat`'s output length is the sum of already-constrained input lengths, so the bound holds by arithmetic with no new constraint.

`populate_len_bytes` / `populate_bytes_le` now assert the populated length lies in range (catches witness-construction bugs).

### `concat` lowering

Per term, the comparison binding the hint output to the input is emitted at the lowest machinery that's statically sound:
- **constant offset + constant length** → fully-static per-word `assert_eq` (constant mask on the boundary word). No multiplexer, no variable shifts, no saturating-diff/select. This is the `circuit_tweaked_keccak` win.
- **constant offset + dynamic length** → constant-offset extraction (`extract_const_range`, shared with `slice_const_range`) feeding the unchanged `assert_slice_eq`. No multiplexer.
- **dynamic offset** → the existing fully-dynamic `slice` + `assert_slice_eq` path.

The output carries `len_range = sum(start)..sum(end)`. The leading-term offset is a folded constant (and `0 + len = len`), so the all-dynamic path is **bit-identical to the previous implementation** — confirmed by `zklogin.snap` being unchanged.

## Soundness

The constant path is a pure structural word comparison over compile-time indices — trivially sound. The unaligned shared-word case correctly splits a boundary word between term `i` (low bytes, masked) and term `i+1` (high bytes, unaligned extraction), tiling `[0, total)` with no gap. Ranges only ever originate as full (`start = 0`), constant (`start = end`), or sums of enforced ranges, so any range the code relies on is genuinely constrained.

## Commits

1. **Add `len_range` field and constructors** — field + constructors + invariant/populate asserts; migrate the constant-length term constructions in `hash_based_sig/hashing/*` to `new_const_len`. Behavior-preserving.
2. **Use `len_range` to elide concat machinery** — factor `extract_const_range`, rewrite `concat`, re-bless `hashsign.snap`.

## Tests

- `fixed_byte_vec`: invariant panics (`new_const_len`/`new_with_len_range` over capacity, out-of-range `populate_len_bytes`) and range propagation.
- `concat`: aligned/unaligned constant terms, empty/single, const↔dynamic mixes (each branch), a constant-path mutation-rejection test, and a proptest over constant-length terms with many non-8-multiple lengths.
- Full suite green: `binius-circuits`, `binius-prover --test shift`, `binius-examples` (incl. zklogin end-to-end), all 9 circuit snapshots, `cargo clippy --all --tests --benches --examples`.

## Deferred (called out for review)

The issue also lists two payoffs I intentionally left out, since the headline win is fully captured by the constant `concat` path and these would add churn/risk for little:
- **`Keccak256::new` / `Sha256::new` `len_check` elision** — one `icmp`+`assert` per hash; eliding it soundly requires threading the (genuinely enforced) range into the hash ctors, which I'd rather not do speculatively.
- **Per-word `assert_slice_eq` elision for tight-but-non-constant ranges** — premature; the cases that dominate (`circuit_tweaked_keccak`) are all-constant and already fully optimized.

Happy to fold either in if you'd prefer them in this PR.

### DIFF
--- a/crates/circuits/src/concat.rs
+++ b/crates/circuits/src/concat.rs
@@ -1,22 +1,33 @@
 // Copyright 2025 Irreducible Inc.
-use binius_core::word::Word;
+use binius_core::{consts::WORD_SIZE_BYTES, word::Word};
 use binius_frontend::{CircuitBuilder, Wire, hints::ByteVecConcatHint};
 
 use crate::{
-	fixed_byte_vec::ByteVec,
+	fixed_byte_vec::{ByteVec, extract_const_range},
 	slice::{assert_slice_eq, slice},
 };
 
 /// Computes the concatenation of a list of [`ByteVec`]s as a new [`ByteVec`].
 ///
 /// The returned vector has:
-/// - capacity (number of data wires) equal to the sum of the inputs' capacities, and
-/// - runtime length equal to the sum of the inputs' `len_bytes` values.
+/// - capacity (number of data wires) equal to the sum of the inputs' capacities,
+/// - runtime length equal to the sum of the inputs' `len_bytes` values, and
+/// - `len_range` equal to the sum of the inputs' ranges (`sum(start)..sum(end)`).
 ///
 /// The output data wires are populated by [`ByteVecConcatHint`]; soundness is enforced by
-/// extracting each input's range from the output via [`slice()`] and asserting equality with the
-/// input data using [`assert_slice_eq`]. Bytes of `output.data` beyond `output.len_bytes` are
-/// unconstrained.
+/// constraining each input's region of the output to equal that input's data. How that constraint
+/// is emitted depends on what is known at circuit-build time:
+///
+/// - When a term's byte offset into the output is a compile-time constant (i.e. every preceding
+///   term has a constant length), its region is extracted with constant shifts and masks (no
+///   multiplexer or dynamic-shift machinery).
+/// - When that term *also* has a constant length, the comparison degenerates further to a plain
+///   per-word `assert_eq` (with a constant mask on the final partial word), dropping the
+///   saturating-diff / variable-shift / select machinery of [`assert_slice_eq`] entirely.
+/// - Otherwise (a dynamic offset, once some preceding term has a dynamic length) the term falls
+///   back to the fully-dynamic [`slice()`] + [`assert_slice_eq`] path.
+///
+/// Bytes of `output.data` beyond `output.len_bytes` are unconstrained.
 pub fn concat(b: &CircuitBuilder, inputs: &[ByteVec]) -> ByteVec {
 	let dimensions: Vec<usize> = inputs.iter().map(|v| v.data.len()).collect();
 	let mut hint_inputs: Vec<Wire> = inputs.iter().flat_map(|v| v.data.iter().copied()).collect();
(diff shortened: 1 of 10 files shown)


.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"eb1e874d-fb13-4ab2-be0e-d17033154ff3","pid":964,"procStart":"482275","acquiredAt":1781283682907}
```
jimpo: Delete this file, it shouldn't be here. (You can use the `.claude/local` dir for temp files, it is gitignored)

crates/circuits/src/fixed_byte_vec.rs
```diff
@@ -32,41 +48,94 @@ pub struct ByteVec {
 	/// The data wires, each holding up to 8 bytes. The number of wires determines
 	/// the capacity: capacity = data.len() * 8.
 	pub data: Vec<Wire>,
+	/// Compile-time bound on `len_bytes`: `len_range.start <= len_bytes <= len_range.end`. See the
+	/// struct docs for the invariants and how the range is established.
+	pub len_range: Range<usize>,
```
jimpo: Perhaps `RangeInclusive` is a better fit than `Range` semantically?

crates/circuits/src/fixed_byte_vec.rs
```diff
@@ -32,41 +48,94 @@ pub struct ByteVec {
...
+	/// * If `len_bytes` lies outside `self.len_range`.
 	pub fn populate_len_bytes(&self, w: &mut WitnessFiller, len_bytes: usize) {
+		self.assert_len_in_range(len_bytes);
 		w[self.len_bytes] = Word(len_bytes as u64);
 	}
 
+	/// Asserts that a concrete byte length lies within the compile-time `len_range`.
+	fn assert_len_in_range(&self, len_bytes: usize) {
+		assert!(
+			self.len_range.start <= len_bytes && len_bytes <= self.len_range.end,
```
jimpo: With `RangeInclusive`, this could be `self.len_range.contains(len_bytes)`.

crates/circuits/src/fixed_byte_vec.rs
jimpo: Restrict even more to `self.len_range.end()`

crates/circuits/src/fixed_byte_vec.rs
```diff
@@ -162,66 +229,96 @@ impl ByteVec {
...
-		let start_word_idx = range.start / WORD_SIZE_BYTES;
-		// Word index containing the last byte of the sliced data.
-		let last_word_index = (range.end - 1) / WORD_SIZE_BYTES;
-		let byte_offset = range.start % WORD_SIZE_BYTES;
-
-		// Calculate number of output words (rounded up to word boundary)
-		let num_output_words = slice_len.div_ceil(WORD_SIZE_BYTES);
-
-		// Validate that range.end <= self.len_bytes at runtime
+		// Validate that range.end <= self.len_bytes at runtime.
```
jimpo: Skip this check if `range.end == self.len_range.end()`

crates/circuits/src/fixed_byte_vec.rs
```diff
@@ -162,66 +229,96 @@ impl ByteVec {
...
+) -> Vec<Wire> {
+	assert!(range.start <= range.end, "invalid range: start > end");
+	assert!(
+		range.end <= data.len() * WORD_SIZE_BYTES,
+		"range.end {} exceeds capacity {}",
+		range.end,
+		data.len() * WORD_SIZE_BYTES
+	);
+
+	let slice_len = range.end - range.start;
```
jimpo: nit: use `range.len()`

crates/circuits/src/fixed_byte_vec.rs
```diff
@@ -162,66 +229,96 @@ impl ByteVec {
...
+					// Combine the two parts (XOR is cheaper than OR).
+					b.bxor(shifted_current, shifted_next)
+				} else {
+					shifted_current
+				}
+			})
+			.collect()
+	};
+
+	// Apply byte mask to final word if slice doesn't end on word boundary.
```
jimpo: We don't need to mask the leftover bytes. `ByteVec` makes no guarantees about the byte values after the end if its length. If tests expect 0s, update them accordingly.

crates/circuits/src/concat.rs
```diff
@@ -1,74 +1,139 @@
...
+/// `len` are ignored, matching [`assert_slice_eq`]'s semantics).
+fn assert_const_slice_eq(
+	b: &CircuitBuilder,
+	name: &str,
+	output: &[Wire],
+	offset: usize,
+	input: &[Wire],
+	len: usize,
+) {
+	// `extract_const_range` already zeroes the trailing bytes of its final word, so we only need to
```
jimpo: Invalid assumption, see other comment

jimpo (changes requested):
I see no changes to the `slice` module. As a follow-up PR, optimize the `slice` module to reduce multiplexing based on the len_range field.

.claude/scheduled_tasks.lock
benji-potamus: Done — removed it and added `.claude/scheduled_tasks.lock` to my global git exclude so the harness stops snapshotting it into commits.

crates/circuits/src/fixed_byte_vec.rs
benji-potamus: Done — `len_range` is now `RangeInclusive<usize>`.

crates/circuits/src/fixed_byte_vec.rs
benji-potamus: Done — `assert_len_in_range` now uses `self.len_range.contains(&len_bytes)`.

crates/circuits/src/fixed_byte_vec.rs
benji-potamus: Done — `slice_const_range` now bounds against `self.len_range.end()` rather than capacity, so it rejects slicing into guaranteed-garbage capacity. (Updated the `exceeds_capacity` test's expected panic message to match.)

crates/circuits/src/fixed_byte_vec.rs
benji-potamus: Gated the skip on **const-length** rather than `range.end == self.len_range.end()` literally — the latter is unsound for dynamic-length `ByteVec`s. The runtime check is `range.end <= len_bytes`; `range.end == len_range.end()` only gives `len_bytes <= len_range.end()` (the wrong direction), and `len_range.start()` is a populate-time assert, not a circuit constraint, so it can't backstop a malicious prover. Concretely, on a `new_witness` vec it would let `test_slice_const_range_bounds_check_fail`'s bogus witness (a 16-byte slice of 10-byte data) verify. So I skip iff `len_range.start() == len_range.end()`, where `len_bytes` is structurally pinned to the constant and the new compile-time bound above makes `range.end <= len_bytes` provably redundant. Added `test_slice_const_range_const_len_skips_runtime_check` for the skip path; `bounds_check_fail` still guards the dynamic path.

crates/circuits/src/fixed_byte_vec.rs
benji-potamus: Done — uses `range.len()`.

crates/circuits/src/fixed_byte_vec.rs
benji-potamus: Done — dropped the final-word mask in `extract_const_range`. `slice_const_range` and concat no longer zero the trailing bytes past the length, and I updated the slice tests/docs that assumed zero padding.

crates/circuits/src/concat.rs
benji-potamus: Done — since `extract_const_range` no longer zeroes the extracted word's high bytes, `assert_const_slice_eq` now masks **both** sides of the partial boundary word before comparing (it previously relied on the extracted side already being zero).

benji-potamus:
Pushed `76223e6f` addressing all 8 review comments (replies inline on each thread).

- `len_range` is now `RangeInclusive<usize>` (`.contains()` / `.start()` / `.end()` throughout).
- Dropped the trailing-byte mask in `extract_const_range`; `assert_const_slice_eq` now masks **both** sides of the partial boundary word.
- `slice_const_range` bounds against `len_range.end()` instead of capacity.
- The runtime range-check skip is gated on **const-length**, not the literal `range.end == len_range.end()` (which would be unsound for dynamic-length vecs — details in that thread).
- `range.len()` nit; removed `.claude/scheduled_tasks.lock`.

`hashsign`/`ethsign` snapshots unchanged (the boundary mask moves rather than disappears, so the AND count holds). Local: full `binius-circuits` suite, `clippy --all --tests --benches --examples`, `rust-doc`, and `fmt` all green.

jimpo (approved):

PR #1526, opened 2026-06-13
Make Divisible a supertrait of PackedField and Field

benji-potamus:
Resolves [BINIUS-87](https://linear.app/jimpo/issue/BINIUS-87/add-divisibleselfscalar-as-parent-trait-to-packedfield) (PR 1 of 2).

Makes `Divisible` the single source of element access for packed fields: `PackedField` now requires `Divisible<Self::Scalar>` and `Field` requires `Divisible<Self>`. This is the "make an always-definable capability a parent trait" pattern — every implementer supplies it trivially, and callers/downstream signatures pay nothing.

## Commits

1. **Add `Divisible` impls for `PackedPrimitiveType` and field types** — additive, no behavior change. A reflexive `Divisible<Self>` (width 1) for every binary field via the `binary_field!` macro, and a generic `Divisible<Scalar> for PackedPrimitiveType<U, Scalar>` delegating to the underlier's `Divisible<Scalar::Underlier>`.
2. **Make `Divisible` a supertrait of `PackedField` and `Field`** — adds the supertrait bounds; removes `PackedField::{get, broadcast}` in favor of the `Divisible` methods (`get_unchecked`/`set_unchecked` and the `iter*`/`from_fn` helpers stay).

## The `set` collision

`Divisible::set(self, …) -> Self` (by-value) and a mutating `PackedField::set(&mut self, …)` share a name. Once `Divisible` is a supertrait, method resolution prefers the by-value receiver, so a bare `packed.set(i, v)` silently binds to `Divisible::set` and discards the result (caught early: it broke ~96 `binius-field` tests via `from_scalars`/`set_single`).

Per the issue discussion, `set` is the one method with a large mutation-rewrite ripple, so this PR **keeps `PackedField::set(&mut self)`** (its default now delegates to `Divisible::set`) and disambiguates the handful of in-place `.set()` call sites with UFCS `PackedField::set(&mut x, …)` — a mechanical, semantics-preserving change. A `# Note` on the method documents the shadowing. **PR 2 (stacked) removes `PackedField::set`.**

UFCS'd in-place sites: `binius-math` (`batch_invert`, `field_buffer`, `ntt/neighbors_last`), `binius-prover` (`and_reduction/ntt_lookup`, `shift/phase_1`), plus internal `set_single`/`from_scalars` and a couple of tests/benches. `get`/`broadcast` callers resolve to `Divisible` automatically in generic `P: PackedField` code (via the bound); a few concrete-typed sites in `binius-field` gained a `use Divisible`.

## Test plan

- `cargo test --workspace` — green (no behavior change; the silent-`.set()` audit was validated by the test suite catching the one site, `ntt_lookup`, that a build warning didn't).
- `cargo +nightly-2026-01-01 fmt --check` and `cargo clippy --all --tests --benches --examples -- -D warnings` — clean.

### DIFF
--- a/crates/field/src/packed.rs
+++ b/crates/field/src/packed.rs
@@ -18,7 +18,7 @@ use binius_utils::{
 use bytemuck::Zeroable;
 
 use super::{PackedExtension, Random, arithmetic_traits::Square};
-use crate::{BinaryField, Field, WideMul, field::FieldOps};
+use crate::{BinaryField, Divisible, Field, WideMul, field::FieldOps};
 
 /// A packed field represents a vector of underlying field elements.
 ///
@@ -45,9 +45,15 @@ pub trait PackedField:
 	+ Random
 	+ WideMul<Output: Debug + Send + Sync + 'static>
 	+ 'static
+	// A packed field divides into its `WIDTH` scalars. Element access (`get`), broadcast, and the
+	// scalar iterators are provided by this supertrait; `set` is kept here as an in-place `&mut`
+	// convenience (see [`Self::set`]).
+	+ Divisible<Self::Scalar>
 {
 	/// Base-2 logarithm of the number of field elements packed into one packed element.
-	const LOG_WIDTH: usize;
+	///
+	/// This is the number of scalars the packed field divides into, i.e. its `Divisible` log-count.
+	const LOG_WIDTH: usize = <Self as Divisible<Self::Scalar>>::LOG_N;
 
 	/// The number of field elements packed into one packed element.
 	///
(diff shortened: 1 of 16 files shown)


crates/field/src/packed.rs
```diff
@@ -45,6 +45,10 @@ pub trait PackedField:
 	+ Random
 	+ WideMul<Output: Debug + Send + Sync + 'static>
 	+ 'static
+	// A packed field divides into its `WIDTH` scalars. Element access (`get`), broadcast, and the
+	// scalar iterators are provided by this supertrait; `set` is kept here as an in-place `&mut`
+	// convenience (see [`Self::set`]).
+	+ Divisible<Self::Scalar>
 {
 	/// Base-2 logarithm of the number of field elements packed into one packed element.
 	const LOG_WIDTH: usize;
```
jimpo: This could be `= <Self as Divisible<Self::Scalar>>::LOG_N;`

jimpo (approved):

crates/field/src/packed.rs
benji-potamus: Done — `LOG_WIDTH` now defaults to `<Self as Divisible<Self::Scalar>>::LOG_N`, and I dropped the explicit `const LOG_WIDTH` from both impls (the scalar-field blanket and `PackedPrimitiveType`), since `Divisible::LOG_N` is exactly that count. Full gate green (test/fmt/clippy/rustdoc).

PR #1527, opened 2026-06-14
Remove PackedField::set in favor of Divisible::set

benji-potamus:
Resolves [BINIUS-87](https://linear.app/jimpo/issue/BINIUS-87/add-divisibleselfscalar-as-parent-trait-to-packedfield) (PR 2 of 2; follows #1526).

Removes the temporary `PackedField::set(&mut self)` that #1526 kept to avoid a large call-site churn. Now that `Divisible` is `PackedField`'s supertrait, a bare `packed.set(i, v)` resolves unambiguously to `Divisible::set`, so the UFCS disambiguation is gone and the in-place call sites move to the functional `x = x.set(i, v)` form.

Converted sites: `binius-field` (`set_single`, `from_scalars`, and two tests + a bench), `binius-math` (`batch_invert`, `field_buffer` truncate + chunk write-back, `ntt/neighbors_last`), `binius-prover` (`and_reduction/ntt_lookup`, `shift/phase_1`).

## Note on the audit

Removing `PackedField::set` rebinds **every** bare `packed.set(i, v)` to the by-value `Divisible::set` — including sites outside any `use Divisible` scope, because the method comes in via the supertrait bound. This is a silent change (no compile error; the returned value is just discarded if not assigned). One such site — the chunk write-back in `field_buffer.rs` — was not part of the previous PR's UFCS set and was caught here by `test_chunk_mut`. I then audited every remaining `.set(` workspace-wide; all others are `FieldBuffer` / `FieldSliceMut` / `ValueVec` methods, unaffected.

## Test plan

- `cargo test --workspace` — green.
- `cargo +nightly-2026-01-01 fmt --check`, `cargo clippy --all --tests --benches --examples -- -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --document-private-items --workspace`, and an aarch64 `cargo clippy --all-targets -- -D warnings` (with `-Ctarget-feature=+neon,+aes,+sha2,+sha3`) — all clean.

### DIFF
--- a/crates/field/benches/packed_field_element_access.rs
+++ b/crates/field/benches/packed_field_element_access.rs
@@ -34,9 +34,9 @@ fn benchmark_set_impl<P: PackedField>(group: &mut BenchmarkGroup<'_, WallTime>, 
 	group.throughput(Throughput::Elements(BATCH_SIZE as _));
 	group.bench_function(id, |b| {
 		b.iter(|| {
-			indices_values
-				.iter()
-				.for_each(|(i, val)| PackedField::set(&mut value, *i, *val));
+			for (i, val) in &indices_values {
+				value.set(*i, *val);
+			}
 
 			value
 		})
--- a/crates/field/src/arch/aarch64/m128.rs
+++ b/crates/field/src/arch/aarch64/m128.rs
@@ -243,9 +243,9 @@ impl Divisible<u128> for M128 {
 	}
 
 	#[inline]
-	fn set(self, index: usize, val: u128) -> Self {
+	fn set(&mut self, index: usize, val: u128) {
 		match index {
-			0 => Self::from(val),
+			0 => *self = Self::from(val),
 			_ => panic!("index out of bounds"),
 		}
 	}
@@ -291,14 +291,14 @@ impl Divisible<u64> for M128 {
 	}
 
 	#[inline]
-	fn set(self, index: usize, val: u64) -> Self {
-		unsafe {
+	fn set(&mut self, index: usize, val: u64) {
+		*self = unsafe {
 			match index {
 				0 => Self(vsetq_lane_u64(val, self.0, 0)),
 				1 => Self(vsetq_lane_u64(val, self.0, 1)),
 				_ => panic!("index out of bounds"),
 			}
-		}
+		};
 	}
 
 	#[inline]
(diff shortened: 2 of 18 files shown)


jimpo (changes requested):
Actually, change the method signature of `Divisible::set` to take `&mut self` instead of consuming `self` and returning Self. This will make it match the old `PackedField::set` method signature.

benji-potamus:
Done — `Divisible::set` now takes `&mut self` and mutates in place (returning `()`), matching the old `PackedField::set` signature.

What changed:
- Trait method is now `fn set(&mut self, index: usize, val: T)` across every `Divisible` impl (memcast, bitmask sub-byte, the hand-written `SmallU` impls, the reflexive `Divisible<Self>` for field types, `ScaledUnderlier`, `PackedPrimitiveType`, and all 23 SIMD impls in `m128`/`m256`/`m512` for x86_64 and aarch64). SIMD lane-delegating impls now do `get` → in-place sub-`set` → in-place lane-`set`.
- The internal `memcast::set`/`bitmask::set` *helper* free functions still return the modified value by-value (they're construction helpers, not the trait method); their bodies just call the new in-place `set`.
- Call sites that previously did `x = x.set(i, v)` / `*x = x.set(i, v)` are now plain in-place `x.set(i, v)` (batch_invert, neighbors_last, phase_1, ntt_lookup, field_buffer, packed_ghash/packed_binary_field tests, and the access bench). `PackedField::{set_single, from_scalars}` likewise mutate a local in place.

Full local gate green on the new tip (`6f6d426f`): `cargo build --all-targets`, `cargo nextest run`, `cargo test --doc`, `cargo +nightly-2026-01-01 fmt --check`, x86_64 `clippy --all --tests --benches --examples -D warnings`, rustdoc `-D warnings`, and the aarch64 cross-clippy with crypto features (`+neon,+aes,+sha2,+sha3`) — the aarch64 run caught one mangled `u128` set impl that the x86_64 build couldn't see, now fixed. Re-requesting review.

jimpo (approved):

benji-potamus:
Fixed the arm64 + wasm-tests failures from the previous push (`8aefc371`):
- **aarch64** `m128.rs` set impls did `let v: uintNxM_t = self.into()` — now that `self: &mut`, that needed `From<&mut M128>`; changed to `(*self).into()`.
- A few `divisible.rs` unit tests still used the old by-value `Divisible::set` API (`let modified = …::set(val, …)`); converted to in-place.
- One `m512` `u8` set impl was missed in the first pass; fixed to match the others.

Re-gated locally with real exit-code capture this time (the prior run's pipe-to-`tail` had masked failures): `cargo test --workspace` (all green), aarch64 `clippy --all-targets` with crypto features, wasm32 `binius-field` test build, fmt, x86_64 clippy, rustdoc.

PR #1528, opened 2026-06-14
Move get_unchecked and set_unchecked from PackedField to Divisible

benji-potamus:
Resolves [BINIUS-89](https://linear.app/jimpo/issue/BINIUS-89/move-packedfield-get-unchecked-and-set-unchecked-to-divisible).

Follows the BINIUS-87 cleanup (#1526, #1527): now that `Divisible` is a supertrait of `PackedField`, the unchecked scalar accessors belong on `Divisible` alongside `get`/`set`/`broadcast`.

## Change
- Move `get_unchecked`/`set_unchecked` onto the `Divisible` trait as the **required primitives**, and make `get`/`set` defaults that bounds-check then call them (the `Vec` pattern):
  ```rust
  unsafe fn get_unchecked(self, index: usize) -> T;
  unsafe fn set_unchecked(&mut self, index: usize, val: T);
  fn get(self, index: usize) -> T { assert!(index < Self::N, ..); unsafe { self.get_unchecked(index) } }
  fn set(&mut self, index: usize, val: T) { assert!(index < Self::N, ..); unsafe { self.set_unchecked(index, val) } }
  ```
  Each `Divisible` impl's former `get`/`set` body becomes its `*_unchecked` impl, so the bounds check lives only in the safe wrappers and the unchecked path is check-free.
- Make the unchecked impls genuinely zero-check: `N = 1` impls drop the index `match` (index is 0 by the precondition); the multi-index SIMD impls (`m128`/`m256`/`m512` x86_64 + aarch64) replace the `_ => panic!()` arm with `core::hint::unreachable_unchecked()`; the `memcast` and `ScaledUnderlier` bodies use `get_unchecked`/`get_unchecked_mut`. Delegating impls (`M256`/`M512` → `M128`, `ScaledUnderlier` → inner, `PackedPrimitiveType` → underlier) call the inner `*_unchecked`, so the path is check-free end to end.
- Remove `get_unchecked`/`set_unchecked` from the `PackedField` trait and its two impls (the blanket `impl<F: Field>` and `PackedPrimitiveType`); both inherit the `Divisible` methods via the supertrait. `PackedField`'s internal callers (`into_iter`, `iter`, `try_from_fn`) and the `get_packed_slice_unchecked`/`set_packed_slice_unchecked` helpers resolve through the supertrait bound with no call-site churn. The `RandomAccessSequence`/`RandomAccessSequenceMut` impls for `PackedSlice`/`PackedSliceMut` are a different trait and untouched.

## Test plan
- `cargo test --workspace` (CI) — green.
- `cargo +nightly-2026-01-01 fmt --check`, `cargo clippy --all --tests --benches --examples -- -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --document-private-items --workspace`, aarch64 `cargo clippy --target aarch64-unknown-linux-gnu --all-targets -- -D warnings` (with `-Ctarget-feature=+neon,+aes,+sha2,+sha3`), and a `wasm32-wasip1` test build — all clean.

### DIFF
--- a/crates/field/src/arch/aarch64/m128.rs
+++ b/crates/field/src/arch/aarch64/m128.rs
@@ -235,19 +235,13 @@ impl Divisible<u128> for M128 {
 	}
 
 	#[inline]
-	fn get(self, index: usize) -> u128 {
-		match index {
-			0 => self.into(),
-			_ => panic!("index out of bounds"),
-		}
+	unsafe fn get_unchecked(self, _index: usize) -> u128 {
+		self.into()
 	}
 
 	#[inline]
-	fn set(&mut self, index: usize, val: u128) {
-		match index {
-			0 => *self = Self::from(val),
-			_ => panic!("index out of bounds"),
-		}
+	unsafe fn set_unchecked(&mut self, _index: usize, val: u128) {
+		*self = Self::from(val);
 	}
 
 	#[inline]
@@ -280,23 +274,23 @@ impl Divisible<u64> for M128 {
 	}
 
 	#[inline]
-	fn get(self, index: usize) -> u64 {
+	unsafe fn get_unchecked(self, index: usize) -> u64 {
 		unsafe {
 			match index {
 				0 => vgetq_lane_u64(self.0, 0),
 				1 => vgetq_lane_u64(self.0, 1),
-				_ => panic!("index out of bounds"),
+				_ => core::hint::unreachable_unchecked(),
 			}
 		}
 	}
 
 	#[inline]
-	fn set(&mut self, index: usize, val: u64) {
+	unsafe fn set_unchecked(&mut self, index: usize, val: u64) {
 		*self = unsafe {
 			match index {
 				0 => Self(vsetq_lane_u64(val, self.0, 0)),
 				1 => Self(vsetq_lane_u64(val, self.0, 1)),
-				_ => panic!("index out of bounds"),
+				_ => core::hint::unreachable_unchecked(),
 			}
 		};
 	}
@@ -336,7 +330,7 @@ impl Divisible<u32> for M128 {
 	}
 
 	#[inline]
-	fn get(self, index: usize) -> u32 {
+	unsafe fn get_unchecked(self, index: usize) -> u32 {
 		unsafe {
 			let v: uint32x4_t = self.into();
 			match index {
@@ -344,13 +338,13 @@ impl Divisible<u32> for M128 {
 				1 => vgetq_lane_u32(v, 1),
 				2 => vgetq_lane_u32(v, 2),
 				3 => vgetq_lane_u32(v, 3),
-				_ => panic!("index out of bounds"),
+				_ => core::hint::unreachable_unchecked(),
 			}
 		}
 	}
 
 	#[inline]
-	fn set(&mut self, index: usize, val: u32) {
+	unsafe fn set_unchecked(&mut self, index: usize, val: u32) {
 		*self = unsafe {
 			let v: uint32x4_t = (*self).into();
 			match index {
@@ -358,7 +352,7 @@ impl Divisible<u32> for M128 {
 				1 => Self::from(vsetq_lane_u32(val, v, 1)),
 				2 => Self::from(vsetq_lane_u32(val, v, 2)),
 				3 => Self::from(vsetq_lane_u32(val, v, 3)),
-				_ => panic!("index out of bounds"),
+				_ => core::hint::unreachable_unchecked(),
 			}
 		};
 	}
@@ -398,7 +392,7 @@ impl Divisible<u16> for M128 {
 	}
 
 	#[inline]
-	fn get(self, index: usize) -> u16 {
+	unsafe fn get_unchecked(self, index: usize) -> u16 {
 		unsafe {
 			let v: uint16x8_t = self.into();
 			match index {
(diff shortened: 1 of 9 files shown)


crates/field/src/underlier/divisible.rs
```diff
@@ -49,6 +49,24 @@ pub trait Divisible<T>: Copy {
 	/// Panics if `index >= Self::N`.
 	fn set(&mut self, index: usize, val: T);
 
+	/// Get element at index (LSB-first ordering) without bounds checking.
+	///
+	/// # Safety
+	///
+	/// The caller must ensure that `index < Self::N`.
+	unsafe fn get_unchecked(self, index: usize) -> T {
+		self.get(index)
```
jimpo: These implementations are super inefficient. Leave no default implementation, but instead you can implement `get` in terms of `get_unchecked` and `set` in terms of `set_unchecked`, by doing an assertion on index, then an unsafe call.

jimpo (changes requested):

crates/field/src/underlier/divisible.rs
benji-potamus: Makes sense — I'll flip it to the `Vec` pattern: make `get_unchecked`/`set_unchecked` the required trait methods (no default) and provide `get`/`set` as defaults that `assert!(index < Self::N)` then call the unchecked form in an `unsafe` block. Every `Divisible` impl's current `get`/`set` bodies become the `*_unchecked` impls (so the per-element work stays but the bounds check moves out of the hot path), and the delegating SIMD impls (`M256`/`M512` → `M128`, `ScaledUnderlier`) will call the inner `get_unchecked`/`set_unchecked` so the unchecked path is truly check-free end to end. Reworking + re-gating now.

benji-potamus: Done (pushed). Flipped to the `Vec` pattern: `get_unchecked`/`set_unchecked` are now the required trait methods (no default), and `get`/`set` are defaults that `assert!(index < Self::N)` then call the unchecked form in an `unsafe` block. Every `Divisible` impl's prior `get`/`set` body became its `*_unchecked` impl, and the delegating impls (`M256`/`M512` → `M128`, `ScaledUnderlier` → inner, `PackedPrimitiveType` → underlier) now call the inner `*_unchecked` inside `unsafe {}` so the unchecked path is check-free end to end. Verified locally with fmt, `clippy --all --tests --benches --examples -D warnings`, the aarch64 cross-clippy (`+neon,+aes,+sha2,+sha3`), and rustdoc — all clean; CI is running the full test suite.

crates/field/src/arch/aarch64/m128.rs
```diff
@@ -235,15 +235,15 @@ impl Divisible<u128> for M128 {
 	}
 
 	#[inline]
-	fn get(self, index: usize) -> u128 {
+	unsafe fn get_unchecked(self, index: usize) -> u128 {
 		match index {
```
jimpo: get rid of the match. `index` must be 0 by unchecked precondition

crates/field/src/arch/aarch64/m128.rs
```diff
@@ -235,15 +235,15 @@ impl Divisible<u128> for M128 {
...
 			0 => self.into(),
 			_ => panic!("index out of bounds"),
 		}
 	}
 
 	#[inline]
-	fn set(&mut self, index: usize, val: u128) {
+	unsafe fn set_unchecked(&mut self, index: usize, val: u128) {
```
jimpo: same

crates/field/src/arch/aarch64/m128.rs
jimpo: use `unreachable_unchecked!` here now

jimpo (changes requested):
Optimized the get_unchecked/set_unchecked methods using the techniques in the review comments. `unreachable_unchecked` is especially applicable.

crates/field/src/arch/aarch64/m128.rs
benji-potamus: Will do — applying this across all the `*_unchecked` impls: drop the `match` for `N = 1` impls (index is 0 by the unchecked precondition) and replace the `_ => panic!()` arm in the multi-index matches with `core::hint::unreachable_unchecked()`. I'll also switch the `memcast`/`bitmask` macro bodies from panicking `[index]` to `get_unchecked`/`get_unchecked_mut` so those paths are check-free too. Pushing shortly.

crates/field/src/arch/aarch64/m128.rs
benji-potamus: Done (872ebce9) — both `N = 1` `M128` impls (x86_64 + aarch64) drop the `match` entirely and do the index-0 op directly, since `index` is 0 by the unchecked precondition.

crates/field/src/arch/aarch64/m128.rs
benji-potamus: Done (872ebce9) — every multi-index `_ => panic!()` arm in the SIMD `get_unchecked`/`set_unchecked` impls (m128/m256/m512 x86_64 + aarch64 m128) is now `_ => core::hint::unreachable_unchecked()`. Also applied unchecked indexing (`get_unchecked`/`get_unchecked_mut`) to the `memcast` and `ScaledUnderlier` bodies so the whole unchecked path is branch/bounds-check free. native + aarch64 clippy, fmt, rustdoc all clean.

jimpo (approved):

PR #1530, opened 2026-06-14
BINIUS-93: add public-derived wires to the IronSpartan frontend

benji-potamus:
Closes [BINIUS-93](https://linear.app/jimpo/issue/BINIUS-93/add-public-derived-wires-to-the-ironspartan-frontend)

## Summary

Adds a `Derived` wire kind to the IronSpartan constraint system. When all inputs to `add`/`mul`/`hint` are public-derivable (`Constant`, `InOut`, or `Derived`), the output is `Derived` — no constraint is emitted and no private commitment is needed. Derived wires live in the public segment; the verifier recomputes them independently via a new `InstanceGenerator` before calling `verify`.

## Commits

1. **add Derived wire kind and public-derivable inference to spartan-frontend** — `WireKind::Derived`, shared `is_public_derivable` classifier, `ConstraintBuilder`/`WitnessGenerator` dual-allocator branching, derived liveness scan in `finalize`, `WitnessLayout` gains `n_derived`/`derived_index_map`, `WitnessWire` carries kind. Updates `wire_elimination` tests whose `n_private` counts shift.

2. **add InstanceGenerator for verifier-side public reconstruction** — `InstanceGenerator`: a `CircuitBuilder` impl run by the verifier to produce `[constants | inout | derived]`. Calls `f` for derived hints only; returns placeholder private wires without calling `f` on garbage secrets. Four new unit tests: sync, derived-elision, mixed, pruned-intermediate.

3. **build verifier public via InstanceGenerator in spartan-prover integration test** — Updates `integration_test.rs` to construct the public vector via `InstanceGenerator` instead of `witness.public()`. Adds a soundness negative test: prover tampers a derived wire; verifier rejects.

## Test results

- `binius-spartan-frontend`: 26 tests pass (4 new InstanceGenerator tests)
- `binius-spartan-verifier`: full suite pass (proof_size 46848 → 46784, one fewer derived slot committed)
- `binius-spartan-prover`: integration test + soundness test pass
- `cargo clippy --all --tests --benches --examples -- -D warnings`: clean
- `cargo +nightly-2026-01-01 fmt`: clean

---

### DIFF
--- a/crates/spartan-frontend/src/circuit_builder.rs
+++ b/crates/spartan-frontend/src/circuit_builder.rs
@@ -488,6 +488,124 @@ impl<'a, F: Field> CircuitBuilder for WitnessGenerator<'a, F> {
 	}
 }
 
+/// Wire type for [`InstanceGenerator`]: holds the field value of a public wire (a constant, inout,
+/// or derived value the verifier can compute) and is `None` for private/precommit wires, whose
+/// values the verifier does not know. A wire being `Some` mirrors [`WireKind::is_public`], so the
+/// derived-vs-private branching stays aligned with the other builders.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PublicWire<F: Field>(Option<F>);
+
+/// Reconstructs the public input vector `[constants | inout | derived]` on the verifier side by
+/// re-running the circuit function.
+///
+/// Implements [`CircuitBuilder`] with [`PublicWire`] as the wire type, structurally mirroring
+/// [`WitnessGenerator`], but it only computes the *public* segment. The verifier lacks the secret
+/// (precommit) values, so private wires carry no value; by the soundness invariant a derived wire
+/// never reads a secret, so every derived value is computable from the real public inputs alone.
+/// Derived wires are allocated in the same order as [`ConstraintBuilder`], keeping their ids
+/// aligned so [`WitnessLayout::get`] resolves each to the correct public slot.
+#[derive(Debug)]
+pub struct InstanceGenerator<'a, F: Field> {
+	derived_alloc: WireAllocator,
+	public: Vec<F>,
+	layout: &'a WitnessLayout<F>,
+}
+
+impl<'a, F: Field> InstanceGenerator<'a, F> {
+	pub fn new(layout: &'a WitnessLayout<F>) -> Self {
+		let mut public = zeroed_vec(layout.public_size());
+		public[..layout.constants.len()].copy_from_slice(&layout.constants);
+
+		Self {
+			derived_alloc: WireAllocator::new(WireKind::Derived),
+			public,
+			layout,
+		}
+	}
+
+	/// Writes a real inout value into its public slot.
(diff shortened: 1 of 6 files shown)


crates/spartan-frontend/src/circuit_builder.rs
```diff
@@ -301,12 +337,15 @@ pub struct WitnessError {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct WitnessWire<F: Field>(F);
+pub struct WitnessWire<F: Field> {
+	val: F,
+	kind: WireKind,
```
jimpo: Why do we need this? What if only WitnessGenerator uses it, and not InstanceGenerator? InstanceGenerator could use a different type like:

```
// A value if public, None if private
struct PublicWire<F>(Option<F>);
```

crates/spartan-frontend/src/constraint_system.rs
```diff
@@ -15,10 +16,24 @@ use smallvec::{SmallVec, smallvec};
...
 	Private,
 }
 
+/// Returns whether a wire kind is public-derivable: its value is determined once the public inputs
+/// (constants and inout) are known, so the verifier can recompute it without any secret data.
+///
+/// Shared by all `CircuitBuilder` implementations so they make identical derived-vs-private
+/// decisions. The output of a binary op (or hint) is [`WireKind::Derived`] iff every input is
+/// public-derivable, and [`WireKind::Private`] otherwise.
+pub(crate) fn is_public_derivable(kind: WireKind) -> bool {
```
jimpo: Make this a pub method `WireKind::is_public(self) -> bool`

benji-potamus: Done — replaced the free `is_public_derivable` function with a `pub fn WireKind::is_public(self) -> bool` method and updated all call sites to use it. (1804be5f)

crates/spartan-frontend/src/circuit_builder.rs
benji-potamus: Done. `InstanceGenerator` now uses a dedicated `struct PublicWire<F>(Option<F>)` as its wire type — `Some(v)` is a public/derived value, `None` is a private/precommit wire (the verifier has no secret for it). `is_some()` is exactly `is_public`. `WitnessWire { val, kind }` is now used only by `WitnessGenerator`, and the `placeholder_private` helper was removed. (1804be5f)

jimpo: Perhaps instead of a full `WireKind`, just use `WitnessSegment`

benji-potamus: Done. `WitnessWire` now stores a `WitnessSegment` instead of a full `WireKind` — it only ever needed the public-vs-not distinction (`is_public`), which is exactly "is this in the public segment". Added `WireKind::segment()` to map kind → segment in one place, and `WireKind::is_public()` now delegates to it. (8e5b1614)

jimpo (approved):

PR #1534, opened 2026-06-15
BINIUS-85: Add Maskable trait for branchless conditional lane masking

benji-potamus:
Closes BINIUS-85.

## Summary

Adds a `Maskable<T>` trait to the field crate and migrates the shift protocol's branchless masking off direct underlier access.

**New trait** (`crates/field/src/underlier/maskable.rs`):
```rust
pub trait Maskable<T>: Divisible<T> + Copy + Zeroable {
    type Mask: Send + Sync;
    fn make_mask(selectors: impl Iterator<Item = bool>) -> Self::Mask;
    fn select(&self, mask: &Self::Mask) -> Self;
}
```
`Maskable<T>` is now a parent trait of both `Field` (as `Maskable<Self>`) and `PackedField` (as `Maskable<Self::Scalar>`), mirroring the existing `Divisible` pattern.

**Implementations:**
- `BinaryField` types via the `binary_field!` macro: `Mask = $typ` (the underlier integer), `make_mask` uses `fill_with_bit(0/1)`, `select` is `Self(self.0 & *mask)` — same AND-with-bitmask strategy as `PackedPrimitiveType`.
- `PackedPrimitiveType<U, Scalar>`: `Mask = U` (the packed underlier), `make_mask` uses `<U as Divisible<Scalar::Underlier>>::from_iter(selectors.map(|s| fill_with_bit(u8::from(s))))`, `select` is a single bitwise AND — `from_underlier(self.to_underlier() & *mask)`.

The `Send + Sync` bound on `Mask` (beyond jimpo's sketch) is needed so a precomputed `Vec<Self::Mask>` mask table can be shared across rayon threads.

**Prover migration** (`crates/prover/src/protocols/shift/phase_1.rs`):
- Precomputes the mask table via `make_mask` and accumulates with `select`, replacing manual underlier bit manipulation.
- Drops `WithUnderlier` bounds from `phase_1.rs`, `shift/prove.rs`, `prover/src/prove.rs`, and `prover/src/zk_config.rs` (3 impl sites) — the direct underlier access is now fully encapsulated in `Maskable`.

## Performance

Measured on the `shift_reduction` benchmark (log2_size=12, sequential, `RUSTFLAGS="-C target-cpu=native"`):

| | Median | Range |
|---|---|---|
| `main` (8884039d) | 86.8 ms | 82–95 ms (noisy) |
| this branch | 77.2 ms | 76–79 ms |

The branch is within noise — the codegen is identical (`select` lowers to `from_underlier(acc & mask)`, same precomputed table as before). The ~11% difference is measurement noise on a loaded machine; do not read it as a speedup claim.

---

### DIFF
--- a/crates/field/src/arch/portable/packed.rs
+++ b/crates/field/src/arch/portable/packed.rs
@@ -22,7 +22,7 @@ use rand::{
 };
 
 use crate::{
-	BinaryField, Divisible, ExtensionField, Field, PackedField, WideMul,
+	BinaryField, Divisible, ExtensionField, Field, Maskable, PackedField, WideMul,
 	arithmetic_traits::{InvertOrZero, Square},
 	field::FieldOps,
 	underlier::{NumCast, UnderlierType, WithUnderlier},
@@ -468,6 +468,34 @@ where
 	#[inline]
 	fn from_iter(iter: impl Iterator<Item = Scalar>) -> Self {
 		<U as Divisible<Scalar::Underlier>>::from_iter(iter.map(Scalar::to_underlier)).into()
+	}
+}
+
+// Lane masking lowers to a single bitwise AND against an all-ones/all-zeros per-lane mask, the same
+// operation the shift protocol previously performed by reaching into the underlier directly.
+impl<U, Scalar> Maskable<Scalar> for PackedPrimitiveType<U, Scalar>
+where
+	U: UnderlierType + Divisible<Scalar::Underlier>,
+	Scalar: BinaryField,
+{
+	type Mask = U;
+
+	#[inline]
+	fn make_mask(selectors: impl Iterator<Item = bool>) -> U {
+		// Build a per-lane all-ones/all-zeros sub-underlier for each scalar slot and pack into U.
+		// A selected lane produces fill_with_bit(1) (every bit set); unselected gives ZERO.
+		<U as Divisible<Scalar::Underlier>>::from_iter(
+			selectors
+				.take(<Self as Divisible<Scalar>>::N)
+				.map(|selected| {
+					<Scalar::Underlier as UnderlierType>::fill_with_bit(u8::from(selected))
+				}),
+		)
+	}
+
+	#[inline]
+	fn select(&self, mask: &U) -> Self {
+		Self::from_underlier(self.to_underlier() & *mask)
 	}
 }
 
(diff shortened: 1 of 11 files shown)


crates/field/src/binary_field.rs
```diff
@@ -278,6 +278,22 @@ macro_rules! binary_field {
 			}
 		}
 
+		// As a width-one packed field, a field element is masked by its single selector: kept
+		// when selected, otherwise zeroed.
+		impl $crate::Maskable<$name> for $name {
```
jimpo, line 284: This should use the same sort of underlier bitmask masking strategy as `PackedPrimitiveType`.

crates/field/src/arch/portable/packed.rs
```diff
@@ -471,6 +471,38 @@ where
...
+
+	#[inline]
+	fn make_mask(selectors: impl Iterator<Item = bool>) -> U {
+		// A selected lane is all-ones (every bit set) and an unselected lane all-zeros, so ANDing a
+		// value against the mask keeps or zeroes each lane. Build it by setting the selected scalar
+		// lanes of a zeroed value to the all-ones scalar, then peel to the underlier so the mask
+		// can be precomputed once and reused across many `select` calls.
+		let all_ones =
+			Scalar::from_underlier(<Scalar::Underlier as UnderlierWithBitOps>::fill_with_bit(1));
+		let mut mask = Self::zeroed();
```
jimpo: Instead use `U::from_iter` (on Divisible), and generate each sub-underlier in the iterator with `fill_with_bit`.

jimpo (changes requested):

crates/field/src/binary_field.rs
benji-potamus: Done — `type Mask = $typ` (the underlier integer type in the macro) with `fill_with_bit(0/1)` for `make_mask` and `Self(self.0 & *mask)` for `select`. Same AND-with-bitmask strategy as `PackedPrimitiveType`.

crates/field/src/arch/portable/packed.rs
benji-potamus: Done — replaced the loop with `<U as Divisible<Scalar::Underlier>>::from_iter(selectors.take(N).map(|s| fill_with_bit(u8::from(s))))`.

jimpo (approved):

PR #1535, opened 2026-06-15
BINIUS-69: Combine non-ZK with ZK BaseFold compiler (C1+C2+tests)

benji-potamus:
Partially closes BINIUS-69 (step E — delete non-ZK family and rename `BaseFoldZK*` → `BaseFold*` — will follow in a second commit once this soundness-critical core is approved).

## What this PR does

Four commits on top of `main` (which already contains BINIUS-94 = #1532, the FRI skip-window prerequisite):

**Commit 1 — Generalize batched BaseFold inner-round count**
Replaces the hardcoded `batch_challenge: F` parameter in `verify/prove_mlecheck_basefold_zk_batch` with `inner_challenges: &[F]` (a slice). Currently all callers pass `&[γ]` (all-ZK) or `&[]` (zero-ZK), so behaviour is unchanged, but the core no longer assumes a fixed single masking round.

**Commit 2 — Add `is_zk` to `OracleSpec`; make masking conditional on ZK oracles**
Foundation for the heterogeneous opening:
- `OracleSpec` gets `is_zk: bool` with `new()` (false) / `new_zk()` (true) constructors.
- Both compilers map `is_zk` → `{log_msg_len+1, batch=1}` / `{log_msg_len, batch=0}` in `PartialOracleSpec`.
- `CommittedOracleData.mask` becomes `Option`; `send_oracle` branches to `commit_masked` (ZK) vs `commit_interleaved` (non-ZK).
- σ_i inner products and the `(1−γ)π + γω` blinding are computed only for ZK oracles; γ is sampled only when `n_zk > 0`.
- Behaviour-preserving on the all-ZK path (every current oracle is ZK).

**Commit 3 — C1: decouple combined-opening dimension from FRI codeword dim**
`max_n = max_i oracle_specs[i].log_msg_len` (message dim, can exceed `rs_code().log_dim()` when non-ZK oracles are larger). Removes the `max_n == D` assumption from `basefold.rs` and the prover. Still behaviour-preserving on all-ZK.

**Commit 4 — C2: heterogeneous mixed ZK/non-ZK opening + guardrail tests**
The soundness core:
- **Compilers**: non-ZK oracles use `log_batch_size = Some(0)` (no FRI interleave batch for now — see below) and `skip_batch_challenges = 1` when ZK oracles are present, so the FRI routes γ only to ZK oracles and non-ZK oracles skip the γ slot (per #1532's `skip_batch_challenges` mechanism).
- **Basefold assertion**: `max_log_batch_size` → `max_inner_challenges = max(skip + batch)` (the inner-challenge count the FRI actually draws).
- **`wiring.rs`**: missed `is_zk` field added (non-ZK, caught by reverse-dep build).
- **Three new guardrail tests** (D from the plan): `test_basefold_zk_channel_mixed_zk_non_zk` (one non-ZK + one ZK oracle), `test_basefold_zk_channel_zero_zk` (two non-ZK oracles, γ never sampled), `test_basefold_zk_channel_multi_zk_shared_gamma` (three ZK oracles, single shared γ).

## Design decision on non-ZK batch size

Non-ZK oracles currently use `log_batch_size = 0` (no FRI interleaving). This sidesteps a subtle soundness constraint: when multiple non-ZK oracles have different sizes, the optimizer may choose different `log_batch_size` values for each. For the combined MLE-check's inner challenges to be consistent across oracles, **all non-ZK oracles must use the same batch fold variables** — which holds only when they share the same `log_batch_size`. The simplest correct fix is `Some(0)` for all non-ZK. Restoring the proof-size optimization via flexible non-ZK batching is filed as a follow-up to BINIUS-69.

## What's NOT in this PR (step E)

Step E — deleting `basefold_channel.rs` (non-ZK) from both crates and renaming `BaseFoldZK*` → `BaseFold*` everywhere — will follow in a second commit after this core is approved. That rename touches ~20 files and changes the non-ZK proof format (eagerly per-oracle FRI → single batched ZK channel with no masking), which was pre-approved in the BINIUS-69 thread (Q2).

---

### DIFF
--- a/crates/iop-prover/src/basefold.rs
+++ b/crates/iop-prover/src/basefold.rs
@@ -179,11 +179,13 @@ where
 /// * `witness` - the combined oracle multilinear `𝛑` with `log_len = 𝐧`
 /// * `eval_point` - the point `r` with `len = 𝐧`, in low-to-high variable order
 /// * `eval_claim` - the combined target `s' = 𝛑(r)`
-/// * `batch_challenge` - the masking challenge `γ`; folds each interleaved `[π_i ‖ ω_i]` codeword
-///   down to the codeword of `π_i'` in the FRI inner (unbatch) round
+/// * `inner_challenges` - the masking challenge(s) `γ`, one per inner (unbatch) round; each folds
+///   every interleaved `[π_i ‖ ω_i]` codeword down to the codeword of `π_i'`. There are
+///   `max_log_batch_size` of them: a single `γ` when any oracle is masked, empty when none is.
 /// * `outer_challenges` - the batching challenges `r'` (`len = log_n_oracles`); combine the `k`
 ///   lifted codewords in the FRI outer (oracle-combine) rounds
-/// * `fri_folder` - the combined FRI fold prover, with `n_rounds == 𝐧 + 1 + log_n_oracles`
+/// * `fri_folder` - the combined FRI fold prover, with
+///   `n_rounds == 𝐧 + max_log_batch_size + log_n_oracles`
 /// * `transcript` - the prover transcript
 ///
 /// The final FRI value equals the final MLE-check value `𝛑(r)` (see
@@ -193,7 +195,7 @@ pub fn prove_mlecheck_basefold_zk_batch<'a, F, P, NTT, MerkleScheme, MerkleProve
 	witness: FieldBuffer<P>,
 	eval_point: &[F],
 	eval_claim: F,
-	batch_challenge: F,
+	inner_challenges: &[F],
 	outer_challenges: &[F],
 	mut fri_folder: FRIFoldProver<'a, F, P, NTT, MerkleProver>,
 	transcript: &mut ProverTranscript<Challenger_>,
@@ -210,14 +212,20 @@ where
 
 	let n_vars = witness.log_len();
 	assert_eq!(eval_point.len(), n_vars);
-	// The FRI folder has one inner (unbatch) round, `log_n_oracles` outer (oracle-combine) rounds,
-	// and `𝐧` standard fold rounds.
-	assert_eq!(n_vars + 1 + outer_challenges.len(), fri_folder.n_rounds());
+	// The FRI folder has `max_log_batch_size` inner (unbatch) rounds, `log_n_oracles` outer
+	// (oracle-combine) rounds, and `𝐧` standard fold rounds.
+	assert_eq!(
+		n_vars + inner_challenges.len() + outer_challenges.len(),
+		fri_folder.n_rounds()
+	);
 
-	// Inner (unbatch) round: fold every interleaved (π_i ‖ ω_i) codeword at the masking challenge.
-	fri_folder.receive_challenge(batch_challenge);
+	// Inner (unbatch) rounds: fold every interleaved (π_i ‖ ω_i) codeword at the masking challenge(s).
+	// Empty when there are no masked oracles.
+	for &inner_challenge in inner_challenges {
+		fri_folder.receive_challenge(inner_challenge);
+	}
 	// Outer rounds: combine the k lifted codewords at the batching challenges r'. These carry no
-	// sumcheck round-polynomial; the folder applies them lazily with γ at the first commit round.
+	// sumcheck round-polynomial; the folder applies them lazily at the first commit round.
 	for &outer_challenge in outer_challenges {
 		fri_folder.receive_challenge(outer_challenge);
 	}
@@ -518,7 +526,7 @@ mod test {
 			witness_prime,
 			&evaluation_point,
 			eval_claim,
-			batch_challenge,
+			&[batch_challenge],
 			&[],
 			fri_folder,
 			&mut prover_transcript,
@@ -538,7 +546,7 @@ mod test {
 			&[retrieved_commitment],
 			eval_claim,
 			&evaluation_point,
-			batch_challenge_v,
+			&[batch_challenge_v],
 			&[],
 			&mut verifier_transcript,
 		)?;
--- a/crates/iop-prover/src/basefold.rs
+++ b/crates/iop-prover/src/basefold.rs
@@ -212,8 +212,7 @@ where
 
 	let n_vars = witness.log_len();
 	assert_eq!(eval_point.len(), n_vars);
-	// The FRI folder has `max_log_batch_size` inner (unbatch) rounds, `log_n_oracles` outer
-	// (oracle-combine) rounds, and `𝐧` standard fold rounds.
+	// n_vars = D (rs_code.log_dim()), inner_challenges.len() = max_inner_challenges, outer = log_n.
 	assert_eq!(n_vars + inner_challenges.len() + outer_challenges.len(), fri_folder.n_rounds());
 
 	// Inner (unbatch) rounds: fold every interleaved (π_i ‖ ω_i) codeword at the masking
--- a/crates/iop-prover/src/basefold.rs
+++ b/crates/iop-prover/src/basefold.rs
@@ -184,8 +184,8 @@ where
 ///   `max_log_batch_size` of them: a single `γ` when any oracle is masked, empty when none is.
 /// * `outer_challenges` - the batching challenges `r'` (`len = log_n_oracles`); combine the `k`
 ///   lifted codewords in the FRI outer (oracle-combine) rounds
-/// * `fri_folder` - the combined FRI fold prover, with
-///   `n_rounds == 𝐧 + max_log_batch_size + log_n_oracles`
+/// * `fri_folder` - the combined FRI fold prover, with `n_rounds == 𝐧 + max_log_batch_size +
+///   log_n_oracles`
 /// * `transcript` - the prover transcript
 ///
 /// The final FRI value equals the final MLE-check value `𝛑(r)` (see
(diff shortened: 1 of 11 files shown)


jimpo:
The `log_batch_size = 0` restriction for non-ZK oracles is wrong. With https://github.com/binius-zk/binius64/pull/1532 merged, it should be unnecessary by setting `skip_batch_challenges` appropriately.

benji-potamus:
On it — dropping the `log_batch_size = Some(0)` restriction. Non-ZK oracles will use `PartialOracleSpec.log_batch_size = None` (optimizer-chosen) + `skip_batch_challenges = (n_zk > 0) as usize` to skip the ZK γ slot.

Confirming the **channel plumbing** before I touch the soundness core, since the FRIParams knob is only half of it — the batched MLE-check's `eval_point`/`inner_challenges` split has to move too (today both assume non-ZK batch = 0). In every real config there's **≤1 non-ZK oracle** (binius64 outer = one `OracleSpec::new`; spartan = all `new_zk`; `zk_config` = that one non-ZK + the 3 ZK spartan oracles), so I'm implementing exactly your 20/10 example, not a general multi-non-ZK mapping.

Derived mechanics — one non-ZK oracle (optimizer batch `b`) + ZK oracles; `D = rs_code().log_dim()` (the optimizer's chosen reduced dim), `max_n` = max message dim, `b = max_n − D`:

1. Phase-A sumcheck stays `max_n`-variate → `point` (len `max_n`). The non-ZK oracle's `b` batch dims are its top `b` vars = the first `b` sumcheck rounds = `point[D..max_n]`.
2. `inner_challenges = [γ] ++ point[D..max_n]`, routed by skip windows (ZK `inner[0..1] = γ`, non-ZK `inner[1..1+b]`); `max_inner = 1 + b`.
3. Combined buffer built at **`D`** (fold the non-ZK message's top `b` vars at `point[D..max_n]`; ZK messages padded as today), `eval_point = point[..D]`. `prove/verify_mlecheck_basefold_zk_batch` stay as-is (they already require `eval_point.len() == D`).
4. `s' = Σ_i e[i]·α_i·eq_ind_zero(point[n_i′..D])`, with `n_i′ = D` for the full-width non-ZK oracle and `n_i` for each ZK oracle.

Does that match your intent — in particular the non-ZK batch challenges coming from the **Phase-A sumcheck point** (`point[D..max_n]`) rather than being freshly sampled? If so I'll wire it up and add a mixed test where the optimizer actually picks `b > 0`. Doing the mechanical compiler-spec half meanwhile.

jimpo:
Let's make this general and keep it robust to multiple ZK oracles and multiple non-ZK oracles.

We have `k` oracles, some are ZK, some aren't. After FRIParams selection, we have chosen a `log_dim` for the RS code.

Each FRI oracle additionally has:

* log_msg_len
* log_batch_size
* skip_batch_challenges

And for each, we can compute `log_lift = log_dim - (log_msg_len - log_batch_size)`.

The total number of inner challenges is `n_inner_challenges = max(log_batch_size + skip_batch_challenges)` across all oracles. For each, ZK oracle, log_batch_size = 1 and skip_batch_challenges = 0. For each non-ZK oracle, we need to ensure that `skip_batch_challenges = n_inner_challenges - log_batch_size` to ensure the batching challenges are a suffix of the inner_challenges. This means the FRIParams selection should be ZK-aware: it needs to know to choose skip_batch_challenges in this way. In FRIParams, it can determine `n_inner_challenges = max(log_batch_size among the non-ZK oracles) + max(log_batch_size among the ZK oracles — either 0 or 1)`.

When constructing the combined buffer, the individual oracle messages need to be accumulated in, taking into account their log_lift and repetition amounts. For all the ZK oracles, we begin by folding with the batch challenge. Then they have a `log_lift` (which corresponds to zero padding) and a repetition 2^(n_inner_challenges - 1) times. For the non-ZK oracles, they may have a `log_lift` (also zero-padding) and a repetition 2^`skip_batch_challenges` times. So that means splitting the combined buffer into 2^skip_batch_challenges chunks, and adding in the scaled oracle message into the first 2^(log_dim - log_lift) spots.

Does that design make sense? I think you probably want to start the implementation over, dropping #1535 because it's messy.

benji-potamus:
Makes sense — agreed on starting fresh (I'll abandon #1535 and rebuild cleanly on current main, general over any mix of ZK/non-ZK oracles). Playing back the core to be sure before I write soundness-critical code:

**ZK-aware FRIParams selection.** Per-oracle `is_zk`; ZK ⇒ `log_batch_size = 1`, non-ZK ⇒ optimizer-chosen. After batch selection, `n_inner = max(log_batch_size over non-ZK) + max(log_batch_size over ZK)` (the ZK term is 0 or 1). Then ZK ⇒ `skip = 0` (its γ batch is `inner[0]`, the prefix); non-ZK ⇒ `skip = n_inner − log_batch_size` (its batch is the **suffix** `inner[skip..n_inner]`). So every oracle covers all `n_inner` inner dims = its own batch window + repetition over the rest (ZK repeats `2^(n_inner−1)`, non-ZK repeats `2^skip`), which is what makes the inner challenges consistent across heterogeneous batch sizes — thanks, that resolves the worry that pushed the earlier PR to `batch=0`.

**One thing to pin before I build it:** the combined-buffer / MLE-check dimension. Two readings of your buffer construction:

- **(i)** the combined MLE-check buffer is `log_dim`-variate; the `n_inner` inner folds (γ + the batch suffixes) are the FRI **inner rounds**, and `prove/verify_mlecheck_basefold_zk_batch` keep `eval_point.len() == log_dim` (the channel just supplies `inner_challenges` of length `n_inner` extracted from the Phase-A point + γ); or
- **(ii)** the combined buffer is `(log_dim + n_inner)`-variate and the MLE-check itself runs the `n_inner` extra rounds (mlecheck core changes).

Your "split into `2^skip` chunks, add the message into the first `2^(log_dim − log_lift)` spots, repeat" reads like the buffer carries the inner dims explicitly — i.e. **(ii)**, an `(log_dim + n_inner)`-variate buffer. Can you confirm it's (ii) (so I generalize the mlecheck round count to `n_inner + log_dim`), or (i)?

Starting now on the unambiguous half — `PartialOracleSpec`/FRIParams ZK-awareness (the `is_zk` flag + the `skip = n_inner − batch` rule) and the compiler specs — and will wire the combined buffer per your answer.

jimpo:
Let me clarify and redefine two things:

* `n_inner = max(log_batch_size over non-ZK)`oracles
* `n_batch_challenges = n_inner + max(log_batch_size over ZK)`

The combined buffer is then `log_dim + n_inner`-variate and the MLE-check runs the `n_inner` rounds.

The order of sampling challenges must then be:

* The ZK batch challenge gamma (sampled in Phase A)
* The outer challenges (ceil(log2(k)) - many)
* Then the inner batch challenges
* Then the log_dim reduced folding challenges.

benji-potamus:
Got it — (ii): combined buffer `(log_dim + n_inner)`-variate, MLE-check runs `n_inner + log_dim` rounds, with `n_inner = max(batch over non-ZK)` and `n_batch_challenges = n_inner + max(batch over ZK)`. And the sampling order γ (Phase A) → outer → inner batch → reduced folding — noting that reorders **outer before inner** vs. today's FRI fold, so the `FRIFold`/batch machinery picks up that reordering too, not just the channel. Building it now on a fresh branch; I'll post the new PR with the full diff + the round-order/soundness notes before merge.

benji-potamus:
Foundation's landed (is_zk on OracleSpec/PartialOracleSpec, ZK-aware skip selection, conditional + ZK-only masking) and is green/behaviour-preserving on the all-ZK path. Before the MLE-check core I traced `FRIFold` to pin the buffer/round mechanics, and hit a fork I'd rather confirm than guess on:

**What the current `FRIFold` does** (`iop-prover/src/fri/fold.rs`, `iop/src/fri/verify.rs`): the **first fold consumes the inner challenges (`max(skip + batch)` of them) AND the outer challenges together**, leaving the combined codeword at `rs_code.log_dim() = reduced_log_msg_len`; then `sum(fold_arities)` standard rounds reduce it to the terminal dim. Each oracle is lifted into the combined codeword by repeating it `2^log_lift` times (`log_lift = log_dim − (log_msg_len − log_batch_size)`).

So **today the non-ZK batch folds happen in the first fold via the skip windows** — which is exactly what my foundation's `skip = n_batch_challenges − log_batch_size` selection wires up. That's model **(i)**: combined MLE-check buffer is `reduced`-variate (`eval_point.len() == rs_code.log_dim()`), and `inner_challenges = [γ, …non-ZK batch…]` (length `n_batch_challenges`) is passed to the mlecheck — γ sampled in Phase A, the non-ZK batch challenges sampled after the outer round.

For model **(ii)** as you described it (combined buffer `reduced + n_inner`-variate, the `n_inner` non-ZK batch dims folding **inside** the MLE-check as standard rounds via the repetition), `FRIFold` would need to (a) size `rs_code.log_dim() = reduced + n_inner`, and (b) make the first fold consume **only γ + outer** (not the non-ZK batch), so the `n_inner` batch dims become standard fold rounds in lockstep with the MLE-check. That's a genuine change to the first-fold + params sizing.

Both give the same transcript order (γ → outer → inner batch → reduced folding); they differ only in whether the inner batch folds **in the first fold** (i) or **as standard MLE-check rounds** (ii).

Which do you want?
- **(i)** — keep `FRIFold` as-is; non-ZK batch folds in the first fold via skip windows, MLE-check buffer at `reduced`, `inner_challenges = [γ, batch…]`. Much smaller change, and it's what the foundation already supports.
- **(ii)** — modify `FRIFold` (first fold takes only γ+outer; rs_code at `reduced + n_inner`); batch folds as standard MLE-check rounds.

I lean (i) since it needs no FRIFold surgery and the skip-window routing already passes its tests — unless (ii)'s explicit `reduced + n_inner` buffer is load-bearing for the soundness argument. Your call.

jimpo:
Post the Foundation as a draft PR so I can see if you're on the right track

benji-potamus:
Foundation draft is up: **#1539**. It's the behaviour-preserving groundwork only — `is_zk` on `OracleSpec`/`PartialOracleSpec` + the ZK-aware skip selection (ZK → skip 0, non-ZK → `n_batch_challenges − batch` suffix). All-ZK byte-identical; `cargo test -p binius-iop -p binius-iop-prover` + workspace check + fmt all green.

The MLE-check core isn't in it — it's gated on your **(i) vs (ii)** call above (does the non-ZK batch fold in the FRI first fold via the skip windows this foundation already wires up — model (i), no `FRIFold` change — or as standard MLE-check rounds — model (ii), needs `FRIFold` surgery + `rs_code` at `reduced + n_inner`). Take a look at #1539 and let me know if the skip-selection direction matches what you intended.

benji-potamus:
Closing this in favor of the rebuilt foundation in #1539 (per your "start over, drop #1535" note). The general design is being built up from there.

PR #1536, opened 2026-06-15
BINIUS-88: remove Copy supertrait from Divisible&lt;T&gt;

benji-potamus:
Resolves [BINIUS-88](https://linear.app/jimpo/issue/BINIUS-88/remove-copy-supertrait-from-divisiblet).

## What this does

Drops the `Copy` supertrait from `Divisible<T>`. The bound wasn't necessary — the only thing that required `Self: Copy` was that `get`/`get_unchecked` took `self` by value. Per the issue's suggestion, those now take `&self`, and the supertrait is relaxed to the minimal `Sized` (still needed because `broadcast`/`from_iter` return `Self`).

## Changes (single commit, `binius-field` only)

- `trait Divisible<T>: Copy` → `trait Divisible<T>: Sized`.
- `fn get(self, ..)` / `unsafe fn get_unchecked(self, ..)` → `&self`, across the trait and every impl (portable, x86_64 m128/m256/m512, aarch64 m128, reflexive `binary_field!`/self macros, `SmallU`, `ScaledUnderlier`).
- Call sites adjusted to pass references (`&value`, `&self.0`, `(*self).into()` on the aarch64 SIMD paths where an owned `From` conversion is needed).
- One internal helper bound added: `mapget::value_iter` gains `Big: Clone` — its owning closure must be `Clone` to satisfy the iterator's `+ Clone` contract, which `Copy` previously provided for free. Every underlier is `Clone`, so this is invisible to callers and stays inside the crate.

`set`/`set_unchecked` already took `&mut self`, so no change there. The removal needed no `+ Copy`/`+ Clone` bounds added to any downstream generic — the supertrait really was unnecessary.

## Verification

Workspace build + tests, x86_64 clippy, aarch64 clippy (`+neon,+aes,+sha2,+sha3`), rustdoc, wasm32-wasip1 build, and fmt all green.

### DIFF
--- a/crates/field/src/arch/aarch64/m128.rs
+++ b/crates/field/src/arch/aarch64/m128.rs
@@ -235,8 +235,8 @@ impl Divisible<u128> for M128 {
 	}
 
 	#[inline]
-	unsafe fn get_unchecked(self, _index: usize) -> u128 {
-		self.into()
+	unsafe fn get_unchecked(&self, _index: usize) -> u128 {
+		(*self).into()
 	}
 
 	#[inline]
@@ -274,7 +274,7 @@ impl Divisible<u64> for M128 {
 	}
 
 	#[inline]
-	unsafe fn get_unchecked(self, index: usize) -> u64 {
+	unsafe fn get_unchecked(&self, index: usize) -> u64 {
 		unsafe {
 			match index {
 				0 => vgetq_lane_u64(self.0, 0),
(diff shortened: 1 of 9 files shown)


jimpo (approved):

benji-potamus:
CI `rust-checks-x86_64` failed at the Compile step on an AVX-512-only path that my local gate can't compile (no AVX-512 hardware here). The `m512` `Divisible::<uN>` impls extract an `M128` lane and then sub-index it, so two call sites passed an owned value where `&self` is now required:

- `get_unchecked`: `Divisible::<uN>::get_unchecked(lane, ..)` → `&lane`
- `set_unchecked`: `Divisible::<M128>::get_unchecked(*self, ..)` → `&*self`

(The `m256`/`m128` impls use direct extract intrinsics rather than lane re-indexing, so they were unaffected.) Fixed in `d087aa1b`. Grepped the rest of the arch tree for any other by-value `Divisible::<_>::get(_unchecked)` UFCS calls — none remain. Re-running CI.

PR #1538, opened 2026-06-15
BINIUS-99: remove InOut lazy evaluation from Spartan wrapper logic

benji-potamus:
Closes [BINIUS-99](https://linear.app/jimpo/issue/BINIUS-99/remove-the-inout-lazy-evaluation-from-spartan-wrapper-logic).

## Summary

The three `CircuitWire` backends in the Spartan wrapper (`BuilderWire`, `WrappedWire`, `WitnessGenWire`) each reimplemented "lazy inout": tracking values as `Constant`/`InOut`/`Private` and materializing inout wires only when mixed with a private wire. BINIUS-93 moved exactly that public-derivable elision into the frontend builders (derived wires get a public slot only when a surviving constraint references them), so the wrapper layer was redundant.

Each wire is now a thin newtype that delegates straight to its frontend builder; inout/precommit wires are allocated eagerly and the builder decides derived-vs-private:

- `BuilderWire(ConstraintWire)` → `ConstraintBuilder`
- `WrappedWire(PublicWire)` → `InstanceGenerator`
- `WitnessGenWire(WitnessWire)` → `WitnessGenerator`

The bespoke `InOutSegmentBuilder`, `WitnessGeneratorWithAlloc`, and `LazyInOut` are gone (net −645/+294 lines).

## Commits

1. **add allocate-and-write inout/precommit methods to witness/instance generators** — fold the inout/precommit `WireAllocator`s (previously in `WitnessGeneratorWithAlloc`) into `WitnessGenerator`, add a matching inout allocator + `next_inout`/`next_precommit` to `InstanceGenerator`, and add `PublicWire::value`. Pure frontend additions; existing `write_inout`/`write_precommit`/`placeholder_precommit` (the constraint-replay flow) stay.
2. **remove InOut lazy evaluation from Spartan wrapper channels** — the wrapper rewrite. The three channels now run identical arithmetic so the public vector (inout + derived) stays aligned across symbolic/instance/witness; `ZKWrappedVerifierChannel::new` gains `&outer_layout` (mirroring the prover side), threaded through `verifier::zk_config`.
3. **drop dead f_op parameter from CircuitWire::combine** — no impl uses the field-level closure anymore; `CircuitElem` keeps it for the all-constant fast path.

The three channels are coupled by the public-vector alignment invariant (the end-to-end `wrapper_integration_test` spans them), so commit 2 is necessarily atomic.

## Design decisions worth a look

A few choices here are mine, not dictated by the issue — flagging them so you can react in review:

- **`assert_zero` on public wires.** The symbolic builder now records a constraint for *every* wire (public-derivable or private) and the concrete verifier no-ops; the outer verifier enforces it over the reconstructed public segment. This replaces the old eager wrapper-side value check — rejection moves from the wrapper to the outer verifier, and the outer CS grows slightly. The alternative (symbolic no-op + concrete eager check) preserves the old behavior more literally; I chose the uniform path since derived wires make the special-casing unnecessary.
- **`compute_public_value` allocates a single inout wire.** The closure is an arbitrary native computation the CS can't replay, so its result enters as a supplied public input (inout), never a derived wire. This keeps the "one inout instead of a sub-circuit" optimization the HACK exists for; it now allocates eagerly rather than lazily.
- **`CircuitWire` could eventually go away.** With the three impls reduced to identical `builder_op` delegations, `CircuitElem` could be made generic over the `CircuitBuilder` directly and the trait dropped. Out of scope here (the issue framed the fix as newtypes), but noting it as a possible follow-up.

## Testing

`cargo test` for `binius-spartan-frontend`, `binius-spartan-verifier`, `binius-spartan-prover`, and the real ZK roundtrip in `binius-prover` (`prove_verify`, incl. `test_zk_prove_verify_sha256_preimage` and the signature-of-knowledge tests) all pass; clippy + fmt clean on the affected crates.

### DIFF
--- a/crates/spartan-frontend/src/circuit_builder.rs
+++ b/crates/spartan-frontend/src/circuit_builder.rs
@@ -555,6 +555,14 @@ impl<'a, F: Field> InstanceGenerator<'a, F> {
 		PublicWire(Some(value))
 	}
 
+	/// The reconstructed public vector `[constants | inout | derived]` so far, of length
+	/// `1 << layout.log_public()`. Once every inout and alive-derived wire has been written, this
+	/// is the final public vector; reading it by reference (rather than consuming via
+	/// [`Self::build`]) lets callers keep the generator alive for wires that still reference it.
+	pub fn public(&self) -> &[F] {
+		&self.public
+	}
+
 	/// Returns the reconstructed public vector `[constants | inout | derived]`, of length
 	/// `1 << layout.log_public()`, ready to pass to `Verifier::verify`.
 	pub fn build(self) -> Vec<F> {
--- a/crates/spartan-frontend/src/circuit_builder.rs
+++ b/crates/spartan-frontend/src/circuit_builder.rs
@@ -495,6 +495,15 @@ impl<'a, F: Field> CircuitBuilder for WitnessGenerator<'a, F> {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct PublicWire<F: Field>(Option<F>);
 
+impl<F: Field> PublicWire<F> {
+	/// The wire's value if it is public-derivable (constant, inout, or derived), else `None` for a
+	/// private/precommit wire whose value the verifier does not know.
+	#[inline]
+	pub fn value(self) -> Option<F> {
+		self.0
+	}
+}
+
 /// Reconstructs the public input vector `[constants | inout | derived]` on the verifier side by
 /// re-running the circuit function.
 ///
--- a/crates/spartan-prover/src/wrapper/replay_channel.rs
+++ b/crates/spartan-prover/src/wrapper/replay_channel.rs
@@ -4,7 +4,7 @@
 //! [`WitnessGenerator`], filling both inout and private wires in the outer witness.
 
 use std::{
-	cell::{Cell, RefCell},
+	cell::RefCell,
 	marker::PhantomData,
 	rc::{Rc, Weak},
 	vec::IntoIter as VecIntoIter,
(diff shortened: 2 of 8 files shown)


crates/spartan-frontend/src/circuit_builder.rs
```diff
@@ -363,7 +363,9 @@ impl<F: Field> WitnessWire<F> {
 /// the first constraint violation as an error for debugging.
 #[derive(Debug)]
 pub struct WitnessGenerator<'a, F: Field> {
+	inout_alloc: WireAllocator,
```
jimpo: inout_alloc and precommit_alloc should go directly on ReplayChannel, not WitnessGenerator.

crates/spartan-frontend/src/circuit_builder.rs
```diff
@@ -506,6 +534,7 @@ pub struct PublicWire<F: Field>(Option<F>);
 /// aligned so [`WitnessLayout::get`] resolves each to the correct public slot.
 #[derive(Debug)]
 pub struct InstanceGenerator<'a, F: Field> {
+	inout_alloc: WireAllocator,
```
jimpo: This should similarly go on the ZKWrappedVerifierChannel

crates/spartan-frontend/src/circuit_builder.rs
```diff
@@ -546,6 +591,14 @@ impl<'a, F: Field> InstanceGenerator<'a, F> {
 		PublicWire(Some(value))
 	}
 
+	/// The reconstructed public vector `[constants | inout | derived]` so far, of length
+	/// `1 << layout.log_public()`. Once every inout and alive-derived wire has been written, this
+	/// is the final public vector; reading it by reference (rather than consuming via
+	/// [`Self::build`]) lets callers keep the generator alive for wires that still reference it.
+	pub fn public(&self) -> &[F] {
```
jimpo, line 562: This doesn't seem necessary. You can consume the InstanceGenerator in ZKWrappedVerifierChannel::finish

jimpo (changes requested):

crates/spartan-frontend/src/circuit_builder.rs
benji-potamus: Done — moved `inout_alloc`/`precommit_alloc` onto `ReplayChannel`. `WitnessGenerator` keeps only the existing `write_inout`/`write_precommit` (write-by-wire); the channel owns the interaction-order allocation and threads the wire in. Pushed in 8670dc5.

crates/spartan-frontend/src/circuit_builder.rs
benji-potamus: Done — same treatment: `inout_alloc`/`precommit_alloc` now live on `ZKWrappedVerifierChannel`, and `InstanceGenerator` keeps only `write_inout`/`placeholder_precommit`. (The only frontend addition left is `PublicWire::value`, now its own commit.) Pushed in 8670dc5.

crates/spartan-frontend/src/circuit_builder.rs
benji-potamus: I tried consuming it first and it panics with `combine ... after the channel is closed`. The reason: the oracle relations queued onto `inner_channel` carry transparent closures that hold `Weak` refs to the `InstanceGenerator` (the sampled-challenge wires are anchored to it), and those closures are only evaluated later, inside `inner_channel.finish()` — *after* the public vector has already been handed to `outer_verifier.verify`. Consuming the generator up front (`build()`/`try_unwrap`) drops it while those closures still upgrade it.

This matches the pre-existing `finish()` on `main`, which also read the public values by borrow rather than consuming, with the comment: *"Read the materialized public values by borrow rather than consuming the shared builder: the queued oracle relations (whose opening is deferred to `inner_channel.finish`) still hold references to it."* `public(&self)` is just the equivalent borrow now that the segment lives in the `InstanceGenerator`.

So consuming would need the transparent-closure evaluation to stop depending on the generator first (a bigger change to that HACK). Happy to go that route or shape `public()` differently if you'd prefer — but as-is the borrow seems structurally required. Left it as `public()` for now.

jimpo (approved):

PR #1539, opened 2026-06-15
BINIUS-69: foundation — is_zk + ZK-aware FRIParams skip selection

benji-potamus:
Draft for direction review (per the thread) — the **behaviour-preserving foundation** of the non-ZK ↔ ZK BaseFold unification ([BINIUS-69](https://linear.app/jimpo/issue/BINIUS-69)). Fresh start replacing the messy #1535. The soundness-critical MLE-check core is **not** here — it's staged locally pending your (i)/(ii) call on the open `FRIFold` question in the #1535 thread.

## What this does (no behaviour change on any current path)

1. **`is_zk` on the channel `OracleSpec`** (`crates/iop/src/channel.rs`) with `OracleSpec::new` (non-ZK) / `new_zk` (ZK) constructors. Production call sites set it: binius64 witness = `new`, the three spartan oracles = `new_zk`, spartan-prover wiring = `new`.

2. **`is_zk` on the FRI `PartialOracleSpec`** (`crates/iop/src/fri/common.rs`), *replacing* its caller-set `skip_batch_challenges`. The compiler maps each oracle: ZK → `log_msg_len + 1`, `log_batch_size = Some(1)`; non-ZK → `log_msg_len`, `log_batch_size = None` (optimizer-chosen).

3. **ZK-aware FRIParams skip selection** in `choose_batch_size_and_arities_multi`: after resolving batch sizes, it computes `n_inner = max(batch over non-ZK)`, `n_batch_challenges = n_inner + max(batch over ZK ∈ {0,1})`, and sets each oracle's `skip_batch_challenges` — **ZK → 0** (γ at the prefix), **non-ZK → `n_batch_challenges − batch`** (suffix routing) — so the inner challenges stay consistent across a heterogeneous batch.

## Behaviour-preserving + green

- **All-ZK path byte-identical** (every current config is all-ZK or single-non-ZK + all-ZK).
- The raw-FRI batched-skip tests (`fri/tests.rs` 245/388/538, `common.rs` optimizer test) now read the *computed* per-oracle skip rather than a hardcoded prefix; `test_commit_prove_verify_batched_mixed_skip` reframed in `is_zk` terms with **identical routing**.
- `cargo check --workspace --all-targets` ✅ · `cargo test -p binius-iop -p binius-iop-prover` ✅ · `cargo +nightly-2026-01-01 fmt` clean.

## What follows (staged locally, not in this PR)

- Conditional masking (`commit_masked` ZK / `commit_interleaved` non-ZK) + σ/γ only for ZK oracles — **done, green** on the all-ZK path.
- The MLE-check core (general Phase B combined buffer + the γ/outer/inner/standard fold), pending your **(i) vs (ii)** decision on whether the non-ZK batch folds in the FRI first fold (skip windows, what this foundation supports) or as standard MLE-check rounds (needs `FRIFold` surgery).

### DIFF
--- a/crates/iop/src/fri/common.rs
+++ b/crates/iop/src/fri/common.rs
@@ -8,7 +8,7 @@ use binius_math::{ntt::DomainContext, reed_solomon::ReedSolomonCode};
 use binius_utils::checked_arithmetics::log2_ceil_usize;
 use getset::{CopyGetters, Getters};
 
-use crate::merkle_tree::MerkleTreeScheme;
+use crate::{channel::OracleSpec, merkle_tree::MerkleTreeScheme};
 
 /// Parameters for an FRI interleaved code proximity protocol.
 ///
@@ -26,7 +26,7 @@ pub struct FRIParams<F> {
 	#[getset(get = "pub")]
 	rs_code: ReedSolomonCode<F>,
 	/// Guaranteed to be non-empty.
-	input_oracles: Vec<OracleSpec>,
+	input_oracles: Vec<CodewordSpec>,
 	/// log2 the maximum message length of all input oracles, after lifting each to the reduced
 	/// dimension. Equals `rs_code.log_dim() + max(skip_batch_challenges + log_batch_size)` over
 	/// the input oracles (the inner challenges the first fold must draw).
@@ -42,10 +42,23 @@ pub struct FRIParams<F> {
 	n_test_queries: usize,
 }
 
+/// Specification of one committed codeword batched into the first-round FRI oracle.
+///
+/// Each input oracle commits an interleaved Reed–Solomon codeword whose own dimension is
+/// `rs_code.log_dim() - log_lift`; it is lifted to the shared reduced dimension by duplicating
+/// each entry `2^log_lift` times before the oracles are batched together. This is distinct from
+/// [`crate::channel::OracleSpec`], the higher-level description of an oracle to be committed (its
+/// message length and whether it is masked); a `CodewordSpec` is the resolved, FRI-level layout
+/// the [`FRIParams`] selection computes from a batch of those.
 #[derive(Debug, Clone)]
-pub struct OracleSpec {
-	/// log2 the message length, i.e. the Reed–Solomon code dimension plus the batch size.
-	pub log_msg_len: usize,
+pub struct CodewordSpec {
+	/// log2 the number of times each committed codeword entry is duplicated to lift it to the
+	/// shared first-round (reduced) dimension.
+	///
+	/// The committed codeword's Reed–Solomon dimension is `rs_code.log_dim() - log_lift`, so its
+	/// message length is `rs_code.log_dim() - log_lift + log_batch_size`. It is `0` when the
+	/// oracle already sits at the reduced dimension.
+	pub log_lift: usize,
 	/// log2 the interleaved batch size.
 	pub log_batch_size: usize,
 	/// The number of leading inner challenges this oracle skips before its batch fold.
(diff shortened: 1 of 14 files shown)


crates/iop/src/fri/common.rs
```diff
@@ -73,10 +80,14 @@ pub struct PartialOracleSpec {
 	///
 	/// This field is Some if the log_batch_size is fixed, and None if it's flexible.
 	pub log_batch_size: Option<usize>,
```
jimpo: Remove `log_batch_size` from PartialOracleSpec. We know that it's Some(1) if is_zk and None otherwise.

crates/iop/src/fri/common.rs
```diff
@@ -65,6 +65,13 @@ pub struct OracleSpec {
...
 /// are instantiated.
+///
+/// Unlike [`OracleSpec`], this does not carry `skip_batch_challenges`: that value depends on the
+/// batch sizes chosen across *all* oracles, so it is computed by the ZK-aware [`FRIParams`]
+/// selection (see [`FRIParams::optimal_for_batch`]). Instead the caller declares, via `is_zk`,
+/// whether the oracle is a masked ZK BaseFold oracle (fixed `log_batch_size = 1`, folded first by
+/// the shared masking challenge γ) or an ordinary oracle whose batch fold is routed to a suffix of
+/// the inner challenges.
 #[derive(Debug, Clone)]
 pub struct PartialOracleSpec {
```
jimpo: Instead of this struct, use `binius_iop::channel::OracleSpec`, which has a log_msg_len and an is_zk.

crates/iop/src/fri/common.rs
jimpo: Rename this to CodewordSpec, to distinguish further from `binius_iop::channel::OracleSpec`.

jimpo: Instead of storing the `log_msg_len`, store `log_lift: usize`. Then the FRIParams, or whatever owns it, can determine what it needs to from the log_dim of the RS code, the log_lift, and batch size.

jimpo (changes requested):

crates/iop/src/fri/common.rs
benji-potamus: Done — `PartialOracleSpec` is gone. `FRIParams::optimal_for_batch` now takes `&[binius_iop::channel::OracleSpec]` directly and derives each oracle's committed message length (`+1` for a ZK oracle's interleaved mask) and batch size internally. Both BaseFold compilers pass their `channel::OracleSpec`s straight through with no intermediate conversion.

crates/iop/src/fri/common.rs
benji-potamus: Done — removed. The batch size is no longer carried on the input at all: the selection (`choose_batch_size_and_arities_multi`) derives `Some(1)` for ZK oracles and a flexible (`None`) batch for non-ZK ones from the `is_zk` flag.

crates/iop/src/fri/common.rs
benji-potamus: Done — the resolved FRI-level spec is now `CodewordSpec`, distinct from `channel::OracleSpec`.

benji-potamus: Done — `CodewordSpec` stores `log_lift` instead of `log_msg_len`. `FRIParams` owns the reduced `rs_code().log_dim()`, and each consumer recovers an oracle's own dimension as `rs_code().log_dim() - log_lift` (see `fri/verify.rs`, `fri/fold.rs`, `fri/commit.rs`), so the spec only carries the lift, batch size, and skip.

benji-potamus:
Refactored the oracle-spec model per review, and rebased onto current `main` (`6aa2d6d0`). Head is now `e0cabd16`.

**What changed**
- Deleted `PartialOracleSpec`. `FRIParams::optimal_for_batch` now takes `&[binius_iop::channel::OracleSpec]` (`{ log_msg_len, is_zk }`) and derives each oracle's committed length (`+1` for a ZK mask) and batch size (`Some(1)` for ZK, flexible `None` for non-ZK) internally. Both BaseFold compilers pass their `channel::OracleSpec`s straight through.
- Renamed the resolved FRI spec `OracleSpec` → `CodewordSpec`, and replaced its `log_msg_len` field with `log_lift`. `FRIParams` owns the reduced `rs_code().log_dim()`; consumers recover an oracle's own dimension as `log_dim - log_lift` (`fri/verify.rs`, `fri/fold.rs`, `fri/commit.rs`).
- `new_batch(rs_code, Vec<CodewordSpec>, fold_arities, n_test_queries)` is now the fundamental multi-oracle constructor; `new` and `optimal_for_batch` both delegate to it. The FRI-layer fold tests (which exercise explicit batch sizes and lifting) construct `CodewordSpec`s directly through it.

All four threads addressed and resolved. `cargo test -p binius-iop -p binius-iop-prover` green (32 tests), clippy `-D warnings` clean, fmt clean — verified on the foundation commit in isolation. Still a draft, foundation only; the heterogeneous Phase A/B core stays staged for the follow-up.

jimpo (approved):

PR #1537, opened 2026-06-15
Add GhashSq256b field (BINIUS-56)

benji-potamus:
Closes [BINIUS-56](https://linear.app/jimpo/issue/BINIUS-56/create-ghashsq-field).

Adds `GhashSq256b`, a `GF(2^256)` binary field implemented as a degree-two extension of `BinaryField128bGhash`, defined by the irreducible polynomial `Y² + Y + X⁻¹` over GHASH. It is backed by `M256` (low 128 bits = coefficient of `1`, high 128 bits = coefficient of `Y`) — the same lane layout as `PackedBinaryGhash2x128b` — matching the `{1, Y}` basis.

It implements `Field`, `BinaryField`, `TowerField`, and `ExtensionField` over both `BinaryField128bGhash` and `BinaryField1b`.

### Commits
1. **Make M256 usable as a binary field underlier** — the trait impls `M256` needs to back a scalar `binary_field!` field. On x86_64 native `M256`: `Hash`, `LowerHex`, `From<M128>`, `NumCast<M256> for M128`, and a const `from_u128s`. For `ScaledUnderlier<M128, 2>` (the `M256` representation off the AVX2 path): `From<M128>`/`From<U1>` low-limb embeddings, an `M128` `NumCast` extraction, a `Hash` derive, and a `LowerHex` generalized over the limb type.
2. **Add GhashSq256b field** — the field, its arithmetic, extension impls, and tests.

### Multiplication
Uses the `mul_m256i_hybrid` algorithm from `binius_arith_bench::ghash_sq`: the two GHASH products that share the 256-bit CLMUL (`t₀ = x₀·y₀`, `t₂ = x₁·y₁`) are batched into a single `PackedBinaryGhash2x128b` multiply — one `vpclmulqdq` over both lanes on AVX2 — while the third Karatsuba product `t₁` is a scalar GHASH multiply. Expressing it through the packed field keeps the code portable (and SIMD-accelerated on every target) with no `unsafe` or arch cfgs. Inversion is norm-based: `N = a² + ab + b²·X⁻¹ ∈ GHASH`, `u⁻¹ = ((a+b) + b·Y)·N⁻¹`.

### Notes for review
[...]

PR #1540, opened 2026-06-15
Change BinaryField128bGhash underlier to M128 (BINIUS-101)

benji-potamus:
Closes [BINIUS-101](https://linear.app/jimpo/issue/BINIUS-101/change-binaryfield128bghash-underlier-to-m128).

Changes `BinaryField128bGhash`'s underlier from `u128` to `M128` (the architecture-chosen 128-bit underlier — a SIMD register on x86_64/aarch64, `u128` elsewhere).

Per the discussion on the issue, this keeps construction ergonomic (option C): `BinaryField128bGhash::new` still takes `u128`, so no caller migration is needed — the only downstream changes are 3 sites that did `u128` arithmetic on `.val()`.

### Commits
1. **Add Hash and LowerHex impls to M128** — the two trait impls a SIMD type needs to back a `binary_field!` field (x86_64 + aarch64 structs; on wasm/portable `M128 = u128` already has them).
2. **Change BinaryField128bGhash underlier to M128** — everything else.

### What the change entails
- **`new` moved out of `binary_field!`**: the macro no longer generates `new` (internal uses switched to the tuple constructor). Each field now defines its own `new` — `BinaryField1b::new(U1)`, `AESTowerField8b::new(u8)`, and `BinaryField128bGhash::new(u128)` (converts to `M128` via a new `const arch::m128_from_u128`, since `From<u128>` isn't `const`).
- **`u128` conversions**: cfg-gated `From<u128>` / `From<BinaryField128bGhash> for u128` on x86_64/aarch64 (on wasm/portable the macro's `From<M128>` already is the `u128` conversion).
- **`Divisible`**: reflexive `Divisible<M128>` (needed for the width-1 packed field) and a reinterpreting `Divisible<M128> for u128` (symmetric to the existing `M128: Divisible<u128>`; needed because the portable `PackedBinaryGhash1x128b` packs the `M128`-underlier scalar in a `u128`).
- **Portable nibble inversion**: `nibble_invert_128b`'s bound relaxed from `WithUnderlier<Underlier = u128>` to `WithUnderlier` + `Underlier: From<u128> + Into<u128>`, converting at the boundary.
- **`ghash.rs`**: `mul_x`/`mul_inv_x` compute over `u128`; the 256-entry AES lookup table stays `[u128; 256]` (wrapped at the lookup site); `min_tower_level` uses `==` instead of a pattern match (`M128` isn't structurally matchable).
- **Downstream (3 sites)**: `binius-math::trace_one_element` is unaffected (`new(u128)`); `field_buffer` test and `prover::ring_switch` (×2) convert `.val()` (now `M128`) to `u128` for byte/integer ops.

### Verification
- Full workspace `cargo clippy --all --tests --benches --examples -- -D warnings`: green.
- `cargo test -p binius-field -p binius-math`: pass. binius-field: 213 lib tests pass.
- Cross-arch: aarch64 `cargo check` (with `+neon,+aes`) and wasm32 build both clean.

### Notes for review
- The reinterpreting `Divisible<M128> for u128` is the one non-obvious addition; it mirrors the existing reverse impl and is only reachable on x86_64/aarch64 (on wasm/portable `M128 = u128`, so the reflexive impl covers it).
- Touches the same `BinaryField128bGhash` foundation as BINIUS-56 (`GhashSq256b`, PR #1537). Whichever merges second needs a trivial rebase fixup in `ghash_sq.rs` (it constructs via the tuple/`from_coeffs`, and `new` is no longer macro-generated).

### DIFF
--- a/crates/field/src/binary_field.rs
+++ b/crates/field/src/binary_field.rs
@@ -82,11 +82,10 @@ macro_rules! binary_field {
 		#[repr(transparent)]
 		$vis struct $name(pub(crate) $typ);
 
+		// NOTE: `new` is intentionally NOT generated here. Each field defines its own `new` so it
+		// can take an ergonomic constructor type independent of the underlier (e.g.
+		// `BinaryField128bGhash::new` takes `u128` even though its underlier is `M128`).
 		impl $name {
-			pub const fn new(value: $typ) -> Self {
-				Self(value)
-			}
-
 			pub const fn val(self) -> $typ {
 				self.0
 			}
@@ -226,8 +225,8 @@ macro_rules! binary_field {
 		}
 
 		impl Field for $name {
-			const ZERO: Self = $name::new(<$typ as $crate::underlier::UnderlierType>::ZERO);
-			const ONE: Self = $name::new(<$typ as $crate::underlier::UnderlierType>::ONE);
+			const ZERO: Self = $name(<$typ as $crate::underlier::UnderlierType>::ZERO);
+			const ONE: Self = $name(<$typ as $crate::underlier::UnderlierType>::ONE);
 			const CHARACTERISTIC: usize = 2;
 			const ORDER_EXPONENT: usize = <$typ as $crate::underlier::UnderlierType>::BITS;
 			const MULTIPLICATIVE_GENERATOR: $name = $name($gen);
@@ -367,7 +366,7 @@ macro_rules! impl_field_extension {
 				if elem.0 >> $subfield_name::N_BITS
 					== <$typ as $crate::underlier::UnderlierType>::ZERO
 				{
-					Ok($subfield_name::new(<$subfield_typ>::num_cast_from(elem.val())))
+					Ok($subfield_name(<$subfield_typ>::num_cast_from(elem.val())))
 				} else {
 					Err(())
 				}
@@ -377,7 +376,7 @@ macro_rules! impl_field_extension {
 		impl From<$subfield_name> for $name {
 			#[inline]
 			fn from(elem: $subfield_name) -> Self {
-				$name::new(<$typ>::from(elem.val()))
+				$name(<$typ>::from(elem.val()))
 			}
 		}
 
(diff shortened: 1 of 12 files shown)


crates/field/src/binary_field.rs
```diff
@@ -713,7 +716,7 @@ pub(crate) mod tests {
 
 		#[test]
 		fn test_inverse_128b(val in 1u128..) {
-			let x = BinaryField128bGhash::new(val);
```
jimpo, line 716: this seems like an unnecessary change

jimpo (commented):
This looks good overall, but the config-gating on M128 conversions and such is a bit annoying. Maybe that would all be simpler if M128 in the portable build was actually a wrapper struct around u128 (and it would have to implement UnderlierType and all that), instead of being a direct type alias to u128.

benji-potamus:
Agreed — making the portable `M128` a `#[repr(transparent)] struct M128(u128)` (impl'ing `UnderlierType` and friends) would make `M128` uniformly distinct from `u128`, so all the `#[cfg(any(x86_64, aarch64))]` gating (the `From<u128>` conversions, `m128_from_u128`, the reinterpreting `Divisible<M128> for u128`) just drops out. It's contained to `binius-field` — `M128` is the underlier, not a field, so there's no downstream churn; only 2 portable files reference it directly plus the gates here.

The cost is writing the full `UnderlierType`/bit-ops/`Divisible`/`From`/`Hash`/`LowerHex`/`Pod` surface for the portable newtype (mostly one-line delegations to `u128`, ~200 lines, mirroring the x86 `M128`).

Happy to do it — want it folded into this PR, or merge this as-is and land the portable-`M128`-newtype as a follow-up cleanup (it's independently useful)? Leaning toward just doing it here unless you'd rather keep this PR scoped to the GHASH underlier change.

jimpo (approved):

benji-potamus:
Sounds good — follow-up PR it is. I'll open the portable-`M128`-newtype cleanup (which removes the cfg-gating added here) once this lands. Tracking it as a separate issue. This PR is otherwise ready (CI green).

PR #1542, opened 2026-06-17
Make portable M128 a wrapper struct over u128 (BINIUS-104)

benji-potamus:
Closes [BINIUS-104](https://linear.app/jimpo/issue/BINIUS-104/make-portable-m128-a-wrapper-struct-over-u128). Follow-up to BINIUS-101 (#1540).

Replaces the portable `pub type M128 = u128` alias with a `#[repr(transparent)] struct M128(u128)`, so `M128` is a distinct type from `u128` on **every** target (mirroring the SIMD `M128` structs on x86_64/aarch64/wasm32). New `crates/field/src/arch/portable/m128.rs` carries the full underlier surface — mostly derives + `impl_divisible_memcast!`/`bitmask!`/`self!`, plus one-line delegations to `u128`.

`M128` is now the underlier for **all** portable packed fields (`packed_128`, `packed_aes_128`, `packed_ghash_128`); the scaled 256/512 fields remain `ScaledUnderlier<M128, N>`. There are no `u128`-backed packed fields.

With `M128` uniformly distinct from `u128`:
- the `#[cfg(any(x86_64, aarch64))]` gates on the `u128` ↔ `BinaryField128bGhash` conversions in `ghash.rs` are dropped (they no longer collide with the `binary_field!`-generated `From` impls);
- `m128_from_u128` collapses to a single non-cfg `const fn`;
- the now-dead reinterpreting `Divisible<M128> for u128` is removed from `x86_64/m128.rs` + `aarch64/m128.rs`;
- the latent `pub use u128 as M128` (a dormant E0252) in the portable-fallback arch branch is removed.

### GHASH software kernels as free functions
Portable `PackedBinaryGhash1x128b` is `M128`-backed. The catch: where a SIMD `M128` exists, `BinaryField128bGhash`'s underlier *is* that SIMD type, so a width-1 `PackedPrimitiveType` over the distinct portable struct `M128` can't `get`/`set_single` — and `mod portable` is dead-compiled there. So the GHASH software kernels are free functions — `ghash_mul`/`ghash_square` and a new `ghash_invert_or_zero(scalar)` — which the x86_64/aarch64 `packed_ghash_128`s call directly for their software fallbacks (no-CLMUL mul/square, and invert which has no CLMUL path), instead of constructing a portable packed type. The portable `M128` `PackedBinaryGhash1x128b` then lives in a small `portable_backend` submodule gated to the no-SIMD-`M128` arches, mirroring how `{x86_64,aarch64}/mod.rs` gate their own `packed_ghash_128`.

### Verification
- x86_64: `cargo test -p binius-field` (213 lib tests) + `cargo clippy -p binius-field --tests --benches -- -D warnings`: green.
- wasm32-unknown-unknown: `cargo build -p binius-field`: green.
- wasm32-wasip1: full `wasm-tests` set (`cargo test -p binius-math -p binius-field -p binius-prover -p binius-verifier` via wasmtime): green — exercises the live portable `M128` GHASH/scaled fields.
- aarch64: `cargo clippy`/`build` with `+neon,+aes` and `build` with `-aes` (the no-SIMD path): green.

The portable-fallback `else` arch branch has no CI target, so it is reasoned-correct rather than build-verified.

### DIFF
--- a/crates/field/src/arch/mod.rs
+++ b/crates/field/src/arch/mod.rs
@@ -27,28 +27,11 @@ cfg_if! {
 
 		mod wasm32;
 		pub use wasm32::{packed_ghash_128, packed_ghash_256};
-		pub use portable::{packed_128::{self, M128}, packed_256::{self, M256}, packed_512, packed_aes_128, packed_aes_256, packed_aes_512, packed_ghash_512};
+		pub use portable::{m128::M128, packed_128, packed_256::{self, M256}, packed_512, packed_aes_128, packed_aes_256, packed_aes_512, packed_ghash_512};
 	} else {
 		mod portable;
-		pub use u128 as M128;
-		pub use portable::{packed_128::{self, M128}, packed_256::{self, M256}, packed_512, packed_aes_128, packed_aes_256, packed_aes_512, packed_ghash_128, packed_ghash_256, packed_ghash_512};
+		pub use portable::{m128::M128, packed_128, packed_256::{self, M256}, packed_512, packed_aes_128, packed_aes_256, packed_aes_512, packed_ghash_128, packed_ghash_256, packed_ghash_512};
 	}
-}
-
-/// Builds an [`M128`] from a `u128` in a `const` context.
-///
-/// `M128` is the architecture-chosen 128-bit underlier — a SIMD register on x86_64/aarch64, plain
-/// `u128` elsewhere. `From<u128>` is not `const`, so this helper is used where a const `M128`
-/// constant is needed (e.g. a field's multiplicative generator).
-#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-pub(crate) const fn m128_from_u128(value: u128) -> M128 {
-	M128::from_u128(value)
-}
-
-/// Builds an [`M128`] from a `u128` in a `const` context. See the SIMD variant for details.
-#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
-pub(crate) const fn m128_from_u128(value: u128) -> M128 {
-	value
 }
 
 pub use arch_optimal::*;
--- a/crates/field/src/arch/portable/packed_128.rs
+++ b/crates/field/src/arch/portable/packed_128.rs
@@ -1,11 +1,11 @@
 // Copyright 2024-2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
+use super::m128::M128;
 use crate::arch::{
 	BitwiseAndStrategy,
 	portable::packed_macros::{portable_macros::*, *},
 };
-
-pub type M128 = u128;
 
 define_packed_binary_fields!(
 	underlier: M128,
--- a/crates/field/src/arch/portable/packed_aes_128.rs
+++ b/crates/field/src/arch/portable/packed_aes_128.rs
@@ -1,5 +1,7 @@
 // Copyright 2024-2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
+use super::m128::M128;
 use crate::{
 	arch::{
 		PairwiseTableStrategy,
@@ -9,7 +11,7 @@ use crate::{
 };
 
 define_packed_binary_fields!(
-	underlier: u128,
+	underlier: M128,
 	packed_fields: [
 		packed_field {
 			name: PackedAESBinaryField16x8b,
--- a/crates/field/src/arch/portable/packed_ghash_128.rs
+++ b/crates/field/src/arch/portable/packed_ghash_128.rs
@@ -4,20 +4,11 @@
 //! Portable implementation of packed GHASH field operations.
 
 use super::{
+	m128::M128,
 	nibble_invert_128b::nibble_invert_128b,
-	packed_macros::{portable_macros::*, *},
-	univariate_mul_utils_128::{Underlier64bLanes, Underlier128bLanes, bmul64},
+	univariate_mul_utils_128::{Underlier64bLanes, Underlier128bLanes, bmul64, spread_bits_64},
 };
-use crate::{
-	Divisible,
-	arithmetic_traits::{
-		TaggedInvertOrZero, TaggedMul, TaggedSquare, impl_invert_with, impl_mul_with,
-		impl_square_with,
-	},
-	ghash::BinaryField128bGhash,
-	packed::PackedField,
-	underlier::WithUnderlier,
-};
+use crate::{ghash::BinaryField128bGhash, underlier::WithUnderlier};
 
 /// Strategy for GHASH field arithmetic operations.
 pub struct GhashStrategy;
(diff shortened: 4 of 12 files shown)


benji-potamus:
Fixed a `wasm-tests` CI failure (the only red check): on the non-SIMD arches `OptimalPackedB128` is the `u128`-backed `PackedBinaryGhash1x128b`, so `OptimalPackedB1 = PackedSubfield<OptimalPackedB128, BinaryField1b>` resolves to a `u128`-backed 1b packed field — which stopped existing once `PackedBinaryField128x1b` became `M128`-backed. Added the `u128`-backed 1b base back (cfg-gated to the non-SIMD arches), alongside the `M128`-backed canonical one.

Re-verified locally: x86_64 test+clippy, aarch64 check, wasm32-unknown-unknown build, and the full `wasm-tests` set (`binius-math`/`field`/`prover`/`verifier` on wasm32-wasip1 via wasmtime, incl. `prove_verify`) all green.

benji-potamus:
Fixed the `rust-checks-arm64` failure (was the only red check). It was the `Run doctests` step: rustdoc compiles doctests with **baseline** target features (it doesn't inherit `RUSTFLAGS=-Ctarget-cpu=native`), so on arm64 it builds without `aes` and takes `aarch64/mod.rs`'s `else` branch, where `arch::M128` came from `pub use super::portable::packed_128::M128` — which I'd made private when moving `M128` into `portable/m128.rs`. Two fixes:
- `aarch64/mod.rs` (no-SIMD branch) now re-exports `M128` from `portable::m128` (its new home).
- The cfg gates on the portable struct-`M128` GHASH/1b bases were too coarse (`not(any(x86_64, aarch64))`); tightened to `not(any(all(x86_64, sse2), all(aarch64, neon, aes)))` so the bases are present on no-SIMD arm64 (e.g. the baseline rustdoc build) but excluded where a SIMD `M128` already provides them. The gate now mirrors the SIMD-`M128` cfgs in `{x86_64,aarch64}/mod.rs`.

Verified locally: x86_64 test(213)+clippy, wasm32-unknown-unknown build, full `wasm-tests` (4 crates via wasmtime), aarch64 build in both `+neon,+aes` and `-aes` configs, and the binius-field doctests compile under baseline aarch64 (`-aes`).

crates/field/src/arch/portable/packed_128.rs
```diff
@@ -20,3 +20,36 @@ define_packed_binary_fields!(
...
+// struct, distinct from `u128`).
+#[cfg(not(any(
+	all(target_arch = "x86_64", target_feature = "sse2"),
+	all(
+		target_arch = "aarch64",
+		target_feature = "neon",
+		target_feature = "aes"
+	)
+)))]
+mod u128_base {
```
jimpo: Why do we need this? Is it actually any faster than the one defined above? It should be exactly the same.

crates/field/src/arch/portable/packed_ghash_128.rs
```diff
@@ -141,6 +141,90 @@ impl TaggedInvertOrZero<GhashStrategy> for PackedBinaryGhash1x128b {
...
+// back to the portable `u128`-newtype struct as `arch::M128`. There `BinaryField128bGhash`'s
+// underlier is that `M128`, so the scaled fields `PackedBinaryGhash2x128b`/`4x128b`
+// (`ScaledUnderlier<M128, N>`) need a `PackedPrimitiveType<M128, _>` base. Where a SIMD `M128`
+// exists (x86_64+sse2, aarch64+neon+aes) the arch's own `packed_ghash_128` provides that base and
+// `mod portable` is dead-compiled with `M128` a distinct SIMD type, so these impls would not
+// type-check — hence the gate mirrors the SIMD-`M128` cfgs in `{x86_64,aarch64}/mod.rs`.
+// `Mul`/`Square` reuse the `u128` software kernels via `M128: Underlier128bLanes` (a no-op
+// delegation); `InvertOrZero` defers to the `u128` `PackedBinaryGhash1x128b` above. The standalone
+// `PackedBinaryGhash1x128b` itself stays `u128`-backed — the SIMD arches invert via it.
+#[cfg(not(any(
```
jimpo, line 136: Can't we drop the cfg gate?

jimpo (changes requested):

crates/field/src/arch/portable/packed_128.rs
benji-potamus: You're right that it was an exact duplicate — restructured to remove it. The optimal/`PackedExtension` tower on these targets is u128-backed (because the optimal ghash `PackedBinaryGhash1x128b` is — see the other thread), so `OptimalPackedB1 = PackedSubfield<OptimalPackedB128, BinaryField1b>` resolves to `PackedPrimitiveType<u128, BinaryField1b>`. I've made the canonical `PackedBinaryField128x1b` (and `PackedAESBinaryField16x8b`) u128-backed so they *are* those optimal-tower types — no separate duplicate. The `M128`-underlier field that the scaled `PackedBinaryField256x1b`/`512x1b` build on is now a clearly-labelled `m128_base` sub-module, parallel to the GHASH one. (Same swap was needed for AES — `OblongZerocheckProver` instantiates an optimal AES packed field whose subfield tower must stay u128-backed.)

crates/field/src/arch/portable/packed_ghash_128.rs
benji-potamus: Dropped it — you're right, it compiles unconditionally. These are bare `Mul`/`Square`/`InvertOrZero` trait impls (not the `define_packed_binary_field!` macro), so they don't pull in the `get`/`set_single` machinery that would require `M128: Divisible<BinaryField128bGhash::Underlier>` on the dead-compiled SIMD path — verified building x86_64 (`-Ctarget-cpu=native`) and aarch64 (`+neon,+aes`). (The macro-based standalone `PackedBinaryGhash1x128b` does need that and so stays u128-backed — that's why this base delegates its invert to it rather than being the standalone type itself.)

crates/field/src/arch/portable/packed_128.rs
```diff
@@ -1,14 +1,19 @@
...
-pub type M128 = u128;
-
+// `PackedBinaryField128x1b` is packed over `u128`, the same underlier as the optimal 128-bit ghash
+// field `PackedBinaryGhash1x128b`, so `OptimalPackedB1 = PackedSubfield<OptimalPackedB128,
+// BinaryField1b>` resolves to exactly this type. (On the SIMD arches the arch's own `M128`-backed
+// `packed_128` shadows this; here `arch::M128` is the `u128`-newtype struct.) The `M128`-underlier
+// 1b base the scaled `PackedBinaryField256x1b`/`512x1b` fields build on is `m128_base` below.
 define_packed_binary_fields!(
-	underlier: M128,
+	underlier: u128,
```
jimpo: Change this to M128 and remove the m128_base module below

crates/field/src/arch/portable/packed_aes_128.rs
```diff
@@ -8,6 +10,10 @@ use crate::{
 	arithmetic_traits::{impl_invert_with, impl_mul_with, impl_square_with},
 };
 
+// Packed over `u128` (the optimal-tower underlier on these portable targets), so its
+// `PackedExtension`/`PackedSubfield` tower stays `u128`-backed and consistent with the optimal
+// GHASH/1b fields. The `M128`-underlier base the scaled `PackedAESBinaryField32x8b`/`64x8b` fields
+// build on is `m128_base` below.
 define_packed_binary_fields!(
 	underlier: u128,
```
jimpo: Use M128 and remove the `m128_base` module below

crates/field/src/arch/portable/packed_ghash_128.rs
jimpo: Change to M128

benji-potamus:
Reworked to use `M128` for **all** portable underliers — no `u128`-backed packed fields, no `m128_base`/`u128_base` duplicates (supersedes my earlier thread replies). `packed_128`, `packed_aes_128`, `packed_ghash_128` are all `M128`-backed; the scaled 256/512 fields are `ScaledUnderlier<M128, N>` as before. `Divisible<M128> for u128` is gone (now dead).

One wrinkle worth flagging: the GHASH software invert/mul/square had been reached by *constructing the portable packed type* from the SIMD arches' fallback paths, which is what had forced `u128`. I made those kernels free functions (`ghash_mul`/`ghash_square` + new `ghash_invert_or_zero`) so x86_64/aarch64 call them directly. The `M128`-backed `PackedBinaryGhash1x128b` then only needs to exist where portable is the GHASH backend (no SIMD `M128`) — gated the same way `{x86_64,aarch64}/mod.rs` gate their own `packed_ghash_128`, because where a SIMD `M128` exists `BinaryField128bGhash`'s underlier *is* that SIMD type and a width-1 `PackedPrimitiveType` over the distinct portable struct couldn't `get`/`set_single`.

Verified green: x86_64 test(213)+clippy, wasm32-unknown-unknown build, wasm32-wasip1 full `wasm-tests` (math/field/prover/verifier via wasmtime), aarch64 build+clippy with `+neon,+aes` and build with `-aes`.

crates/field/src/arch/portable/packed_128.rs
benji-potamus: Done in a67bfa16 — `packed_128` is `M128`-backed and the `m128_base` module is gone (it's the only 1b-128 field now). `OptimalPackedB1` resolves straight to it.

crates/field/src/arch/portable/packed_aes_128.rs
benji-potamus: Done in a67bfa16 — `packed_aes_128` is `M128`-backed, `m128_base` removed.

crates/field/src/arch/portable/packed_ghash_128.rs
benji-potamus: Done in a67bfa16 — `PackedBinaryGhash1x128b` is `M128`-backed. It lives in a `portable_backend` submodule cfg-gated to the no-SIMD-`M128` arches (mirroring `{x86_64,aarch64}/mod.rs`), since where a SIMD `M128` exists the scalar's underlier *is* that SIMD type and a width-1 `PackedPrimitiveType` over the distinct portable struct can't `get`/`set_single`. The GHASH software kernels are now free functions (`ghash_mul`/`ghash_square`/`ghash_invert_or_zero`) that the SIMD arches' fallbacks call directly, so no portable packed type is referenced from there. See the summary comment for details.

crates/field/src/arch/mod.rs
```diff
@@ -27,30 +27,22 @@ cfg_if! {
...
 /// Builds an [`M128`] from a `u128` in a `const` context.
 ///
-/// `M128` is the architecture-chosen 128-bit underlier — a SIMD register on x86_64/aarch64, plain
-/// `u128` elsewhere. `From<u128>` is not `const`, so this helper is used where a const `M128`
-/// constant is needed (e.g. a field's multiplicative generator).
-#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+/// `M128` is the architecture-chosen 128-bit underlier — a SIMD register on x86_64/aarch64, a
+/// `u128` newtype elsewhere. `From<u128>` is not `const`, so this helper is used where a const
+/// `M128` constant is needed (e.g. a field's multiplicative generator).
 pub(crate) const fn m128_from_u128(value: u128) -> M128 {
```
jimpo: Can we remove this and just use `M128::from_u128` directly now?

benji-potamus: Done in 2abc2cd7 — removed `m128_from_u128`; `ghash.rs` (the only user, in the `binary_field!` generator and `new`) now calls `M128::from_u128` directly.

jimpo (approved):

PR #1546, opened 2026-06-20
BINIUS-108: remove UnderlierType::get_subvalue and set_subvalue

benji-potamus:
Closes BINIUS-108.

## Summary

Removes `UnderlierType::get_subvalue` and `UnderlierType::set_subvalue`, which were thin wrappers around `Divisible::<T>::get` / `Divisible::<T>::set` with an extra `debug_assert!` on the index. Call sites are updated to call `Divisible` directly.

**Replacement strategy per call site:**

- **`underlier_with_bit_ops.rs` (production)**: Both call sites are inside `unsafe` functions where the index bounds have been established by the function's own preconditions — replaced with `Divisible::<T>::get_unchecked` / `set_unchecked` and documented Safety comments.
- **`binary_field.rs` (`get_base_unchecked` macro)**: Already an `unsafe fn` whose contract guarantees `i < Self::N` — replaced with `Divisible::get_unchecked`.
- **`arch/portable/packed.rs` (Display impl)**: Index ranges over `0..width` (the count of packed scalars) — replaced with `Divisible::get_unchecked`.
- **`arch/x86_64/packed_ghash_256.rs` / `packed_ghash_512.rs`**: Literal indices (0, 1, 2, 3) all within bounds — replaced with `Divisible::get_unchecked` / `set_unchecked`.

**Tests**: Renamed `test_get_subvalue` → `test_divisible_get_u32` and `test_set_subvalue_*` → `test_divisible_set_*`. Updated to use `Divisible::get` / `Divisible::set` (safe versions) since the test indices are always in range. The unnecessary `unsafe` blocks around the test calls are also removed.

---

### DIFF
--- a/crates/field/src/arch/portable/packed.rs
+++ b/crates/field/src/arch/portable/packed.rs
@@ -138,7 +138,11 @@ impl<U: UnderlierType + Divisible<Scalar::Underlier>, Scalar: BinaryField> Debug
 		let width = checked_int_div(U::BITS, Scalar::N_BITS);
 		let values_str = (0..width)
 			// Safety: `i` ranges over `0..width`, the number of scalars packed in `U`.
-			.map(|i| Scalar::from_underlier(unsafe { self.0.get_subvalue(i) }))
+			.map(|i| {
+				Scalar::from_underlier(unsafe {
+					Divisible::<Scalar::Underlier>::get_unchecked(&self.0, i)
+				})
+			})
 			.map(|value| format!("{value}"))
 			.collect::<Vec<_>>()
 			.join(",");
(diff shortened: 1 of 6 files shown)


jimpo (changes requested):
You use the fully qualified path for Divisible instead of `use` in many places. Import Divisible, unless in a macro definition (where the fully qualified path makes it more robust).

benji-potamus:
Fixed — added `use` imports in all three non-macro files:
- `packed.rs`: `Divisible` was already in the top-level `use crate::{..., Divisible, ...}` — just removed the `crate::underlier::` prefix from the call site.
- `packed_ghash_256.rs`: added `Divisible` to the `underlier::{...}` import, dropped prefix from all call sites.
- `packed_ghash_512.rs`: `Divisible` was already imported — dropped prefix from all call sites.

The fully qualified path in `binary_field.rs` is inside a macro definition and is left as-is.

jimpo (approved):

benji-potamus:
CI failed on x86_64: `unused import: 'UnderlierType'` in `packed_ghash_256.rs`. When I added `Divisible` to the import I kept the existing `UnderlierType`, but my changes removed its only use (the `get_subvalue` calls, which required `UnderlierType` in scope). Fixed by dropping it. New head: `0f0e81a7f5f1`.

PR #1547, opened 2026-06-20
BINIUS-107: add InvertOrZero::invert

benji-potamus:
Closes BINIUS-107.

## Summary

Adds an `unsafe fn invert` to the `InvertOrZero` trait:

```rust
/// Returns the multiplicative inverse.
///
/// ## Safety
/// Requires that `self` is non-zero. Behavior is undefined otherwise.
unsafe fn invert(self) -> Self { self.invert_or_zero() }
```

The default body delegates to `invert_or_zero` (sound for non-zero inputs); specialized impls can later override it to exploit the non-zero precondition. Since `InvertOrZero` is a parent trait of `Field`/`PackedField`, every field and packed field gets it for free.

`InvertOrZero` gains a `Sized` bound — required for the default method body to move `self` and return `Self`. Every implementor is already `Sized`.

## Naming: `Field::invert` → `Field::try_invert`

Adding `InvertOrZero::invert` collides with the existing `Field::invert(&self) -> Option<Self>` (every field type would have two `invert` methods, making call sites ambiguous). This PR renames the checked, `Option`-returning method to `Field::try_invert` and updates all call sites — a pure mechanical rename, no behavior change. The actual per-site migration to `invert_or_zero` / `unsafe invert` and removal of `try_invert` happens in the follow-up #BINIUS-112.

This yields a clear trifecta:
- `invert_or_zero(self) -> Self` — returns zero on zero input
- `unsafe invert(self) -> Self` — UB on zero input (caller guarantees non-zero)
- `try_invert(&self) -> Option<Self>` — checked, `None` on zero

---

### DIFF
--- a/crates/field/src/arithmetic_traits.rs
+++ b/crates/field/src/arithmetic_traits.rs
@@ -63,6 +63,18 @@ pub(crate) use impl_trivial_wide_mul;
 pub trait InvertOrZero {
 	/// Returns the inverted value or zero in case when `self` is zero
 	fn invert_or_zero(self) -> Self;
+
+	/// Returns the multiplicative inverse.
+	///
+	/// ## Safety
+	/// Requires that `self` is non-zero. Behavior is undefined otherwise.
+	#[inline]
+	unsafe fn invert(self) -> Self
+	where
+		Self: Sized,
+	{
+		self.invert_or_zero()
+	}
 }
 
 /// Multiplication that is parameterized with some some strategy.
--- a/crates/field/benches/binary_field.rs
+++ b/crates/field/benches/binary_field.rs
@@ -80,7 +80,7 @@ impl FieldOperation for InvertOp {
 	type Result<F> = Option<F>;
 
 	fn call<F: Field>(lhs: F, _: F) -> Option<F> {
-		lhs.invert()
+		lhs.try_invert()
 	}
 }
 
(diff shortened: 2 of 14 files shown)


jimpo (approved):

crates/field/src/arithmetic_traits.rs
```diff
@@ -60,9 +60,18 @@ macro_rules! impl_trivial_wide_mul {
 pub(crate) use impl_trivial_wide_mul;
 
 /// Value that can be inverted
-pub trait InvertOrZero {
+pub trait InvertOrZero: Sized {
```
jimpo: Why does this need Sized?

benji-potamus: The *trait* doesn't — only the default `invert` body does, because it moves `self` by value and returns `Self`, which needs `Self: Sized` to typecheck generically (the required `invert_or_zero` has no body, so its declaration is exempt — the bound is only checked at impl sites).

I've moved it off the trait and onto the method as `unsafe fn invert(self) -> Self where Self: Sized` (pushed `8f5b3929`), so the trait's bound surface is unchanged. The only alternative that needs no `Sized` at all is making `invert` a *required* method, but that forces every impl (the macros + the manual ghash/aes/1b impls) to repeat `{ self.invert_or_zero() }`, so I kept it as a defaulted method.

jimpo (approved):

PR #1554, opened 2026-06-21
BINIUS-106: don't require power-of-two padding for the value vector

benji-potamus:
Removes the requirement that the committed value vector (`committed_total_len`) be padded to a power of two ([BINIUS-106](https://linear.app/jimpo/issue/BINIUS-106/dont-require-witness-padding-for-binius64)). The `ValueVec` now stores only the actual committed values; the power-of-two-length witness/monster multilinears that the polynomial protocols need are zero-padded at the point they are constructed.

## Commit 1 — `prover, verifier: don't assume committed witness length is a power of two`

Behaviour-preserving while the frontend still pads, so this commit is a no-op on every current path:

- `prover/src/fold_word.rs`: `fold_words` accepts a non-power-of-two `words` slice — it folds into a buffer of `log2_ceil(words.len())` variables, treating the high words up to that rounded-up length as zero (the iteration only covers the words actually present).
- `prover/src/prove.rs`: `pack_witness` takes the unpadded `combined_witness` and zero-pads each word as it packs into field elements (`.get(..).copied().unwrap_or(Word::ZERO)`); the shift reduction receives the unpadded `combined_witness` directly. The `pad_committed_witness` helper is gone.
- `prover/.../shift/key_collection.rs`: `key_ranges` is sized to `committed_total_len` (no `next_power_of_two`).
- `prover/.../shift/monster.rs`: `build_monster_multilinear` builds its multilinear zero-padded to `log2_ceil(committed_total_len)` words — it was `strict_log_2` on a pre-padded `key_ranges`.
- `verifier/.../shift/verify.rs`: derives the witness-MLE variable count with `log2_ceil_usize(...).max(LOG_WORDS_PER_ELEM)` instead of `strict_log_2` (same value as before).

## Commit 2 — `frontend: don't pad the committed value vector to a power of two`

Flips the producer now that the consumers tolerate it:

- `frontend/.../value_vec_alloc.rs`: drop the final `next_power_of_two()` on `committed_total_len`.
- `core/constraint_system.rs`: relax `ValueVecLayout::validate`; `core/error.rs`: drop the now-unused `ValueVecLenNotPowerOfTwo`.
- `frontend/stat.rs`: report the committed *polynomial* size (`committed_total_len.next_power_of_two()`).

## Behaviour-preserving + green

Proofs are unchanged — the committed witness *polynomial* is identical (`log2_ceil(raw) == log2(next_pow2(raw))`, and the padding words were already zero). The example snapshot tests (which round-trip prove→verify) pass without modification.

```
cargo test -p binius-core -p binius-frontend -p binius-prover   ✅
cargo test -p binius-examples                                   ✅ (snapshots unchanged)
cargo clippy --all --tests --benches --examples -- -D warnings  ✅
cargo +nightly-2026-01-01 fmt --check                           ✅
```

### DIFF
--- a/crates/core/src/constraint_system.rs
+++ b/crates/core/src/constraint_system.rs
@@ -668,8 +668,6 @@ pub struct ValueVecLayout {
 	pub offset_witness: usize,
 	/// The total number of committed values in the values vector. This does not include any
 	/// scratch values.
-	///
-	/// This must be a power-of-two.
 	pub committed_total_len: usize,
 	/// The number of scratch values at the end of the value vec.
 	pub n_scratch: usize,
@@ -680,14 +678,9 @@ impl ValueVecLayout {
 	///
 	/// Specifically checks that:
 	///
-	/// - the total committed length is a power of two.
 	/// - the public segment (constants and inout values) is padded to the power of two.
 	/// - the public segment is not less than the minimum size.
 	pub fn validate(&self) -> Result<(), ConstraintSystemError> {
-		if !self.committed_total_len.is_power_of_two() {
-			return Err(ConstraintSystemError::ValueVecLenNotPowerOfTwo);
-		}
-
 		if !self.offset_witness.is_power_of_two() {
 			return Err(ConstraintSystemError::PublicInputPowerOfTwo);
 		}
@@ -777,7 +770,6 @@ impl DeserializeBytes for ValueVecLayout {
 /// as the primary data structure for both constraint evaluation and polynomial commitment.
 ///
 /// Between these sections, there may be padding regions to satisfy alignment requirements.
-/// The total size is always a power of two as required for technical reasons.
 #[derive(Clone, Debug)]
 pub struct ValueVec {
 	layout: ValueVecLayout,
--- a/crates/prover/src/fold_word.rs
+++ b/crates/prover/src/fold_word.rs
@@ -9,7 +9,7 @@ use binius_field::{
 	},
 };
 use binius_math::FieldBuffer;
-use binius_utils::{checked_arithmetics::log2_strict_usize, rayon::prelude::*};
+use binius_utils::{checked_arithmetics::log2_ceil_usize, rayon::prelude::*};
 
 /// Computes a [`FieldBuffer`] where each element is the inner product of the bits of a word and a
 /// vector of field elements.
@@ -20,9 +20,11 @@ use binius_utils::{checked_arithmetics::log2_strict_usize, rayon::prelude::*};
 /// This implementation uses the [Method of Four Russians] to optimize the computation by
 /// precomputing a small lookup table and looking up into it using bitwise chunks of the words.
 ///
+/// The returned buffer has `log2_ceil(words.len())` variables. `words` need not have a power-of-two
+/// length; the high words up to that rounded-up length are treated as zero.
+///
 /// ## Preconditions
 /// * `vec` contains exactly [`binius_core::consts::WORD_SIZE_BITS`] elements
-/// * `words` has a power-of-two length
 ///
 /// [Method of Four Russians]: <https://en.wikipedia.org/wiki/Method_of_Four_Russians>
 pub fn fold_words<F, P>(words: &[Word], vec: &[F]) -> FieldBuffer<P>
@@ -56,11 +58,13 @@ where
 	P: PackedField<Scalar = F>,
 	T: Transformation<u64, F>,
 {
-	assert!(words.len().is_power_of_two()); // precondition
+	// `words` need not have a power-of-two length; the high words up to the next power of two are
+	// treated as zero, so the remaining slots after the last real word are zero-filled by resize.
+	let log_n = log2_ceil_usize(words.len());
+	let capacity = 1 << log_n.saturating_sub(P::LOG_WIDTH);
 
-	let log_n = log2_strict_usize(words.len());
-
-	let values = words
+	let mut values = Vec::<P>::with_capacity(capacity);
+	words
 		.par_chunks(P::WIDTH)
 		.map(|word_chunk| {
 			P::from_scalars(word_chunk.iter().map(|&word| transform.transform(&word.0)))
@@ -74,6 +78,7 @@ where
 mod tests {
 	use binius_core::consts::WORD_SIZE_BITS;
 	use binius_math::test_utils::random_scalars;
+	use binius_utils::checked_arithmetics::log2_strict_usize;
 	use binius_verifier::config::B128;
 	use rand::prelude::*;
 
--- a/crates/prover/src/fold_word.rs
+++ b/crates/prover/src/fold_word.rs
@@ -69,7 +69,8 @@ where
 		.map(|word_chunk| {
 			P::from_scalars(word_chunk.iter().map(|&word| transform.transform(&word.0)))
 		})
-		.collect::<Vec<_>>();
+		.collect_into_vec(&mut values);
+	values.resize(capacity, P::default());
 
 	FieldBuffer::new(log_n, values.into_boxed_slice())
 }
--- a/crates/prover/src/protocols/shift/key_collection.rs
+++ b/crates/prover/src/protocols/shift/key_collection.rs
@@ -236,7 +236,7 @@ fn update_with_operand(
 
 /// Constructs a `KeyCollection` from a constraint system.
 pub fn build_key_collection(cs: &ConstraintSystem) -> KeyCollection {
-	// Initialize a temporary list of builder keys lists, one for each word
+	// Initialize a temporary list of builder keys lists, one for each committed word.
 	let mut builder_key_lists: Vec<Vec<BuilderKey>> = (0..cs.value_vec_layout.committed_total_len)
 		.map(|_| Vec::new())
 		.collect();
--- a/crates/prover/src/protocols/shift/monster.rs
+++ b/crates/prover/src/protocols/shift/monster.rs
@@ -6,9 +6,9 @@ use binius_field::{AESTowerField8b, BinaryField, Field, PackedField};
 use binius_math::{
 	BinarySubspace, FieldBuffer, multilinear::eq::eq_ind_partial_eval, univariate::lagrange_evals,
 };
-use binius_utils::{checked_arithmetics::strict_log_2, rayon::prelude::*};
+use binius_utils::{checked_arithmetics::log2_ceil_usize, rayon::prelude::*};
 use binius_verifier::{
-	config::{LOG_WORD_SIZE_BITS, WORD_SIZE_BITS},
+	config::{LOG_WORD_SIZE_BITS, LOG_WORDS_PER_ELEM, WORD_SIZE_BITS},
 	protocols::shift::{BITAND_ARITY, INTMUL_ARITY, evaluate_h_op},
 };
 use bytemuck::zeroed_vec;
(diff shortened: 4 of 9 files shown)


crates/prover/benches/shift_reduction.rs
```diff
@@ -90,6 +90,16 @@ fn bench_prove_and_verify(c: &mut Criterion) {
 		let intmul_evals = [F::random(&mut rng); 4];
 		let key_collection = build_key_collection(&cs);
 
+		// The committed value vector is not padded to a power of two, but the shift reduction folds
+		// the witness as a power-of-two-length multilinear, so pad it here just as the prover does
+		// internally.
+		let padded_witness = {
```
jimpo: No, prove should take the unpadded Vec.

crates/prover/src/protocols/shift/key_collection.rs
```diff
@@ -236,10 +236,14 @@ fn update_with_operand(
...
-	let mut builder_key_lists: Vec<Vec<BuilderKey>> = (0..cs.value_vec_layout.committed_total_len)
-		.map(|_| Vec::new())
-		.collect();
+	// Initialize a temporary list of builder keys lists, one for each word.
+	//
+	// The committed values are not padded to a power of two in the layout, but the witness is
+	// folded as a power-of-two-length multilinear during the shift reduction. Pad the per-word key
+	// lists (and thus `key_ranges`) up to that length so the monster multilinear matches the
+	// witness polynomial; the padding words carry no keys and contribute zero.
+	let n_words = cs.value_vec_layout.committed_total_len.next_power_of_two();
```
jimpo: power-of-two padding shouldn't be necessary here.

jimpo:
Follow the new guidance on code commits: https://github.com/binius-zk/binius64/pull/1555

benji-potamus:
Thanks — followed the new AGENTS.md guidance. One code comment referenced change history ("the committed length is no longer padded…"); I reworded it to describe only the current behavior. The remaining comments already describe the current contract, and the change-history context lives in the commit descriptions / PR body per the guidance.

Also rebased onto latest `main` (picking up #1555). Force-pushed to `ca34587`; CI re-running.

crates/core/src/constraint_system.rs
```diff
@@ -669,7 +669,8 @@ pub struct ValueVecLayout {
 	/// The total number of committed values in the values vector. This does not include any
 	/// scratch values.
 	///
-	/// This must be a power-of-two.
+	/// This is not padded to a power of two; the prover zero-pads the committed witness up to a
```
jimpo: Remove this comment.

crates/core/src/constraint_system.rs
```diff
@@ -776,8 +772,9 @@ impl DeserializeBytes for ValueVecLayout {
 /// [`ConstraintSystem`]. It follows the layout defined by [`ValueVecLayout`] and serves
 /// as the primary data structure for both constraint evaluation and polynomial commitment.
 ///
-/// Between these sections, there may be padding regions to satisfy alignment requirements.
-/// The total size is always a power of two as required for technical reasons.
+/// Between these sections, there may be padding regions to satisfy alignment requirements. The
+/// committed length is not padded to a power of two; the prover zero-pads the committed witness up
+/// to a power-of-two length when it builds the witness polynomial.
```
jimpo:
```suggestion
/// Between these sections, there may be padding regions to satisfy alignment requirements.
```

crates/prover/src/prove.rs
```diff
@@ -367,23 +372,30 @@ fn compute_batched_transparent<P: PackedField<Scalar = B128>>(
 	rs_eq_ind
 }
 
+/// Materializes the committed witness zero-padded to `2^log_witness_words` words.
+///
+/// The [`ValueVec`] stores only the unpadded committed values, but the witness polynomial that the
+/// prover commits to (and folds during the shift reduction) is a power-of-two-length multilinear.
+/// The padding words are zero, matching the high indices of the witness multilinear that the
+/// verifier assumes to be zero.
+fn pad_committed_witness(witness: &ValueVec, log_witness_words: usize) -> Vec<Word> {
```
jimpo: Remove this

crates/prover/src/prove.rs
```diff
@@ -103,7 +103,12 @@ impl IOPProver {
 
 		// [phase] Setup - initialization and constraint system setup
 		let setup_guard = tracing::debug_span!("Prepare Witness").entered();
-		let witness_packed = pack_witness::<P>(self.log_witness_elems, &witness)?;
+		// The `ValueVec` stores only the unpadded committed values, but the witness polynomial that
+		// the prover commits to and folds has a power-of-two length. Materialize the zero-padded
+		// witness once here and reuse it for both packing and the shift reduction.
+		let log_witness_words = self.log_witness_elems + LOG_WORDS_PER_ELEM;
+		let padded_witness = pad_committed_witness(&witness, log_witness_words);
+		let witness_packed = pack_witness::<P>(self.log_witness_elems, &padded_witness);
```
jimpo: Just do the padding inside of the `pack_witness` loop.

jimpo (changes requested):

crates/prover/benches/shift_reduction.rs
benji-potamus: Done — the bench (and the integration test) now pass `value_vec.combined_witness()` straight to `prove`. The witness is zero-padded to the power-of-two fold length inside the shift reduction (`phase_2`), not by the caller.

crates/prover/src/protocols/shift/key_collection.rs
benji-potamus: Done — `build_key_collection` now sizes `key_ranges` to `committed_total_len`. The power-of-two-ness moved into the multilinear construction where it's actually needed: `build_monster_multilinear` now builds its buffer zero-padded to `log2_ceil(committed_total_len)` (it was `strict_log_2` on the pre-padded `key_ranges`), and `phase_2` pads the witness before `fold_words`.

crates/core/src/constraint_system.rs
benji-potamus: Removed.

crates/core/src/constraint_system.rs
benji-potamus: Applied.

crates/prover/src/prove.rs
benji-potamus: Removed `pad_committed_witness`; the shift reduction now takes the unpadded `combined_witness` directly.

crates/prover/src/prove.rs
benji-potamus: Done — `pack_witness` now takes the unpadded `combined_witness` and zero-pads each word as it packs into field elements (`combined_witness.get(..).copied().unwrap_or(Word::ZERO)`).

crates/prover/src/protocols/shift/phase_2.rs
```diff
@@ -67,7 +71,13 @@ where
...
 
 	let r_j_tensor = eq_ind_partial_eval::<F>(&r_j);
-	let r_j_witness = fold_words::<_, P>(words, r_j_tensor.as_ref());
+
+	// The committed words are not padded to a power of two; fold the witness as a
+	// power-of-two-length multilinear with the variable count the verifier uses for the opening.
+	let log_witness_words = log2_ceil_usize(words.len()).max(LOG_WORDS_PER_ELEM);
+	let mut padded_words = vec![Word::ZERO; 1 << log_witness_words];
+	padded_words[..words.len()].copy_from_slice(words);
+	let r_j_witness = fold_words::<_, P>(&padded_words, r_j_tensor.as_ref());
```
jimpo: fold_words shouldn't need to take a power-of-two length slice. It can document that the remaining words (up to the length of the second argument) are zero. That just shortens the iteration.

jimpo (changes requested):

crates/prover/src/protocols/shift/phase_2.rs
benji-potamus: Done — `fold_words` (via `fold_words_with_transform`) now accepts a non-power-of-two `words` slice: it derives the output length from `log2_ceil(words.len())` and treats the high words up to that length as zero (the `P::from_fn` index loop just covers the words that are actually present). `phase_2` and the test's `evaluate_witness` no longer pad — they pass the unpadded witness straight in, so both revert to their pre-PR call sites and the padding lives entirely in `fold_words`.

benji-potamus:
Rebased onto current `main` (`afcec7f4`, post-#1560) to keep it up-to-date. No code changes — the GHASH commits that landed since the original push (#1557–#1560) touch entirely different crates (binius-field). Verified locally: `cargo test -p binius-core -p binius-frontend -p binius-prover` and `cargo clippy --all … -D warnings` both green on the rebased stack. CI re-running on the new head (`9edc345f`).

crates/prover/src/fold_word.rs
```diff
@@ -56,14 +58,19 @@ where
...
-
-	let values = words
-		.par_chunks(P::WIDTH)
-		.map(|word_chunk| {
-			P::from_scalars(word_chunk.iter().map(|&word| transform.transform(&word.0)))
+	// `words` need not have a power-of-two length; the high words up to the next power of two are
+	// treated as zero, so the iteration only covers the words that are actually present.
+	let log_n = log2_ceil_usize(words.len());
+
+	let values = (0..1 << log_n.saturating_sub(P::LOG_WIDTH))
```
jimpo: The old code was better. But modify it so that it preallocated a Vec::with_capacity(...), then uses `collect_into_vec`, then resizes the Vec with 0s up to max capacity after the par loop.

crates/prover/src/protocols/shift/monster.rs
```diff
@@ -187,33 +187,45 @@ where
...
-							.map(|(operand_index, acc)| {
-								let index = key.id as usize
-									+ operand_index * SHIFT_VARIANT_COUNT * WORD_SIZE_BITS;
-								acc * scalars[index]
-							})
-							.sum::<F>()
+	// The committed words are not padded to a power of two, but the monster multilinear is folded
+	// alongside the witness as a power-of-two-length multilinear. Build it zero-padded up to that
+	// length; the padding words carry no keys and contribute zero.
+	let word_eval = |word_idx: usize| -> F {
```
jimpo: The old code was better. But modify it so that it preallocated a Vec::with_capacity(...), then uses `collect_into_vec`, then resizes the Vec with 0s up to max capacity after the par loop.

crates/prover/src/prove.rs
```diff
@@ -369,10 +369,10 @@ fn compute_batched_transparent<P: PackedField<Scalar = B128>>(
 
 fn pack_witness<P: PackedField<Scalar = B128>>(
 	log_witness_elems: usize,
-	witness: &ValueVec,
+	combined_witness: &[Word],
```
jimpo: Keep name `witness` , just change the type to &[Word]

crates/prover/src/prove.rs
jimpo: Different strategy for this loop. First, use `witness.as_chunks::<2>`. Iterate over the full chunks and collect_into_vec. Handle the possible odd remaining Word, then resize with 0s for the rest.

jimpo (changes requested):

crates/prover/src/fold_word.rs
benji-potamus: Done — reverted to the `par_chunks` + `from_scalars` approach. Now preallocates with `Vec::with_capacity(capacity)`, collects with `collect_into_vec`, then `resize`s with `P::default()` up to the power-of-two capacity to zero-pad the trailing words.

crates/prover/src/protocols/shift/monster.rs
benji-potamus: Done — same pattern as `fold_words_with_transform`: preallocate with `Vec::with_capacity(capacity)`, use `key_collection.key_ranges.par_chunks(P::WIDTH).map(|chunk| P::from_scalars(...)).collect_into_vec(...)`, then `resize` with zeros. The `word_eval` closure is gone — the per-range computation is inlined into `par_chunks`.

crates/prover/src/prove.rs
benji-potamus: Done — parameter renamed to `witness: &[Word]`.

crates/prover/src/prove.rs
benji-potamus: Done — now uses `witness.as_chunks::<2>()` to split into word pairs. The full pairs go through `par_chunks(P::WIDTH).map(|word_pairs| P::from_scalars(...)).collect_into_vec(...)`. The possible odd remaining word is handled separately with a single `push`. Then `resize` to zero-pad up to the power-of-two polynomial length. The `spare_capacity_mut` + unsafe `set_len` is gone.

jimpo (approved):

PR #1561, opened 2026-06-22
BINIUS-128: define GHASH multiplication via WideMul

benji-potamus:
Defines multiplication for every GHASH packing (and the `BinaryField128bGhash` scalar) as `reduce(wide_mul)`, so the widening multiply is the single source of truth for both `Mul` and `WideMul` ([BINIUS-128](https://linear.app/jimpo/issue/BINIUS-128)). Builds on #1560 (scalar `WideMul`).

## What changed

- **`GhashMulStrategy`** (`arch/strategies.rs`): a blanket `impl<P: WideMul> TaggedMul<GhashMulStrategy> for P` computing `reduce(wide_mul(a, b))`. Every GHASH packing now passes `mul: (GhashMulStrategy)` to the packing macro (the macro itself is unchanged). The bespoke per-arch `TaggedMul` impls and the now-redundant `mul_clmul` (x86_64 + aarch64 arithmetic modules) are deleted.

- **No more `TrivialWideMul` on any GHASH packing.** The multi-lane fallbacks (portable 256/512, and the non-VPCLMULQDQ x86_64 256/512 paths) now use a new deferring **`ScaledGhashWideMul`** (`arch/portable/arithmetic/ghash_scaled.rs`): it splits the packing into 128-bit lanes via `Divisible<M128>` and applies the width-1 GHASH `WideMul` per lane — CLMUL-accelerated where available, portable schoolbook otherwise. The vectorized VPCLMULQDQ 256/512 path keeps `GhashClMulWideMul` (does both/all lanes in parallel — faster than splitting).

- **wasm32** width-1 GHASH switches from `TrivialWideMul` to the portable deferring `GhashWideMul` (its SIMD `M128` already implements `Underlier128bLanes`).

- **Scalar `BinaryField128bGhash::mul`** is now `Self::reduce(Self::wide_mul(self, rhs))`.

## Behavior-preserving

`reduce(wide_mul)` is algorithmically identical to the old fused `mul_clmul` (same 4 multiply + 2 reduction CLMULs, same reduction order); the scaled fallback reproduces the old per-lane element-wise multiply exactly. Verified against the scalar reference by the existing proptests.

## Verification

- `cargo test -p binius-field --lib` native (236 passed) — covers 128 CLMUL, 256 vectorized `reduce(wide_mul)`, and 512 **scaled fallback** (this CPU has pclmulqdq+vpclmulqdq+avx2 but no avx512f).
- Same with all CLMUL features disabled (227 passed) — covers portable-128 + scaled-256/512 via the portable width-1 field.
- Cross-compile matrix (per `CONTRIBUTING.md`): aarch64 `+neon,+aes`; x86_64 `+pclmulqdq`, `+avx2,+vpclmulqdq`, `+avx512f,+vpclmulqdq`, and full portable fallback; wasm32-wasip1 / wasm32-unknown-unknown — all clean.
- `cargo clippy -p binius-field` on both the CLMUL and portable paths, `cargo +nightly-2026-01-01 fmt --check`, `cargo doc -D warnings`, and a full `cargo build --workspace` — all green.

### DIFF
--- a/crates/field/src/arch/portable/arithmetic/ghash_scaled.rs
+++ b/crates/field/src/arch/portable/arithmetic/ghash_scaled.rs
@@ -1,0 +1,142 @@
+// Copyright 2026 The Binius Developers
+
+//! Scaled deferred-reduction GHASH widening multiply for multi-lane packings.
+//!
+//! A 256- or 512-bit GHASH packing is a vector of independent 128-bit GHASH elements. When no
+//! wide-vector carryless multiply is available (e.g. no VPCLMULQDQ, or the portable build), the
+//! widening multiply is built by applying the width-1 GHASH [`WideMul`] — itself CLMUL-accelerated
+//! when possible, portable schoolbook otherwise — to each 128-bit lane. This keeps every GHASH
+//! packing on a *deferring* `WideMul` (never the eager `TrivialWideMul`), so a sum of products
+//! reduces once per lane at the end.
+//!
+//! [`ScaledGhashWideMul`] is the wrapper passed as the `wide_mul` strategy of the packing macro. A
+//! single generic [`WideMul`] impl covers every multi-lane packing — both the portable
+//! `ScaledUnderlier` packings and the x86_64 SIMD `M256`/`M512` ones — by splitting the underlier
+//! into its 128-bit lanes via [`Divisible`]. The const parameter `N` is the lane count, which must
+//! equal `<U as Divisible<M128>>::N` for the underlying packed type.
+
+use std::{
+	array,
+	iter::Sum,
+	ops::{Add, AddAssign, Sub, SubAssign},
+};
+
+use bytemuck::TransparentWrapper;
+
+use crate::{
+	BinaryField,
+	arch::{M128, PackedPrimitiveType},
+	arithmetic_traits::WideMul,
+	underlier::{Divisible, UnderlierType},
+};
+
+/// The unreduced product of a multi-lane GHASH packing: one independent width-1 widening product
+/// per 128-bit lane. Lanes accumulate and reduce independently, mirroring the packing's structure.
+/// The const parameter `N` is the lane count.
+#[derive(Clone, Copy, Debug)]
+pub struct ScaledWideGhashProduct<O, const N: usize>([O; N]);
+
+impl<O: Copy + Default, const N: usize> Default for ScaledWideGhashProduct<O, N> {
+	#[inline]
+	fn default() -> Self {
+		Self([O::default(); N])
+	}
+}
+
+impl<O: Copy + Add<Output = O>, const N: usize> Add for ScaledWideGhashProduct<O, N> {
+	type Output = Self;
+
+	#[inline]
+	fn add(self, rhs: Self) -> Self {
+		Self(array::from_fn(|i| self.0[i] + rhs.0[i]))
+	}
+}
+
+impl<O: Copy + Add<Output = O>, const N: usize> AddAssign for ScaledWideGhashProduct<O, N> {
+	#[inline]
+	fn add_assign(&mut self, rhs: Self) {
+		*self = *self + rhs;
+	}
+}
+
+impl<O: Copy + Sub<Output = O>, const N: usize> Sub for ScaledWideGhashProduct<O, N> {
+	type Output = Self;
+
+	#[inline]
+	fn sub(self, rhs: Self) -> Self {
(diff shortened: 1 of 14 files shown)


crates/field/src/arch/portable/arithmetic/ghash_scaled.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// Generates [`WideMul`](crate::arithmetic_traits::WideMul) for `ScaledGhashWideMul<$packed>`,
+/// applying the width-1 GHASH widening multiply to each of `$n` 128-bit lanes.
+///
+/// The lane type is [`crate::arch::M128`] (the active arch's 128-bit underlier, which is exactly
+/// what every multi-lane GHASH packing's underlier is [`Divisible`](crate::underlier::Divisible)
+/// into), and the per-lane field is `PackedPrimitiveType<M128, BinaryField128bGhash>` — the
+/// width-1 GHASH packing, whose `WideMul` is CLMUL-accelerated where available and portable
+/// schoolbook otherwise.
+macro_rules! impl_scaled_ghash_wide_mul {
```
jimpo: We need to find a way to get rid of this macro with a generic impl

jimpo: Why doesn't something generic like the way `ScaledStrategy` works, work here for WideMul?

jimpo (changes requested):

crates/field/src/arch/portable/arithmetic/ghash_scaled.rs
benji-potamus: Done — replaced the macro with a single generic impl (just pushed):

```rust
impl<U, Scalar> WideMul for ScaledGhashWideMul<PackedPrimitiveType<U, Scalar>>
where
    U: UnderlierType + Divisible<M128>,
    Scalar: BinaryField,
    Lane<Scalar>: WideMul,
    <Lane<Scalar> as WideMul>::Output: Copy + Default,
```

It works like `ScaledStrategy`, but keys on `Divisible<M128>` rather than `ScaledUnderlier<U, N>`. The reason it can't key on `ScaledUnderlier` (the way `ScaledStrategy` does) and still be a *single* impl: the portable 256/512 packings are `ScaledUnderlier<M128, N>`, but the **x86_64 SIMD packings are flat `M256`/`M512`** (`__m256i`/`__m512i`), which aren't `ScaledUnderlier`. Both are `Divisible<M128>`, so keying on that covers all multi-lane GHASH packings with one impl. It splits via `Divisible::value_iter` and rebuilds via `Divisible::from_iter` (no `unsafe`).

The one wrinkle from going through `Divisible`: the lane count is `Divisible::N` (a value, not a type-level const I can size an array with on stable), so the per-lane product buffer is a fixed `[O; MAX_LANES]` with `MAX_LANES = 4` (widest GHASH packing is 512-bit). A narrower packing fills a prefix; `value_iter`/`from_iter` bound the work to its actual `N`, so the unused tail is never multiplied or reduced. Happy to switch to the exact-`N` `ScaledUnderlier`-keyed form instead if you prefer — that'd be two impls (one generic for the portable packings, one for the x86 SIMD ones).

jimpo: Can't you make `ScaledGhashWideMul` have a generic `const N: usize` parameter (using PhantomData if necessary) instead of using the hardcoded maximum? Then assert N equal to Divisible::N

benji-potamus: Done — just pushed (86682b5a).

`ScaledGhashWideMul<T, const N: usize>` and `ScaledWideGhashProduct<O, const N: usize>` now carry the lane count as a const generic. The output array is `[O; N]` — no `MAX_LANES` buffer and no unused tail. A `debug_assert_eq!(N, <U as Divisible<M128>>::N)` in `wide_mul` catches any mismatch at debug time (using an associated const from a generic bound in a `const {}` block requires `generic_const_exprs`, so this stays as a runtime assertion in debug builds).

For the packing macro call sites (`wide_mul: (SomeType)`), the macro expands as `SomeType<$packed_name>`, so a two-param `ScaledGhashWideMul<T, N>` can't be passed as a single token-tree argument. The fix is two named convenience aliases — `Scaled2xGhashWideMul<T>` and `Scaled4xGhashWideMul<T>` — which the packing files import and pass directly. The x86 `GhashWideMul<T>` aliases are updated to point at the appropriate lane-count alias.

jimpo (approved):

PR #1562, opened 2026-06-22
BINIUS-130: blake3_compress_2x_seq — two sequential BLAKE3 compressions via 2x lanes

benji-potamus:
Implements [BINIUS-130](https://linear.app/jimpo/issue/BINIUS-130/blake3-double-compression) (sub-issue of BINIUS-129).

## What

Adds `binius_circuits::blake3::blake3_compress_2x_seq`, a circuit gadget that evaluates **two sequential** BLAKE3 compressions — `C2 = compress(C1, block2, …)` where `C1 = compress(cv, block1, …)` — while paying the gate cost of a **single** compression by reusing the parallel 7-round core of `blake3_compress_2x`.

## How

The two compressions have a data dependency (C2's input chaining value is C1's output), so they can't naively run as independent parallel lanes. The trick:

- **Lane 1 (high 32 bits) = C1**, input CV = the passed-in `cv` (genuine, by construction).
- **Lane 0 (low 32 bits) = C2**, input CV = C1's output, supplied by a custom **`Blake3CompressHint`** that precomputes the first compression off-circuit.
- One `blake3_compress_2x` call advances both lanes.
- The hint is **bound**: lane-0's input CV (the hint) is constrained word-for-word against lane-1's *in-circuit* output (`out >> 32`), so a dishonest hint can't pass. The hint value is masked to its low 32 bits first, so lane 1 is provably computed from the genuine `cv` regardless of the hint's high bits.

The result is returned packed as requested: **C2 in the low 32 bits, C1 in the high 32 bits** of each of the 8 output wires — which is exactly what `blake3_compress_2x` already produces, so it's returned directly.

Each compression takes its own `counter / block_len / flags` (mirroring `blake3_compress` ×2), packed into the 2x lanes.

## Notes

- The pure-Rust reference compression (`ref_compress`) is lifted out of the test module to module scope so the hint and the tests share one implementation.
- Tests check the gadget against the reference for chained two-block inputs and for distinct per-compression counters/lengths/flags (including a counter straddling the 32-bit boundary). `populate_wire_witness` enforces the constraints, so the tests exercise the hint binding, not just the happy path.
- Interface (lane mapping, output packing, per-compression params) confirmed with @jimpo on the Linear thread.

### DIFF
--- a/crates/circuits/src/blake3/compress.rs
+++ b/crates/circuits/src/blake3/compress.rs
@@ -1,4 +1,5 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 //! BLAKE3 compression primitive.
 //!
 //! A BLAKE3 block is 64 bytes (16 × 32-bit words). The compression function mixes an
@@ -9,12 +10,13 @@
 //!
 //! [reference implementation]: https://github.com/BLAKE3-team/BLAKE3/blob/master/src/portable.rs
 
-use std::array;
+use std::{array, iter};
 
 use binius_core::word::Word;
-use binius_frontend::{CircuitBuilder, Wire};
+use binius_frontend::{CircuitBuilder, Hint, Wire};
 
 use super::{IV, MSG_SCHEDULE};
+use crate::util::clear_high_bits;
 
 /// BLAKE3 compression function.
 ///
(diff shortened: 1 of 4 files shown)


benji-potamus:
CI note: **`rust-checks-x86_64` failed, but it looks like flaky infra, not a real failure.** On this exact commit (`8a6a7149`):

- ✅ `rust-checks-arm64` passed
- ✅ `rust-checks-x86_64-portable` passed
- ✅ all `circuits-snapshot (*)`, `rust-doc`, `format`, `wasm-*` passed

The only difference for the red job is the `x86_64` (native) runner. Locally on x86_64 I ran the same steps green: `cargo test --workspace --no-fail-fast` (47 test suites, 0 failures) and `cargo clippy --workspace --all-targets -- -D warnings` (clean).

A re-run of the `rust-checks-x86_64` job should go green — I don't have permission to re-run it myself. Flagging so the red check isn't mistaken for a real break during review.

crates/circuits/src/blake3/compress.rs
```diff
@@ -466,4 +590,140 @@ mod tests {
...
+		block2: [u32; 16],
+		counter1: u64,
+		block_len1: u32,
+		flags1: u32,
+		counter2: u64,
+		block_len2: u32,
+		flags2: u32,
+	) -> ([u32; 8], [u32; 8]) {
+		let builder = CircuitBuilder::new();
+		let cv_wires: [Wire; 8] = std::array::from_fn(|_| builder.add_witness());
```
jimpo: `use std::array`

crates/circuits/src/blake3/compress.rs
```diff
@@ -179,69 +180,192 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
...
+///
+/// # Returns
+///
+/// The two output chaining values packed into 8 wires: the second compression's output in the
+/// low 32 bits of each wire, the first compression's output in the high 32 bits.
+#[allow(clippy::too_many_arguments)]
+pub fn blake3_compress_2x_seq(
+	builder: &CircuitBuilder,
+	cv: [Wire; 8],
+	block1: [Wire; 16],
```
jimpo: Use array parameter to reduce the number of arguments: eg. blocks: [[Wire; 16]; 2], counters: [u64; 2]

crates/circuits/src/blake3/compress.rs
```diff
@@ -179,69 +180,192 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
...
-		v[b] = (v[b] ^ v[c]).rotate_right(12);
-		v[a] = v[a].wrapping_add(v[b]).wrapping_add(my);
-		v[d] = (v[d] ^ v[a]).rotate_right(8);
-		v[c] = v[c].wrapping_add(v[d]);
-		v[b] = (v[b] ^ v[c]).rotate_right(7);
+	let mask_lo32 = builder.add_constant(Word(0xFFFF_FFFF));
+	// The hinted chaining value, restricted to the low 32 bits. Masking makes the high lane of
+	// `merged_cv` below provably equal to `cv` regardless of what the prover put in the hint's
+	// high bits, so the first compression (high lane) is computed from the genuine `cv`.
+	let c1_hint: [Wire; 8] = array::from_fn(|i| builder.band(cv1[i], mask_lo32));
```
jimpo: The hint should return wires with the low32 bits being the compression output and the high32 bits being the input. Then assert_eq on SRL(SLL(hint input)) (instead of and mask) with SRL(hint output).

jimpo (changes requested):

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — added `use std::array;` to the test module and changed all `std::array::from_fn` to `array::from_fn`.

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — collapsed to `blocks: [[Wire; 16]; 2]`, `counters: [Wire; 2]`, `block_lens: [Wire; 2]`, `flags: [Wire; 2]`. Removed the `#[allow(clippy::too_many_arguments)]` on the public function; kept it on the test helper `run_compress_2x_seq` which still takes individual scalars.

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — `Blake3CompressHint::execute` now packs `out[i] as u64 | ((cv[i] as u64) << 32)` into each output word (low 32 = C1 output, high 32 = cv input). The extraction in `blake3_compress_2x_seq` is now `builder.shr(builder.shl(cv1[i], 32), 32)` instead of `builder.band(cv1[i], mask_lo32)`. The `merged_cv` construction still uses the actual `cv` wires for the high lane so the first compression is provably computed from the genuine input.

crates/circuits/src/blake3/compress.rs
```diff
@@ -179,69 +180,191 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
...
+///
+/// # Returns
+///
+/// The two output chaining values packed into 8 wires: the second compression's output in the
+/// low 32 bits of each wire, the first compression's output in the high 32 bits.
+pub fn blake3_compress_2x_seq(
+	builder: &CircuitBuilder,
+	cv: [Wire; 8],
+	blocks: [[Wire; 16]; 2],
+	counters: [Wire; 2],
```
jimpo: We only need one counter because this is a sequence (one is the incr of the other)

crates/circuits/src/blake3/compress.rs
```diff
@@ -179,69 +180,191 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
...
-		v[d] = (v[d] ^ v[a]).rotate_right(8);
-		v[c] = v[c].wrapping_add(v[d]);
-		v[b] = (v[b] ^ v[c]).rotate_right(7);
+	let mask_lo32 = builder.add_constant(Word(0xFFFF_FFFF));
+	// The hint packs `low32 = C1 output, high32 = cv input`. Extract C1 output via SHL+SHR so
+	// the high lane of `merged_cv` is provably equal to `cv` regardless of the hint's high bits.
+	let c1_hint: [Wire; 8] = array::from_fn(|i| builder.shr(builder.shl(cv1[i], 32), 32));
+
+	// Pack two lane values into one wire: low 32 bits = lane 0, high 32 bits = lane 1.
+	let pack = |lo: Wire, hi: Wire| builder.bor(lo, builder.shl(hi, 32));
```
jimpo: bor -> bxor

crates/circuits/src/blake3/compress.rs
```diff
@@ -179,69 +180,191 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
...
+
+	// Lane 0 = second compression, lane 1 = first compression.
+	let merged_cv: [Wire; 8] = array::from_fn(|i| pack(c1_hint[i], cv[i]));
```
jimpo: Don't compute merged_cv. The hint should return merged_cv. And instead, use assert_eq to constrain the hint output to correspond with the input `cv` (matching top 32 bits to bottom 32 bits).

crates/circuits/src/blake3/compress.rs
```diff
@@ -179,69 +180,191 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
...
+	let merged_block: [Wire; 16] = array::from_fn(|i| pack(blocks[1][i], blocks[0][i]));
```
jimpo: We _cannot_ assume the top 32 bits of each block wire are 0. So need to clear them with shifts. SRL(SLL(block1)) XOR SLL(block0)

crates/circuits/src/blake3/compress.rs
```diff
@@ -179,69 +180,191 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
...
+
+	let split = |c: Wire| (builder.band(c, mask_lo32), builder.shr(c, 32));
```
jimpo: Using band for masking is less efficient than SLL -> SRL. Maybe make a helper gadget called "clear_high_bits(Builder, Wire, n: usize)" that does this pattern, so it can be reused.

jimpo (changes requested):

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — the public signature now takes a single `counter: Wire`. The second compression's counter is `counter + 1`, computed in-circuit (`builder.iadd(counter, add_constant_64(1))`, carry discarded). Updated `compress_2x_seq_distinct_params` to use `counter = 0x0000_0001_FFFF_FFFF` so the in-circuit `+1` carries across the 32-bit boundary.

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — `pack` now uses `bxor`. The low operand is cleared to its low 32 bits first and `shl(hi, 32)` zeroes the high operand's low bits, so the two halves are disjoint and XOR == OR.

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — the hint now returns `merged_cv` directly (each word: low 32 = C1 output, high 32 = the `cv` input word), and it's fed straight into `blake3_compress_2x` with no in-circuit `pack` for the CV. Added an `assert_eq` binding the high 32 bits of each merged word to the input `cv` (top↔bottom), alongside the existing low-32 ↔ in-circuit-C1-output binding — so both the first compression's input and its output are constrained.

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — `merged_block[i]` is now `clear_high_bits(blocks[1][i], 32) XOR shl(blocks[0][i], 32)`, i.e. `SRL(SLL(block1)) XOR SLL(block0)`. Applied the same clearing to the `block_lens` and `flags` low operands, which carried the same unstated "high bits are zero" assumption.

crates/circuits/src/blake3/compress.rs
benji-potamus: Done — added `clear_high_bits(builder, w, n)` (= `shr(shl(w, n), n)`) in a new `crates/circuits/src/util.rs`, used throughout this gadget in place of `band` masking. Left `ripemd.rs:165` (same `shr(shl(w, 32), 32)` pattern) untouched to keep this PR focused — happy to migrate it to the helper in a follow-up.

benji-potamus:
Round 3 pushed (`20fffd7b`, rebased onto latest `main` / #1564). All five review comments addressed:

- **Single counter** — signature now takes one `counter: Wire`; the second compression uses `counter + 1` (64-bit in-circuit `iadd`).
- **`bor` → `bxor`** in `pack`, with the low operand pre-cleared so the lanes are disjoint.
- **Hint returns `merged_cv` directly** — fed straight into `blake3_compress_2x`; added an `assert_eq` binding each word's high 32 bits to the input `cv` (top↔bottom), keeping the existing low-32 ↔ in-circuit-C1-output binding.
- **Clear block high bits** — `SRL(SLL(block1)) XOR SLL(block0)`; same clearing applied to `block_lens`/`flags`.
- **`clear_high_bits(builder, w, n)` helper** in new `crates/circuits/src/util.rs`, replacing `band` masking throughout.

Locally green: `fmt --check`, `clippy -D warnings`, `cargo test -p binius-circuits blake3` (14/14 — `compress_2x_seq_distinct_params` now uses a counter whose `+1` carries across the 32-bit boundary).

jimpo (approved):

PR #1563, opened 2026-06-23
BINIUS-131: blake3_compress_opt — single compression with rounds split across two 2x lanes

benji-potamus:
Implements [BINIUS-131](https://linear.app/jimpo/issue/BINIUS-131/blake3-optimized-compress) (sub-issue of BINIUS-129).

## What

Adds `binius_circuits::blake3::blake3_compress_opt`, a circuit gadget that computes **one** BLAKE3 compression at the gate cost of **4 rounds** rather than 7, by splitting the 7 rounds across the two parallel lanes of the 2x circuit framework.

Same signature as `blake3_compress`; returns the same 8-word output chaining value.

## How

BLAKE3's 7 rounds are split into two sequential halves:

- **Lane 1 (high 32 bits)** — runs rounds 0–3 starting from the initial 16-word state `v_init`.
- **Lane 0 (low 32 bits)** — runs rounds 4–6 (plus one padding round, totalling 4 steps for symmetry) starting from the intermediate state `v_mid` after round 3, supplied by a `Blake3HalfCompressHint`.

Because the two lanes run **different round schedules** within each step, the 16 message words are re-packed per-step: for step _i_, position _k_ of the merged message wire carries `block[MSG_SCHEDULE[i+4][k]]` in the low half and `block[MSG_SCHEDULE[i][k]]` in the high half (see `round_opt_step`). The underlying `g()` function is lane-agnostic (`iadd_32`/`rotr32`/`bxor` all handle 64-bit packed words), so it applies correctly to both lanes.

**Feed-forward** is extracted from lane 0 after step 2 (round 6), before the padding step runs, via `bxor(v[i], v[i+8]) & 0xFFFF_FFFF`.

**Hint binding**: after step 3 (round 3 for lane 1), lane 1's final 16-word state is asserted word-for-word equal to `v_mid` from the hint, proving the hint was computed from the genuine initial state. The hint output packs `v_mid[i] | (v_init[i] << 32)` per word (same `shr(shl(…, 32), 32)` extraction pattern as BINIUS-130).

## Gate-count estimate

| Function | G calls | Msg packing | Hint overhead | Total (est.) |
|---|---|---|---|---|
| `blake3_compress` | 7 × 8 = 56 | 0 | — | ~56G + small |
| `blake3_compress_opt` | 4 × 8 = 32 | 16 BOR+SHL × 4 steps | SHL×16 + binding×16 | ~32G + ~96 ops |

The 32 G calls in `blake3_compress_opt` run in the parallel-halves context, so each handles both the "first-half" and "second-half" computation simultaneously. Net G-operation reduction vs `blake3_compress`: **~43%** (32 vs 56 G calls), partially offset by message packing and binding overhead of ~96 wire operations.

## Verification

```
cargo test -p binius-circuits blake3   ✅  (15/15, including 3 new compress_opt tests)
cargo clippy -p binius-circuits -- -D warnings  ✅
cargo +nightly-2026-01-01 fmt --check  ✅
```

The three new tests (`compress_opt_zero_block`, `compress_opt_nontrivial`, `compress_opt_counter_across_32bit_boundary`) exercise `populate_wire_witness` end-to-end (hint binding included) and assert equality against the pure-Rust reference for each test vector.

### DIFF
--- a/crates/circuits/src/blake3/compress.rs
+++ b/crates/circuits/src/blake3/compress.rs
@@ -12,7 +12,7 @@
 use std::array;
 
 use binius_core::word::Word;
-use binius_frontend::{CircuitBuilder, Wire};
+use binius_frontend::{CircuitBuilder, Hint, Wire};
 
 use super::{IV, MSG_SCHEDULE};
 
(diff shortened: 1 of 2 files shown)


jimpo:
The 2x_seq variant for compression is good enough.

PR #1566, opened 2026-06-23
BINIUS-133: rewrite sha256 Compress struct as a function gadget

benji-potamus:
Rewrites the SHA-256 `Compress` struct as a free-function gadget, mirroring the established `blake3_compress` pattern ([BINIUS-133](https://linear.app/jimpo/issue/BINIUS-133/rewrite-compress-struct-in-sha2), sub-issue of BINIUS-126).

## What

- **`Compress` struct + `impl` → `sha256_compress(builder, state_in, m) -> State`.** A plain function that runs the message schedule + 64 rounds and returns the updated state, instead of a struct that stored `state_in` / `state_out` / `m`.
- **`Compress::populate_m` → free function `populate_message_schedule(w, &m, bytes)`.** The method only existed to reach the `m` wires the struct held onto; as a free function it takes those wires explicitly, so nothing needs to be retained just to fill the witness.
- **Call sites updated:**
  - `Sha256` now stores `block_messages: Vec<[Wire; 16]>` (the per-block message wires) instead of `compress: Vec<Compress>`, and fills them with `populate_message_schedule` during witness population.
  - `sha256_fixed` calls `sha256_compress` directly and returns its state.
  - The three `compress.rs` unit tests and `prover/tests/prove_verify.rs` updated to the function API.

## Why

`blake3` already exposes its compression as a free function (`blake3_compress`) with no wrapping struct; this brings `sha256` in line. The struct's only role beyond computing the output state was holding `m` for later witness population, which the free `populate_message_schedule` covers without retaining a per-block object.

Behavior-preserving — the constraint construction is identical; only the API shape changes.

## Verification

```
cargo test -p binius-circuits sha256                       ✅ 33 passed
cargo test -p binius-prover --test prove_verify sha256     ✅ 2 passed
prek run --stage pre-push --all-files clippy               ✅ Passed
prek run --all-files rustfmt                               ✅ Passed
```

### DIFF
--- a/crates/circuits/src/sha256/compress.rs
+++ b/crates/circuits/src/sha256/compress.rs
@@ -1,6 +1,6 @@
 // Copyright 2025 Irreducible Inc.
 use binius_core::word::Word;
-use binius_frontend::{CircuitBuilder, Wire};
+use binius_frontend::{CircuitBuilder, Wire, WitnessFiller};
 
 const IV: [u32; 8] = [
 	0x6a09e667, 0xbb67ae85, 0x3c6ef372, 0xa54ff53a, 0x510e527f, 0x9b05688c, 0x1f83d9ab, 0x5be0cd19,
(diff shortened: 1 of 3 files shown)


crates/circuits/src/sha256/compress.rs
```diff
@@ -58,81 +58,81 @@ impl State {
...
+}
 
-			// Write it to the witness.  Word is a thin wrapper around u64.
-			w[self.m[i]] = Word(limb);
-		}
+/// Populates the 16 message-schedule wires of a [`sha256_compress`] block from 64 message bytes.
+///
+/// The bytes are packed as 16 big-endian 32-bit words, one per wire, with the high 32 bits left
+/// zero — matching the precondition on `m` documented in [`sha256_compress`].
+pub fn populate_message_schedule(w: &mut WitnessFiller, m: &[Wire; 16], bytes: [u8; 64]) {
```
jimpo: This is the 64-byte message, not the message schedule

crates/circuits/src/sha256/compress.rs
```diff
@@ -58,81 +58,81 @@ impl State {
...
+	for i in 0..16 {
```
jimpo: iterate .chunks of bytes, convert to [u8; 4] with try_from, then u32::from_be_bytes

jimpo (changes requested):

crates/circuits/src/sha256/compress.rs
benji-potamus: Good catch — these 16 wires are the message block (W[0..15]), not the expanded 64-word schedule. Renamed the helper to `populate_message_block` and fixed its doc, plus the `block_messages` field doc in `mod.rs` that had the same "message-schedule" misnomer. Pushed.

crates/circuits/src/sha256/compress.rs
benji-potamus: Done — now `for (wire, chunk) in m.iter().zip(bytes.chunks_exact(4))` with `u32::from_be_bytes(chunk.try_into().unwrap())`. Pushed.

jimpo (approved):

PR #1567, opened 2026-06-24
BINIUS-134: add sha256_compress_2x parallel two-lane compression gadget

benji-potamus:
Adds `sha256_compress_2x`, a SHA-256 compression gadget that runs **two independent compressions in parallel** by packing two 32-bit lanes into each 64-bit wire — mirroring the established `blake3_compress_2x` pattern ([BINIUS-134](https://linear.app/jimpo/issue/BINIUS-134/write-sha256-compress-2x-circuit-gadget), sub-issue of BINIUS-126). Stacked on #1566 (now merged).

## What

- **`sha256_compress_2x(builder, state_in, m) -> State`.** Each 64-bit input wire packs two 32-bit lanes: bits `[0:32]` are lane 0, bits `[32:64]` are lane 1. SHA-256's mixing is built entirely from parallel-halves gates (`iadd_32`, `rotr32`, `srl32`) plus lane-agnostic `bxor`/`band`, so the message schedule and all 64 rounds run **both** compressions at the gate cost of a single one.
- **Round constants are the only lane-specific detail** — they are replicated into both 32-bit halves (`kt | (kt << 32)`) so each lane adds the same `K[t]`.
- **Shared core extracted as `compress_inner`.** `sha256_compress` and `sha256_compress_2x` now differ only in how they materialize the round-constant wires (low 32 bits vs. both halves); everything else flows through the common core. `round` was refactored to take the round constant and message word as explicit `Wire` arguments rather than indexing `K`/`w` internally.
- **Tests:** a pure-Rust `ref_compress` reference plus two-lane tests that pack distinct lane-0/lane-1 inputs and check both lanes decode correctly.

## Performance

The whole point of the gadget: two compressions at the AND-constraint cost of one. AND constraints dominate the (power-of-two-padded) proving cost.

| | AND constraints |
|---|---|
| `2× sha256_compress` (two separate compressions) | 1824 |
| `1× sha256_compress_2x` (two compressions, two-lane) | **912** |

→ **~2× throughput / 50% fewer AND constraints** for paired compressions.

## Verification

```
cargo test -p binius-circuits sha256                       ✅ 35 passed
prek run --stage pre-push --all-files clippy               ✅ Passed
prek run --all-files rustfmt                               ✅ Passed
```

### DIFF
--- a/crates/circuits/src/sha256/compress.rs
+++ b/crates/circuits/src/sha256/compress.rs
@@ -77,6 +77,54 @@ impl State {
 ///
 /// The updated 8-word state.
 pub fn sha256_compress(builder: &CircuitBuilder, state_in: State, m: [Wire; 16]) -> State {
+	// Round constants live in the low 32 bits; the high 32 bits stay zero, matching the
+	// single-lane convention that every wire's high half is empty.
+	let k: [Wire; 64] = std::array::from_fn(|t| builder.add_constant(Word(K[t] as u64)));
+	compress_inner(builder, state_in, m, &k)
+}
+
+/// SHA-256 compression function running two independent compressions in parallel.
+///
+/// Each 64-bit input wire packs two 32-bit lanes: bits `[0:32]` hold the lane-0 word, bits
+/// `[32:64]` hold the lane-1 word. SHA-256's mixing is built entirely from the parallel-halves
+/// gates ([`iadd_32`](CircuitBuilder::iadd_32), [`rotr32`](CircuitBuilder::rotr32),
+/// [`srl32`](CircuitBuilder::srl32)) plus lane-agnostic `bxor`/`band`, so the schedule and all 64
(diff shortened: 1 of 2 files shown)


benji-potamus:
**Perf methodology** (for the speedup numbers in the description): the metric is **AND-constraint count**, obtained statically from the circuit (not a wall-clock benchmark), since for these gadgets the power-of-two-padded AND-constraint count is the dominant proving cost and is fully deterministic — so there's no machine/iteration/noise to report.

- `sha256_compress` alone: 912 AND constraints → two separate calls = **1824**.
- `sha256_compress_2x`: **912** AND constraints for the same two compressions.

The two-lane gadget does both compressions within one set of parallel-halves gates, so the AND cost is identical to a single compression — a clean ~2× / 50% reduction for paired work. End-to-end prove-time impact depends on how many SHA-256 compressions a circuit can pair up.

jimpo (approved):

PR #1569, opened 2026-06-24
BINIUS-135: blanket B1 packed arithmetic impls + export M512

benji-potamus:
Part of [BINIUS-135](https://linear.app/jimpo/issue/BINIUS-135/lift-b1-packed-field-definitions-into-binius-fieldpacked-binary-field).

Every `BinaryField1b` packing shares the same arithmetic — addition is bitwise XOR (already blanket for every `PackedPrimitiveType` in `packed.rs`), multiplication is bitwise AND, and squaring/inversion are the identity. This replaces the per-type `Mul`/`Square`/`InvertOrZero`/`WideMul` impls that the `define_packed_binary_field` macro generated for each B1 type with a **single blanket impl over `PackedPrimitiveType<U, BinaryField1b>`** in `packed_binary_field`.

### Changes
- Blanket `Mul` (bitwise AND) / `Square` / `InvertOrZero` (identity) / `WideMul` (trivial) for all `PackedPrimitiveType<U, BinaryField1b>`.
- Each B1 `packed_*` module now holds just a `pub type` alias instead of a macro invocation (net −190 LOC).
- Removed `BitwiseAndStrategy` + `bitwise_and_arithmetic.rs` — they were B1-only and now dead.
- **`M512` is exported from `binius_field::arch`** in every arch config (previously only reachable as `arch::packed_512::M512`, and not at all on x86_64 `avx512f` where `m512` was `pub(super)`).

### Verification
- `binius-field` lib tests (238) green; workspace `cargo clippy --all --tests --benches --examples -- -D warnings` green; fmt clean.
- Cross-compiled: x86_64 native / `+avx512f` / sse2-baseline, aarch64 (`+neon,+aes`), wasm32 — all build.

### Deferred — needs your call on structure
The M128/M256/M512 B1 aliases are still defined in **arch-specific** `packed_128/256/512` modules (now as plain `pub type`), not consolidated into one `arch/mod.rs` definition as the issue envisions ("create type aliases for the versions over M128/M256/M512 … exported from arch"). I held off because consolidating entangles with where the **`M256`/`M512` underlier aliases and `m256_from_u128s`** live — they currently sit *inside* the `packed_256`/`packed_512` modules (and `m256_from_u128s` is used by `ghash_sq.rs`). Collapsing the per-arch alias files cleanly probably wants those underlier aliases moved to `m256.rs`/`m512.rs` (mirroring `m128.rs`) first. Two questions before I do the consolidation as a follow-up: (1) move `M256`/`M512` + `m256_from_u128s` into dedicated `m256`/`m512` modules? (2) lift the small `1x1b`..`64x1b` aliases too, or leave them in the portable modules (they're arch-independent)?


crates/field/src/arch/portable/packed_1.rs
```diff
@@ -1,23 +1,6 @@
...
+// Copyright 2026 The Binius Developers
 
-use crate::{
-	arch::{
-		BitwiseAndStrategy,
-		portable::packed_macros::{portable_macros::*, *},
-	},
-	underlier::U1,
-};
+use crate::{BinaryField1b, arch::PackedPrimitiveType, underlier::U1};
```
jimpo: Do all  of these type aliases in the single `packed_binary_field.rs` file.

jimpo (changes requested):

crates/field/src/arch/portable/packed_1.rs
benji-potamus: Done — all B1 type aliases (`1x1b`..`512x1b`) now live in `packed_binary_field.rs`. The per-arch `packed_{1..512}` B1 modules are deleted. Per your other note, the underlier types are lifted so each arch module (`portable`, `x86_64`, `aarch64`) exports `M128`/`M256`/`M512` and `arch/mod.rs` re-exports the active set; `M256`/`M512` + `m256_from_u128s` moved out of the old `packed_256`/`packed_512` modules into dedicated `portable/m256.rs` + `portable/m512.rs`. Force-pushed.

jimpo (approved):

PR #1573, opened 2026-06-24
BINIUS-43: replace ReplayChannel with a recorded gate sequence

benji-potamus:
Closes [BINIUS-43](https://linear.app/jimpo/issue/BINIUS-43/replace-replaychannel-with-recorded-gate-sequence-for-witness-gen). Fresh take on the post-BINIUS-99/103 foundation, replacing PR #1524.

## What

The ZK wrapper's prover re-executed the inner IOP verifier (via `ReplayChannel`) purely to populate the outer witness — re-deriving values already fixed by the symbolic build of the outer constraint system.

Record the inner verifier's derived arithmetic **once**, during the symbolic build, as a flat `GateSequence` (`Add`/`Mul`/`Generic`), then replay it on the prover to fill the witness. **`ReplayChannel` is deleted.** The verifier is unchanged — it keeps reconstructing the public inputs via `InstanceGenerator`; the gate sequence is prover-only.

## Commits (3)

1. **`require 'static on CircuitBuilder::hint`** — so a hint closure can be recorded into a replayable gate.
2. **`add CircuitBuilder::hint_varsize`** — a hint whose output arity is a runtime `out_len` (closure `&[F] -> Vec<F>`). `compute_public_value` is a one-output `hint_varsize`.
3. **the bulk** — gate recording, prover replay, `CircuitBuilderWithAlloc`, the verifier wiring.

## How

- **`spartan-frontend/gate.rs`** (new): `Gate` (`Add`/`Mul`/`Generic`) + `GateSequence`. `GateRecordingConstraintBuilder` records the sequence while delegating wire allocation to an inner `ConstraintBuilder`, so recorded ids match the layout. `add`/`mul`/`hint`/`hint_varsize` each record a gate; the inner builder's closure is ignored, so they pass it a no-op and keep the real `f` for the gate's eval.
- **`WitnessGenerator::apply_gates`**: replays the gates over a partially-filled witness, computing the derived and private wires (a side table holds wire-eliminated intermediates).
- **`compute_public_value`** is now a one-output `hint_varsize` rather than a bespoke channel method — it emits no constraints, and its output is a derived public wire when the inputs are public, so it sorts after the transcript inout wires (keeping the prover channel's sequential inout allocation aligned). It's sound regardless: the verifier recomputes the value via `InstanceGenerator`. The closure is `Fn(&[F]) -> F + 'static` so the recording backend can box it for replay.
- **Prover** — `ZKWrappedProverChannel` allocates the inout (transcript) and precommit (OTP key) wires on its own internal `WitnessGenerator` and writes their values directly as the inner proof runs; `finish` then `apply_gates` to fill the rest and proves the outer CS. No interaction buffer.
- **`CircuitBuilderWithAlloc` trait** (only `alloc_inout` + `alloc_precommit`) makes `IronSpartanBuilderChannel` generic over its backend; `new` takes the builder by value (`ConstraintBuilder::new()` for the verifier setup, `GateRecordingConstraintBuilder::new()` for the prover setup).

## Validation

End-to-end green (each commit builds + tests; plus clippy, fmt, rustdoc on the tip):
- `wrapper_integration_test` (pow7 wrapper, prove + verify),
- full `prove_verify` suite (7/7), incl. `test_zk_prove_verify_sha256_preimage` — the real binius64 path through `monster_eval` / oracle relations / transparent closures — and the serialized round-trip,
- gate-sequence unit tests: `apply_gates` is bit-identical to the direct `WitnessGenerator` path.


jimpo:
This looks overcomplicated still. And I changed my mind about using the gate sequence on the ZKWrappedVerifier side --- revert to the InstanceGenerator on the verifier side and use the gates only on the prover side.

You don't need InOut and Precommit gate types. The ZKWrapped prover channels can allocate those wires on its internal allocators the way it used to, without first storing them to an interaction. The gates are just meant to substitute in for the "verify" logic on the inner Verifier, which happens in ZKWrappedProverChannel::finish

benji-potamus:
Got it — reverting the verifier side and slimming the gates. Plan:

1. **Verifier:** restore `ZKWrappedVerifierChannel` to the `InstanceGenerator` path; delete the option-B `replay_public` / `OpaqueBuilder` machinery. Gates live on the prover side only.
2. **Gates:** drop the `Input` / `Precommit` variants — `GateSequence` keeps only `Add` / `Mul` / `Generic`, i.e. just the inner verifier's "verify" arithmetic.
3. **Prover:** `ZKWrappedProverChannel` allocates the inout/precommit wires on its own internal allocators (a `WitnessGenerator`) and writes each sent/sampled/observed value — and the OTP keys — straight in as it goes, in allocation order, so the ids line up with the recorded derived gates. No `interaction` vector. `finish` then just replays the `Add`/`Mul`/`Generic` gates over that partially-filled witness to fill the derived wires and proves the outer CS.

That removes the whole verifier-side reconstruction and the interaction-stream indirection. Will push the rework on this branch. Shout if (3) isn't the allocator wiring you had in mind.

benji-potamus:
Started the rework and hit a real snag on (3) — want your call before I build it.

`compute_public_value` allocates an **InOut** wire, and it has to stay InOut (it's a verifier-recomputed public *instance* value, not something any constraint derives). In the spartan `IOPVerifier::verify` it isn't the last inout-allocating op: `lib.rs:204` is the `compute_public_value`, but `:218 recv_one` and `:234 verify_oracle_relations` allocate more inout wires **after** it.

So if `ZKWrappedProverChannel` allocates inout wires inline only for the prover's transcript ops (it never runs `compute_public`), every inout id after the monster-eval is shifted by one vs the symbolic build, and the recorded `Add`/`Mul`/`Generic` gates resolve to the wrong wires. (Precommit ids are fine — `compute_public` allocates no precommit.) The old `ReplayChannel` only dodged this by re-running the verifier live, so it allocated the `compute_public` inout wire inline at the right point — which is exactly the live re-run the gates are meant to replace.

How do you want the channel to keep inout ids aligned across that gap? Two options I see:

1. **Keep a minimal allocation-order record** — essentially today's `Input`/`Precommit` markers, but *only* the wire ids, no stored values. `finish` walks them to allocate inout/precommit wires in the correct interleaved order while the channel writes the transcript values straight into the witness as they arrive — so the `interaction` Vec you flagged goes away, but a slim per-external-input gate stays.
2. **Make `compute_public` sort last** in the inout segment (allocate its wires after all transcript inout, or give it its own trailing index range), so the channel's plain sequential allocation lines up and no markers are needed.

I'd lean (2) if compute_public can cleanly be made to sort last — it keeps the gate sequence purely `Add`/`Mul`/`Generic`. But it touches how inout ids are assigned. Which way do you want, or is there a framing I'm missing?

jimpo:
compute_public_value shouldn't make an InOut wire, it should make a derived public wire, right? At that's handled by a gate on the prover side, so I don't see how it could get out of sync.

benji-potamus:
Yes — that's the clean fix, thanks. A derived public wire sorts after all inout in the public segment, so the inout id-space stays transcript-only and the prover channel's plain sequential allocation lines up; the `Generic` gate fills the derived wire in `finish`. (And it's sound regardless — the verifier recomputes the value via `InstanceGenerator`, so the wire kind only changes layout position.) Reworking with `compute_public` → derived public wire, gates purely `Add`/`Mul`/`Generic`, verifier back on `InstanceGenerator`. Will push.

benji-potamus:
Reworked and pushed (rebased onto latest main). Summary of the change vs the previous version:

- **Verifier reverted** to the `InstanceGenerator` path — the `replay_public` / `OpaqueBuilder` machinery is gone; `ZKWrappedVerifierChannel` and `verifier/zk_config.rs` are back to their pre-PR form. Gates are prover-only.
- **Gates slimmed** to `Add` / `Mul` / `Generic` (no `Input` / `Precommit` variants).
- **`compute_public_value` → derived public wire** (per your note) — that keeps the inout id space transcript-only, so the prover channel's sequential inout/precommit allocation lines up with the symbolic build and the interleaving concern goes away.
- **`ZKWrappedProverChannel`** now writes each transcript value + OTP key straight into its own `WitnessGenerator` as the inner proof runs; `finish` replays the derived gates via `WitnessGenerator::apply_gates`. No interaction `Vec`.

Green locally: `wrapper_integration_test`, the full `prove_verify` suite (incl. `test_zk_prove_verify_sha256_preimage` and the serialized round-trip), plus the gate unit tests, clippy, fmt, and rustdoc. CI running.

crates/spartan-frontend/src/gate.rs
```diff
@@ -0,0 +1,376 @@
...
+		const IN: usize,
+		const OUT: usize,
+	>(
+		&mut self,
+		inputs: [Self::Wire; IN],
+		f: H,
+	) -> [Self::Wire; OUT] {
+		// Allocate the outputs via the inner builder's closure-free path, keeping `f` to box into
+		// the recorded gate's eval.
+		let outputs = self.inner.alloc_hint(inputs);
```
jimpo: alloc_hint seems unnecessary for the WrapperBuilder trait. `self.inner.hint(...)` should work, except that it consumes f. So instead, call `self.inner.hint(...)` with a no-op closure, because we know it gets ignored by `ConstraintBuilder` anyway.

crates/spartan-frontend/src/circuit_builder.rs
```diff
@@ -39,13 +42,48 @@ pub trait CircuitBuilder {
...
+/// precommit wires (transcript inputs and one-time-pad keys) and `compute_public_value` results.
+///
+/// The wrapper builder channel
+/// ([`IronSpartanBuilderChannel`](https://docs.rs/binius-spartan-verifier)) is generic over this
+/// trait, so the same channel either builds only the constraint system ([`ConstraintBuilder`]) or
+/// also records a replayable [`GateSequence`]
+/// ([`RecordingBuilder`](crate::gate::RecordingBuilder)).
+pub trait WrapperBuilder: CircuitBuilder + Sized {
+	/// Constructs a fresh builder.
+	fn fresh() -> Self;
```
jimpo: Get rid of the constructor on trait. IronSpartanBuilderChannel::new can take the builder instance by value. (So call it like IronSpartanBuilderChannel::new(ConstraintBuilder::new()).

crates/spartan-frontend/src/circuit_builder.rs
```diff
@@ -39,13 +42,48 @@ pub trait CircuitBuilder {
...
+	fn channel_inout(&mut self) -> Self::Wire;
+
+	/// Allocates a precommit wire (a one-time-pad key).
+	fn channel_precommit(&mut self) -> Self::Wire;
+
+	/// Allocates a derived public wire holding `f(inputs)`. Recording builders capture `f` for
+	/// replay; symbolic builders ignore it (the value is supplied at instance/replay time, by the
+	/// verifier's `InstanceGenerator` or a recorded gate). The wire is derived, not inout, so it
+	/// sorts after the transcript inout wires. See the channel's `compute_public_value`.
+	fn channel_compute_public<G: Fn(&[Self::Field]) -> Self::Field + 'static>(
```
jimpo: This is weird. It seems like what we really want is to add a method `hint_varsize` on the `CircuitBuilder` trait, which takes a hint with a closure that's a slice and returns a Vec, a the method also takes a parameter for the size of the Vec that will get returned. Do that as a commit before the other changes in this PR. (so multi-commit PR)

jimpo (changes requested):
Getting there, but I think the WrappedBuilder trait is unnatural. Also, the code is hard to review in one big commit. Split it into at least 3 commits: one for the small change to the hint signature (requiring 'static), one to add hint_varsize, and one for the bulk of the changes (though if you can split that up naturally too, even better).

After the inline comments, WrappedBuilder should have only two methods: alloc_inout and alloc_precommit. At that point, rename it to CircuitBuilderWithAlloc that has only those two methods, and implement it on ConstraintBuilder and GateRecordingConstraintBuilder (rename from RecordingBuilder).

crates/spartan-frontend/src/gate.rs
benji-potamus: Done. `GateRecordingConstraintBuilder::hint` (renamed from `RecordingBuilder`) now calls `self.inner.hint(inputs, |_| [F::ZERO; OUT])` and keeps the real `f` for the recorded gate's eval — `alloc_hint` is gone. `hint_varsize` does the same with `|_: &[F]| Vec::new()`.

crates/spartan-frontend/src/circuit_builder.rs
benji-potamus: Done. The trait constructor is gone; `IronSpartanBuilderChannel::new` now takes the builder by value — call sites are `IronSpartanBuilderChannel::new(ConstraintBuilder::new())` (verifier setup) and `IronSpartanBuilderChannel::new(GateRecordingConstraintBuilder::new())` (prover setup). The trait is renamed `CircuitBuilderWithAlloc` with only `alloc_inout` + `alloc_precommit`.

crates/spartan-frontend/src/circuit_builder.rs
benji-potamus: Done. Added `hint_varsize(&mut self, inputs: &[Wire], out_len: usize, f: impl Fn(&[F]) -> Vec<F> + 'static) -> Vec<Wire>` to `CircuitBuilder` as its own commit (the 2nd of 3). `channel_compute_public` is removed; `compute_public_value` is now a one-output `hint_varsize` (`builder.hint_varsize(&inputs, 1, move |vals| vec![f(vals)])`). It emits no constraints and its outputs are derived when the inputs are public, so a `compute_public_value` over public inputs lands in the public segment as before.

benji-potamus:
Addressed the review — replies inline. Summary:

- **`hint_varsize`** added to `CircuitBuilder` as its own commit; `channel_compute_public` removed and `compute_public_value` is now a one-output `hint_varsize`.
- **`WrapperBuilder` → `CircuitBuilderWithAlloc`**, only `alloc_inout` + `alloc_precommit`; trait constructor gone, `IronSpartanBuilderChannel::new` takes the builder by value.
- **`RecordingBuilder` → `GateRecordingConstraintBuilder`**; its `hint`/`hint_varsize` call `inner.hint*` with a no-op closure (no more `alloc_hint`).
- **Split into 3 commits**: `hint 'static` → `add hint_varsize` → bulk. Each commit builds + tests green; rebased onto latest `main`.

CI running.

PR #1577, opened 2026-06-24
Add binius-m4 crate with batch witness ValueTable

tcoratger:
## Summary

First piece of the **Binius64 M4** (data-parallel) proof system: the batch witness type. No proving or commitment yet — this is the witness layer only.

`ValueTable` holds `K = 2^k` independent instances of a single circuit, stored back-to-back in one flat buffer in **instance-major** order (instance index in the high-order positions, within-instance word index in the low-order positions). This is the layout a later batch commitment reads directly as a multilinear over `k + m_0` variables.

## What's here

- **New crate `binius-m4`** (`crates/m4/`) with `ValueTable`:
  - `populate` runs the existing single-instance witness generation **per instance in parallel**, each instance writing its own disjoint buffer chunk.
  - Accessors for a per-instance slice, a reconstructed single-instance `ValueVec` (for constraint checking), and the flat instance-major buffer.
  - Strongly-typed failures: `PopulateInstanceError` pairs the failing instance index with the wrapped frontend assertion error, and implements `Error::source()`.
- **`binius-frontend`**: re-export `PopulateError` (was `pub` but trapped in a private module) so the wrapped error type is nameable downstream.

## Design notes

- **Instance-major = instance-high / wire-low** layout. Implemented and documented here; validating it against the commitment is left to a follow-up.
- **No inter-instance padding**: stride is exactly one instance's committed-word count. Padding each instance to a power of two is a commitment-layout concern for the follow-up, deliberately not baked in here.
- `log_instances` (not `n_instances`) as the parameter makes a non-power-of-two batch unrepresentable.

## Tests

`cargo test -p binius-m4` (5 tests): per-instance equivalence to the single-instance flow (incl. a proptest), flat-buffer-equals-concatenation, the degenerate single-instance batch, and an unsatisfiable instance reporting its index with the exact wrapped assertion failure. fmt- and clippy-clean.


crates/m4/src/value_table.rs
```diff
@@ -0,0 +1,422 @@
...
+		//      \_ stride _/
+		let mut data = vec![Word::ZERO; n_instances * stride];
+
+		// Populate each instance into its own slice of the buffer, concurrently.
+		// The chunks are disjoint, so the instances never contend for the same words.
+		data.par_chunks_mut(stride)
+			.enumerate()
+			.try_for_each(|(instance, chunk)| {
+				// A fresh witness vector for this instance, including transient scratch space.
+				let mut filler = circuit.new_witness_filler();
```
jimpo, line 98: Noting that this does an allocation.

You could use [`for_each_init`](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.for_each_init) to keep the same WitnessFiller between iterations.

crates/m4/src/value_table.rs
```diff
@@ -0,0 +1,422 @@
...
+				// The caller assigns this instance's input wires.
+				// Different instances generally receive different inputs.
+				fill(instance, &mut filler);
+
+				// Evaluate the circuit gate by gate to derive the remaining committed values.
+				// A failed assertion means this instance's inputs do not satisfy the circuit.
+				// Record which instance failed.
+				circuit
+					.populate_wire_witness(&mut filler)
+					.map_err(|source| PopulateInstanceError { instance, source })?;
```
jimpo, line 109: This could be an assert. I'd changed the convention in the codebase to more asserts/panics and less error throwing. See [Error Handling](https://github.com/binius-zk/binius64/blob/main/CONTRIBUTING.md#error-handling). But that's not consistently changed everywhere yet.

Your call.

jimpo (approved):
Sweet!

My main concern is whether we should have the table in this memory order or the transposed (wire-major) order. That's what I'd originally written in the ticket.

This is clearly easier for witness generation though. I'd say let's keep it instance-major for now, and then once we finish more of the prover we can revisit.

Inline comments can be handled as follow-ups.

crates/m4/src/value_table.rs
tcoratger: Addressed in #1579 — switched to `for_each_init` so each worker thread reuses one `WitnessFiller` instead of allocating per instance.

PR #1578, opened 2026-06-24
BINIUS-140: use blake3_compress_2x_seq in blake3_fixed (~42% fewer AND gates)

benji-potamus:
## Summary

[BINIUS-140](https://linear.app/jimpo/issue/BINIUS-140/fixed-length-blake3-hash-gadget-for-length-1024-bytes) asks for the single-chunk `blake3_fixed` gadget (≤1024 bytes) to use `blake3_compress_2x_seq` for efficiency. The gadget already existed on `main` (the `sha256_fixed`-style signature was already in place) but compressed blocks **sequentially, one `blake3_compress` per 64-byte block** (up to 16). This PR pairs adjacent block compressions onto the shared 2× parallel core, roughly halving the dominant compression cost.

Two cohesive commits:

1. **Share the block counter across `blake3_compress_2x_seq` compressions.** The primitive hardcoded `counter2 = counter + 1` for the second compression. That is wrong for chaining two blocks *within a single chunk*: BLAKE3 blocks in a chunk all carry the **same** chunk counter (the existing `blake3_fixed` correctly passes `counter = 0` to every block). Sequential CV-chaining (`C2 = compress(C1, …)`) only ever happens within a chunk, so the counter is now shared by both compressions rather than incremented. No non-test callers existed, so the API change is contained to this commit.

2. **Pair sequential block compressions in `blake3_fixed`.** A single loop of `ceil(n_blocks / 2)` iterations, each calling `blake3_compress_2x_seq`. Per-block message words / lengths / flags are built as vectors and padded with one unused dummy block when `n_blocks` is odd, so every iteration pairs uniformly. The running chaining value is the second compression's output (low 32 bits of each word); for a trailing odd block the second lane is the dummy, so the result is taken from the first compression's output (high 32 bits) instead. Behavior is unchanged — verified against `blake3::hash`.

## Performance (AND constraints, the dominant gate; MUL is 0 throughout)

| message len | blocks | before | after | Δ |
|---|---|---|---|---|
| 64 B | 1 | 634 | 681 | +7.4% |
| 128 B | 2 | 1220 | 689 | −43.5% |
| 512 B | 8 | 4742 | 2756 | −41.9% |
| 1024 B | 16 | 9442 | 5512 | −41.6% |

Multi-block messages drop ~42%. A single-block message (≤64 B) costs ~47 more AND constraints, because it now runs `2x_seq` with a dummy second lane rather than a lone `blake3_compress` — the cost of the uniform single-loop structure.

## Testing

- `cargo test -p binius-circuits` — all 351 pass.
- `blake3_fixed` `block_boundaries` extended to cover odd block counts (3, 5, 7 blocks) that exercise the trailing-odd-block path.
- `blake3_compress_2x_seq` tests updated to the shared-counter semantics.
- `cargo +nightly-2026-01-01 fmt --check` and full-workspace `cargo clippy --all --tests --benches --examples -- -D warnings` clean.

### DIFF
--- a/crates/circuits/src/blake3/mod.rs
+++ b/crates/circuits/src/blake3/mod.rs
@@ -16,6 +16,8 @@
 use binius_core::word::Word;
 use binius_frontend::{CircuitBuilder, Wire};
 
+use crate::util::clear_high_bits;
+
 pub mod compress;
 
 pub use compress::{blake3_compress, blake3_compress_2x, blake3_compress_2x_seq};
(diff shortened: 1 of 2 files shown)


crates/circuits/src/blake3/mod.rs
```diff
@@ -115,40 +117,62 @@ pub fn blake3_fixed(builder: &CircuitBuilder, message: &[Wire], len_bytes: usize
...
+			counter,
+			[block_len_at(block_idx), block_len_at(block_idx + 1)],
+			[flags_at(block_idx), flags_at(block_idx + 1)],
+		);
+		// The chaining value after both blocks is the second compression's output, packed into the
+		// low 32 bits of each word; extract it as the running chaining value.
+		cv = std::array::from_fn(|i| clear_high_bits(builder, out[i], 32));
+		block_idx += 2;
+	}
+	if block_idx < n_blocks {
```
jimpo: Simplify this by just calling blake3_compress_2x_seq for the last one. That is, do just one loop, with ceil(n_blocks / 2) iterations. Then, depending whether n_blocks is even or odd, choose the high or low lanes of the output wires. Then a lot of the closure stuff you had to introduce here will go away.

jimpo (changes requested):

crates/circuits/src/blake3/mod.rs
benji-potamus: Done in fbe9ed6. Now a single loop of `ceil(n_blocks / 2)` iterations, always calling `blake3_compress_2x_seq`. The blocks/lengths/flags are built as per-block vectors and padded with one unused dummy block when `n_blocks` is odd, so every iteration pairs uniformly. The final chaining value is the second compression's output (low 32 bits) for even `n_blocks`, or the first compression's output (high 32 bits) when the last block is the odd one. The three closures and the separate trailing single-`blake3_compress` branch are gone.

One tradeoff to flag: a single-block message (≤64 B) now goes through `2x_seq` with a dummy second lane instead of a lone `blake3_compress`, so it costs ~47 more AND constraints (634 → 681). Multi-block counts are unchanged from before (e.g. 1024 B still 9442 → 5512). Happy to special-case `n_blocks == 1` back to a single `blake3_compress` if you'd rather not pay that, but I read your comment as preferring the uniform loop.

jimpo (approved):

PR #1583, opened 2026-06-25
BINIUS-125: align ValueVec words to a 16-byte boundary to speed up copy

tcoratger:
Closes [BINIUS-125](https://linear.app/jimpo/issue/BINIUS-125/align-words-in-valuevec-to-16-byte-boundary-to-speed-up-copy).

## What

Force the word storage inside `ValueVec` to start on a 16-byte boundary so the frequent bulk copies on the prover's hot path take the aligned SIMD `memcpy` path.

## Why

The value vector is copied constantly — cloned wholesale, packed into field elements, and sliced out to owned `Vec`s.

A `Word` is 8 bytes, so a plain `Vec<Word>` is only 8-byte aligned. Its buffer lands on a 16-byte boundary only about half the time (by luck of the allocator), and the other half every `memcpy` pays a misalignment prologue instead of running the aligned SIMD loop.

## How

- Store words two-at-a-time inside a `#[repr(C, align(16))]` `WordPair` block. A `Vec<WordPair>` therefore always starts on a 16-byte boundary — guaranteed, not by luck — at **zero memory overhead** (two 8-byte words exactly fill the 16-byte block).
- The new buffer derefs to `&[Word]` via a zero-copy `bytemuck` reinterpret, so the public `ValueVec` API is completely unchanged. The section offsets are even, so the inner sub-slices (`public`, `witness`, …) stay 16-byte aligned too.
- `Word` is now `#[repr(transparent)]` and derives `Pod`/`Zeroable` so word slices can be reinterpreted as bytes and back.

## Trade-off

`new_from_data` now copies the public + private words into one fresh aligned buffer instead of reusing the incoming public `Vec`. That's one extra copy at witness-reconstruction setup, traded for aligned copies on the hot path forever after.

## Tests

Added to the core test module:
- alignment + zero-fill + length across empty / odd / even / multi-block lengths,
- mutable-view writes are visible through the shared view without touching the padding word,
- a proptest asserting `new_from_data` preserves both sections byte-for-byte, keeps 16-byte alignment for any word count, and zero-fills the scratch tail.

`cargo test -p binius-core`, `cargo clippy -p binius-core --all-targets`, and `cargo build --workspace` all pass.

### DIFF
--- a/crates/core/src/constraint_system.rs
+++ b/crates/core/src/constraint_system.rs
@@ -3,10 +3,11 @@
 
 use std::{
 	borrow::Cow,
-	ops::{Index, IndexMut},
+	ops::{Deref, DerefMut, Index, IndexMut},
 };
 
 use binius_utils::serialization::{DeserializeBytes, SerializationError, SerializeBytes};
+use bytemuck::{Pod, Zeroable};
 use bytes::{Buf, BufMut};
 
 use crate::{consts, error::ConstraintSystemError, word::Word};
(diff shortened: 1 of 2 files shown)


jimpo (approved):
Looks like a nice easy way to address this problem!

One limitation is that one AVX2 or AVX512 machines, it might be even better to have 32 or 64 byte alignment.

But I'm in favor of just going ahead with this as is.

----

I was thinking of other solutions to this, and I don't think we should do this yet, but just to give some more thoughts:

Another solution is to instantiate ValueVec generically over `BV: Divisible<u64>`. (Side note on planning refactors: `UnderlierType` trait will get renamed to `BitVector` soon).

Then all of the logic touching witness generation would have to use `get/set` on `Divisible` instead of indexing `u64` directly, but it would mean that everything is packed into the type that we'll ultimately use it in.

But let's not do that yet, adding a Compute abstraction will affect the decision anyway.

jimpo:
Needs a rebase please

tcoratger:
> Needs a rebase please

Done

PR #1584, opened 2026-06-25
BINIUS-143: move verify.rs helpers onto their constraint-system types

tcoratger:
Closes [BINIUS-143](https://linear.app/jimpo/issue/BINIUS-143).

`crates/core/src/verify.rs` held two operand-evaluation primitives that are not verification: `eval_shifted_word` and `eval_operand`. They're the hot-path helpers the prover uses to materialize operand images (`crates/prover/src/prove.rs`), yet lived in a module documented as constraint *checking*. This moves each onto the type it belongs to:

- `eval_shifted_word` → `ShiftVariant::apply`
- `eval_operand` → `ValueVec::eval_operand` (+ `ShiftedValueIndex::eval` per term)

Per review, the verification entry points (`verify_and_constraint`, `verify_mul_constraint`, `verify_constraints`) **stay as standalone functions** in `verify.rs`: naive verification is conceptually separate from the constraint-system definition, is never called by a SNARK verifier, and M4 will want to verify a `ValueTable` from another crate without making `ConstraintSystem` aware of tables. They now call `witness.eval_operand(..)` instead of the moved free function.

Verification: `cargo check --workspace --all-targets`, clippy (core + prover), and nightly `rustfmt` all clean; `binius-core` tests and the prover `shift` integration test pass.


jimpo (commented):
Hmm, I think I prefer the separate methods. Except for `ShiftVariant::apply` and `ValueVec::eval_operand`, those are good changes.

It's mostly a matter of taste, but I see the naive verification as separate from the ConstraintSystem definition. (Especially given that naive verification is never even called by a SNARK verifier). Another point is that in M4, we'll need to verify a ValueTable -- and that's in another crate, so  I'd prefer standalone functions for it rather than methods, which make ConstraintSystem aware of tables.

Do you feel strongly?

tcoratger:
> Hmm, I think I prefer the separate methods. Except for `ShiftVariant::apply` and `ValueVec::eval_operand`, those are good changes.
> 
> It's mostly a matter of taste, but I see the naive verification as separate from the ConstraintSystem definition. (Especially given that naive verification is never even called by a SNARK verifier). Another point is that in M4, we'll need to verify a ValueTable -- and that's in another crate, so I'd prefer standalone functions for it rather than methods, which make ConstraintSystem aware of tables.
> 
> Do you feel strongly?

Absolutely no problem, just a proposal that I've seen while reading the code, let's close for now

jimpo:
Can you slim it down to the `ShiftVariant` and `ValueVec` changes? I think those two are good.

tcoratger:
> Can you slim it down to the `ShiftVariant` and `ValueVec` changes? I think those two are good.

Sure, done

tcoratger:
@jimpo Just rebased and reopen the Linear issue for tracking

jimpo (approved):

jimpo:
Sorry for the delay. Approved, needs rebase onto main.

tcoratger:
> Sorry for the delay. Approved, needs rebase onto main.

Sure, done

PR #1585, opened 2026-06-25
BINIUS-83: replace IPVerifierChannel::compute_public_value with a general compute

tcoratger:
Closes [BINIUS-83](https://linear.app/jimpo/issue/BINIUS-83).

`IPVerifierChannel::compute_public_value` ran a single native closure `FnOnce(&[F]) -> F` and returned one `Elem`. That shape was rigid: tied to one concrete field, single-output, and both call sites had to expand their challenge tensors *outside* the channel and pass the whole expansion through.

This replaces it with a `compute` method backed by a new `FieldFn` trait:

```rust
// binius_field::util
pub trait FieldFn<G: Field> {
    fn n_outputs(&self) -> usize;
    fn call<F: FieldOps<Scalar = G> + From<G>>(&self, inputs: &[F]) -> Vec<F>;
}

fn compute(&mut self, inputs: &[Self::Elem], f: impl FieldFn<F>) -> Vec<Self::Elem>;
```

The function is generic in the field it runs in, multi-output, and exposes its output arity so the symbolic builder channel can lay out result wires without running it. The two call sites become `FieldFn` structs that expand their eq-indicator tensors internally, so callers pass the raw challenge (`r_x`, `r_x'`, `r_y`), not its `2^n`-element expansion.

- `compute_public_value` -> `compute` (all 8 channel implementors)
- spartan `public_eval` closure -> `wiring::PublicWiringEvalFn`
- shift `monster_eval` closure -> `MonsterEvalFn`

Scope: the field crate (`FieldFn`), the channel trait + 8 implementors, and the two call sites in `binius-spartan-verifier` and `binius-verifier`. `binius-spartan-prover`'s own local `eval_operand` is unrelated and untouched. While here, the single-use `extract_public_values` helpers in the replay and zk-wrapped channels were inlined.

Two deviations from the issue sketch, both deliberate:
- The trait is parameterized by the scalar field `G` (`FieldFn<G>`), not the bare `FieldFn`. The shift site needs `FieldOps<Scalar = G> + From<G>` to embed subspace / constraint-system elements, which a bare `FieldFn<F: FieldOps>` can't express.
- The `+ 'static` bound is dropped. No implementor stores the function, and `'static` would force the `FieldFn` structs to clone the constraint data instead of borrowing it.

Verification: `cargo check --workspace --all-targets`, clippy on the six touched crates, and nightly `rustfmt` all clean; tests pass for field, ip, iop, spartan-verifier, verifier (incl. shift), and the spartan-prover wrapper integration suite (symbolic-builder / replay / zk-wrapped circuit consistency).


jimpo:
I'll take a look after https://github.com/binius-zk/binius64/pull/1573 lands -- that's a big important change that conflicts.

crates/field/src/util.rs
```diff
@@ -4,6 +4,23 @@ use std::iter;
...
 use crate::{Field, field::FieldOps};
 
+/// An arithmetic function over field elements, generic in the field it runs in.
+///
+/// One implementation evaluates in any field `F` whose scalar is the base field `G`:
+/// - natively, in the verifier's own field;
+/// - symbolically, over a circuit-element type that also implements [`FieldOps`].
+///
+/// The output length is fixed and known from [`Self::n_outputs`] without running the function.
+pub trait FieldFn<G: Field> {
```
jimpo: in other places in math, the convention is to use `F: Field, E: FieldOps<Scalar=F>`

jimpo (commented):
Could you try this again on main? 🙏 

Let's do a scoped down version. Don't rename compute_public_value to `compute` -- it should still only be for public values. But do replace the closure argument to `compute_public_value` with `FieldFn`.

crates/field/src/util.rs
tcoratger: Fixed

tcoratger:
> Could you try this again on main? 🙏
> 
> Let's do a scoped down version. Don't rename compute_public_value to `compute` -- it should still only be for public values. But do replace the closure argument to `compute_public_value` with `FieldFn`.

Should be done, let me know what you think :)

jimpo (approved):
Looks good! The context for why I think this is a good change now is that a channel for full recursion will need the expression evaluation logic for the computation.

PR #1586, opened 2026-06-25
BINIUS-69: heterogeneous mixed ZK/non-ZK BaseFold opening

benji-potamus:
## Summary

Generalizes the batched ZK BaseFold channel + MLE-check to a **mixed ZK / non-ZK** batch, per [BINIUS-69](https://linear.app/jimpo/issue/BINIUS-69/combine-non-zk-with-zk-basefold-compiler) and the design settled in the Linear thread (materialized combined buffer with lift/repeat placement; `log_lift` from `FRIParams`; full flexible non-ZK batch). Built on top of the merged foundation (#1539).

The all-ZK path remains **byte-identical** — the generalization doesn't disturb the existing protocol.

## Commits

1. **generalize batched ZK BaseFold to mixed ZK/non-ZK oracles** — conditional masking (`γ: Option`, σ/blinding only for ZK oracles); `max_n = max_i(log_msg_len_i)`; non-ZK messages/claims pass through unmasked and commit via `commit_interleaved` (`CommittedOracleData.mask` is now `Option`); the combined buffer places each oracle with zero-pad over `log_lift` and **repeat** over `log_repeat = max_n − n_i − log_lift`, with `s'` picking up `eq_ind_zero` over the lift dims only.
2. **minor refactors** (@jimpo) — review cleanups to taste.
3. **fix FRI challenge ordering for heterogeneous batch; enable mixed/zero-ZK tests** — see below.

## The FRI challenge-ordering fix (commit 3)

When a non-ZK oracle gets `log_batch_size > 0`, its batch-fold challenges come from the **leading `max_n − D` MLE rounds** (`D = rs_code().log_dim()`), which arrive *after* the outer (oracle-combine) challenges in the original feed order — so `FRIFolder`'s `[inner-batch | outer]` split mis-routed them. The fix feeds the outer challenges to the folder **after** those leading rounds, in both `prove_mlecheck_basefold_zk_batch` and `verify_mlecheck_basefold_zk_batch` (threading `reduced_log_dim = D` through). For an all-ZK batch `n_leading = max_n − D = 0`, so the outers still precede every MLE round and the transcript is unchanged.

## Tests

All 40 `binius-iop` / `binius-iop-prover` library tests pass, including the now-enabled heterogeneous cases:
- `test_basefold_zk_channel_mixed_zk_non_zk` — `[(8, non-ZK), (6, ZK)]`
- `test_basefold_zk_channel_zero_zk` — all non-ZK (γ never sampled)
- `test_basefold_zk_channel_mixed_invalid_proof` — tampered mixed batch is rejected

### DIFF
--- a/crates/iop/src/basefold_zk_channel.rs
+++ b/crates/iop/src/basefold_zk_channel.rs
@@ -6,8 +6,6 @@
 //! using FRI commitment and ZK BaseFold opening protocols. Unlike [`super::basefold_channel`],
 //! this channel always applies zero-knowledge blinding to all oracles.
 
-use std::iter;
-
 use binius_field::BinaryField;
 use binius_ip::{
 	channel::IPVerifierChannel,
@@ -23,6 +21,7 @@ use binius_transcript::{
 	fiat_shamir::{CanSample, Challenger},
 };
 use binius_utils::{DeserializeBytes, checked_arithmetics::log2_ceil_usize};
+use itertools::izip;
 
 use crate::{
 	basefold,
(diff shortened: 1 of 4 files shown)


jimpo:
This all looks good, and the planned FRI challenge fix looks right. I pushed some minor cleanups, according to my taste.

crates/iop/src/basefold_zk_channel.rs
```diff
@@ -242,6 +264,7 @@ where
 		&point,
 		gamma,
 		&outer_challenges,
+		fri_params.rs_code().log_dim(),
```
jimpo: The callee already gets `fri_params`, so this argument is redundant

jimpo (changes requested):

crates/iop/src/basefold_zk_channel.rs
benji-potamus: Good catch — dropped the redundant argument; `verify_mlecheck_basefold_zk_batch` now derives `D = fri_params.rs_code().log_dim()` internally. (Kept it on the prover's `prove_mlecheck_basefold_zk_batch`, which receives `fri_folder` rather than `fri_params` and can't derive it.) Also rebased onto current main (`45172d79`). All 40 iop/iop-prover lib tests still green; fmt + clippy clean.

jimpo (approved):

PR #1589, opened 2026-06-25
BINIUS-146: use hint_varsize for compute_public_value (derived public wire)

benji-potamus:
Closes [BINIUS-146](https://linear.app/jimpo/issue/BINIUS-146/use-hint-varsize-for-compute-public-value-derived-public-wire).

Salvaged from the cancelled BINIUS-43 / PR #1573 (the ReplayChannel → recorded-gate-sequence rewrite that wasn't a net win) — this is just the standalone-valuable piece.

## What

`compute_public_value` on the ZK wrapper's channels allocated its output as an **InOut** wire — a public input the verifier must supply — even though the value is *computed* from other public inputs. Make it a **derived** public wire instead: computed from the public inputs, emitting no constraint, sorted after the transcript inout wires.

## Commits (2)

1. **add `CircuitBuilder::hint_varsize`** — the runtime-arity sibling of `hint` (a `&[Wire]` input slice + runtime `out_len`, closure `&[F] -> Vec<F>`). Outputs are derived iff every input is public, else private; emits no constraints. Implemented on `ConstraintBuilder`, `WitnessGenerator`, and `InstanceGenerator`.
2. **use it for `compute_public_value`** — each of the three wrapper channels (symbolic `IronSpartanBuilderChannel`, prover `ReplayChannel`, verifier `ZKWrappedVerifierChannel`) now routes `compute_public_value` through its backend's `hint_varsize`, yielding a one-output derived wire instead of a channel-allocated inout. Because all three already allocate derived wires via the backend during the shared circuit traversal (exactly as `add`/`mul`/`hint` do), the wire ids stay aligned — `ReplayChannel` and the rest of the interaction path are untouched.

No `'static` bound is added to `hint` (per review): nothing records the closure for replay, so `compute_public_value`'s closure stays borrow-friendly.

## Validation (native)

- `spartan-frontend` lib (27, incl. a new `hint_varsize` sync test asserting `InstanceGenerator` matches the witness public segment),
- `spartan-verifier` lib (10), `wrapper_integration_test` (1),
- full `prove_verify` (7/7), incl. `test_zk_prove_verify_sha256_preimage` — the real path through `monster_eval` / `compute_public_value` — and the serialized round-trip,
- clippy `-D warnings` + rustdoc `-D warnings` clean on the tip.

### DIFF
--- a/crates/spartan-verifier/src/wrapper/builder_channel.rs
+++ b/crates/spartan-verifier/src/wrapper/builder_channel.rs
@@ -107,14 +107,24 @@ impl<F: Field> IPVerifierChannel<F> for IronSpartanBuilderChannel<F> {
 
 	fn compute_public_value(
 		&mut self,
-		_inputs: &[Self::Elem],
-		_f: impl FnOnce(&[F]) -> F,
+		inputs: &[Self::Elem],
+		f: impl FnOnce(&[F]) -> F,
 	) -> Self::Elem {
 		// The closure is an arbitrary native computation the constraint system cannot replay, so
-		// its result enters as a single inout wire (a public input the verifier supplies), rather
-		// than a sub-circuit's worth of constraints. The value is filled concretely by the
-		// instance/witness channels; symbolically we only allocate the wire.
-		self.alloc_inout_elem()
+		// its result enters as a single derived public wire (a one-output `hint_varsize`,
+		// computed from the public inputs, emitting no constraints) rather than a sub-circuit's
+		// worth of constraints. Symbolically we only allocate the wire; the value is filled
+		// concretely by the instance/witness channels, which recompute it via their own
+		// `hint_varsize`.
+		let out_wire = {
+			let mut builder = self.builder.borrow_mut();
+			let input_wires: Vec<_> = inputs
+				.iter()
+				.map(|elem| elem.to_wire(&mut builder))
+				.collect();
+			builder.hint_varsize(&input_wires, 1, move |vals| vec![f(vals)])[0]
+		};
+		CircuitElem::wire(&self.builder, out_wire)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


crates/spartan-verifier/src/wrapper/builder_channel.rs
```diff
@@ -107,14 +107,27 @@ impl<F: Field> IPVerifierChannel<F> for IronSpartanBuilderChannel<F> {
...
+		// its result enters as a single derived public wire (a one-output `hint_varsize`,
+		// computed from the public inputs, emitting no constraints) rather than a sub-circuit's
+		// worth of constraints. Symbolically we only allocate the wire; the value is filled
+		// concretely by the instance/witness channels, which recompute it via their own
+		// `hint_varsize`.
+		let out_wire = {
+			let mut builder = self.builder.borrow_mut();
+			let input_wires: Vec<_> = inputs
+				.iter()
+				.map(|elem| match elem {
```
jimpo: this pattern repeats in a few places. Perhaps we can have a helper method like CircuitElem::wire(&Bmut B)?

benji-potamus: Done — added `CircuitElem::to_wire(&mut B) -> B::Wire` (materializes a `Constant` via `builder.constant`, returns a `Wire`'s wire) and used it in all three channels' `compute_public_value`. Named it `to_wire` since `wire` is already the constructor. Pushed in 228bc8a6 (rebased onto latest main). Re-verified green: prove_verify 7/7, wrapper_integration_test, spartan-frontend/verifier libs, clippy/rustdoc `-D`.

(I considered routing through the existing `combine_varlen`, but it also constant-folds the all-inputs-constant case at the `F` level, and its two `FnOnce` closures would both need to capture `f` — so the small `to_wire` helper keeps the always-allocate semantics and reads cleaner here.)

jimpo (approved):

PR #1590, opened 2026-06-26
BINIUS-147: replace TaggedMul/TaggedSquare/TaggedInvertOrZero with the wrapper-struct pattern

benji-potamus:
Closes [BINIUS-147](https://linear.app/jimpo/issue/BINIUS-147/replace-taggedmultaggedsquaretaggedinvertorzero).

Replaces the three marker-parameterized arithmetic-strategy traits — `TaggedMul<S>`, `TaggedSquare<S>`, `TaggedInvertOrZero<S>` — with the wrapper-struct pattern already used by `WideMul`/`TrivialWideMul`. Each strategy is now a `#[repr(transparent)] #[derive(TransparentWrapper)] struct Foo<T>(T)` that implements the **normal** `Mul`/`Square`/`InvertOrZero` trait directly, and the `impl_{mul,square,invert}_with!` macros wire it into the canonical op by `wrap → op → peel` (mirroring how `define_packed_binary_field!` already consumes the `WideMul` wrapper). Pure aesthetics/refactor — no behavior change. Removes three redundant parallel traits and lets one wrapper carry several ops for a shared algorithm family.

Three commits, one per trait (as requested):

1. **`TaggedMul` → `Mul` wrapper** — `Pairwise`, `Gfni`, `Scaled`, `MulFromWideMul` become `TransparentWrapper` structs implementing `std::ops::Mul`; `impl_mul_with!`'s `@` arm consumes the wrapper. `TaggedMul` deleted.
2. **`TaggedSquare` → `Square` wrapper** — adds `Square` to `Pairwise`/`Scaled`; new wrappers `PairwiseTable`, `ReuseMultiply`, the generic `Divide<SubU, T>`, per-arch `Ghash`, `GhashClMulSquare`, and aarch64 `Aes`. `TaggedSquare` + the square-only markers (`ReuseMultiplyStrategy`, `DivideStrategy`, `GhashClMulSquareStrategy`) deleted.
3. **`TaggedInvertOrZero` → `InvertOrZero` wrapper** — adds `InvertOrZero` to `Pairwise`/`Scaled`/`Gfni`/`PairwiseTable`/per-arch `Ghash`/aarch64 `Aes`; new invert-only `Ghash256`/`Ghash512`. `TaggedInvertOrZero` + the last markers (`PairwiseTableStrategy`, `GfniStrategy`, `ScaledStrategy`, per-arch `GhashStrategy`, `AesStrategy`, `Ghash256Strategy`, `Ghash512Strategy`) deleted. After this, `grep -rn 'Tagged' crates/field` is empty.

Each arch's per-width strategy aliases (`Aes{Square,Invert}{1,16,32,64}x`, `Ghash{Square,Invert}{1,2,4}x`) are now generic `<T>` and resolve to wrappers, exactly like the existing `…WideMul…x<T>` aliases.

### Notes
- The `impl_*_with!` `@` arms capture the strategy as raw token-trees (`$($strategy:tt)*`, not `:ty`/`:path`) so the macro can append `<$name>` — a matched type fragment is opaque to the trailing `<`. Documented in a comment on `impl_mul_with!`.
- The `benchmark_alternative_strategies` bench variants that referenced strategy/op combinations with no impl (`packed`/`pairwise_table` for multiply, `packed` for square/invert) are dropped — they never compiled (that feature was already broken; it builds again now).
- Out of scope but noticed: `PackedStrategy`, `SimdStrategy`, `GfniSpecializedStrategy256b/512b` are now confirmed dead (zero references) — left for a follow-up to keep this PR scoped to the `Tagged*` removal.

### Verification (each commit, full arch matrix)
Portable `cargo test -p binius-field` (212) · native clippy `-D warnings` with `--tests --benches --examples` · aarch64 (`+neon,+aes`) clippy `--tests --benches` · `wasm32-unknown-unknown` build · full `cargo build --workspace` · fmt. The x86_64 gfni/avx512 paths (gfni `Gfni`/`Aes`, vpclmul `Ghash256/512`) are CI-gated as usual; they're identical wrapper translations of the old marker impls.


crates/field/src/arch/aarch64/packed_aes_128.rs
```diff
@@ -1,40 +1,44 @@
...
-pub type AesInvert16x = AesStrategy;
+/// Invert wrapper for the `PackedAESBinaryField16x8b` packing.
+pub type AesInvert16x<T> = Aes<T>;
 
-/// Strategy for aarch64 AES square/invert, both backed by `vqtbl` lookup tables.
-pub struct AesStrategy;
+/// Square wrapper for aarch64 AES, backed by `vqtbl` lookup tables.
+#[repr(transparent)]
+#[derive(TransparentWrapper)]
+pub struct Aes<T>(T);
```
jimpo: Rename to NeonTableLookupArithmetic

crates/field/src/arch/aarch64/packed_ghash_128.rs
```diff
@@ -7,38 +7,43 @@
...
-/// Strategy for aarch64 GHASH field arithmetic operations.
-pub struct GhashStrategy;
+/// Square wrapper for aarch64 GHASH field arithmetic.
+#[repr(transparent)]
+#[derive(TransparentWrapper)]
+pub struct Ghash<T>(T);
 
-// Implement TaggedSquare for GhashStrategy
-impl TaggedSquare<GhashStrategy> for PackedPrimitiveType<M128, BinaryField128bGhash> {
+impl Square for Ghash<PackedPrimitiveType<M128, BinaryField128bGhash>> {
```
jimpo: Split the wrapper types. This should be `Square for GhashClMulArithmetic<...>` and below, `InvertOrZero for GhashItohTsujii`

crates/field/src/arch/portable/packed_ghash_128.rs
```diff
@@ -52,20 +51,27 @@ impl Underlier128bLanes for M128 {
...
+impl Square for Ghash<PackedPrimitiveType<M128, BinaryField128bGhash>> {
 	#[inline]
 	fn square(self) -> Self {
-		ghash_square(self.0).into()
+		Self::wrap(ghash_square(Self::peel(self).0).into())
 	}
 }
 
-impl TaggedInvertOrZero<GhashStrategy> for PackedPrimitiveType<M128, BinaryField128bGhash> {
+impl InvertOrZero for Ghash<PackedPrimitiveType<M128, BinaryField128bGhash>> {
```
jimpo: This module should export the same `GhashItohTsujii` as the x86_64 case.

crates/field/src/arch/aarch64/packed_ghash_128.rs
```diff
@@ -7,38 +7,43 @@
...
+		Self::wrap(PackedPrimitiveType::from_underlier(super::arithmetic::ghash::square_clmul(
+			Self::peel(self).to_underlier(),
+		)))
 	}
 }
 
-// Implement TaggedInvertOrZero for GhashStrategy (Itoh-Tsujii — no CLMUL invert)
-impl TaggedInvertOrZero<GhashStrategy> for PackedPrimitiveType<M128, BinaryField128bGhash> {
+// Implement InvertOrZero for the GHASH packing (Itoh-Tsujii — no CLMUL invert)
+impl InvertOrZero for Ghash<PackedPrimitiveType<M128, BinaryField128bGhash>> {
```
jimpo: This should be in the non-x86-specific `itoh_tsujii` module, and just type aliased in this module.

crates/field/src/arch/portable/packed_ghash_128.rs
```diff
@@ -52,20 +51,27 @@ impl Underlier128bLanes for M128 {
 	}
 }
 
-impl TaggedSquare<GhashStrategy> for PackedPrimitiveType<M128, BinaryField128bGhash> {
+/// Square wrapper for portable GHASH field arithmetic.
+#[repr(transparent)]
+#[derive(TransparentWrapper)]
+pub struct Ghash<T>(T);
```
jimpo: This same struct is defined identically in the wasm32 case. Define it in arithmetic::ghash and use the same one in both.

crates/field/src/arch/x86_64/packed_ghash_128.rs
```diff
@@ -79,36 +80,42 @@ impl ghash::ClMulUnderlier for M128 {
...
 
-/// Strategy for x86_64 GHASH field arithmetic operations.
-pub struct GhashStrategy;
+/// Square wrapper for x86_64 GHASH field arithmetic.
+#[repr(transparent)]
+#[derive(TransparentWrapper)]
+pub struct Ghash<T>(T);
 
-// Implement TaggedSquare for GhashStrategy
+// Implement Square for the GHASH packing
```
jimpo: No, two separate strategies, and the `cfg_if` just selects between them. GhashClMul and the `GhashSoftMul` that should be used in the portable and wasm32 cases.

crates/field/benches/packed_field_invert.rs
```diff
@@ -18,30 +18,54 @@ fn invert_main<T: PackedField>(val: T) -> T {
...
-			arithmetic_traits::TaggedInvertOrZero,
+			arch::{Pairwise, PairwiseTable},
+			arithmetic_traits::InvertOrZero,
 		};
+		use bytemuck::TransparentWrapper;
 
-		fn invert_pairwise<T: TaggedInvertOrZero<PairwiseStrategy>>(val: T) -> T {
-			val.invert_or_zero()
+		/// Marker trait for packed types whose `Pairwise` wrapper supports inversion.
+		trait PairwiseInvert: Sized {
```
jimpo: Can we get rid of all this Pairwise stuff now? Seems like it's never used.

crates/field/src/arch/portable/packed_aes_128.rs
```diff
@@ -2,5 +2,5 @@
 // Copyright 2026 The Binius Developers
 
 pub type AesWideMul16x<T> = crate::arch::ElementwiseWideMul<T>;
-pub type AesSquare16x = crate::arch::PairwiseTableStrategy;
-pub type AesInvert16x = crate::arch::PairwiseTableStrategy;
+pub type AesSquare16x<T> = crate::arch::PairwiseTable<T>;
```
jimpo: Use the `Divide` strategy instead of `PairwiseTable`

crates/field/src/arch/portable/packed_aes_128.rs
```diff
@@ -2,5 +2,5 @@
...
+pub type AesInvert16x<T> = crate::arch::PairwiseTable<T>;
```
jimpo: Use the `Divide` strategy instead of `PairwiseTable`

crates/field/src/arch/portable/packed_aes_8.rs
```diff
@@ -2,5 +2,5 @@
 // Copyright 2026 The Binius Developers
 
 pub type AesWideMul1x<T> = crate::arch::portable::pairwise_table_arithmetic::AesLookupWideMul<T>;
-pub type AesSquare1x = crate::arch::PairwiseTableStrategy;
-pub type AesInvert1x = crate::arch::PairwiseTableStrategy;
+pub type AesSquare1x<T> = crate::arch::PairwiseTable<T>;
```
jimpo: Don't do pairwise. Make a strategy called `BytewiseLookup` for PackedPrimitiveType AES 1x8b.

I don't want any more packed arithmetic defined in terms of scalar arithmetic.

jimpo (changes requested):

crates/field/src/arch/aarch64/packed_aes_128.rs
benji-potamus: Done — renamed the wrapper to `NeonTableLookupArithmetic` (e3c12d50).

crates/field/src/arch/aarch64/packed_ghash_128.rs
benji-potamus: Done — split into a `GhashClMul` square wrapper (now in `aarch64::arithmetic::ghash`, calling the PMULL `square_clmul`) and the shared `GhashItohTsujii` for invert. I named the square wrapper `GhashClMul` (matching the x86_64 wrapper and `GhashSoftMul`/`GhashItohTsujii`) rather than `GhashClMulArithmetic`, for naming consistency — easy to switch to the longer name if you'd prefer it.

crates/field/src/arch/portable/packed_ghash_128.rs
benji-potamus: Done — portable now aliases `GhashInvert1x` to the same shared `GhashItohTsujii`. The old scalar-bridge is gone: on the portable target the packing is reflexively `Divisible<GhashB128>`, so the generic wrapper applies directly.

crates/field/src/arch/aarch64/packed_ghash_128.rs
benji-potamus: Done — `GhashItohTsujii` now lives in `portable::arithmetic::itoh_tsujii` as a single generic `InvertOrZero` impl (over `P: Copy + Square + Mul + Divisible<GhashB128>`). Every arch — x86_64, aarch64, portable, wasm32, and the 2x/4x packings (the old `Ghash256`/`Ghash512` invert wrappers are gone too) — type-aliases its `GhashInvert*` to it.

crates/field/src/arch/portable/packed_ghash_128.rs
benji-potamus: Done — the software square is now `GhashSoftMul` in `portable::arithmetic::ghash`, used by both the portable and wasm32 packings (and the x86_64 non-PCLMULQDQ fallback). The duplicated per-arch `Ghash` square structs are removed.

crates/field/src/arch/x86_64/packed_ghash_128.rs
benji-potamus: Done — `GhashSquare1x` is now a `cfg_if` between two separate strategies at the alias level: `GhashClMul` when PCLMULQDQ is present, otherwise the shared `GhashSoftMul`. No cfg inside a `square` method anymore.

crates/field/benches/packed_field_invert.rs
benji-potamus: Done — removed the `Pairwise`/`PairwiseTable` benchmark variants from all three benches (multiply/square/invert; they're now just `main`), and deleted the now-unused `Pairwise` wrapper + `pairwise_arithmetic` module.

crates/field/src/arch/portable/packed_aes_128.rs
benji-potamus: Done — `AesSquare16x` now uses `Divide<u8, _>`, which recurses one byte at a time into the 1×8b `BytewiseLookup`. I also switched the x86_64 non-GFNI fallbacks (16x/32x/64x) to `Divide<u8, _>` and removed `PairwiseTable` entirely, so no packed AES arithmetic is defined via per-scalar iteration anywhere.

crates/field/src/arch/portable/packed_aes_128.rs
benji-potamus: Done — `AesInvert16x` uses `Divide<u8, _>` as well (added an `InvertOrZero` impl to `Divide`, mirroring its `Square` impl).

crates/field/src/arch/portable/packed_aes_8.rs
benji-potamus: Done — added `BytewiseLookup` for the 1×8b AES packing: it looks up square/invert directly in the byte tables keyed by the `u8` underlier (no `PackedField::get` scalar iteration). Wider packings reach it via `Divide<u8, _>`, so packed AES arithmetic is no longer defined in terms of scalar arithmetic.

crates/field/src/arch/aarch64/arithmetic/ghash.rs
jimpo, line 201: Follow up: drop this struct and just use `GhashClMul` (above) for WideMul as well as Square.

jimpo (approved):

PR #1592, opened 2026-06-26
BINIUS-149: tabulate the inner equality block in the AND-reduction round message

tcoratger:
Closes [BINIUS-149](https://linear.app/jimpo/issue/BINIUS-149).

Computes the AND-reduction round message the same way, but turns its expensive multiplies into table lookups.

### The problem

The round message weights every vertex by an equality factor.
That is **one $\mathbb{F}_{2^{128}}$ multiply per chunk, per output row**.
At $2^{15}$ chunks $\times$ 64 rows that is ~2M generic GHASH multiplies — the round's bottleneck.

### The idea

The equality weight factors. Split the big-field coordinates into an inner block (the innermost 4) and an outer block:

```
eq_big[s*16 + b] = eq_outer[s] * eq_inner[b]
```

The inner block has only 16 values, so bake them into a small table:

```
convert[b][v] = eq_inner[b] * phi(v)      // 16 x 256 = 64 KB
```

Now the inner weight is a lookup, not a multiply. Only the outer weight is still multiplied in.

### Per output row

- **before:** 16 $\mathbb{F}_{2^{128}}$ multiplies (one per inner value)
- **after:** 16 lookups + XORs, then 1 multiply by the outer weight

→ a $16\times$ cut in the dominant operation.

### The change

```
OblongZerocheckProver::new:
- univariate_round_message_extension_domain(.., &eq_ind_partial_eval(&big), ..)   // full eq buffer, multiply per vertex
+ eq_convert::univariate_round_message(.., &big, ..)                              // factor + tabulate the inner block
```

### How this relates to Flock §4.3 — and what it is *not*

§4.3 is about **pinning** coordinates to fixed friendly constants. This PR does **not** pin.

- It borrows only the lookup-table trick from §4.3's medium level.
- The table weights here are the existing **random** `eq_inner[b]`, not the pinned powers $\gamma^b$.
- So the table is rebuilt each proof, not baked once at startup.

In short: this reuses §4.3's plumbing, not its idea. Actual friendly challenges (pinning) are a follow-up.

- The challenge distribution is untouched; the verifier samples exactly as before.
- The message is bit-identical, so the transcript matches today's protocol exactly.
- §4.3's $\mathbb{F}_2$-independence condition is only needed for pinning, which is not done here.

The pinning follow-up (fixed startup table, 4 fewer challenges in the transcript) is where that soundness check would be owed.

### Scope

- **new:** `eq_convert.rs` (the table + round message) and its bench
- **changed:** one call site in `prover.rs`
- the old round message is untouched — it stays as the test/bench reference oracle
- instances with fewer than 4 big coordinates fall back to it
- distinct from BINIUS-148 (the §4.2 LDE-table shrink): same function, different sub-region — merges need coordinating

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-prover`, nightly `fmt` — clean
- proptests: optimized == reference, on both the inner-block and fallback paths
- full `tests/prove_verify.rs` (SHA-256 prove → verify, including ZK) passes

### Benchmark

`cargo bench --bench eq_convert`, $2^{24}$, round-1 message, Apple silicon (8 performance cores), rebased on the rewritten NTTLookup (#1591):

```
                          reference   eq_convert   speedup
multi-threaded (8 cores)   3.25 ms     2.64 ms      1.23x
single-threaded           16.77 ms    15.13 ms      1.11x
```

The multi-threaded win is larger because the 64 KB table and the $16\times$-smaller outer buffer stay in cache, where the reference's full eq buffer does not.

### DIFF
--- a/crates/prover/Cargo.toml
+++ b/crates/prover/Cargo.toml
@@ -58,6 +58,10 @@ name = "and_reduction"
 harness = false
 
 [[bench]]
+name = "eq_convert"
+harness = false
+
+[[bench]]
 name = "shift_reduction"
 harness = false
 
--- a/crates/prover/benches/eq_convert.rs
+++ b/crates/prover/benches/eq_convert.rs
@@ -1,0 +1,85 @@
+// Copyright 2025 Irreducible Inc.
+//! Head-to-head benchmark for the AND-reduction univariate round message.
+//!
+//! Compares the unoptimized equality weighting, which pays one large-field multiply per hypercube
+//! vertex, against the convert-table path, which tabulates the innermost equality coordinates and
+//! replaces those multiplies with lookups and XORs.
+
+use std::{iter, iter::repeat_with};
+
+use binius_core::word::Word;
+use binius_field::{AESTowerField8b, PackedAESBinaryField64x8b, Random};
+use binius_math::{BinarySubspace, multilinear::eq::eq_ind_partial_eval};
+use binius_prover::and_reduction::{
+	NTTLookup, eq_convert::univariate_round_message,
+	sumcheck_round_messages::univariate_round_message_extension_domain,
+};
+use binius_verifier::{config::B128, protocols::bitand::SKIPPED_VARS};
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
+use rand::prelude::*;
+
+fn bench(c: &mut Criterion) {
+	// Problem size: 2^24 total variables, so 2^18 words per column.
+	let log_num_rows = 24;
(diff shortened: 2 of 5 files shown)


jimpo:
Ah, I did the same thing in https://github.com/binius-zk/binius64/pull/1597! But also several other refactors to tighten the inner loop.

tcoratger:
> Ah, I did the same thing in #1597! But also several other refactors to tighten the inner loop.

Cool, let me close this one to avoid duplicates!

PR #1593, opened 2026-06-26
BINIUS-120: scaffold the M4 batch prover and verifier with PCS commit/open

tcoratger:
Closes [BINIUS-120](https://linear.app/jimpo/issue/BINIUS-120).

Adds the prover+verifier scaffold that commits the 2-D batch witness and opens it at a verifier-chosen random point.

### The problem

The batch witness can already be packed and committed (BINIUS-142), but nothing yet *opens* it.
A commitment with no evaluation proof anchors nothing.
And there is no verifier path at all, so we cannot check that a committed batch opens to the value the prover claims.

### The idea

Stand up the smallest faithful PCS round-trip on the existing BaseFold stack:
- pack the table to one $\mathbb{F}_{2^{128}}$ multilinear and commit it as the trace oracle,
- draw a random point from the transcript *after* committing, so Fiat-Shamir binds it to the commitment,
- open the commitment at that point; the verifier mirrors and checks.

The opened value is the standard inner product:

```
committed(point) = <committed, eq_ind(point)>
```

### The change

Both operations live on `BatchPcsParams`, the one type the prover and verifier share, so both derive the oracle shape and FRI query count identically:

```
params.prove_evaluation::<P>(&table, &mut prover_transcript)  -> (point, eval)
params.verify_evaluation(&mut verifier_transcript)            -> Result<(point, eval)>
```

- **new:** `crates/m4/src/pcs.rs` — `BatchPcsParams::{new, prove_evaluation, verify_evaluation}`
- **changed:** `crates/m4/src/lib.rs` exports `BatchPcsParams`; `Cargo.toml` gains `binius-hash` / `binius-ip` / `binius-ip-prover`, and `binius-transcript` moves to a normal dependency.

This opens the **packed $\mathbb{F}_{2^{128}}$ multilinear directly**.
It does **not** ring-switch down to the bit witness — that belongs with the later reductions (BINIUS-121 BitAnd, BINIUS-119 shift), which produce the claim those open against.
So this is the PCS plumbing the reductions plug into, not a full circuit prove/verify.

### Soundness

- Reuses BaseFold/FRI in the unique-decoding regime; nothing new assumed.
- The evaluation point is sampled only after the commitment, so it cannot be chosen to fit the data.
- The FRI query count is `calculate_n_test_queries(96, log_inv_rate)`, targeting 96-bit soundness to match the Binius64 verifier.

### Scope

- **new:** the prover+verifier commit/open scaffold and its tests
- **changed:** m4 exports and dependencies
- **left untouched:** the witness table and packer (BINIUS-142); no other crate touched

### Verification

- nightly `cargo fmt --all --check`, `cargo clippy -p binius-m4 --all-targets` — clean
- `cargo test -p binius-m4` — 15 pass, including:
  - round-trip proptest: the verifier accepts, and the opened value equals the table's direct MLE at the point
  - tamper test: a flipped proof byte is rejected with the exact variant `BaseFold(Verification(FRI(MerkleError(InvalidProof))))`


crates/m4/src/pcs.rs
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	/// The committed-multilinear shape of the batch.
+	pub fn layout(&self) -> BatchCommitLayout {
+		self.layout
+	}
+
+	/// The single committed oracle: the packed batch witness.
+	fn oracle_specs(&self) -> Vec<OracleSpec> {
+		vec![self.layout.oracle_spec()]
```
jimpo: oracle_spec() doesn't belong on layout -- layout doesn't "know" about the protocol, in theory. Also, I see that it uses OracleSpec::new_zk currently. ZK-ness is a higher level choice, and one we shouldn't default to right now for M4.

crates/m4/src/pcs.rs
```diff
@@ -0,0 +1,329 @@
...
+	/// So Fiat-Shamir binds it to the committed data.
+	///
+	/// This opens the packed B128 multilinear directly.
+	/// It does not ring-switch down to the bit witness.
+	/// That step belongs with the later reductions.
+	///
+	/// # Returns
+	///
+	/// The random point and the evaluation the prover commits to at it.
+	pub fn prove_evaluation<P, Challenger_>(
```
jimpo: Please begin a `binius-m4-prover` crate for the prover code. Separate prover/verifier crates is important, and best to start with it.

crates/m4/src/pcs.rs
```diff
@@ -0,0 +1,329 @@
...
+	/// Returns an error if the commitment opening does not verify.
+	pub fn verify_evaluation<Challenger_>(
+		&self,
+		transcript: &mut VerifierTranscript<Challenger_>,
+	) -> Result<(Vec<B128>, B128), binius_iop::channel::Error>
+	where
+		Challenger_: Challenger,
+	{
+		// Build the matching verifier over the same oracle shape and query count.
+		let merkle_scheme = BinaryMerkleTreeScheme::<B128, StdHashSuite>::new();
```
jimpo: Lift this up to a setup phase.

crates/m4/src/pcs.rs
```diff
@@ -0,0 +1,329 @@
...
+		transcript: &mut ProverTranscript<Challenger_>,
+	) -> (Vec<B128>, B128)
+	where
+		P: PackedField<Scalar = B128>,
+		Challenger_: Challenger,
+	{
+		// Build the BaseFold prover over an evaluation domain large enough for the codeword.
+		let subspace = BinarySubspace::<B128>::with_dim(self.log_code_len());
+		let domain_context = GenericOnTheFly::generate_from_subspace(&subspace);
+		let ntt = NeighborsLastSingleThread::new(domain_context);
```
jimpo: Lift this up to a setup phase. And this should not be the SingleThreaded NTT on the prover side.

jimpo (changes requested):
Try to model this off of the Prover/Verifier structs in `binius-prover` / `binius-verifier`.

crates/m4/src/pcs.rs
tcoratger: Right, just removed.

crates/m4/src/pcs.rs
tcoratger: Just created a `binius-m4-prover` crate.

crates/m4/src/pcs.rs
tcoratger: Done, `Prover::setup(verifier: &Verifier)` now does the one-time work and uses the multithreaded NTT.

crates/m4/src/pcs.rs
tcoratger: Fixed

tcoratger:
> Try to model this off of the Prover/Verifier structs in `binius-prover` / `binius-verifier`.

Just fixed that so that we have a prover and a verifier crate in order to separate things properly and be consistent with the rest of the codebase.

crates/m4-verifier/src/verify.rs
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	/// Verifies a batch-witness commitment opened at a verifier-chosen random point.
+	///
+	/// The verifier receives the commitment, redraws the same point, reads the claim, and checks
+	/// the opening.
+	///
+	/// # Returns
+	///
+	/// The random point and the evaluation the opening proves at it.
```
jimpo, line 99: Why do we return these values from the verifier & prover? I think the return type should just be Result<()>

jimpo (approved):
Nice!!

PR #1596, opened 2026-06-27
BINIUS-127: define GhashSq256b multiplication in terms of WideMul

tcoratger:
Closes [BINIUS-127](https://linear.app/jimpo/issue/BINIUS-127/define-mul-in-terms-of-widemul-across-all-fields).

Makes the widening multiply the single source of truth for `GhashSq256b` — the last field still doing it the other way round. `Mul` becomes `reduce(wide_mul)`: bit-identical results, plus a real deferred-reduction path for inner products.

### The problem

A binary-field multiply is two steps: a raw product, then a reduction back to size. The reduction is the expensive part.

GHASH and AES fields already define `Mul = reduce(wide_mul)`, so a sum of products can accumulate raw and reduce once (BINIUS-128, BINIUS-137). $\mathbb{F}_{2^{256}}$ was the exception: it used the trivial widening multiply, so it reduced on **every** term of an inner product.

### The idea

$\mathbb{F}_{2^{256}}$ here is a degree-two extension of GHASH, with $Y^2 = Y + X^{-1}$:

```
z0 = x0*y0 + (x1*y1)*X^-1
z1 = (x0+x1)*(y0+y1) + x0*y0
```

That is three GHASH products (a Karatsuba split: the diagonal pair `[x0*y0, x1*y1]` batched into one packed multiply, plus the cross term).

The reduction and the `X^-1` scaling are both $\mathbb{F}_2$-linear, so:

$$\mathrm{reduce}(A) + \mathrm{reduce}(B) = \mathrm{reduce}(A + B)$$

So an inner product can keep the three products **unreduced**, XOR-accumulate them, and reduce once at the end.

### Per inner product term

- **before:** 3 GHASH reductions per term (trivial widening = a full multiply each)
- **after:** XOR-accumulate raw; 3 GHASH reductions total, once

→ reductions drop from $3N$ to $3$ for an $N$-term sum.

### The change

```
- impl_trivial_wide_mul!(GhashSq256b);              // WideMul defined via Mul
- fn mul(..) { Karatsuba, reduces every product }
- impl Mul { from_coeffs(mul(..)) }

+ struct WideGhashSqProduct { diag, cross }          // the three unreduced products
+ impl WideMul for GhashSq256b { wide_mul; reduce }  // single source of truth
+ impl Mul { reduce(wide_mul(self, rhs)) }
```

### Soundness — preserved

- The math is the existing Karatsuba identity, unchanged.
- Deferring is exact because reduction and `X^-1` scaling are $\mathbb{F}_2$-linear.
- A standalone multiply does the same work as before (no regression); the win is only in sums.

### Scope

- **changed:** `crates/field/src/ghash_sq.rs` — the widening multiply, `Mul`, and a proptest for the deferred-reduction identity.
- **bench:** added `GhashSq256b` to `crates/field/benches/ghash_widening.rs`.
- **untouched:** `square` keeps its own optimized path; `B1` stays the trivial leaf (its multiply is a single AND — nothing to defer).
- This is the final field, so every non-trivial field now defines `Mul = reduce(wide_mul)`.

### Verification

- nightly `fmt --all`, `clippy -p binius-field --all-targets`, `check --workspace --all-targets` — clean
- `cargo test -p binius-field ghash_sq` — 15 pass, including a proptest that `reduce(sum_i wide_mul) == sum_i a_i*b_i`

### Benchmark

Inner product `sum_i a_i*b_i` over $\mathbb{F}_{2^{256}}$, deferred path (`ghash_widening`):

| n (terms) | before (trivial) | after (real WideMul) | speedup |
|---|---|---|---|
| 16 | 86.6 ns | 63.7 ns | 1.36× |
| 256 | 1.371 µs | 0.987 µs | 1.39× |
| 4096 | 23.42 µs | 16.34 µs | 1.43× |

A standalone multiply is unchanged — the gain is purely in batched sums, exactly where the deferred reduction applies.

### DIFF
--- a/crates/field/benches/ghash_widening.rs
+++ b/crates/field/benches/ghash_widening.rs
@@ -8,7 +8,7 @@
 
 use std::hint::black_box;
 
-use binius_field::{PackedField, WideMul, arch::OptimalPackedB128};
+use binius_field::{GhashSq256b, PackedField, WideMul, arch::OptimalPackedB128};
 use criterion::{
 	BenchmarkGroup, Criterion, Throughput, criterion_group, criterion_main, measurement::WallTime,
 };
@@ -47,6 +47,12 @@ fn bench_ghash_widening(c: &mut Criterion) {
 	let label = format!("OptimalPacked_{}x128b", OptimalPackedB128::WIDTH);
 	for &n in &[16, 256, 4096] {
 		bench_at_n::<OptimalPackedB128>(&mut group, &label, n);
+	}
+
+	// GF(2^256) is a degree-two extension of GHASH, so one multiply is three GHASH products.
+	// Deferring their reduction across an inner product is where the widening multiply pays off.
+	for &n in &[16, 256, 4096] {
+		bench_at_n::<GhashSq256b>(&mut group, "GhashSq256b", n);
 	}
 
 	group.finish();
--- a/crates/field/src/ghash_sq.rs
+++ b/crates/field/src/ghash_sq.rs
@@ -34,7 +34,7 @@ use super::{
 use crate::{
 	BinaryField128bGhash, Divisible, Field, PackedBinaryGhash2x128b, PackedField,
 	arch::{M128, M256, m256_from_u128s},
-	arithmetic_traits::{InvertOrZero, Square, impl_trivial_wide_mul},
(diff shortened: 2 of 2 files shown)


jimpo (approved):
Beautiful! ❤️

PR #1599, opened 2026-06-27
BINIUS-151: fix the circuit-test recompile (run-layer -E filter, not --exclude)

benji-potamus:
Closes [BINIUS-151](https://linear.app/jimpo/issue/BINIUS-151/reduce-the-ci-cost-of-circuit-tests). Pure `ci.yml`, one commit.

Fixes a regression from #1598: skipping `binius-circuits` tests on the arm64/portable legs with `cargo nextest run --workspace --exclude binius-circuits` changed the workspace **package set**, so Cargo's v2 feature resolver unified features differently than the `Compile` step (`cargo build --workspace --all-targets`) and recompiled shared deps (`sha2`/`sha3`/`rand`/…) for several minutes — more than it saved.

Now the test step keeps `--workspace` (same build set → no recompile) and just skips *running* the circuit tests on the non-x86 legs via nextest's filterset `-E 'not package(binius-circuits)'`.

**Simplified per review:** dropped the path-filtering (the `changes` job + `ci-all-checks` skip-handling). Validated by `yaml.safe_load` + `prek run typos`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,13 +46,15 @@ jobs:
       fail-fast: false
       matrix:
         arch:
+          # binius-circuits tests are architecture-independent (the crate depends only on
+          # binius-frontend + binius-core, which contain no arch-specific code), so they run only on
+          # the x86_64 leg. The other legs skip them with nextest's run-layer filter (`-E`) rather
+          # than `cargo --exclude`: keeping the build set at `--workspace` matches the Compile step's
+          # feature unification, so the dependency tree is not recompiled. See BINIUS-151.
           - name: x86_64-portable
             runner: ubuntu-24.04
             rustflags: # portable
-            # binius-circuits tests are architecture-independent (it depends only on
-            # binius-frontend + binius-core, which contain no arch-specific code), so they run
-            # only on the x86_64 leg below. See BINIUS-151.
-            test_args: "--exclude binius-circuits"
+            test_args: "-E 'not package(binius-circuits)'"
           - name: x86_64
             runner: ubuntu-24.04
             rustflags: "-Ctarget-cpu=native"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -16,6 +16,35 @@ concurrency:
...
+      pull-requests: read
+    outputs:
+      circuits: ${{ steps.filter.outputs.circuits }}
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v6
+      - name: Filter circuit-relevant paths
+        id: filter
+        if: github.event_name == 'pull_request'
+        uses: dorny/paths-filter@v3
```
jimpo: Is this necessary? This looks simpler: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-excluding-paths

jimpo (changes requested):
This is too complicated. Simplify it.

benji-potamus:
Simplified — dropped the path-filtering entirely (the `changes` job + `ci-all-checks` skip-handling + the conditional test-step logic). This PR is now just the one-commit recompile fix: keep `cargo nextest run --workspace` and skip *running* circuit tests on the non-x86 legs via `-E 'not package(binius-circuits)'`, so the build set matches the Compile step and there's no feature-unification recompile. 14 lines, one file.

If you still want unaffected PRs to skip the heavy `circuits-snapshot` job, I can do that as a minimal follow-up (gate just that one job) — but it needs the `ci-all-checks` skipped-as-success handling, so I left it out here rather than carry the complexity. Just say the word.

Rebased onto latest main; CI re-running.

jimpo (commented):
No, go back to the version that takes test_args as a matrix entry and just sets it from the different jobs. Just change the flag to the new one.

benji-potamus:
Done — reverted to the per-arch `test_args` matrix entry (your structure from #1598), just with the flag changed: `--exclude binius-circuits` → `-E 'not package(binius-circuits)'`. The test step is back to `cargo nextest run --workspace --no-fail-fast ${{ matrix.arch.test_args }}`. Rebased onto latest main; diff is now a 7-line flag swap.

jimpo (approved):

PR #1600, opened 2026-06-27
BINIUS-57: PackedField for GhashSq256b in a sliced layout

tcoratger:
Closes [BINIUS-57](https://linear.app/jimpo/issue/BINIUS-57/packedfields-for-ghashsq-with-sliced-layout).

Adds `PackedField` implementations whose scalar is `GhashSq256b`, stored in a sliced (struct-of-arrays) layout so a batch multiply costs three packed GHASH multiplies total instead of per element.

### The problem

A `GhashSq256b` element is a degree-1 polynomial in $Y$ over GHASH: $a + b\,Y$, with $Y^2 = Y + X^{-1}$.
Multiplying two of them is multiplying two such polynomials — the schoolbook way costs **four** $\mathbb{F}_{2^{128}}$ multiplies per element.
The hardware multiply (`PCLMUL`) is the expensive operation, and it batches best over a long run of same-kind values. An element-by-element layout `[a b][a b]...` never gives it that run.

### The idea

Store the two coordinates in **separate arrays** and pack `WIDTH` elements as `[P; 2]`:

```
coords[0] = [a0 a1 a2 ...]   // every element's "1"-part
coords[1] = [b0 b1 b2 ...]   // every element's "Y"-part
```

Then Karatsuba over $Y$ needs only **three** products, each run once across the whole batch:

```
t0 = coords[0] * coords[0]'              // a*e per lane
t2 = coords[1] * coords[1]'              // b*f per lane
t1 = (coords[0]+coords[1]) * (..)'       // holds the cross term
```

Fold $Y^2 = Y + X^{-1}$ back into the basis:

```
z.coords[0] = t0 + t2 * X^-1
z.coords[1] = t1 + t0     // the two t2 cancel in characteristic two
```

### Per element

- **before (schoolbook):** 4 GHASH multiplies per element
- **after (sliced Karatsuba):** 3 packed GHASH multiplies for the entire batch

→ the multiply cost amortizes over `WIDTH`, and the reduction collapses into two XOR-and-scale lines.

### The change

- **new:** `crates/field/src/packed_ghash_sq.rs` — `PackedSlicedGhashSq<P>`, generic over any `PackedField<Scalar = BinaryField128bGhash>`, holding `[P; 2]`. Concrete aliases `PackedGhashSq1x256b`, `PackedGhashSq2x256b`, `PackedGhashSq4x256b`.
- Implements the full `PackedField` surface: add/sub/mul/square/invert-or-zero, broadcast/get/set, interleave/unzip/spread, `square_transpose`, and `Divisible`/`Maskable`/`WideMul`.
- **wired:** `lib.rs` registers and re-exports the module; the generic `packed` test suite now runs all three widths.

### Soundness

- Squaring uses that the cross term vanishes in characteristic two: $(a + b\,Y)^2 = (a^2 + b^2 X^{-1}) + b^2\,Y$.
- Inversion goes through the norm of the degree-2 extension: the conjugate of $a + b\,Y$ under $Y \mapsto Y + 1$ is $(a+b) + b\,Y$, its norm $N = a^2 + ab + b^2 X^{-1}$ lies in GHASH, and the inverse is the conjugate times $N^{-1}$. A zero input gives norm 0, so `invert_or_zero` returns 0.

### Scope

- new field type and its tests only; no existing type changed.
- no protocol or call-site changes — this only adds packings.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-field --all-targets`, nightly `fmt` — clean
- `cargo test -p binius-field` — including: every width's mul/square/invert pinned lane-by-lane against the scalar `GhashSq256b` reference; get/set/broadcast round trips; arithmetic identities; interleave/unzip inverse; and `square_transpose` against an independent naive oracle.


crates/field/src/packed_ghash_sq.rs
```diff
@@ -0,0 +1,679 @@
...
+/// Each lane `i` holds the element `coords[0].get(i) + coords[1].get(i) * Y`.
+///
+/// Multiplication is the sliced Karatsuba algorithm over GHASH:
+/// for `x = x_0 + x_1*Y` and `y = y_0 + y_1*Y` with `Y^2 = Y + X^-1`,
+/// - `z_0 = x_0*y_0 + (x_1*y_1)*X^-1`
+/// - `z_1 = (x_0+x_1)*(y_0+y_1) + x_0*y_0`
+///
+/// Each of the three GHASH products runs once across all lanes, so the cost amortizes over `WIDTH`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+pub struct PackedSlicedGhashSq<P: PackedField<Scalar = BinaryField128bGhash>>([P; 2]);
```
jimpo: Ah, could you actually try do instead do this as a standard `PackedPrimitiveType<ScaledUnderlier<U, 2>, GhashSq>`? I think the right approach for the field crate is for all packed fields to be views over bitvectors (in this way).

Basically, you'll implement Mul, Square, and InvertOrZero strategies generically when `PackedPrimitiveType<U, Ghash>: PackedField`. Then you can do some underlier casting to get in into that form.

jimpo (changes requested):

crates/field/src/packed_ghash_sq.rs
jimpo: Oh, wait, now I remember the problem. The iterator order for `ScaledUnderlier` is wrong. We would actually need a different underlier wrapper that has a strided iteration order.... 

Unclear if that's actually worth it. Maybe we just merge this and later see if it needs to be changed.

tcoratger: Oops I wasn't familiar with the framework. I just pushed a new version respecting `PackedPrimitiveType<ScaledUnderlier<U, 2>, GhashSq>` in the last commit.

Let me know which one you prefer and if you prefer the old one, I can just revert the last commit. But the new one looks good and more aligned with the rest of the codebase. Normally it should come back to the same and I've included `split` and `join` methods to switch between the representations easily :)

tcoratger:
@jimpo With https://github.com/binius-zk/binius64/pull/1747, we can close this one no?

jimpo:
Yes, that one is almost ready for merge. Sorry to replace this -- this is an important new abstraction (sliced packed fields) that I wanted to think through.

PR #1602, opened 2026-06-27
ci: don't compile benchmarks in rust-checks

benji-potamus:
Second of three independent CI compile-time PRs (work order: disable-debuginfo → this → lld linker).

`rust-checks` builds with `--all-targets`, which includes benchmarks. CI never *runs* benches, and each criterion bench is a separate binary that links the whole crate under test — so codegen + linking them is pure overhead, paid on all three arch legs.

This drops benches from the **Compile** step only (`--lib --bins --tests --examples`), while keeping **clippy** at `--all-targets`. Clippy is front-end only (no codegen/link), so it still type-checks the benches — guarding against bench bitrot (API drift) for a small fraction of a full build. The only gap is codegen-/link-only failures in benches (rare, and benches aren't shipped).

Validated `yaml.safe_load` + `prek typos`; confirmed `cargo build -p binius-field --lib --bins --tests --examples` is valid locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,8 +99,12 @@ jobs:
           tool: cargo-nextest
       - name: Install lld
         run: sudo apt-get update -qq && sudo apt-get install -y lld
+      # Don't *build* benchmarks: each criterion bench is a separate binary linking the whole crate,
+      # and CI never runs them (`cargo bench`), so codegen + linking them is pure overhead. Clippy
+      # below still checks them via `--all-targets` — that's front-end only (no codegen/link), so it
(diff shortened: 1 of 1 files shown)


jimpo (approved):

benji-potamus:
Heads up: I rebased this onto latest main to pick up #1603 (lld), which dismissed your approval. The only rebase conflict was trivial — where the new "Install lld" step met the bench comment; kept both. Content is otherwise exactly what you approved on `b9a52148` (Compile drops benches, clippy `--all-targets` still checks them). CI green on arm64/portable, x86_64 finishing. Mind re-approving `41064bb1` when convenient?

PR #1604, opened 2026-06-27
BINIUS-154: logUp* verifier

tcoratger:
Closes [BINIUS-154](https://linear.app/jimpo/issue/BINIUS-154/verifier-code-for-logup).

Adds the verifier for the logUp* indexed-lookup reduction (one table, one looker column).
Verifier only — no prover yet, as scoped by the issue. The transcript layout it expects is documented so the prover can match it.

### The problem

An indexed lookup checks $(I^* T)[i] = T[I[i]]$: a long index column $I$ reads into a small table $T$.

The naive proof commits the looked-up vector $I^* T$ — that is $2^n$ *big* field elements, the dominant cost when the table is small but the lookups are many.

We want to verify a claim about $I^* T$ **without ever committing it**.

### The idea

Duality turns the expensive pullback evaluation into a cheap pushforward one:

$$
(I^* T)(r) = \langle I^* T, eq_r \rangle = \langle T, I_* eq_r \rangle = \langle T, Y \rangle
$$

$Y = I_* eq_r$ is table-sized ($2^m$ entries), so we work with $Y$ instead of $I^* T$.

The verifier then runs two checks:

```
1. Y is the true pushforward, via the logUp fraction identity for a random c:
     sum_i eq_r[i]/(c - I[i])  =  sum_j Y[j]/(c - j)
2. the product claim <T, Y> = e
```

Each side of the fraction identity is a sum of fractions, collapsed to one root fraction by a fractional-addition GKR circuit; equality is the cross-multiplication of the two roots.

### The batching (the point of this issue)

The table-side GKR and the product sumcheck both end in an evaluation of $Y$.
Run separately that is two distinct $Y$ evaluations at two points.

Both, at their last step, split the leaf multilinears on the highest variable and combine the halves. So they share the same $m-1$ low variables and the same final line-fold:

```
S_1 = sum_{x'} eq(x'; Z) * (Y_0*D_1 + Y_1*D_0)(x') = num_1(Z)   (last fraction layer)
S_2 = sum_{x'} eq(x'; Z) * (D_0*D_1)(x')           = den_1(Z)
S_3 = sum_{x'} (T_0*Y_0 + T_1*Y_1)(x')             = e          (product <T,Y>)
```

These three ride one degree-3 sumcheck over $m-1$ variables, then one shared line-fold over the top variable. Result: a **single** evaluation of $Y$ (and of $T$) at one point — not two.

### The change

- **new:** `crates/ip/src/logup_star/` — `verify.rs` (orchestration), `final_layer.rs` (the batched layer + index-denominator embedding), `output.rs`, `error.rs`, `mod.rs` (protocol docs).
- **changed:** one line in `crates/ip/src/lib.rs` to expose the module.

The module lives in `binius-ip` (next to `fracaddcheck`/`sumcheck`, its dependencies), not the `binius_iop` path in the original ticket text — that naming predates the IP/IOP split.

### Index embedding

Table positions $j$ and committed index values $I[i]$ share one domain.
In the binary field a position embeds through the $\mathrm{GF}(2)$-linear basis, so the table-side denominator multilinear is $J(x) = \sum_t \mathrm{basis}(t)\,x_t$ — verifier-computable, matching the encoding the Spartan verifier already uses.

### Soundness

- The fraction identity catches a wrong $Y$ except with probability $(n+m)/|F|$ (Lemma 2 of [eprint 2025/946](https://eprint.iacr.org/2025/946)).
- The GKR circuits and batched sumcheck add the usual sumcheck soundness error.
- Cross-multiplication assumes the root denominators are nonzero, which holds except with probability $(n+m)/|F|$ over the random $c$.

### Scope

- Verifier only; the prover (BINIUS-153) is a follow-up. Mergeable without it, per the issue.
- End-to-end tests need the prover, so only the pure pieces are unit-tested: the index-denominator embedding (against a brute-force multilinear, $m = 1..6$) and the empty-table guard.

### Verification

- `cargo clippy -p binius-ip --all-targets` — clean
- `cargo test -p binius-ip logup_star` — 2 tests pass


crates/ip/src/logup_star/final_layer.rs
```diff
@@ -0,0 +1,245 @@
...
+use binius_math::{line::extrapolate_line, multilinear::eq::eq_ind};
+
+use super::error::{Error, VerificationError};
+use crate::{
+	channel::IPVerifierChannel,
+	sumcheck::{self, BatchSumcheckOutput},
+};
+
+/// The output of the batched final layer.
+pub(super) struct FinalLayer<F> {
```
jimpo: style: I generally dislike `pub(super)` and `pub(crate)`. Just make it `pub` or private. This is a perfect example of using `pub` but simply not making the module pub in the parent.

I'll document that in the CONTRIBUTING.md

crates/ip/src/logup_star/verify.rs
```diff
@@ -0,0 +1,189 @@
...
+where
+	F: Field + ExtensionField<BinaryField1b>,
+	C: IPVerifierChannel<F>,
+	C::Elem: From<F>,
+{
+	let m = table_n_vars;
+	let n = eval_point.len();
+
+	// The table-side GKR circuit needs at least one variable to split on.
+	if m == 0 {
```
jimpo: This isn't a VerificationError (it's an error on a known input parameter). Just assert `!eval_point.is_empty()` and document the precondition. See https://github.com/binius-zk/binius64/blob/main/CONTRIBUTING.md#precondition-contracts

jimpo (commented):
This looks great, just two small stylistic notes.

crates/ip/src/logup_star/final_layer.rs
tcoratger: Ok noted, just fixed

crates/ip/src/logup_star/verify.rs
tcoratger: Fixed

jimpo (approved):

PR #1605, opened 2026-06-27
BINIUS-156: replace ScaledWideMul with a generic Divide WideMul strategy

benji-potamus:
Closes [BINIUS-156](https://linear.app/jimpo/issue/BINIUS-156/replace-scaledwidemul-with-a-generic-divide-strategy-widemul-recursive).

Makes `Divide<SubU, T>` the single "split the underlier into `SubU` lanes, apply the sub-packing's op, recombine" fallback for **all** of square, invert, **and** widening multiply, then deletes `ScaledWideMul` and uses `Divide` everywhere it was used.

### Approach

`Divide` already carried `Square`/`InvertOrZero` (which stream through `Divisible` and need no lane count). `WideMul` is different: it must *defer reduction*, so it materializes one unreduced product per lane that lanes accumulate (`Add`/`Sub`/`Sum`) and reduce independently — i.e. it needs a concrete `[O; N]` output. Since the crate doesn't enable `generic_const_exprs`, an associated const can't be an array length, so `Divide` gains a **const generic lane count `N`** (defaulting to `1`, ignored by square/invert):

```rust
pub struct Divide<SubU, T, const N: usize = 1>(T, PhantomData<SubU>);

impl<U, SubU, F, const N: usize> WideMul for Divide<SubU, PackedPrimitiveType<U, F>, N>
where U: Divisible<SubU>, PackedPrimitiveType<SubU, F>: WideMul, ...
{
    type Output = LaneWideProduct<<PackedPrimitiveType<SubU, F> as WideMul>::Output, N>;
    // wide_mul: one unreduced product per SubU lane; reduce: per-lane reduce + recombine
}
```

A **single generic impl** covers every case — no per-underlier impls, no cfg-gating. The per-lane deferred products live in the new `LaneWideProduct<O, N>` (the `[O; N]` accumulator that was `ScaledWideProduct`). A `debug_assert_eq!(N, <U as Divisible<SubU>>::N)` guards the lane count, as the old `ScaledWideMul` did.

### Two commits

1. **Add `WideMul` to `Divide`** — purely additive: the generic impl + `LaneWideProduct`. Compiles on its own.
2. **Cut over + delete** — repoint every fallback at `Divide<SubU, T, N>` and delete `ScaledWideMul` / `Scaled2xWideMul` / `Scaled4xWideMul` / `ScaledWideProduct` (the unrelated `Scaled<T>` for `ScaledUnderlier` square/invert/mul is kept).

Fallbacks now:
- GHASH 256/512 (x86_64 non-VPCLMULQDQ + portable) → `Divide<M128, T, {2,4}>`
- Portable AES 256/512 → `Divide<M128, T, {2,4}>`
- **x86_64 AES 256/512 (non-GFNI)** → `Divide<M128, T, 2>` and `Divide<M256, T, 2>` for **all three** ops, making the fallback **recursive** (M512→M256→M128→u8) and replacing the flat `ElementwiseWideMul` + `Divide<u8>`.

### Verification (each commit, full arch matrix)

Portable `cargo test -p binius-field` (199, exercising the `Divide` fallbacks at runtime incl. the recursive AES path) · native clippy `-D warnings` with `--tests --benches --examples` · aarch64 (`+neon,+aes`) clippy · `wasm32-unknown-unknown` build · full `cargo build --workspace` · fmt. The x86_64 GFNI/VPCLMULQDQ runtime paths are CI-gated as usual and are unchanged by this PR (only the `else` fallbacks change).

### DIFF
--- a/crates/field/src/arch/strategies.rs
+++ b/crates/field/src/arch/strategies.rs
@@ -1,7 +1,12 @@
 // Copyright 2024-2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::marker::PhantomData;
+use std::{
+	array,
+	iter::Sum,
+	marker::PhantomData,
+	ops::{Add, AddAssign, Sub, SubAssign},
+};
 
 use bytemuck::TransparentWrapper;
 
@@ -15,16 +20,24 @@ use crate::{
 /// Pairwise strategy. Apply the result of the operation to each packed element independently.
 pub struct PairwiseStrategy;
 
-/// Square wrapper that squares by dividing the underlier into `SubU`-sized lanes, squaring each
-/// lane as a `PackedPrimitiveType<SubU, F>`, and recombining. A generic fallback for packings that
-/// lack a specialized full-width square. The sub-underlier `SubU` is carried as a `PhantomData`
-/// type parameter so the packing type `T` stays last for the macro's `Divide<SubU, $name>` form.
+/// Strategy that splits the underlier into `SubU`-sized lanes, applies the sub-packing
+/// `PackedPrimitiveType<SubU, F>`'s op to each lane, and recombines — a generic fallback for
+/// packings that lack a specialized full-width [`Square`], [`InvertOrZero`], or [`WideMul`]. The
+/// sub-underlier `SubU` is a `PhantomData` parameter so the packing type `T` stays last for the
+/// macro's `Divide<SubU, $name, N>` form.
+///
+/// `N` is the lane count: callers always pass `N = <U as Divisible<SubU>>::N` (or the literal it
+/// works out to). `Square`/`InvertOrZero` stream through [`Divisible`] and ignore `N`, but it is
+/// still required so every `Divide` instantiation names its lane count explicitly. `WideMul` must
+/// defer reduction, so it materializes one unreduced product per lane in an `N`-element
+/// [`LaneWideProduct`] — and an associated const can't be an array length without
+/// `generic_const_exprs`, which is why `N` is a const generic rather than read from `Divisible`.
 #[repr(transparent)]
 #[derive(TransparentWrapper)]
 #[transparent(T)]
-pub struct Divide<SubU, T>(T, PhantomData<SubU>);
+pub struct Divide<SubU, T, const N: usize>(T, PhantomData<SubU>);
 
-impl<U, SubU, F> Square for Divide<SubU, PackedPrimitiveType<U, F>>
+impl<U, SubU, F, const N: usize> Square for Divide<SubU, PackedPrimitiveType<U, F>, N>
 where
 	U: UnderlierType + Divisible<SubU>,
 	SubU: UnderlierType,
@@ -43,7 +56,7 @@ where
 	}
 }
 
-impl<U, SubU, F> InvertOrZero for Divide<SubU, PackedPrimitiveType<U, F>>
+impl<U, SubU, F, const N: usize> InvertOrZero for Divide<SubU, PackedPrimitiveType<U, F>, N>
 where
 	U: UnderlierType + Divisible<SubU>,
 	SubU: UnderlierType,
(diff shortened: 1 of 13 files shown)


crates/field/src/arch/x86_64/packed_aes_256.rs
```diff
@@ -9,8 +9,11 @@ cfg_if! {
 		pub type AesSquare32x<T> = crate::arch::ReuseMultiply<T>;
 		pub type AesInvert32x<T> = super::gfni::gfni_arithmetics::Gfni<T>;
 	} else {
-		pub type AesWideMul32x<T> = crate::arch::ElementwiseWideMul<T>;
-		pub type AesSquare32x<T> = crate::arch::Divide<u8, T>;
-		pub type AesInvert32x<T> = crate::arch::Divide<u8, T>;
+		// Divide into two `M128` lanes (the 16×8b AES packing) for all three ops; that packing in
+		// turn divides to `u8`, so the fallback is recursive rather than flat over bytes.
+		use crate::arch::{Divide, M128};
+		pub type AesWideMul32x<T> = Divide<M128, T, 2>;
```
jimpo: Continue using u8 as the subunderlier in this case, to remove the recursion step.

crates/field/src/arch/x86_64/packed_aes_256.rs
```diff
@@ -9,8 +9,11 @@ cfg_if! {
...
+		pub type AesSquare32x<T> = Divide<M128, T>;
```
jimpo: Use `Divide<M128, T, 2>` specifying the value N (well, actually `Divide<u8, T, 32>` per the comment above). Even though the const N doesn't get used in this case, it's confusing without it. In fact, remove the default value for N in Divide, and always pass in N = <T as Divisible<U>>::N.

crates/field/src/arch/strategies.rs
```diff
@@ -15,16 +20,23 @@ use crate::{
...
+/// `Square`/`InvertOrZero` stream through [`Divisible`] and ignore `N`. `WideMul` must defer
+/// reduction, so it materializes one unreduced product per lane in an `N`-element
+/// [`LaneWideProduct`]; `N` is the lane count and must equal `<U as Divisible<SubU>>::N`. It is a
+/// const generic (defaulting to `1`) rather than read from the associated const because an
+/// associated const cannot be an array length without `generic_const_exprs`.
 #[repr(transparent)]
 #[derive(TransparentWrapper)]
 #[transparent(T)]
-pub struct Divide<SubU, T>(T, PhantomData<SubU>);
+pub struct Divide<SubU, T, const N: usize = 1>(T, PhantomData<SubU>);
```
jimpo: remove the default = 1. Caller should always provide N = <T as Divisible<SubU>>::N (or the constant that it happens to be, not necessarily the expression).

jimpo (changes requested):

crates/field/src/arch/x86_64/packed_aes_256.rs
benji-potamus: Done — the AES fallbacks now divide straight into `u8` lanes (`Divide<u8, T, 32>` for 32x, `Divide<u8, T, 64>` for 64x, `Divide<u8, T, 16>` for 16x), with no `M128`/`M256` intermediate, so the path is flat rather than recursive. This made `ElementwiseWideMul` exactly equal to `Divide<u8, _, N>`'s widening multiply, so I deleted it and routed every AES square/invert/widen fallback (both arches) through `Divide<u8, T, N>`.

crates/field/src/arch/x86_64/packed_aes_256.rs
benji-potamus: Done — `AesWideMul32x = Divide<u8, T, 32>` (and the `N` default is removed — see the strategies.rs thread).

crates/field/src/arch/strategies.rs
benji-potamus: Done — removed the `= 1` default; `Divide<SubU, T, const N: usize>` now requires `N` at every instantiation. Each call site passes the literal lane count: `Divide<u8, T, {16,32,64}>` for AES, `Divide<M128, T, {2,4}>` for GHASH — including the `Square`/`InvertOrZero` aliases that ignore `N`. (Also folded the two commits into one, since the cutover is now a single cohesive change.)

jimpo (approved):

PR #1614, opened 2026-06-28
BINIUS-152: Process the univariate round inner loop in strides

tcoratger:
Closes [BINIUS-152](https://linear.app/jimpo/issue/BINIUS-152/process-the-univariate-round-inner-loop-in-strides).

Computes the same univariate-skip round message, but walks the output domain in 4 cache-resident passes instead of one wide pass. Output is bit-identical — no protocol change, no new soundness assumption.

### The problem

The hot loop of the AND-reduction round is the univariate low-degree extension.
For each witness word it reads a precomputed NTT table and XOR-accumulates 64 output evaluations.
The table holds all 64 outputs per entry: $8 \text{ bytes} \times 256 \text{ values} \times 64 \text{ bytes} = 128\text{ KB}$.
On the benchmark M4 that is the *entire* 128 KB L1 data cache — the table evicts everything else the loop touches.

### The idea

Never hold all 64 outputs at once. Tile them into 4 limbs of 16 and process one limb per pass.
A width-16 packed field is exactly one 128-bit register, and one limb's table slab is only 32 KB:

```
table: [limb][byte_position][byte_value] -> 16 evals     (limb-major, transposed)
slab(limb) = 8 * 256 * 16 bytes = 32 KB                  (contiguous, fits L1)
```

Each pass reads only its own 32 KB slab. Same total arithmetic, same total byte traffic — only the live working set shrinks to a quarter.

### Per pass

- **before:** one width-64 pass, 128 KB table streamed, 1 KB live accumulator
- **after:** 4 passes of width-16, 32 KB slab each, 256 B live accumulator

→ working set per pass drops from all of L1 to a quarter of it.

### The change

```
NTTLookup<P>(Box<[[P; 256]; 8]>)          // one 64-wide entry, P = 64x8b
NTTLookup<P>(Box<[[[P; 256]; 8]; 4]>)     // limb-major, P = 16x8b

// hot loop: one wide pass  ->  for limb in 0..4 { read slab; fill acc[16*limb .. 16*limb+16] }
```

### How this relates to Flock §4.2 — and what it is *not*

§4.2 motivates the unified lookup table and notes its 128 KB footprint sits right at the M4's L1 boundary.
This PR does **not** change the table's contents or the LDE math — it only re-lays-out and re-orders the reads so each pass is L1-resident.

One deliberate divergence from the ticket: the ticket text and a stale in-tree doc-comment disagreed on layout. The transposed **limb-major** layout (`[[[P;256];8];4]`) is the one that makes a pass touch a contiguous 32 KB slab; the limb-*innermost* alternative would pull a 64-byte line per lookup while using 16 bytes, defeating the purpose. The benchmark confirms limb-major.

### Soundness — preserved

- The round message is bit-identical to today's, verified against a from-scratch NTT and an independent sum claim.
- No challenge, domain, or transcript change.

### Scope

- **changed:** `ntt_lookup.rs` (table layout + width-16), `sumcheck_round_messages.rs` (4-pass hot loop)
- **call sites:** `prove.rs`, `prover.rs`, the bench, and tests switch `PackedAESBinaryField64x8b -> 16x8b`
- **left untouched:** the LDE/NTT math, the protocol, the verifier

### Verification

- nightly `fmt --all`, `cargo check --workspace --all-targets`, `clippy -p binius-prover --all-targets` — clean
- `cargo test -p binius-prover and_reduction` — table-vs-NTT, hot-loop-vs-sum-claim, and full prover↔verifier round-trip all pass

### Benchmark

`univariate_round_message 2^21` on M4 Max, criterion before/after baselines (p ≈ 0):

| variant | before | after | change |
|---|---|---|---|
| single-thread (per-core L1 signal) | 121.75 ms | 84.35 ms | **−30.7%** (+44% thrpt) |
| multi-thread (10 cores) | 16.34 ms | 14.45 ms | **−11.6%** (+13% thrpt) |

Single-thread is the honest measure of the cache effect: a flat 30% kernel speedup. The multi-thread gain is smaller because at 10 cores the loop is partly memory-bandwidth-bound, which the L1 trick doesn't address.

### DIFF
--- a/crates/prover/benches/and_reduction.rs
+++ b/crates/prover/benches/and_reduction.rs
@@ -3,7 +3,7 @@ use std::{iter, iter::repeat_with};
 
 use binius_core::word::Word;
 use binius_field::{
-	Field, PackedAESBinaryField64x8b, Random,
+	Field, PackedAESBinaryField16x8b, Random,
 	linear_transformation::{
 		BytewiseLookupTransformationFactory, LinearTransformationFactory,
 		OutputWrappingTransformationFactory,
@@ -51,11 +51,11 @@ fn bench(c: &mut Criterion) {
 		.reduce_dim(prover_message_domain.dim() - 1)
 		.isomorphic();
 
-	let ntt_lookup = NTTLookup::<PackedAESBinaryField64x8b>::new(&prover_message_domain);
+	let ntt_lookup = NTTLookup::<PackedAESBinaryField16x8b>::new(&prover_message_domain);
 
 	let mut group = c.benchmark_group("evaluate");
 	group.bench_function("NTT lookup precompute", |bench| {
-		bench.iter(|| NTTLookup::<PackedAESBinaryField64x8b>::new(&prover_message_domain));
+		bench.iter(|| NTTLookup::<PackedAESBinaryField16x8b>::new(&prover_message_domain));
 	});
 
 	group.throughput(Throughput::Elements(1 << log_words));
--- a/crates/prover/src/and_reduction/ntt_lookup.rs
+++ b/crates/prover/src/and_reduction/ntt_lookup.rs
@@ -24,25 +24,40 @@
 //! - **Output**: 64 evaluations of the polynomial at specified domain points
 //! - **Optimization**: Precomputes all 256 possible evaluations for each 8-bit position
 //!
-//! ## Performance
+//! ## Striding the output domain
 //!
-//! By precomputing the lookup tables, the NTT operation is reduced to:
-//! - 8 table lookups (one per byte)
-//! - 7 packed field additions
+//! - The 64 output evaluations are split into 4 limbs of 16 evaluations.
+//! - One limb is exactly the 16 scalars held by one 128-bit packed field.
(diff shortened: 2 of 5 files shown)


jimpo:
I tried this. It's better an ARM (for now), but worse on x64. But more importantly, I think there are better ways to reduce register spilling, which I'm working on.

tcoratger:
> I tried this. It's better an ARM (for now), but worse on x64. But more importantly, I think there are better ways to reduce register spilling, which I'm working on.

Mhhh interesting! Feel free to close this one or take anything from it! I've no x64 to test it, but indeed if this is slower on it probably worth checking something else!

jimpo:
@tcoratger I'm pretty sure that if you try this on top of main, it will no longer be faster (probably slower). But if I'm wrong, let me know!

tcoratger:
> @tcoratger I'm pretty sure that if you try this on top of main, it will no longer be faster (probably slower). But if I'm wrong, let me know!

Indeed, I've just checked locally and this PR is slower than `main` now. Let me close.

PR #1622, opened 2026-06-29
BINIUS-164: IntMul Phase-5 forward leaf reconstruction

benji-potamus:
Closes [BINIUS-164](https://linear.app/jimpo/issue/BINIUS-164) — first conformance fix under [BINIUS-163](https://linear.app/jimpo/issue/BINIUS-163) (make IntMul match the overhauled `sec:intmul`).

## What

Spec **Phase 5** (leaf reconstruction): the prover sends the raw per-bit evals `a(i,r_x), c_lo(i,r_x), c_hi(i,r_x)` and the verifier **reconstructs the leaf selectors forward** — `sel(z, g^{2^i}) = z·(g^{2^i}−1)+1` — then binds those to the GKR-verified layer-0 product claims.

The code did the opposite: the prover sent the **selector** values and the verifier **inverted** them (`recover_selectors`/`normalize_a_c_exponent_evals`) to recover the raw bits. So the transcript carried selectors, not raw bits, and the binding direction was reversed.

This swaps to the spec direction:
- **verifier** (`verify_phase_5`): reads the raw per-bit evals, reconstructs the selectors via the new `reconstruct_selectors` (forward), binds the reconstructed selectors to the bivariate-product claims, uses the raw bit-0 evals for the wraparound-parity check, and returns the raw evals. No inversion on the verifier.
- **prover** (`phase5`): sends the raw per-bit evals instead of the selectors.
- **common.rs**: adds the forward `reconstruct_selectors`; `normalize_a_c_exponent_evals`/`recover_selectors` stay (now prover-only).

`IntMulOutput` (raw per-bit evals at the shared point) is **unchanged**, so the downstream shift reduction is unaffected — this is a transcript/binding-shape change, not an output change. For an honest prover the output values are identical to before.

Note: the prover still derives the raw evals by inverting its own selector evals (which the GKR already produced) — that's one cheap field inversion per bit, versus re-evaluating the witness bit-columns. The spec-relevant property (transcript carries raw bits; verifier reconstructs forward) is what changed.

## Test

`prove_and_verify` roundtrip passes (it checks both the per-bit output↔witness consistency and prover↔verifier agreement through the new path); `+ fmt + clippy -D warnings` clean on `binius-verifier`/`binius-prover`. No negative test added — none existed, the binding structure/soundness is unchanged, and a byte-level tamper test is awkward against the flat transcript; flagging as a possible follow-up.

This is the first of a sequence (next: BINIUS-165, recombine `b` via `r_I^b`).

### DIFF
--- a/crates/verifier/src/protocols/intmul/common.rs
+++ b/crates/verifier/src/protocols/intmul/common.rs
@@ -198,3 +198,56 @@ fn recover_selectors<F: FieldOps>(selecteds: Vec<F>, inv_factors: &[F]) -> Vec<F
 		})
 		.collect()
 }
+
+/// Reconstruct the "selected" leaf evaluations from the raw per-bit evaluations.
+///
+/// This is the forward direction of [`normalize_a_c_exponent_evals`]: given the raw bit
+/// evaluations $a(i, r), c_{\textsf{lo}}(i, r), c_{\textsf{hi}}(i, r)$, it returns the selected
+/// leaf values
+///
+/// * $\textsf{select}(a(i, r), g^{2^i})$,
+/// * $\textsf{select}(c_{\textsf{lo}}(i, r), g^{2^i})$,
+/// * $\textsf{select}(c_{\textsf{hi}}(i, r), g^{2^{i + k}})$,
+///
+/// for $i \in \{0, \ldots, 2^k - 1\}$, where $\textsf{select}(S, V) = S \cdot (V - 1) + 1$.
+///
+/// The verifier reconstructs these forward from the prover's raw evaluations and binds them to the
+/// GKR-verified leaf-product claims, rather than receiving them and inverting. $g$ is a constant
+/// multiplicative generator of the field $F$.
+pub fn reconstruct_selecteds<F, E>(
+	k: usize,
+	a_evals: &[E],
+	c_lo_evals: &[E],
+	c_hi_evals: &[E],
+) -> [Vec<E>; 3]
+where
+	F: Field,
+	E: FieldOps<Scalar = F> + From<F>,
+{
+	assert_eq!(a_evals.len(), 1 << k);
+	assert_eq!(c_lo_evals.len(), 1 << k);
+	assert_eq!(c_hi_evals.len(), 1 << k);
+
+	// powers[j] = g^{2^j}, for j in 0..2^{k+1}.
+	let powers: Vec<E> = iterate(F::MULTIPLICATIVE_GENERATOR, |g| g.square())
+		.take(2 << k)
+		.map(E::from)
+		.collect();
+	let (lo_powers, hi_powers) = powers.split_at(1 << k);
+
+	[
+		apply_selectors(a_evals, lo_powers),
+		apply_selectors(c_lo_evals, lo_powers),
+		apply_selectors(c_hi_evals, hi_powers),
+	]
+}
+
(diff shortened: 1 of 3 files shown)


crates/verifier/src/protocols/intmul/common.rs
```diff
@@ -198,3 +198,56 @@ fn recover_selectors<F: FieldOps>(selecteds: Vec<F>, inv_factors: &[F]) -> Vec<F
...
+/// * $\textsf{select}(a(i, r), g^{2^i})$,
+/// * $\textsf{select}(c_{\textsf{lo}}(i, r), g^{2^i})$,
+/// * $\textsf{select}(c_{\textsf{hi}}(i, r), g^{2^{i + k}})$,
+///
+/// for $i \in \{0, \ldots, 2^k - 1\}$, where $\textsf{select}(S, V) = S \cdot (V - 1) + 1$.
+///
+/// The verifier reconstructs the selectors forward from the prover's raw evaluations and binds
+/// them to the GKR-verified leaf-product claims, rather than receiving the selectors and inverting
+/// them. $g$ is a constant multiplicative generator of the field $F$.
+pub fn reconstruct_selectors<F, E>(
```
jimpo: nit: rename to `reconstruct_leaves` or `reconstruct_selecteds`. The "selected" is the output of the select expression, the inputs are the "selectors".

jimpo (approved):

crates/verifier/src/protocols/intmul/common.rs
```diff
@@ -198,3 +198,56 @@ fn recover_selectors<F: FieldOps>(selecteds: Vec<F>, inv_factors: &[F]) -> Vec<F
...
+where
+	F: Field,
+	E: FieldOps<Scalar = F> + From<F>,
+{
+	assert_eq!(a_evals.len(), 1 << k);
+	assert_eq!(c_lo_evals.len(), 1 << k);
+	assert_eq!(c_hi_evals.len(), 1 << k);
+
+	// factors[j] = g^{2^j} - 1, for j in 0..2^{k+1}.
+	let factors: Vec<E> = iterate(F::MULTIPLICATIVE_GENERATOR, |g| g.square())
```
jimpo: Aesthetically, I'd prefer to just compute and collect the generator powers (not subtracting ONE), then subtract one in `apply_selectors`. It's easier to understand.

crates/verifier/src/protocols/intmul/common.rs
benji-potamus: Done in 06a0faca — renamed `reconstruct_selectors` → `reconstruct_selecteds` (and updated the doc/comment terminology), since the returned values are the `select` *outputs*.

crates/verifier/src/protocols/intmul/common.rs
benji-potamus: Done in 06a0faca — `reconstruct_selecteds` now collects the generator powers `g^{2^i}` directly, and `apply_selectors(z, V)` does the `z·(V−1)+1` (subtracting one inside). Reads more clearly.

benji-potamus:
This is approved + all CI green + `mergeable_state: clean` and ready to go. I can't complete the merge from here, though — the repo requires the **merge queue**, and the REST merge endpoint returns `405 "Changes must be made through the merge queue"` (no programmatic enqueue available to me). Could you click **Merge when ready** to enqueue it? `main` just advanced via #1619, but the queue will rebase this on top. Thanks!

PR #1623, opened 2026-06-29
BINIUS-122: build the batched BitAnd witness from the 2-D value table

tcoratger:
Closes [BINIUS-122](https://linear.app/jimpo/issue/BINIUS-122/build-bitand-witness).

Builds the BitAnd check's operand columns for the M4 batch witness — the 2-D analogue of the single-instance `build_bitand_witness`.

### The problem

The single-instance prover turns the AND constraints into three operand columns `A`, `B`, `C` by evaluating each constraint against one witness, and the BitAnd reduction checks `A & B == C` row by row.

The M4 batch holds `K = 2^k` instances of one circuit in a single value table (one batch dimension `k = log_instances`).
Nothing yet builds those operand columns for the batch — so the BitAnd reduction has nothing to consume.

### The idea

Evaluate every AND constraint against every instance, and stack the rows instance-major — the same layout the value table already uses.

```text
       instance 0        instance 1            instance K-1
 A: [ a_0 .. a_{n-1} ][ a_0 .. a_{n-1} ] ... [ a_0 .. a_{n-1} ]
 B: [ b_0 .. b_{n-1} ][ b_0 .. b_{n-1} ] ... [ b_0 .. b_{n-1} ]
 C: [ c_0 .. c_{n-1} ][ c_0 .. c_{n-1} ] ... [ c_0 .. c_{n-1} ]
     \___ n_and ___/
```

The row index splits cleanly:

```text
row = instance * n_and + local_constraint
```

- The high `log_instances` bits select the instance.
- The low bits select the constraint within that instance.
- `n_and` is a power of two for a prepared constraint system, so the batch is a clean hypercube whose high coordinates are the instance index.
- The reduction reads each column as a multilinear over `log_instances + log2(n_and)` bits.

### The change

- **new:** `crates/m4-prover/src/bitand.rs` — `BatchAndCheckWitness::{build, a, b, c, len, is_empty, into_columns}`.
- **changed:** `crates/m4-prover/src/lib.rs` exports `BatchAndCheckWitness`.

An operand is a XOR of shifted committed values. Each instance's block is filled in parallel from that instance's committed-word slice:

```text
A[instance * n_and + j] = eval_operand( table.instance(instance), constraint_j.a )
```

### Performance

Operands are evaluated straight off each instance's committed-word slice, not a reconstructed value vector.

- A value-vector reconstruction allocates and copies a fresh buffer per instance.
- At `K = 2^18` instances that is ~260k allocations purely to read operands.

→ The slice path does **zero** extra per-instance allocation, and the instances fill in parallel over disjoint row blocks.

### Soundness

- The columns are bit-identical to evaluating each instance through the single-instance path, so this changes no protocol and adds no assumption.
- Reading the slice in place is in range because constraints reference only committed values, whose count equals the slice length — they never touch the dropped scratch tail.

### Scope

- This builds the witness only.
- It does not run the reduction or sample challenges — that is the parent BINIUS-121.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-m4-prover --all-targets`, nightly `fmt` — clean.
- `cargo test -p binius-m4-prover` — 15 tests pass, including:
  - a proptest pinning the slice-based evaluator to the core value-vec evaluator on random inputs,
  - the same proptest asserting `A & B == C` holds on every row, padding included,
  - shape / instance-major-layout and `K = 1` degenerate-batch checks.

### DIFF
--- a/crates/m4-prover/src/bitand.rs
+++ b/crates/m4-prover/src/bitand.rs
@@ -1,0 +1,340 @@
+// Copyright 2025 Irreducible Inc.
+
+//! The batched BitAnd-check witness built from a populated batch value table.
+
+use binius_core::{
+	constraint_system::{AndConstraint, ShiftedValueIndex},
+	verify::eval_shifted_word,
+	word::Word,
+};
+use binius_utils::rayon::prelude::*;
+
+use crate::ValueTable;
+
+/// The operand columns of the BitAnd check for a whole batch of instances.
+///
+/// The BitAnd check works on three columns of words, one row per AND constraint.
+/// It enforces the bitwise relation `A & B == C` on every row.
+///
+/// This holds those three columns for a batch of `K = 2^log_instances` instances at once.
+/// The rows are stacked in instance-major order, the same layout the batch witness uses:
+///
+/// ```text
+///         instance 0         instance 1            instance K-1
+///   A: [ a_0 .. a_{n-1} ][ a_0 .. a_{n-1} ] ... [ a_0 .. a_{n-1} ]
+///   B: [ b_0 .. b_{n-1} ][ b_0 .. b_{n-1} ] ... [ b_0 .. b_{n-1} ]
+///   C: [ c_0 .. c_{n-1} ][ c_0 .. c_{n-1} ] ... [ c_0 .. c_{n-1} ]
+///       \___ n_and ___/
+/// ```
+///
+/// The row index splits cleanly into the instance and the per-instance constraint:
+///
+/// ```text
+/// row = instance * n_and + local_constraint
+/// ```
+///
+/// - The high `log_instances` bits select the instance.
+/// - The low bits select the constraint within that instance.
+/// - The reduction reads each column as a multilinear over `log_instances + log(n_and)` bits.
+///
+/// `n_and` is a power of two when the constraints come from a prepared constraint system.
+/// So the batch forms a clean hypercube whose high coordinates are the instance index.
+#[derive(Clone, Debug)]
+pub struct BatchAndCheckWitness {
+	/// Operand `A` of every constraint of every instance, instance-major.
+	a: Vec<Word>,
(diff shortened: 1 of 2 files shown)


crates/m4-prover/src/bitand.rs
```diff
@@ -0,0 +1,340 @@
...
+		let total = table.n_instances() * n_and;
+
+		// One column each for the three operands, laid out instance-major.
+		let mut a = vec![Word::ZERO; total];
+		let mut b = vec![Word::ZERO; total];
+		let mut c = vec![Word::ZERO; total];
+
+		// Fill one instance's contiguous block of rows per parallel task.
+		// The blocks are disjoint, so the instances never contend for the same rows.
+		a.par_chunks_mut(n_and.max(1))
```
jimpo: Are the `max(1)`'s necessary? `and_constraints.len()` is always a power of two, so at least 1 anyway. And the n_instances should also be a power of 2.

crates/m4-prover/src/bitand.rs
```diff
@@ -0,0 +1,340 @@
...
+		&self.c
+	}
+
+	/// The number of rows in each column: `K * n_and`.
+	pub fn len(&self) -> usize {
+		self.a.len()
+	}
+
+	/// Whether the witness has no rows.
+	pub fn is_empty(&self) -> bool {
```
jimpo: This shouldn't be possible.

crates/m4-prover/src/bitand.rs
tcoratger: Fair point, I've removed that, added some assertions at the beginning of the function + unit test to be sure the conditions are ok to get rid of this.

crates/m4-prover/src/bitand.rs
tcoratger: Removed

jimpo (approved):

PR #1625, opened 2026-06-29
BINIUS-165: IntMul recombine the b exponent to a single claim via r_I^b

benji-potamus:
Closes [BINIUS-165](https://linear.app/jimpo/issue/BINIUS-165) — second conformance fix under [BINIUS-163](https://linear.app/jimpo/issue/BINIUS-163) (make IntMul match the overhauled `sec:intmul`), audit divergence **D-V2**. Stacks conceptually after #1622 (D-V1); both reshape Phase 5.

## What

**Spec (Phase 2 → 4 → 5):** after the variable-base subprotocol yields the `2^k` per-bit exponent claims `b̃(i, r₂)`, the verifier samples a recombination point `r_I^b ∈ K^k` and **collapses them into a single claim** `b̃(r_I^b, r₂) = Σ_i eq_k(r_I^b, i)·b̃(i, r₂)`, carrying **one** exponent claim (not `2^k`) into Phases 4–5. Phase 5 then point-switches that single claim to the shared point `r_x`, and binds the prover's raw per-bit evals via `Σ_i eq_k(r_I^b, i)·b̃(i, r_x) = b̃(r_I^b, r_x)`.

**Before:** the code carried all `2^k` `b_evals` from `verify_phase_3` and ran a `2^k`-claim rerandomization in `verify_phase_5` (prover-side `RerandMlecheckProver` over the 64 bit-columns); there was no `r_I^b` sample at all. Soundness-equivalent, but not the spec's transcript shape.

**This PR:**
- **verifier** (`verify_phase_3`): samples `r_I^b` right after the Phase-3 sent evals and recombines the `2^k` claims into `b_recomb = b̃(r_I^b, r₂)` (via `evaluate_inplace_scalars`). `verify_phase_5` carries the single claim through the batch and binds the `2^k` raw per-bit evals the prover sends via `Σ eq(r_I^b,i)·b̃(i,r_x)`.
- **prover** (`phase3`): samples the matching `r_I^b` (same transcript position) and recombines. `phase5`: folds the `b` bit-columns into one field multilinear `B(x) = Σ_i eq(r_I^b,i)·b(i,x)` and re-randomizes it with a **single-claim** `MultilinearEvalProver` (the large-field analogue of the bit-column `RerandMlecheckProver`) in place of the `2^k`-claim rerandomization. It still sends the `2^k` raw per-bit `b(i, r_x)` for Phase-5 leaf reconstruction (computed by evaluating each bit-column at `r_x`).
- **common.rs**: `Phase3Output` carries `r_ib` + the recombined `b_recomb` instead of the `2^k`-vec `b_evals`.

`IntMulOutput` (raw per-bit evals at the shared point) is **unchanged** — the downstream shift reduction is unaffected. For an honest prover the output values are identical to before; this is a transcript/binding-shape change.

Note: materializing the folded `B` in the field gives up the bit-switchover memory optimization the old `2^k`-column rerandomization had (one field multilinear instead of bit-columns). Flagged with a `TODO` for a possible 1-bit-optimized fold; the spec-relevant property (single recombined `b` claim through Phases 4–5) is what changed.

## Test

`prove_and_verify` roundtrip passes (gates prover↔verifier transcript agreement and per-bit output↔witness consistency through the new path); a `debug_assert` checks the recombined rerandomization eval `B(r_x)` equals `Σ eq(r_I^b,i)·b(i,r_x)`. `+ fmt + clippy -D warnings + rustdoc -D warnings` clean on `binius-verifier`/`binius-prover`.

### DIFF
--- a/crates/prover/src/protocols/intmul/prove.rs
+++ b/crates/prover/src/protocols/intmul/prove.rs
@@ -2,6 +2,7 @@
 
 use std::marker::PhantomData;
 
+use binius_core::word::Word;
 use binius_field::{BinaryField, FieldOps, PackedField};
 use binius_ip_prover::{
 	channel::IPProverChannel,
@@ -10,9 +11,12 @@ use binius_ip_prover::{
 use binius_math::{
 	field_buffer::FieldBuffer,
 	inner_product::inner_product_buffers,
-	multilinear::{eq::eq_ind_partial_eval, evaluate::evaluate},
+	multilinear::{
+		eq::{eq_ind_partial_eval, eq_ind_partial_eval_scalars},
+		evaluate::evaluate,
+	},
 };
-use binius_utils::{bitwise::Bitwise, rayon::prelude::*};
+use binius_utils::rayon::prelude::*;
 use binius_verifier::protocols::{
 	intmul::common::{
 		IntMulOutput, Phase1Output, Phase2Output, Phase3Output, Phase4Output, frobenius_twist,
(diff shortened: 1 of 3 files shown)


crates/prover/src/protocols/intmul/prove.rs
```diff
@@ -392,17 +406,28 @@ where
...
 
-		let b_rerand_prover = RerandMlecheckProver::<P, _>::new(
-			b_eval_point,
-			b_evals_pre,
-			b_exponents,
-			self.switchover,
-		)?;
-		let b_sumcheck_prover = MleToSumCheckDecorator::new(b_rerand_prover);
+		let b_tensor = eq_ind_partial_eval_scalars::<F>(r_ib);
+		let b_folded_scalars =
```
jimpo: Use `fold_words` to do this loop, packing directly into a `FieldBuffer<P>`. Edit the comment above -- 1-bit optimized switchover wouldn't help, the MultilinearEvalProver is good here.

crates/prover/src/protocols/intmul/prove.rs
```diff
@@ -417,14 +442,32 @@ where
...
 			.expect("batch_prove with 3 provers returns 3 multilinear_evals vecs");
 
 		assert_eq!(bivariate_evals.len(), 3 << log_bits);
 		assert_eq!(lsb_evals.len(), 3);
-		assert_eq!(b_evals.len(), 1 << log_bits);
+		assert_eq!(b_recomb_evals.len(), 1);
+
+		// The prover still sends the 2^k raw per-bit evals b(i, r_x) for Phase-5 leaf
+		// reconstruction; the verifier binds them via sum_i eq(r_I^b, i) * b(i, r_x) = B(r_x).
+		let b_evals = (0..1 << log_bits)
```
jimpo: Use `fold_across_words`, introduced in #1631

jimpo (changes requested):

crates/prover/src/protocols/intmul/prove.rs
benji-potamus: Done — replaced the manual accumulate with `fold_words::<_, P>(&b_words, &b_tensor)` packing straight into a `FieldBuffer<P>` (converting `b_exponents: &[u64]` to `Vec<Word>` first, since `Word` isn't `Bitwise`), and dropped the switchover TODO comment.

crates/prover/src/protocols/intmul/prove.rs
benji-potamus: Done — the 2^k per-bit output evals `b̃(i, r_x)` now come from `fold_across_words::<_, P>(&b_words, &challenges)` (rebased onto main to pick up #1631) instead of the per-bit `evaluate` loop.

jimpo (approved):

PR #1630, opened 2026-06-29
Remove dead abstractions from binius-hash

tcoratger:
Three small cleanups in `binius-hash`, each deleting an abstraction that no production path exercises.

### The problem

`binius-hash` carries three pieces of API that look load-bearing but aren't:

- `MultiDigest<N>` + `ParallelMultidigestImpl` — a trait pair for N-way SIMD batch hashing.
  Nothing in the workspace implements it for a real hash; the only impl is a mock in tests.
  It is vestigial from the now-removed Vision hash.
- `CompressionFunction` — a marker trait sitting on top of `PseudoCompressionFunction`.
  It is implemented once (for the SHA-256 compressor) and never used as a bound anywhere.
- `ParallelCompressionAdaptor` exposes its inner compressor twice — once as an inherent method, once as the trait method of the same name.

Unexercised generality still has to be read, kept compiling, and reasoned about during audits — for no caller.

### The change

- **`parallel_digest.rs`** — delete `MultiDigest` and `ParallelMultidigestImpl` (trait, wrapper, impls) plus the dead test scaffolding (the mock multi-digest and its helpers). Keep `ParallelDigest` and `ParallelDigestAdapter`.
- **`compress.rs`** — delete the `CompressionFunction` marker trait. `PseudoCompressionFunction` stays.
- **`sha256.rs`** — drop the empty `impl CompressionFunction for Sha256Compression` and its import.
- **`parallel_compression.rs`** — drop the redundant inherent accessor; the trait method remains canonical.
- **`lib.rs`** — stop re-exporting the removed items.

### Scope

- **removed:** `MultiDigest`, `ParallelMultidigestImpl`, `CompressionFunction`, one duplicate inherent accessor, dead mock tests
- **kept:** `ParallelDigest`, `ParallelDigestAdapter`, `PseudoCompressionFunction`, `ParallelPseudoCompression`, `ParallelCompressionAdaptor`
- net: +8 / −252 across 5 files
- no downstream crate referenced any removed item — only `lib.rs` re-exported them

### Note

`MultiDigest` was the intended hook for a future SIMD multi-way SHA-256.
Removing it now is the YAGNI call: it comes back alongside its first real backend, not before.

### Verification

- `cargo clippy -p binius-hash --all-targets --all-features` — clean
- `cargo test -p binius-hash --all-features` — 5 passed
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-hash --no-deps` — clean

### DIFF
--- a/crates/hash/src/compress.rs
+++ b/crates/hash/src/compress.rs
@@ -9,14 +9,10 @@
 //!
 //! [Plonky3]: <https://github.com/plonky3/plonky3>
 
-/// An `N`-to-1 compression function, collision-resistant in a hash tree setting.
+/// An `N`-to-1 compression function, collision-resistant in a hash-tree setting.
 ///
-/// Unlike `CompressionFunction`, it may not be collision-resistant in general.
-/// Instead it is only collision-resistant in hash-tree like settings where
-/// the preimage of a non-leaf node must consist of compression outputs.
+/// - It is not assumed collision-resistant for arbitrary inputs.
+/// - Collision resistance holds only where every preimage is itself a compression output.
 pub trait PseudoCompressionFunction<T, const N: usize>: Clone {
 	fn compress(&self, input: [T; N]) -> T;
 }
-
-/// An `N`-to-1 compression function.
-pub trait CompressionFunction<T, const N: usize>: PseudoCompressionFunction<T, N> {}
--- a/crates/hash/src/compress.rs
+++ b/crates/hash/src/compress.rs
@@ -9,10 +9,13 @@
 //!
 //! [Plonky3]: <https://github.com/plonky3/plonky3>
 
-/// An `N`-to-1 compression function, collision-resistant in a hash-tree setting.
+/// An `N`-to-1 compression function used to build the inner nodes of a hash tree.
 ///
-/// - It is not assumed collision-resistant for arbitrary inputs.
-/// - Collision resistance holds only where every preimage is itself a compression output.
-pub trait PseudoCompressionFunction<T, const N: usize>: Clone {
+/// It folds `N` values into a single value of the same type, so it can be applied level by level:
+/// the children of an inner node are compressed to produce that node.
+pub trait CompressionFunction<T, const N: usize>: Clone {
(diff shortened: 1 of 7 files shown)


crates/hash/src/compress.rs
```diff
@@ -9,14 +9,10 @@
...
 
-/// An `N`-to-1 compression function, collision-resistant in a hash tree setting.
+/// An `N`-to-1 compression function, collision-resistant in a hash-tree setting.
 ///
-/// Unlike `CompressionFunction`, it may not be collision-resistant in general.
-/// Instead it is only collision-resistant in hash-tree like settings where
-/// the preimage of a non-leaf node must consist of compression outputs.
+/// - It is not assumed collision-resistant for arbitrary inputs.
+/// - Collision resistance holds only where every preimage is itself a compression output.
 pub trait PseudoCompressionFunction<T, const N: usize>: Clone {
```
jimpo: Just rename this to `CompressionFunction` now

crates/hash/src/parallel_digest.rs
```diff
@@ -1,70 +1,16 @@
...
 use crate::HashBuffer;
 
-/// An object that efficiently computes `N` instances of a cryptographic hash function
-/// in parallel.
-///
-/// This trait is useful when there is a more efficient way of calculating multiple digests at once,
-/// e.g. using SIMD instructions. It is supposed that this trait is implemented directly for some
-/// digest and some fixed `N` and passed as an implementation of the `ParallelDigest` trait which
-/// hides the `N` value.
-pub trait MultiDigest<const N: usize>: Clone {
```
jimpo, line 25: hmm, I think we actually should keep this. It'll be useful for BINIUS-167

crates/hash/src/compress.rs
tcoratger: Fixed

crates/hash/src/parallel_digest.rs
tcoratger: Fixed

jimpo (approved):

PR #1631, opened 2026-06-29
BINIUS-166: add fold_across_words

benji-potamus:
Implements [BINIUS-166](https://linear.app/jimpo/issue/BINIUS-166/fold-across-words-function).

## Summary

Adds `fold_across_words`, which folds the word matrix against the tensor expansion of a point **by bit position** — the transpose of `fold_words`. For `n` challenges and `2^n` words, it returns `[F; WORD_SIZE_BITS]` where entry `i` is the inner product of the tensor expansion of the point with the bits at position `i` across all words.

## Algorithm

Following the issue sketch:

- The point is split into a `LOG_CHUNK_SIZE` (= 6) prefix and a suffix. The prefix tensor expansion (64 entries) is split into 8 groups of 8 and turned into 8 subset-sum lookup tables ([Method of Four Russians](https://en.wikipedia.org/wiki/Method_of_Four_Russians), as in `BytewiseLookupTransformation`).
- `duplicate_to_fixed_chunks` views the words as chunks of `CHUNK_SIZE` = 64 words. A parallel map-reduce folds each chunk into a `[F; 64]` accumulator and scales it by the suffix tensor expansion weight for that chunk (computed with the packed field `P`), then sums the accumulators — the same map-reduce shape as `univariate_round_message_extension_domain`.
- Within a chunk, each stride of 8 words is bit-transposed (Hacker's Delight 8×8 algorithm) so that byte `i` of output word `k` collects bit `k * 8 + i` across the 8 words; the byte indexes the stride's lookup table.
- When `words.len() < CHUNK_SIZE`, `duplicate_to_fixed_chunks` repeats the words and the zero-padded prefix pairs each repeated copy with a zero weight, leaving the result unchanged.

`duplicate_to_fixed_chunks` is moved from the `and_reduction` sumcheck module into `fold_word` so both can share it (as the issue suggested).

## Testing & benchmarking

- A naive reference implementation is checked against the optimized one across `log_n` values smaller than, equal to, and larger than `LOG_CHUNK_SIZE`.
- A unit test validates the 8×8 bit transpose against a brute-force transpose.
- A `fold_across_words` benchmark measures throughput in words. On this machine it sustains ~155 Melem/s at 2^20 words (`fold_words` is ~290 Melem/s for comparison; the gap is the extra transpose work).

### DIFF
--- a/crates/prover/src/fold_word.rs
+++ b/crates/prover/src/fold_word.rs
@@ -1,15 +1,33 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
-use binius_core::word::Word;
+use std::{array, borrow::Cow, iter};
+
+use binius_core::{
+	consts::{LOG_WORD_SIZE_BITS, WORD_SIZE_BITS, WORD_SIZE_BYTES},
+	word::Word,
+};
 use binius_field::{
 	BinaryField, Field, PackedField,
 	linear_transformation::{
 		BytewiseLookupTransformationFactory, LinearTransformationFactory,
 		OutputWrappingTransformationFactory, Transformation,
 	},
+	util::expand_subset_sums_array,
 };
-use binius_math::FieldBuffer;
+use binius_math::{
+	FieldBuffer,
+	multilinear::eq::{eq_ind_partial_eval, eq_ind_partial_eval_scalars},
+};
 use binius_utils::{checked_arithmetics::log2_ceil_usize, rayon::prelude::*};
+
+/// Base-2 logarithm of the number of words folded together within a single chunk.
+const LOG_CHUNK_SIZE: usize = LOG_WORD_SIZE_BITS;
+/// Number of words folded together within a single chunk.
+const CHUNK_SIZE: usize = 1 << LOG_CHUNK_SIZE;
+/// Number of bits in a byte; [`fold_across_words`] processes each chunk in groups of this many
+/// words, one per byte of the words.
+const BITS_PER_BYTE: usize = WORD_SIZE_BITS / WORD_SIZE_BYTES;
 
 /// Computes a [`FieldBuffer`] where each element is the inner product of the bits of a word and a
 /// vector of field elements.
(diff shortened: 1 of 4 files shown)


crates/prover/src/fold_word.rs
```diff
@@ -75,9 +93,183 @@ where
...
+			.try_into()
+			.expect("prefix_expansion has CHUNK_SIZE = WORD_SIZE_BYTES * BITS_PER_BYTE entries");
+		expand_subset_sums_array(group)
+	});
+
+	// The suffix tensor expansion provides one weight per chunk of CHUNK_SIZE words. Use the packed
+	// field `P` to compute it efficiently.
+	let suffix_weights = eq_ind_partial_eval::<P>(suffix)
+		.iter_scalars()
+		.collect::<Vec<_>>();
```
jimpo: This collect is bad. Instead, for now, just use eq_ind_partial_eval::<F> and par iter over suffix_weights.as_ref().

crates/prover/src/fold_word.rs
```diff
@@ -75,9 +93,183 @@ where
...
+
+/// Folds a single chunk of [`CHUNK_SIZE`] words against the prefix lookup tables, accumulating the
+/// per-bit-position result (before scaling by the suffix weight).
+fn fold_chunk<F: BinaryField>(
+	chunk: &[Word; CHUNK_SIZE],
+	lookups: &[[F; 1 << BITS_PER_BYTE]; WORD_SIZE_BYTES],
+) -> [F; WORD_SIZE_BITS] {
+	let mut acc = [F::ZERO; WORD_SIZE_BITS];
+	for (stride, lookup) in lookups.iter().enumerate() {
+		let stride_words: &[Word; BITS_PER_BYTE] = chunk
```
jimpo: Do the bytemuck reshaping, like in `univariate_round_message_extension_domain` so that you can zip with it.

crates/prover/src/fold_word.rs
```diff
@@ -75,9 +93,183 @@ where
...
+	}
+	acc
+}
+
+/// Bit-transposes [`BITS_PER_BYTE`] words (a `BITS_PER_BYTE`×[`WORD_SIZE_BITS`] bit matrix) into
+/// [`WORD_SIZE_BYTES`] output words (a `WORD_SIZE_BITS`×`BITS_PER_BYTE` bit matrix).
+///
+/// The byte at bit-position `p` — output word `p / BITS_PER_BYTE`, byte `p % BITS_PER_BYTE` — holds
+/// in bit `w` the value of bit `p` of `words[w]`.
+fn transpose_bits(words: &[Word; BITS_PER_BYTE]) -> [u64; WORD_SIZE_BYTES] {
```
jimpo: This is implemented wrong. It should operate in place on a &mut [Word; BITS_PER_BYTE]. Copy into a mutable array on the caller side. It should do bit-twiddling, like in square transpose.

View the input and output arrays as 8x8x8 bit matrices. the first axis is bits in a byte, the second is bytes in a word, the third is words. this permutes out[i]j[j][k] = input[k][j][i]. The j axis doesn't change.

This is what necessesitates the final transpose of the accumulator, as written in the ticket.

jimpo (changes requested):

crates/prover/src/fold_word.rs
benji-potamus: Done — switched to `eq_ind_partial_eval::<F>(suffix)` and par-iter directly over `suffix_weights.as_ref()`, no more collect. I left the `P` type parameter on the signature (per the ticket) as a placeholder for the eventual packed-suffix version; it's currently unused but clippy is fine with it. Happy to drop it if you'd rather not carry an unused param for now.

crates/prover/src/fold_word.rs
benji-potamus: Done — `fold_chunk` now reshapes the chunk with `bytemuck::must_cast_ref::<[Word; CHUNK_SIZE], [[Word; BITS_PER_BYTE]; WORD_SIZE_BYTES]>` and `iter::zip`s the sub-chunks with `lookups`, matching `univariate_round_message_extension_domain`.

crates/prover/src/fold_word.rs
benji-potamus: Reworked: `transpose_bits` now takes `&mut [Word; BITS_PER_BYTE]` and does in-place delta-swap bit-twiddling (3 rounds with masks `0xAA…`/`0xCC…`/`0xF0…`), exchanging the within-byte bit axis with the word axis while leaving the byte-in-word axis `j` fixed — `out[i][j][k] = in[k][j][i]`. `fold_chunk` copies each sub-chunk into a mutable array and accumulates into the transposed layout (`acc[BITS_PER_BYTE * a + j]` holds bit position `BITS_PER_BYTE * j + a`), and `fold_across_words` does the final 8×8 accumulator transpose once after the reduce. Added a unit test asserting the `out[i][j][k] = in[k][j][i]` permutation directly. Throughput at 2²⁰ words is ~161 Melem/s.

jimpo (approved):

PR #1634, opened 2026-06-29
BINIUS-153: implement the logUp* prover for one table, one looker column

tcoratger:
Closes [BINIUS-153](https://linear.app/jimpo/issue/BINIUS-153/implement-logup-for-one-table-one-looker-column).

Adds the prover for logUp*, the counterpart to the verifier merged in #1604.
It produces the transcript the verifier consumes and returns the same reduced claims.

### The problem

The logUp* verifier landed in #1604, but with no prover there was nothing to produce its proofs.
logUp* is an indexed-lookup argument with one notable feature: the looked-up values are never committed.
A prover must build the witnesses and run the protocol so that the looked-up vector `(I^* T)` — which has $2^n$ entries — is never materialized as a commitment.

### The idea

Use the pushforward duality. For the equality indicator `X = eq_r` at the evaluation point,

```
(I^* T)(r) = <I^* T, eq_r> = <T, I_* eq_r> = <T, Y>
```

So instead of the $2^n$-entry pullback `(I^* T)`, the prover commits the pushforward `Y = I_* eq_r`, which has only $2^m$ entries.
In the target regime the table is small ($m \ll n$), so this is the whole cost saving.

Correctness of `Y` is enforced by a logarithmic-derivative (logUp) identity for a random challenge `c`:

```
sum_i  eq_r(i) / (c - I(i))   ==   sum_j  Y(j) / (c - j)
```

Each side is collapsed to a single root fraction by a fractional-addition GKR circuit, and the two roots are cross-multiplied.

### The flow

The prover mirrors the verifier's transcript exactly:

1. sample the logUp challenge `c`;
2. build the looker numerator `eq_r`, the two denominators `c - I` and `c - J`, and the pushforward `Y`, then send the two circuits' root fractions;
3. run the looker-side GKR over all `n` layers down to the index leaf claim;
4. run the first `m-1` table-side GKR layers down to the layer-1 claim;
5. prove the batched final layer.

### The batched final layer

The last table-side GKR layer and the product check `<T, Y> = e` both end in an evaluation of `Y`.
Run naively that is two evaluations at two points.
Instead both are split on the highest variable and proved as one `(m-1)`-variable sumcheck plus one shared line-fold, collapsing them to a single evaluation point:

```
S_1 = sum_{x'} eq(x'; Z) * (Y_0*D_1 + Y_1*D_0)(x')   = layer-1 numerator
S_2 = sum_{x'} eq(x'; Z) * (D_0*D_1)(x')             = layer-1 denominator
S_3 = sum_{x'} (T_0*Y_0 + T_1*Y_1)(x')               = e
```

`S_1` and `S_2` carry the equality factor `eq(.; Z)`, so they run through the decorated MLE-check prover.
`S_3` carries no equality factor, so it runs through a small regular degree-2 product prover.
The three round polynomials are batched with one coefficient, in the order the verifier reconstructs.

### Scope

- **new:** `crates/ip-prover/src/logup_star/` — `mod.rs`, `error.rs`, `witness.rs`, `prove.rs`, `final_layer.rs`.
- **changed:** one line in `crates/ip-prover/src/lib.rs` to register the module.
- **reused:** the fractional-addition prover, the sumcheck provers, and the verifier's `LogupOutput` (re-exported, not redefined).
- **untouched:** the verifier in `binius-ip`.

### Soundness

- The prover sends a transcript byte-identical to what the verifier expects, so no new assumption is introduced.
- The looker numerator leaf is `eq_r`, which the verifier evaluates itself; the prover commits nothing extra of size $2^n$.
- The logUp identity catches a wrong `Y` except with probability $(n + m) / |\mathbb{F}|$, which is the verifier's existing bound.

### Verification

- nightly `cargo fmt --all --check`, `cargo clippy -p binius-ip-prover --all-targets -D warnings`, `cargo check --workspace --all-targets` — all clean.
- Round-trip tests across shapes (including $m \ll n$, $m = n$, $m = 1$, and $n = 0$) assert the prover output equals the verifier output, and that all three reduced claims match independent honest evaluations of `T`, `Y`, and the embedded index column.
- Negative tests: a tampered `eval_claim` is rejected by the verifier, and the index length / range preconditions are checked.

### DIFF
--- a/crates/ip-prover/src/logup_star/final_layer.rs
+++ b/crates/ip-prover/src/logup_star/final_layer.rs
@@ -1,0 +1,304 @@
+// Copyright 2026 The Binius Developers
+
+//! The batched final layer of the table side.
+//!
+//! It fuses two reductions that both end in an evaluation of the pushforward `Y`:
+//!
+//! - the last fractional-addition GKR layer of the table circuit,
+//! - the product sumcheck `<T, Y> = e`.
+//!
+//! Both split their leaf multilinears on the highest variable.
+//! So both can share one `(m-1)`-variable sumcheck and one line-fold over that variable.
+//! That collapses the two `Y` evaluations into a single evaluation point.
+//!
+//! This is the prover mirror of the verifier's final layer in [`binius_ip::logup_star`].
+
+use binius_field::{Field, PackedField};
+use binius_ip::{MultilinearEvalClaim, sumcheck::RoundCoeffs};
+use binius_math::{
+	FieldBuffer, line::extrapolate_line, multilinear::fold::fold_highest_var_inplace,
+};
+use either::Either;
+use itertools::izip;
+
+use super::error::Error;
+use crate::{
+	channel::IPProverChannel,
+	sumcheck::{
+		self, MleToSumCheckDecorator, batch::batch_prove, common::SumcheckProver, frac_add_mle,
+	},
+};
+
+/// The state of a round-by-round sumcheck prover between its two phases.
+///
+/// The prover alternates: it emits the round polynomial, then folds it on a challenge.
+/// So between calls it holds either the running sum claim or the last round's coefficients.
+#[derive(Debug, Clone)]
+enum RoundCoeffsOrSum<F: Field> {
+	/// The coefficients of the round polynomial just emitted, awaiting a fold.
+	Coeffs(RoundCoeffs<F>),
+	/// The running sum claim before a round is executed.
+	Sum(F),
+}
+
+/// The evaluation claims that the batched final layer reduces to.
+pub struct FinalLayerOutput<F> {
+	/// The `m`-coordinate point shared by the table and pushforward evaluation claims.
+	pub table_eval_point: Vec<F>,
+	/// The claimed evaluation of the table multilinear `T` at the point.
+	pub table_eval_claim: F,
+	/// The claimed evaluation of the pushforward multilinear `Y` at the point.
+	pub pushforward_eval_claim: F,
+}
+
+/// A regular sumcheck prover for the leaf product `sum_{x'} (Y_0 * T_0 + Y_1 * T_1)(x')`.
+///
+/// This is the product claim `<T, Y> = e`, rewritten over the low `m-1` variables.
+/// It is obtained by splitting `T` and `Y` on the highest variable.
+///
+/// The composite has degree 2 and carries no equality factor.
+/// That is why it cannot ride along inside the eq-weighted fractional-addition prover.
+///
+/// The four multilinears are held as `[Y_0, Y_1, T_0, T_1]`.
+/// So the final evaluations are exactly the leaf halves the verifier reads, in that order.
+///
+/// The table is small in the logUp* regime, so the round polynomial is summed at the scalar level.
+/// Variables are bound from the highest index to the lowest, matching the other final-layer prover.
+struct LeafProductProver<P: PackedField> {
+	/// The leaf halves `[Y_0, Y_1, T_0, T_1]`, folded in place one variable per round.
+	multilinears: [FieldBuffer<P>; 4],
+	/// The sum claim before a round, or the round polynomial awaiting a fold.
+	last_coeffs_or_sum: RoundCoeffsOrSum<P::Scalar>,
+}
+
+/// Prove the batched final layer of the table side.
+///
+/// Three sum claims are batched over the `m-1`-variable cube, then the highest variable is folded:
+///
+/// ```text
+///     S_1 = sum_{x'} eq(x'; Z) * (Y_0 * D_1 + Y_1 * D_0)(x') = num_1(Z)
+///     S_2 = sum_{x'} eq(x'; Z) * (D_0 * D_1)(x')             = den_1(Z)
+///     S_3 = sum_{x'} (T_0 * Y_0 + T_1 * Y_1)(x')             = e
+/// ```
+///
+/// The claims play two different roles:
+///
+/// - `S_1` and `S_2` are the layer-1 numerator and denominator of the fractional-addition circuit.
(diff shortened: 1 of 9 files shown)


crates/ip-prover/src/logup_star/prove.rs
```diff
@@ -0,0 +1,424 @@
...
+	if index.len() != expected {
+		return Err(Error::IndexLengthMismatch {
+			got: index.len(),
+			expected,
+			n_vars: n,
+		});
+	}
+	// Every index must address a real table position for the embedding and pushforward to be valid.
+	let table_size = 1usize << m;
+	if let Some((position, &value)) = index.iter().enumerate().find(|&(_, &v)| v >= table_size) {
```
jimpo: This is slow and unnecessary. Make it a precondition. You can use `debug_assert` to check the indices are < table size, which will get compiled out in release.

jimpo:
I'm now realizing this does nothing about the pushforward commitments. This protocol should probably encapsulate that too, which requires using an IOP channel (not just an IP channel) to send and verify the commitments. You can do that in a follow-up PR.

tcoratger:
> I'm now realizing this does nothing about the pushforward commitments. This protocol should probably encapsulate that too, which requires using an IOP channel (not just an IP channel) to send and verify the commitments. You can do that in a follow-up PR.

Yes probably this is better to isolate this in a followup to be sure that this change is done properly via the git diff, what do you think?

crates/ip-prover/src/logup_star/prove.rs
```diff
@@ -0,0 +1,424 @@
...
+/// So the returned claim is over `n_layers` variables.
+///
+/// The looker side runs all `n` layers down to the leaf claim.
+/// The table side runs only the first `m-1` layers.
+/// That leaves its leaf layer to be proved in batch with the product check.
+///
+/// # Preconditions
+///
+/// - The prover must hold at least `n_layers` layers.
+fn prove_fracadd_layers<F, P>(
```
jimpo: This looks right, but it belongs as part of the FracAddProver itself. A method like

```rust
	pub fn prove_layers(
		self,
		n_layers: usize,
		claim: (MultilinearEvalClaim<F>, MultilinearEvalClaim<F>),
		channel: &mut impl IPProverChannel<F>,
	) -> (Option<Self>, FracEvalClaim<F>) {
		todo!()
	}
```

and then `FracAddProver::prove` simply delegates to that with `n_layers = k`, and dropping the None prover from the output.

crates/ip-prover/src/logup_star/final_layer.rs
```diff
@@ -0,0 +1,304 @@
...
+	let [y_0, y_1] = split_halves(pushforward);
+	let [d_0, d_1] = split_halves(table_denominator);
+	let [t_0, t_1] = split_halves(table);
+
+	// S_1, S_2: the fractional-addition numerator/denominator, weighted by eq(.; Z).
+	//
+	// The eq factor is kept implicit by the MLE-check prover.
+	// The decorator factors it back in per round as a (X - alpha) multiplier.
+	// That turns the MLE-check into a regular sumcheck the verifier's batch reconstructs.
+	let frac_prover = frac_add_mle::new(
```
jimpo: Hmm, can't you just get this already from `.layer_prover()` on the remaining `FracAddProver`?

crates/ip-prover/src/logup_star/final_layer.rs
```diff
@@ -0,0 +1,304 @@
...
+		[y_0.clone(), y_1.clone(), d_0, d_1],
+		num_claim.point,
+		[num_claim.eval, den_claim.eval],
+	)?;
+	let frac_prover = MleToSumCheckDecorator::new(frac_prover);
+
+	// S_3: the product leaf check, split on the highest variable, carrying no eq factor.
+	let product_prover = LeafProductProver::new([y_0, y_1, t_0, t_1], eval_claim)?;
```
jimpo: Ah, pretty annoying that that is necessary. But yeah.

jimpo: Instead of a custom SumcheckProver, I'd just go with two BivariateProductSumcheckProvers and batch the three of them together.

jimpo (changes requested):

crates/ip-prover/src/logup_star/prove.rs
tcoratger: Done just added the debug assert stuff.

crates/ip-prover/src/logup_star/prove.rs
tcoratger: Done and deleted this now useless `prove_fracadd_layers` helper.

crates/ip-prover/src/logup_star/final_layer.rs
tcoratger: Fixed

crates/ip-prover/src/logup_star/final_layer.rs
tcoratger: Done, just deleted the `LeafProductProver` stuff.

crates/ip-prover/src/logup_star/witness.rs
```diff
@@ -0,0 +1,106 @@
...
+	index: &[usize],
+	table_n_vars: usize,
+) -> FieldBuffer<P>
+where
+	F: Field,
+	P: PackedField<Scalar = F>,
+{
+	// Scatter-add the looker numerator onto its table position, summing collisions.
+	let mut values = vec![F::ZERO; 1usize << table_n_vars];
+	for (eq_i, &j) in eq_r.iter_scalars().zip(index) {
```
jimpo, line 102: Interesting -- we'll want to parallelize this, but it's going to be tricky. Can you make a Linear ticket to track? (for follow-up)

jimpo (approved):

PR #1635, opened 2026-06-29
BINIUS-170: IntMul run the c_hi product tree at base g via the high-half Frobenius twist

benji-potamus:
Closes [BINIUS-170](https://linear.app/jimpo/issue/BINIUS-170) — audit divergence **D-V3** under [BINIUS-163](https://linear.app/jimpo/issue/BINIUS-163) (conform IntMul to `sec:intmul`). Second of the D-V3/4/5 trio.

## What

The spec rewrites the base-`g^{2^64}` high-half claim `D̃_hi(r)` as a base-`g` claim `C̃_hi(φ⁻⁶⁴(r))` so the `c_hi` product tree runs at base `g`, uniformly with `a` and `c_lo`. The code instead ran the whole `c_hi` tree at base `g^{2^64}` and reduced the "scaled" leaf evals locally (high generator powers). Soundness-equivalent, but not spec-literal.

Identity: `D_hi = φ⁶⁴(C_hi)` entrywise (φ(y)=y², so φ⁶⁴(y)=y^{2⁶⁴}), hence `C̃_hi(φ⁻⁶⁴(r)) = φ⁻⁶⁴(D̃_hi(r))`. φ⁶⁴ is an 𝔽₂-automorphism: it distributes over the multilinear sum and fixes the {0,1} eq weights, twisting only the evaluation point.

Two observations keep the change tight:
- **The twist only touches Phase-4 layer 0.** The GKR `eval_point` resets to the shared fold challenges after every layer, so from layer 1 on (and all of Phase 5) `c_hi` is already at the shared point — only its layer-0 root claim + eq weight differ.
- **No shared-primitive change.** `batch_prove` already composes provers with distinct eval points (Phase 5 does it), so Phase-4 layer 0 batches two provers — `[a, c_lo]` at `r` and `[c_hi]` at `φ⁻⁶⁴(r)` — into one sumcheck. The verifier mirrors with two eq weights in the batched-layer check.

Changes:
- **verifier** (`verify_phase_3`): keep the raw `D̃_hi(r)` for the LO·HI product check, then convert to `C̃_hi(φ⁻⁶⁴(r))` and thread the twisted point into `Phase3Output`. `verify_phase_4` / the batched-layer function weight the `c_hi` tuples by `eq(φ⁻⁶⁴(r))` at layer 0. `reconstruct_selecteds`: `c_hi` now uses the low generator powers `g^{2^i}` (base g).
- **prover** (`witness`): `c_hi` tree built at base `g`; the `c_root` product identity uses `D_hi = φ⁶⁴(C_hi_root)`. `phase3`/`phase4` mirror the verifier (convert the claim, two-prover layer 0). `inv_frobenius`/`inv_frobenius_point` exposed from `common`, plus a `frobenius` buffer helper.

No interface change; `IntMulOutput` unchanged.

## Test

`prove_and_verify` roundtrip + the witness tests pass (the roundtrip gates the prover↔verifier transcript through the new twist and the two-prover batching). `+ fmt + clippy -D warnings + rustdoc -D warnings` clean on `binius-verifier`/`binius-prover`.

Last in the trio: BINIUS-171 (D-V4, eq_6 Phase-1 batching).

### DIFF
--- a/crates/prover/src/protocols/intmul/prove.rs
+++ b/crates/prover/src/protocols/intmul/prove.rs
@@ -20,7 +20,7 @@ use binius_utils::rayon::prelude::*;
 use binius_verifier::protocols::{
 	intmul::common::{
 		IntMulOutput, Phase1Output, Phase2Output, Phase3Output, Phase4Output, frobenius_twist,
-		normalize_a_c_exponent_evals,
+		inv_frobenius, inv_frobenius_point, normalize_a_c_exponent_evals,
 	},
 	prodcheck::MultilinearEvalClaim,
 };
@@ -29,7 +29,7 @@ use itertools::{Itertools, chain, izip};
 
 use super::{
 	error::Error,
-	witness::{Witness, two_valued_field_buffer},
+	witness::{Witness, frobenius, two_valued_field_buffer},
 };
 use crate::{
 	fold_word::{fold_across_words, fold_words},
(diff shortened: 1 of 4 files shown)


jimpo:
OK, I see that this is a bad idea actually, because it requires computing the 2^64'th power of the highest C_hi layer to materialize D_hi. Closing.

PR #1637, opened 2026-06-29
Adopt Plonky3-style workspace clippy lints and fix all findings

tcoratger:
@jimpo Feel free to close if you feel this is useless (this is an automatic PR made with automatic fixing). I've just replicated here the setup we are using in Plonky3 which helps us a lot to clean the codebase and adopt the strongest constraints.

Pure cleanup — no behavior change. Brings the workspace's clippy configuration in line with Plonky3 and fixes every finding so the repo passes `cargo clippy --workspace --all-targets -- -D warnings` (the CI command).

### The idea

Plonky3 lints its whole workspace with a strict, shared config. This ports that same config here and clears the resulting backlog.

Root `Cargo.toml` `[workspace.lints]` now mirrors Plonky3 exactly:

```
[workspace.lints]
rust.rust_2018_idioms = { level = "deny", priority = -1 }
rust.unused_must_use = "deny"
rust.rust_2024_incompatible_pat = "warn"
rustdoc.all = "warn"

[workspace.lints.clippy]
all = { level = "warn", priority = -1 }
nursery = { level = "warn", priority = -1 }
semicolon_if_nothing_returned = "warn"
match_bool = "warn"
needless_pass_by_value = "warn"
redundant_pub_crate = "allow"
too_long_first_doc_paragraph = "allow"
unused_peekable = "allow"
tuple_array_conversions = "allow"
transmute_undefined_repr = "allow"
cognitive_complexity = "allow"
```

All 21 crates already opt in with `[lints] workspace = true`, so no per-crate change was needed.
The previous `needless_range_loop = "allow"` is dropped (Plonky3 doesn't allow it either).

### The change

Two buckets of fixes.

Mechanical (largely `clippy --fix`):

- `use_self`, `missing_const_for_fn`, `semicolon_if_nothing_returned`, `redundant_clone`
- elided lifetimes in paths (`<'_>`), required by the now-`deny` `rust_2018_idioms`

By hand:

- `needless_range_loop` → iterators (`iter()/iter_mut().enumerate()/.take()/.skip()`)
- `option_if_let_else`, `needless_pass_by_ref_mut`, `collection_is_never_read`, `or_fun_call`, `branches_sharing_code`
- the `needless_pass_by_value` cascade: `T` / `Vec<T>` → `&T` / `&[T]`, propagated through every call site

The largest cascade is the prove chain and the MLE-check constructors:

```diff
- pub fn prove(&self, witness: ValueVec, ...) -> ...
+ pub fn prove(&self, witness: &ValueVec, ...) -> ...

- pub fn new(..., eval_point: Vec<F>, ...) -> Result<Self, Error>
+ pub fn new(..., eval_point: &[F], ...) -> Result<Self, Error>
```

### Soundness

Pure cleanup. Every change is type-level (passing `&x` where `x` was passed before) or a mechanical lint fix; no computation, challenge, or transcript behavior changes.

One subtlety worth calling out: a few wrapper constructors return `impl MleCheckProver<F>`. Under edition 2024, switching their `eval_point` to `&[F]` would make the returned `impl Trait` capture that lifetime and break callers that pass temporaries. Since the point is copied into an owned `Gruen32` (the prover never borrows it), those returns use precise capturing — `+ use<F, P, M>` — to exclude the lifetime.

A few redundant `.to_vec()` snapshots that only existed to copy data before a now-removed move were dropped. The genuinely load-bearing ones were kept: one releases a `RefCell` borrow before a later `borrow_mut` (would panic otherwise), one feeds a tampering test that mutates the copy.

### Scope

- **changed:** root `Cargo.toml` lint config + clippy fixes across all 21 crates (282 files, mostly one- or two-line edits)
- **left untouched:** all logic, all public behavior; the old reference oracles in tests

### Verification

- `cargo build --workspace --all-targets` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — exits 0
- `cargo test --workspace` — all pass, 0 failures

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,8 +98,27 @@ trait-set = "0.3.0"
 uninit = "0.6.2"
 ureq = "3.1"
 
+[workspace.lints]
+rust.rust_2018_idioms = { level = "deny", priority = -1 }
+rust.unused_must_use = "deny"
+rust.rust_2024_incompatible_pat = "warn"
+rustdoc.all = "warn"
+
 [workspace.lints.clippy]
-needless_range_loop = "allow"
+all = { level = "warn", priority = -1 }
+nursery = { level = "warn", priority = -1 }
+
+# Some pedantic lints
+semicolon_if_nothing_returned = "warn"
+match_bool = "warn"
+needless_pass_by_value = "warn"
+
+redundant_pub_crate = "allow"
+too_long_first_doc_paragraph = "allow"
+unused_peekable = "allow"
+tuple_array_conversions = "allow"
+transmute_undefined_repr = "allow"
+cognitive_complexity = "allow"
 
 [workspace.metadata.typos]
 default.extend-words = { "typ" = "typ", "Toom" = "Toom", "inouts" = "inouts" }
--- a/crates/arith-bench/benches/field_mul.rs
+++ b/crates/arith-bench/benches/field_mul.rs
@@ -45,21 +45,21 @@ fn run_google_mul_benchmark<U, R>(
 	group.bench_function(name, |b| {
 		b.iter(|| {
 			let mut x = black_box(x);
-			for j in 0..N {
-				y[j] = x;
+			for slot in y.iter_mut().take(N) {
+				*slot = x;
 				x = mul_fn(x, x);
 			}
 			for _i in 0..N * N {
-				for j in 0..N {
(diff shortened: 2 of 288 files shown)


jimpo:
Spot checking a few, I don't really like these changes.

If you think there are any good lint rules we're missing, do them in separate commits so I can see the effects in isolation.

PR #1641, opened 2026-06-30
BINIUS-171: IntMul batch Phase-1 Frobenius leaf claims with eq_6(γ,i) (D-V4)

benji-potamus:
Closes [BINIUS-171](https://linear.app/jimpo/issue/BINIUS-171) — audit divergence **D-V4** under [BINIUS-163](https://linear.app/jimpo/issue/BINIUS-163) (conform IntMul to `sec:intmul`). Last of the D-V3/4/5 trio (D-V5 merged in #1633; D-V3 scrapped per the [BINIUS-170 discussion](https://linear.app/jimpo/issue/BINIUS-170)).

## What

The spec's variable-base Frobenius subprotocol batches the `2^k` Frobenius-twisted Phase-1 leaf claims with the multilinear `eq_k(γ, i)`, `γ ∈ K^k` (a single verifier challenge), giving soundness `2^k/|K|`. The code instead batched them with **powers of a single univariate challenge** (`batch_coeff^i`). Soundness-equivalent, but not spec-literal.

This switches the 64 selector claims to `eq_6(γ, i)` weights. The LO·HI (`c_root`) claim still rides the same sumcheck, now combined with the 64-way aggregate by a 2-way univariate `batch_coeff` instead of being the 65th univariate term.

## How — batching inside `SelectorMlecheckProver`

The eq-batching lives in `SelectorMlecheckProver` (its only consumer), rather than in a separate adapter:

- **`SelectorMlecheckProver`** now takes a `weights` vector (one per selector) and exposes a **single** claim (`n_claims() == 1`): `execute()` and `round_claim()` combine its per-selector round polynomials / claims into `Σ_i weights[i] · R_i`, while folding and the final multilinear evaluations stay per-selector internally. Passing `eq_ind_partial_eval_scalars(γ)` as the weights gives the `eq_k(γ, i)` batching.
- **prover `phase3`**: sample `γ = channel.sample_many(k)` before the batched sumcheck, pass the `eq_k(γ, ·)` weights to `SelectorMlecheckProver::new`, then `batch_prove` 2-way `[selector-agg, LO·HI]` (was 65-way).
- **verifier `verify_phase_3`**: sample `γ` first, form the aggregate claim `evaluate_inplace_scalars(twisted_evals, γ)` — the same multilinear-eval convention the D-V2 `b_recomb` already uses, so there's no bit-order risk — and reconstruct the two batched terms 2-way.

No interface change beyond the new `weights` argument; `IntMulOutput` unchanged. (An earlier revision routed the combination through a standalone `CombineClaimsDecorator`; per review it was folded into `SelectorMlecheckProver` directly, since that prover is the sole user.)

## Test

`prove_and_verify` roundtrip + the 3 witness tests pass (the roundtrip gates the full prover↔verifier transcript through the new `γ` sampling and the in-prover batching), and the two `selector_mle` unit tests now check the weighted combination. `fmt --check`, `clippy -p binius-ip-prover -p binius-prover -p binius-verifier --all-targets -D warnings`, and `rustdoc -D warnings` all clean.

This is the final divergence; with it, IntMul matches `sec:intmul` (modulo the D-V3 high-half twist, which is being handled in the spec instead).

### DIFF
--- a/crates/ip-prover/src/sumcheck/selector_mle.rs
+++ b/crates/ip-prover/src/sumcheck/selector_mle.rs
@@ -39,23 +39,34 @@ pub struct SelectorMlecheckProver<'b, P: PackedField, B: Bitwise> {
 	last_coeffs_or_sums: RoundCoeffsOrSums<P::Scalar>,
 	selected: FieldBuffer<P>,
 	gruen32s: Vec<Gruen32<P>>,
+	weights: Vec<P::Scalar>,
 	switchover: BinarySwitchover<'b, P, B>,
 }
 
 impl<'b, F: Field, P: PackedField<Scalar = F>, B: Bitwise> SelectorMlecheckProver<'b, P, B> {
 	/// Constructs a prover, given `bitmasks` as representation of 1-bit columns, `selected` being
-	/// the shared large field multilinear, individual `claims` per selector and `switchover` as
-	/// the round at which 1-bit columns should be folded.
+	/// the shared large field multilinear, individual `claims` per selector, `weights` to combine
+	/// the per-selector round polynomials into one (one weight per claim), and `switchover` as the
+	/// round at which 1-bit columns should be folded.
+	///
+	/// The prover exposes a single claim — the `weights`-combination `Σ_i weights[i] · C_i` of the
+	/// per-selector claims. Supplying the equality-indicator tensor `eq_k(γ, ·)` as the weights
+	/// batches the claims with `eq_k(γ, i)`.
 	pub fn new(
 		selected: FieldBuffer<P>,
 		claims: Vec<Claim<F>>,
 		bitmasks: &'b [B],
+		weights: Vec<F>,
 		switchover: usize,
 	) -> Result<Self, Error> {
 		let n_vars = selected.log_len();
 
 		if claims.iter().any(|claim| claim.point.len() != n_vars) {
 			return Err(Error::MultilinearSizeMismatch);
+		}
+
+		if weights.len() != claims.len() {
+			return Err(Error::EvalClaimsNumberMismatch);
 		}
 
 		if bitmasks.len() != selected.len() {
@@ -75,6 +86,7 @@ impl<'b, F: Field, P: PackedField<Scalar = F>, B: Bitwise> SelectorMlecheckProve
 			last_coeffs_or_sums,
 			selected,
 			gruen32s,
+			weights,
 			switchover,
 		})
 	}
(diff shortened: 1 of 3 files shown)


jimpo (changes requested):
The new decorator is not necessary. Just move the batching logic into `SelectorMlecheckProver` directly -- it's only used in this one place.

jimpo:
In fact, in SelectorMlecheckProver, it would be more efficient to handle it as degree-3 sumcheck of the selection expression times the linear combination of the eq indicators. So no Gruen32, just expand out the point tensors, combine them into one buffer, and do a degree 3 sumcheck (using "X" in Ghash) as the extra evaluation point.

benji-potamus:
Done — dropped the decorator and moved the batching into `SelectorMlecheckProver` directly. It now takes the `weights` vector (one per selector), exposes a single claim (`n_claims() == 1`), and combines its per-selector round polynomials into `Σ_i weights[i] · R_i` in `execute()`/`round_claim()` (internals stay per-claim for folding). `phase3` passes `eq_ind_partial_eval_scalars(γ)` as the weights; `CombineClaimsDecorator` is deleted and the verifier is unchanged. Updated the two `selector_mle` unit tests to pass weights and compare against the weighted reference.

Re-validated on current `main`: `selector_mle` unit tests + `prove_and_verify` roundtrip pass; clippy/fmt/rustdoc `-D warnings` clean. Repushed (head `5dc88cb`).

benji-potamus:
Good call — that's clearly better, and I'd missed this comment until you pointed it out (my mistake). My current push still keeps the 64 per-claim Gruen32s and just combines their round polys by `eq_k(γ,·)`, which is exactly the blowup you're avoiding.

To confirm I've got the approach: `Σ_i eq_k(γ,i)·M(p_i) = Σ_x M(x)·W(x)` with `W(x) = Σ_i eq_k(γ,i)·eq(p_i, x)` materialized as one buffer (expand each twisted point's tensor, accumulate weighted by `eq_k(γ,i)`). Then a single degree-3 sumcheck of `(selected·selector + (1−selector))·W` — no Gruen32, with `X ∈ GHASH` as the extra (4th) evaluation point — same shape as `bivariate_product_mle`'s bivariate×eq trivariate product. Verifier side: the selector sub-claim's round polys become degree 3 and the final binding is `selection_expr(r)·W(r)` with the verifier computing `W(r) = Σ_i eq_k(γ,i)·eq(p_i, r)` itself; the initial batched claim `Σ_i eq_k(γ,i)·twisted_eval_i` and the `γ` sampling are unchanged.

Implementing it now; will repush.

benji-potamus:
As I started building this I hit one construction question I'd rather confirm than guess, since the selection expression is per-selector:

`comp_i(x) = selected(x)·selector_i(x) + (1 − selector_i(x))`, with **both** the selector column `selector_i` and the point `p_i` varying per `i`. So the batched claim is

`Σ_i eq_k(γ,i)·comp_i(p_i) = Σ_x [ selected(x)·SW(x) + W(x) − SW(x) ]`,

where `W(x) = Σ_i eq_k(γ,i)·eq(p_i,x)` (the single combined eq buffer you're describing) and `SW(x) = Σ_i eq_k(γ,i)·eq(p_i,x)·selector_i(x)`.

The `(1−selector_i)` term collapses into `W` cleanly. The wrinkle is the `selected·SW` term: the degree-3 comes from `selected · eq(p_i,·) · selector_i` with `eq·selector_i` left unflattened (degree-2 in the fold variable), which seems to need the per-`i` selector loop to survive — whereas if I precompute `SW` as a flat multilinear and fold it, `selected·SW` is only degree-2. So:

1. Is the intended win "collapse the 64 eq expansions → one `W`, and the `(1−selector)` term → `W`", with the `selected·selector_i·eq(p_i)` term still computed by a per-`i` loop each round (just degree-3 via the `X`-node interpolation instead of Gruen)? Or do you intend a single precomputed `SW` buffer folded alongside — and if so, how do you want the degree handled (it'd drop to degree-2)?
2. Confirming `finish()` still exposes the 64 per-`i` `b(i,r)` evals — the verifier recombines them into `b_recomb` via `r_ib` for Phases 4–5 (`verify.rs:152–164`), so the selectors fold individually regardless.

Once I know how you want `SW` handled I'll build it (incl. the `RoundEvals3` / `X`-node interpolation, which doesn't exist yet — happy to add it). Leaning toward interpretation (1) unless you say otherwise.

jimpo (approved):

jimpo:
@benji-potamus OK, ignore that suggestion about the degree-3 sumcheck, it's not good, merging as is.

benji-potamus:
Sounds good — standing down on the degree-3 rewrite, `5dc88cb3` as-is it is. Thanks for the approval!

PR #1646, opened 2026-06-30
Enable clippy::use_self and replace type-name repetition with Self

tcoratger:
One lint in isolation, in its own commit, so its effect can be reviewed on its own (same approach as the const-fn and redundant-clone PRs).

### The lint

`clippy::use_self` (nursery): inside an `impl` block, flags references to the implementing type written with its full name where `Self` would do — `MyType::new()` -> `Self::new()`, `-> MyType` -> `-> Self`, `x as *const MyType` -> `x as *const Self`. Using `Self` keeps impls robust to renames and makes the "this is the impl's own type" relationship explicit.

Enabled in the workspace config:

```toml
[workspace.lints.clippy]
needless_range_loop = "allow"
missing_const_for_fn = "warn"
redundant_clone = "warn"
use_self = "warn"
```

### The change

46 files, all from `cargo clippy --fix`. Purely syntactic — no behavior change.

### Checked every arch/feature config

`use_self` fires in the feature-gated SIMD code under `arch/x86_64` and `arch/aarch64`, which a plain build on one host never fully compiles. So I ran the fix and `-D warnings` check across every config the CI matrix (`-Ctarget-cpu=native` per leg) can hit:

| config | result |
|---|---|
| native aarch64 (default) | clean |
| aarch64 `+neon,+aes,+sha3` | clean |
| x86_64 default (sse2 / portable leg) | clean |
| x86_64 `+avx,+avx2,+pclmulqdq` (AMD-like, no AVX-512) | clean |
| x86_64 `+avx2,+avx512f,+gfni,+vpclmulqdq,...` (Intel native) | clean |

### Scope

- **changed:** root `Cargo.toml` (one lint line) + `Self` substitutions across the workspace, including the x86_64/aarch64 arch modules

### Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — exits 0 (all five configs above)
- `cargo build --workspace --all-targets` — clean
- `cargo test --workspace` — all pass, 0 failures

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,6 +102,7 @@ ureq = "3.1"
 needless_range_loop = "allow"
 missing_const_for_fn = "warn"
 redundant_clone = "warn"
+use_self = "warn"
 
 [workspace.metadata.typos]
 default.extend-words = { "typ" = "typ", "Toom" = "Toom", "inouts" = "inouts" }
--- a/crates/arith-bench/src/arch/aarch64.rs
+++ b/crates/arith-bench/src/arch/aarch64.rs
@@ -35,7 +35,7 @@ impl Underlier for uint64x2_t {
 
 	fn random(mut rng: impl Rng) -> Self {
 		let value: u128 = rng.random();
-		unsafe { std::mem::transmute::<_, uint64x2_t>(value) }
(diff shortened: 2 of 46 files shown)


jimpo (commented):
I don't like this one 😝

PR #1648, opened 2026-06-30
BINIUS-174: finish() the non-ZK prover and verifier channels

benji-potamus:
Closes [BINIUS-174](https://linear.app/jimpo/issue/BINIUS-174/binius64-non-zk-config-doesnt-finish-on-channel).

The non-ZK Binius64 `Prover`/`Verifier` never called the consuming `finish()` on their channels (the ZK config already does). The blocker was that `IOPProver::prove` *consumed* the channel, leaving the wrapper nothing to finish.

## Change

- **`IOPProver::prove`** now takes `channel: &mut Channel` instead of consuming it (the internal sub-reductions just reborrow `&mut *channel`). `IOPVerifier::verify` already took `&mut Channel`.
- **`Prover::prove`** (`crates/prover/src/prove.rs`) and **`Verifier::verify`** (`crates/verifier/src/verify.rs`) now own the channel and call `finish()` after the protocol — the prover's `finish` returns `()`, the verifier's returns `Result`.
- **`ZKProver::prove`** (`crates/prover/src/zk_config.rs`): the spartan `ZKWrappedProverChannel` implements `IOPProverChannel` on `&mut Self`, so the call site binds the `&mut` reference and passes a mutable borrow of it (the established pattern, also used in the wrapper integration test). Its existing `finish(rng)` is unchanged. I kept this localized rather than moving the wrapper's trait impl to the base type, since that would ripple into the spartan `IOPProver` API (which also takes the wrapper as `&mut Self`) — out of scope for this fix. Happy to revisit if you'd prefer the wrapper impl moved.

## Verification

`cargo test` for prover / verifier / spartan-prover / spartan-verifier / examples — all pass (the prove/verify roundtrips exercise the new `finish()`, whose oracle-count assertions would fire on any mismatch); clippy `--tests -D warnings` clean; fmt clean.

### DIFF
--- a/crates/prover/src/zk_config.rs
+++ b/crates/prover/src/zk_config.rs
@@ -184,11 +184,6 @@ where
 			)
 			.entered();
 
-			// `ZKWrappedProverChannel` implements `IOPProverChannel` on `&mut Self`, so the channel
-			// type is the reference itself; bind it and pass a mutable borrow of that reference.
-			// The borrow ends with this block, leaving `wrapped_channel` owned for `finish`
-			// below.
-			let mut wrapped_channel = &mut wrapped_channel;
 			self.inner_iop_prover
 				.prove::<P, _>(witness, &mut wrapped_channel)?;
 		}
--- a/crates/prover/src/zk_config.rs
+++ b/crates/prover/src/zk_config.rs
@@ -184,6 +184,11 @@ where
 			)
 			.entered();
 
+			// `ZKWrappedProverChannel` implements `IOPProverChannel` on `&mut Self`, so the channel
+			// type is the reference itself; bind it and pass a mutable borrow of that reference.
+			// The borrow ends with this block, leaving `wrapped_channel` owned for `finish`
+			// below.
+			let mut wrapped_channel = &mut wrapped_channel;
 			self.inner_iop_prover
 				.prove::<P, _>(witness, &mut wrapped_channel)?;
 		}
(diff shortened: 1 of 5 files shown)


crates/prover/src/zk_config.rs
```diff
@@ -184,6 +184,11 @@ where
 			)
 			.entered();
 
+			// `ZKWrappedProverChannel` implements `IOPProverChannel` on `&mut Self`, so the channel
```
jimpo: Change so IOPProverChannel is impl'd for ZKWrappedProver channel directly. (In a new commit in this PR).

jimpo (changes requested):

crates/prover/src/zk_config.rs
benji-potamus: Done in a new commit (`9cbfd97`): moved both `IOPProverChannel` and `IPProverChannel` onto `ZKWrappedProverChannel` directly instead of `&mut Self` (the inner reductions need `IPProverChannel` on the same type). Since the spartan `IOPProver` already takes `&mut Channel`, this let me drop the `&mut`-of-`&mut` indirection at both call sites — here it's now plain `prove(witness, &mut wrapped_channel)`, and the wrapper integration test passes `&mut wrapped_prover_channel` directly. All prover/verifier/spartan/examples tests + clippy + fmt green.

jimpo (approved):

PR #1649, opened 2026-06-30
[codex] Add independent hash primitive benchmarks

latifkasuli:
## Summary

Adds benchmark coverage for batches of independent hash primitives:

- `sha256_compressions`: independent `compress(IV, m_i)` SHA-256 compression calls
- `blake3_compressions`: independent BLAKE3 compression calls
- `keccak_permutations`: independent Keccak-f[1600] permutations

The benchmark runner reports witness generation, proof generation, combined witness-generation-plus-proving throughput, verification, proof size, and peak memory. The combined group is intended to make cross-prover comparisons easier for workloads like Flock's independent hash primitive benchmarks.

## Benchmark knobs

- `N_COMPRESSIONS`: SHA-256/BLAKE3 compression count, default `4096`
- `N_PERMUTATIONS`: Keccak-f[1600] permutation count, default `4096`
- `N_PRIMITIVES`: fallback count for either primitive type
- `LOG_INV_RATE`: inverse-rate parameter, default `1`

## Local 4096 results

Measured on Apple M4 Pro, generic ARM64 target, default features, `LOG_INV_RATE=1`.

| Primitive | Witness gen | Proof gen | Witness + prove | Verify | Proof | Peak mem |
|---|---:|---:|---:|---:|---:|---:|
| SHA-256 compressions | 91.47K/s | 6.87K/s | 6.43K/s | 39.70K/s | 14816 B / 14.47 KiB | 519.06 MB |
| BLAKE3 compressions | 275.80K/s | 8.51K/s | 8.20K/s | 58.51K/s | 14816 B / 14.47 KiB | 468.21 MB |
| Keccak-f[1600] permutations | 72.97K/s | 6.87K/s | 6.26K/s | 47.07K/s | 14816 B / 14.47 KiB | 519.09 MB |

The identical proof size is expected here: at `N=4096` and `LOG_INV_RATE=1`, all three circuits round into the same proof-size bucket.

## Methodology notes

- The `witness_generation_and_proving` group is the closest analog to Flock's headline prove path because it includes witness generation plus proving.
- The standalone `witness_generation` group includes `WitnessFiller` allocation, input population, circuit witness evaluation, and conversion into a `ValueVec`. Flock's competitor benches time only `populate_wire_witness` after pre-filling inputs, so this row is a Binius64-internal baseline rather than a strict cross-prover witness-generation comparison.
- The local numbers above are single-threaded, generic ARM64-target results. They are not a native CPU / pinned-performance-core comparison run.

## Test plan

- `cargo test -p binius-examples independent_hashes --lib`
- `cargo bench -p binius-examples --bench sha256_compressions --bench blake3_compressions --bench keccak_permutations --no-run`
- `N_COMPRESSIONS=4096 cargo bench -p binius-examples --bench sha256_compressions -- --noplot`
- `N_COMPRESSIONS=4096 cargo bench -p binius-examples --bench blake3_compressions -- --noplot`
- `N_PERMUTATIONS=4096 cargo bench -p binius-examples --bench keccak_permutations -- --noplot`

### DIFF
--- a/crates/examples/Cargo.toml
+++ b/crates/examples/Cargo.toml
@@ -64,6 +64,18 @@ name = "hashsign"
 harness = false
 
 [[bench]]
+name = "sha256_compressions"
+harness = false
+
+[[bench]]
+name = "blake3_compressions"
+harness = false
+
+[[bench]]
+name = "keccak_permutations"
+harness = false
+
+[[bench]]
 name = "sha256"
 harness = false
 
--- a/crates/examples/benches/blake3_compressions.rs
+++ b/crates/examples/benches/blake3_compressions.rs
@@ -1,0 +1,33 @@
+// Copyright 2026 The Binius Developers
+// Copyright 2025 Irreducible Inc.
+//! Independent BLAKE3 compression benchmark.
+
+#[path = "utils/config.rs"]
+mod config;
+#[path = "utils/independent_hashes.rs"]
+mod independent_hashes;
+#[path = "utils/reporting.rs"]
+mod reporting;
+#[path = "utils/runner.rs"]
+mod runner;
+
+use std::alloc::System;
+
+use binius_examples::circuits::independent_hashes::IndependentBlake3Compressions;
+use criterion::{Criterion, criterion_group, criterion_main};
+use independent_hashes::{IndependentPrimitive, run_independent_hash_benchmark};
+use peakmem_alloc::PeakMemAlloc;
+
+#[global_allocator]
+static BLAKE3_COMPRESSIONS_PEAK_ALLOC: PeakMemAlloc<System> = PeakMemAlloc::new(System);
+
+fn bench_blake3_compressions(c: &mut Criterion) {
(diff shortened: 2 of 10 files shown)


jimpo (approved):
Thanks! I wasn't even aware we had this benchmark infrastructure. 😅 I always use the example CLI.

TBH, I don't think Criterion is the right tool for this. But I get that it's useful so, I'll merge.

latifkasuli:
Post-merge correction to this PR's benchmark table: the proof sizes (14,816 B) and the
prove/verify throughputs were measured on the branch's base, which predates #1648 —
before it, the non-ZK prover/verifier never called `finish()`, so those transcripts were
missing the entire batched PCS opening (~348 KB at N=4096) and verified vacuously with
respect to the committed-oracle evaluation claims. On post-#1648 code the true proof size
at these settings is 362,176 B (~354 KiB). Witness-generation numbers are unaffected.

Separately, as of #1653 (dead-code elimination) these bench circuits needed their outputs
observed to measure real proving — fixed in #1663, which also includes corrected numbers.

PR #1650, opened 2026-06-30
BINIUS-176: use batched prodcheck for intmul fixed-base exponentiations

benji-potamus:
Closes [BINIUS-176](https://linear.app/jimpo/issue/BINIUS-176/intmul-use-batched-prodcheck-for-the-fixed-base-exponentiations).

Replaces IntMul Phase 4's manual GKR walk over the `a`/`c_lo`/`c_hi` fixed-base exponentiation trees with a single **batched prodcheck**, using the prodcheck batching trick (selector-variable interpolation over the three trees).

## Scope (k = 5, per the issue thread)

The prodcheck runs with **k = 5** reduction layers (`log_bits - 1`), so it subsumes only the *all-but-last* GKR layers — the **final (widest) layer stays in Phase 5's batched sumcheck, exactly as today**. **Phase 5 is untouched** (its parity zerocheck + `b`-rerandomization batching, and the `Phase4Output` it consumes, are unchanged). Because the last layer stays in Phase 5, no `fold_across_words` leaf-sending is needed.

## Changes

- **`prodcheck::batch_prove`** (`crates/ip-prover/src/prodcheck.rs`) now carries a **content point**: the signature takes `selector_point` and `content_point` separately (was a single selector `eval_point`, scalar products only). On the first layer the content point seeds each prover's `layer_prover` claim, so the batched reduction supports `n_vars`-variate products (the tree roots, claimed at the Phase-3 point) — the way the single `ProdcheckProver::prove` already does for Phase 1's `b`-tree. The verifier-side `prodcheck::verify` already supported this. New `test_batch_prove_with_content` covers the non-scalar case; existing batch tests pass an empty content point.
- **Prover `phase4`** (`crates/prover/src/protocols/intmul/prove.rs`): builds one `ProdcheckProver` per tree over its all-but-last layer (node-major concatenation → node index in the high bits, matching `compute_b_leaves` and Phase 5's adjacent pairing), samples `log2_ceil_usize(3) = 2` selector challenges, runs `prodcheck::batch_prove(.., selector, content = phase3_point)`, and sends the `3·2^(log_bits-1)` all-but-last-layer evals (folded at the reduced suffix) for Phase 5 — same `Phase4Output` shape. (TODO left in place: these evals should later be pulled directly out of the last prodcheck layer's sumcheck folding.)
- **Verifier `verify_phase_4`** (`crates/verifier/src/protocols/intmul/verify.rs`): mirrors it — samples the 2 selector challenges, forms the eq(selector)-combined root claim at `selector ++ phase3_point`, runs `prodcheck::verify(log_bits - 1, ..)`, parses the reduced point as `[selector(2), suffix(n_vars), bit_index(log_bits-1)]`, receives the `3·2^(log_bits-1)` evals, and binds them to the prodcheck output via `Σ_t eq(selector,t)·Σ_z eq(bit_index,z)·leaf[t][z] == output.eval`. Deletes the now-dead `verify_multi_bivariate_product_mle_layer`.

## Soundness

The combine `assert_zero` ties the prover's sent all-but-last-layer evals to the prodcheck output; the unchanged Phase 5 then pins those same evals to the actual leaf-pair products — the same soundness structure as the previous per-layer GKR walk (whose final-layer evals were likewise bound by one batched equation + consumed by Phase 5). The three trees are interpolated over 2 selector vars padded to 4, with the phantom 4th tree structurally zero (no 4th prover).

## Verification

`cargo test` for `binius-ip-prover` / `binius-ip` / `binius-prover` / `binius-verifier` — all pass, including the intmul roundtrip `protocols::intmul::tests::prove_and_verify` (exercises all phases), the new `test_batch_prove_with_content`, the existing prodcheck batch tests, and the `shift` integration test. clippy `--tests -D warnings` clean; fmt clean.

### DIFF
--- a/crates/ip-prover/src/prodcheck.rs
+++ b/crates/ip-prover/src/prodcheck.rs
@@ -5,8 +5,7 @@ use std::iter;
 use binius_field::{Field, PackedField};
 use binius_ip::{mlecheck, prodcheck::MultilinearEvalClaim};
 use binius_math::{
-	FieldBuffer, inner_product::inner_product, line::extrapolate_line_packed,
-	multilinear::eq::eq_ind_partial_eval,
+	FieldBuffer, line::extrapolate_line_packed, multilinear::eq::eq_ind_partial_eval,
 };
 use binius_utils::rayon::prelude::*;
 use itertools::izip;
@@ -83,6 +82,15 @@ where
 		self.layers.len()
 	}
 
+	/// Consumes the prover and returns its single remaining (widest) layer.
+	///
+	/// # Preconditions
+	/// * `self.n_layers() == 1`
+	pub fn into_final_layer(mut self) -> FieldBuffer<P> {
+		assert_eq!(self.layers.len(), 1, "precondition: exactly one remaining layer");
+		self.layers.pop().expect("layers has exactly one element")
+	}
+
 	/// Pops the last layer and returns an MLE-check prover for it.
 	///
 	/// Returns `(layer_prover, remaining)` where:
@@ -157,6 +165,17 @@ where
 	}
 }
 
+/// Output of [`batch_prove`].
+///
+/// After running `n_layers - 1` reduction layers, each prover retains its final (widest) layer.
+/// `provers` pairs each remaining prover with its reduced evaluation at `eval_point`.
+pub struct BatchProveOutput<P: PackedField> {
+	/// The reduced evaluation point shared by all remaining provers.
+	pub eval_point: Vec<P::Scalar>,
+	/// Each remaining prover (with its final layer) paired with its reduced eval at `eval_point`.
+	pub provers: Vec<(P::Scalar, ProdcheckProver<P>)>,
+}
+
 /// Runs a batched product check protocol for multiple independent prodcheck provers.
 ///
 /// This combines n provers, each for an $m$-variate multilinear, using multilinear interpolation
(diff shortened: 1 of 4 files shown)


jimpo (changes requested):
The verify side looks fine, prover side needs work. Refactor the Witness to get rid of BinaryTree. The Witness should create the a, c_lo, c_hi tree base layers (the leaves). Then it should make ProdcheckProvers out of them and store the ProdcheckProvers in the Witness struct, in place of BinaryTree. Have prodcheck::batch_prove return the remaining provers. In fact, remove the final combining step from batch_prove -- it should return an output struct with the evalpoint, and a Vec<(F, ProdcheckProver<F>)> where the first item in each pair is that prover's reduced eval.

benji-potamus:
Done — reworked the prover side per your review (force-pushed `a40cdd1`). The verifier is byte-for-byte unchanged; the intmul roundtrip still passes.

- **`Witness` (witness.rs)**: removed `BinaryTree` entirely. `Witness` now builds the `2^log_bits` selected leaf layer for each of `a`/`c_lo`/`c_hi` (concatenated, node index in the high bits, like `b_leaves`) and stores a `ProdcheckProver` per tree (`a_prodcheck`/`c_lo_prodcheck`/`c_hi_prodcheck`) plus the roots + the exponent slices Phase 5 needs — mirroring how `b` was already handled. `BinaryTree` + its helpers are gone.
- **`batch_prove` (ip-prover/prodcheck.rs)**: dropped the final eq-weighted combine. It now runs `n_layers - 1` reductions (stopping one short) and returns `BatchProveOutput { eval_point, provers: Vec<(F, ProdcheckProver<P>)> }` — each remaining (single-layer) prover paired with its reduced eval. Added `ProdcheckProver::into_final_layer()` for the caller to take the retained leaf layer. Tests updated to finish the last layer themselves; the content-point test is kept.
- **Phase 4 / Phase 5 (prove.rs)**: Phase 4 calls the new `batch_prove`, takes the three retained leaf layers, sends the same `3·2^(log_bits-1)` all-but-last evals (transcript unchanged), and hands the leaf layers to Phase 5. Phase 5 does its final bivariate layer directly off those leaf layers.

**One design decision worth your eyes** (load-bearing): the prodcheck reduces on the *highest* node bit (strided pairing `z`↔`z+half`), but the verifier's final GKR layer pairs *adjacent* bits `(2z, 2z+1)`. To keep the verifier untouched I lay each leaf layer out in **even-then-odd** bit order (node `p` carries bit `2p` for `p<half`, else bit `2(p−half)+1`), so the first prodcheck split pairs bit `2z` with `2z+1` and Phase 5's `(leaf[z], leaf[z+half])` come out in bit order for `reconstruct_selecteds`. It's one permutation, documented at `constant_base_leaves`. If you'd rather the trees use the natural (b-style) strided layout and adjust the verifier to match, say the word — I kept the verifier fixed since you'd signed off on it.

CI re-running; clippy `-D warnings` + fmt clean locally.

jimpo:
Use the natural layout and adjust the verifier

benji-potamus:
Done — switched to the natural leaf layout and adjusted the verifier (force-pushed `f70833c`).

- **`constant_base_leaves` (witness.rs)**: now lays each tree's leaf layer out in natural bit order (node `p` carries bit `p`), identical to `compute_b_leaves`. The even-then-odd permutation and its `bit_of_node` helper are gone.
- **`verify_phase_5` (verify.rs)**: the final GKR layer now pairs the reconstructed leaf selectors **strided** — `(selected[z], selected[z + 2^(k-1)])` — to match the prodcheck's highest-bit split (node `z` = leaf `z` · leaf `z + half`), instead of the adjacent `(2z, 2z+1)` tuples. Phase 4 is untouched (its binding is purely node-index based).

One consequence worth noting: the bivariate-product prover flattens its `(leaf[z], leaf[z + half])` pairs pair-major, so with the natural layout its leaf evals now come out interleaved as `[bit 0, bit half, bit 1, bit half+1, …]`. The Phase-5 output contract (and downstream shift reduction) wants bit order, so the prover **de-interleaves** them back (`deinterleave_pairs`) before `normalize_a_c_exponent_evals`. So the one unavoidable permutation moved from the witness leaf layout to a reindex of the prover's sumcheck output — the witness and the leaf layout are now the simple/natural ones.

Verified the full lockstep myself (soundness, not just the roundtrip): every sent eval stays bound — the strided pairing reproduces node `z` from the bit-order reconstructed selectors, and the de-interleave only reorders already-bound sumcheck output evals, so the parity LSB and `b`-rerandomization checks still pin the true per-bit evals. `cargo test` for all four crates passes (incl. the intmul roundtrip and prodcheck batch tests); clippy `-D warnings` + fmt clean. CI re-running.

jimpo (approved):

PR #1651, opened 2026-06-30
BINIUS-168: SIMD-optimized MultiDigest for Blake3

tcoratger:
Closes [BINIUS-168](https://linear.app/jimpo/issue/BINIUS-168/blake3-simd-optimized-parallel-digest).

Hashes `N` independent Blake3 messages at once using the official crate's SIMD kernel, and wires it in as the Blake3 Merkle-leaf hasher.
Output is bit-identical to `blake3::hash` — no protocol change, no new soundness assumption.

### The problem

Blake3 leaves were hashed one at a time through the generic per-leaf adapter.
Rayon parallelized *across* leaves, but each leaf was a scalar `blake3::hash`.
So nothing used Blake3's actual vector parallelism, which is *across multiple inputs*.

The one public multi-lane primitive is `blake3::platform::Platform::hash_many`.
It compresses one fixed, block-aligned input per lane, all starting from the IV.
That is exactly the shape of a batch of Merkle leaves (each a single chunk, $\le 1024$ bytes).

### The idea

Treat the `N` leaves as `N` SIMD lanes and drive `hash_many` across them.

- One SIMD pass compresses the leading whole 64-byte blocks of every lane.
- A scalar compression then finishes each lane's trailing partial (or final) block.
- When the length is an exact block multiple, the single SIMD pass already emits the root.

```
lane 0:  [ block | block | block ]  ─┐
lane 1:  [ block | block | block ]   ├─ one hash_many call → 4 chaining values
lane 2:  [ block | block | block ]   │
lane 3:  [ block | block | block ]  ─┘
```

### The const-generic wrinkle

`hash_many` takes the per-lane byte length as a **const generic**, so it must be a compile-time literal.
A single chunk holds at most $1024 / 64 = 16$ whole blocks.
A small macro dispatches the runtime block count in `1..=16` to the right fixed-length call.

### The change

```diff
 impl HashSuite for Blake3HashSuite {
-    type ParLeafHash = ParallelDigestAdapter<blake3::Hasher>;   // scalar, one leaf at a time
+    type ParLeafHash = Blake3ParallelDigest;                    // SIMD batch of MAX_SIMD_DEGREE lanes
 }
```

- The lane count is `MAX_SIMD_DEGREE` (4 on NEON, 8 on AVX2, 16 on AVX-512), so one batch fills the widest vector.
- Sub-block leaves (`< 64` bytes) have no multi-lane kernel, so `digest_with_const_len` routes them to the per-leaf adapter — batching there would only add buffering overhead.

- The digest is bit-identical to `blake3::hash`; the transcript is unchanged.
- The SIMD path is pinned to the scalar reference by tests across every branch: empty, sub-block, exact block multiples, partial trailing block, the single-chunk boundary, multi-chunk, and unequal-length lanes.
- The fallback (unequal lengths or any lane over one chunk) hashes each lane with `blake3::hash`.

### Scope

- **new:** `Blake3MultiDigest<N>` (the SIMD `MultiDigest`), `Blake3ParallelDigest`, and a bench — all in `crates/hash/src/blake3.rs` + `crates/hash/benches/blake3.rs`.
- **changed:** the one `ParLeafHash` wiring in `Blake3HashSuite`.
- **untouched:** the scalar `blake3::hash` stays as the test/bench reference.

### Verification

- `cargo check -p binius-hash --all-targets`, `cargo clippy -p binius-hash --all-targets`, nightly `cargo +nightly fmt --all`, `cargo test -p binius-hash` — all clean (12 tests).
- Note: `cargo check --workspace` currently fails in `binius-prover` (intmul `SelectorMlecheckProver::new` arity) — a pre-existing breakage on `main`, unrelated to this change, which touches `crates/hash` only.

### Benchmark

Apple M-series (NEON, `simd_degree = 4`), 1 MiB of field elements folded into leaves, hashed the way `binary_merkle_tree::build` does (`digest_with_const_len`). Before = old per-leaf adapter; after = `Blake3ParallelDigest`.

| leaf size | before | after | speedup |
|---|---|---|---|
| 16 B (1 elem)  | 144 MiB/s | 144 MiB/s | 1.00× (routed to adapter) |
| 32 B (2 elem)  | 278 MiB/s | 278 MiB/s | 1.00× (routed to adapter) |
| 64 B (4 elem)  | 454 MiB/s | 1.03 GiB/s | 2.32× |
| 128 B (8 elem) | 496 MiB/s | 1.13 GiB/s | 2.33× |
| 256 B (16 elem)| 499 MiB/s | 1.18 GiB/s | 2.42× |
| 512 B (32 elem)| 521 MiB/s | 1.21 GiB/s | 2.37× |
| 1024 B (64 elem)| 533 MiB/s | 1.21 GiB/s | 2.33× |

Leaves of one block or more get 2.3×–2.4×; sub-block leaves stay at parity. The win scales with SIMD width, so AVX2 (8 lanes) and AVX-512 (16 lanes) should gain more than this NEON box.

### DIFF
--- a/crates/hash/src/blake3.rs
+++ b/crates/hash/src/blake3.rs
@@ -2,11 +2,22 @@
 
 //! Blake3 hash and compression functions for use in Merkle tree constructions.
 
+use std::{array, mem::MaybeUninit};
+
+use binius_utils::{FixedSizeSerializeBytes, SerializeBytes, rayon::iter::IndexedParallelIterator};
+use blake3::{
+	BLOCK_LEN, CHUNK_LEN, IncrementCounter, OUT_LEN,
+	platform::{Platform, le_bytes_from_words_32, words_from_le_bytes_32},
+};
 use digest::Output;
 
 use super::{
-	binary_merkle_tree::HashSuite, compress::CompressionFunction,
-	parallel_compression::ParallelCompressionAdaptor, parallel_digest::ParallelDigestAdapter,
+	binary_merkle_tree::HashSuite,
+	compress::CompressionFunction,
+	parallel_compression::ParallelCompressionAdaptor,
+	parallel_digest::{
+		MultiDigest, ParallelDigest, ParallelDigestAdapter, ParallelMultidigestImpl,
+	},
 };
 
 /// A two-to-one compression function that hashes the concatenation of its inputs with Blake3.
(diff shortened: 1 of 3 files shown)


crates/hash/src/blake3.rs
```diff
@@ -29,20 +40,360 @@ pub struct Blake3HashSuite;
...
+
+	fn digest_with_const_len<I: IntoIterator<Item: FixedSizeSerializeBytes>>(
+		&self,
+		n_items_per_input: usize,
+		source: impl IndexedParallelIterator<Item = I>,
+		out: &mut [MaybeUninit<Output<Self::Digest>>],
+	) {
+		// Every leaf serializes to the same fixed byte length.
+		let leaf_len = n_items_per_input * I::Item::BYTE_SIZE;
+		// Below one 64-byte block the SIMD kernel gives no speedup.
```
jimpo: That's unfortunate.

The comment is wrong, it's actually that the `hash_many` function doesn't even allow messages that aren't a multiple of the block length.

crates/hash/src/blake3.rs
```diff
@@ -29,20 +40,360 @@ pub struct Blake3HashSuite;
...
+	}
+
+	fn digest<I: IntoIterator<Item: SerializeBytes>>(
+		&self,
+		source: impl IndexedParallelIterator<Item = I>,
+		out: &mut [MaybeUninit<Output<Self::Digest>>],
+	) {
+		// Send every leaf through the SIMD batch.
+		// The batch picks the fast path or the scalar fallback per group at run time.
+		Blake3MultiParallel::new().digest(source, out);
```
jimpo: I'd say just fall back to the Parallel adapter here. This non-const-len version is never used and could even be removed everywhere if you want.

crates/hash/src/blake3.rs
```diff
@@ -29,20 +40,360 @@ pub struct Blake3HashSuite;
...
+	/// The hashing interface is streaming, so bytes arrive across calls and cannot be hashed
+	/// early. Buffering lets the SIMD batch see each lane's full message at finalization.
+	buffers: [Vec<u8>; N],
+}
+
+impl<const N: usize> Default for Blake3MultiDigest<N> {
+	fn default() -> Self {
+		// Every lane starts with an empty, unallocated buffer.
+		Self {
+			buffers: array::from_fn(|_| Vec::new()),
```
jimpo: Would be nice if the vecs had capacity, or were fixed size arrays. So either that this doesn't have a `Default` impl and instead has a `new` constructor that takes the byte-length (ParallelMultidigestImpl would also need a constructor that takes the MultiDigest impl as an arg). Or Blake3MultiDigest gets a `const M: usize` parameter for the byte length of each input.

crates/hash/src/blake3.rs
```diff
@@ -29,20 +40,360 @@ pub struct Blake3HashSuite;
...
+fn hash_many_blocks<const M: usize, const N: usize>(
+	platform: &Platform,
+	buffers: &[&[u8]; N],
+	flags_start: u8,
+	flags_end: u8,
+	out: &mut [[u8; OUT_LEN]; N],
+) {
+	// Reborrow each lane's leading `M` bytes as a fixed-size block-aligned input.
+	// The uniform-length precondition guarantees every lane holds at least `M` bytes.
+	let inputs: [&[u8; M]; N] = array::from_fn(|i| {
```
jimpo: This suggests that `Blake3MultiDigest` should itself have a `const M: usize` param so its buffer is already in this shape.

crates/hash/src/blake3.rs
```diff
@@ -29,20 +40,360 @@ pub struct Blake3HashSuite;
...
+		// Take the first lane's length as the candidate shared length.
+		let len = refs[0].len();
+
+		// The fast path needs one shared SIMD shape across all lanes:
+		// - every lane fits in one chunk of at most 1024 bytes, and
+		// - every lane has the same length.
+		if len <= CHUNK_LEN && refs.iter().all(|r| r.len() == len) {
+			Self::single_chunk_simd(&refs, len, out);
+		} else {
+			// No shared shape: hash each lane on its own with the scalar reference.
```
jimpo: Get rid of this. `Blake3MultiDigest` should only be used in the const len case.

crates/hash/src/blake3.rs
```diff
@@ -29,20 +40,360 @@ pub struct Blake3HashSuite;
...
+		let n_full = len / BLOCK_LEN;
+		let rem = len % BLOCK_LEN;
+		// Scratch space for one 32-byte chaining value per lane.
+		let mut cv_bytes = [[0u8; OUT_LEN]; N];
+
+		// Case 1: the length is a positive multiple of the block size.
+		// The chunk is all whole blocks, so the last block carries CHUNK_END | ROOT.
+		// One SIMD pass then emits the root digest of every lane directly.
+		if rem == 0 && len > 0 {
+			dispatch_full_blocks!(n_full =>
```
jimpo: Do the whole macro thing to construct the `Blake3MultiDigest` to simplify the code inside of it.

jimpo (changes requested):
Thanks for picking this up! Unfortunate limitation from the `blake3` library, but it doesn't seem worth it to write & maintain new cross-platform SIMD blake3 stuff.

jimpo:
Another approach would be to write another Blake3 impl in pure, portable Rust using vectorized-looking source code (so computing in a fixed-width loop), and trying to get LLVM to generate SIMD code for it. On ARM64, it might even be able to produce SVE2, which would be very fast.

I think that's actually worth an experiment.

crates/hash/src/blake3.rs
tcoratger: Fixed by rewriting the routine rationale

crates/hash/src/blake3.rs
tcoratger: Fixed

crates/hash/src/blake3.rs
tcoratger: Fixed with a const M

crates/hash/src/blake3.rs
tcoratger: Yes exact now

crates/hash/src/blake3.rs
tcoratger: Deleted the scalar fallback branch in compute and the whole equal-length / <= CHUNK_LEN runtime check.

crates/hash/src/blake3.rs
tcoratger: Fixed

tcoratger:
> Another approach would be to write another Blake3 impl in pure, portable Rust using vectorized-looking source code (so computing in a fixed-width loop), and trying to get LLVM to generate SIMD code for it. On ARM64, it might even be able to produce SVE2, which would be very fast.
> 
> I think that's actually worth an experiment.

Great idea, I've tried to explore it here https://github.com/binius-zk/binius64/pull/1660 with a benchmark comparing:
- Main
- This PR (1651)
- The portable PR (1660)

So that we can merge either 1651 or 1660.

tcoratger:
@jimpo Once #1660 is merged, we can close this one!

PR #1653, opened 2026-07-01
BINIUS-177: eliminate dead gates before emitting constraints

tcoratger:
Closes [BINIUS-177](https://linear.app/jimpo/issue/BINIUS-177/eliminate-dead-gates-before-emitting-constraints).

Skips gates whose output nothing observes, so they emit no constraint and claim no witness slot. Bit-exact — no protocol change, no new soundness assumption.

### The problem

The circuit compiler emits a constraint for every gate in the graph, unconditionally.

- Constant propagation folds full-constant gates but leaves them in the graph.
- The used-wire scan runs *after* emission, so it cannot undo a constraint already emitted.

So a gate whose output nothing reads still emits a dead AND/MUL constraint.

And because that dead constraint references the gate's output wire, the used-wire scan then marks that wire as committed — a witness slot spent on a value no one observes.

AND constraints are the prover's baseline cost unit, so both the constraint and the committed wire are pure waste.

### The idea

A gate matters only if its output is observed.

- Observed means it feeds an assertion, a public input/output, or a wire pinned as committed.
- Compute the set of live gates by transitive backward reachability from those roots.
- Skip dead gates at emission — they then cost neither a constraint nor a committed wire.

### The change

```
build():
  const-prop (unchanged)
+ live = dce::live_gates(graph, force_committed, hints)   // backward reachability from roots
  for gate in graph.gates:
+     if live.is_some() && !live.contains(gate) { continue }   // drop dead gates
      constrain(gate, ...)
```

### How this relates to constant folding — and what it is not

This is complementary to folding trivial bitwise/select identities, not a duplicate of it.

- Constant folding rewrites a gate whose *operands* are constant, at build time.
- This removes a whole gate whose *output* is transitively unread, after constant propagation.
- Constant propagation is exactly what *creates* many of these dead gates (it folds a gate to a constant but leaves the original behind).

### Soundness

- A dropped constraint pins a wire that no assertion or public output transitively reads.
- Removing a constraint on an unread wire cannot change what the proof establishes.
- The surviving constraint system is the old one restricted to observable gates.

### Scope

- **new:** `dce.rs` — the `live_gates` reachability pass plus unit tests.
- **changed:** `build()` computes the live set and skips dead gates; a compiler option (default on) gates the pass, with a `BINIUS_DISABLE_DCE` env hatch for A/B measurement.
- **left untouched:** the eval form still evaluates every gate, so witness generation is unchanged; a dropped gate's output simply becomes a scratch value.
- one existing test built an all-dead circuit as scaffolding; it now pins its result as committed so the linear cone it exercises survives.

### Verification

- `cargo test -p binius-frontend` — 150 pass, including two new unit tests (dead gate excluded / committed output kept).
- End to end: built the sha256 and keccak circuits with the pass on, generated real witnesses, and both proved **and** verified.

### Benchmark

AND-constraint count (the prover-cost metric), pass off vs on:

| circuit | DCE off | DCE on | reduction |
|---|---|---|---|
| sha256 | 16,946 | 16,938 | 8 (0.05%) |
| keccak | 6,914 | 6,675 | 239 (3.46%) |
| ethsign | 209,157 | 209,113 | 44 (0.02%) |
| hashsign (XMSS) | 853,400 | 838,164 | 15,236 (1.79%) |

The win is circuit-dependent: negligible for sha256 and ethsign, but a real cut for keccak and the XMSS hash-signature circuit, where const-prop and unused comparisons leave many dead gates behind.

Note: the committed `CircuitStat` snapshots under `crates/examples/snapshots/` record AND counts and will need re-blessing, since those counts legitimately drop.

### DIFF
--- a/crates/examples/snapshots/bitcoin_headers.snap
+++ b/crates/examples/snapshots/bitcoin_headers.snap
@@ -5,24 +5,24 @@ Gates & Instructions
 └─ Number of evaluation instructions: 72,938
 
 Constraints
-├─ AND constraints: 31,880 used (97.3% of 2^15)
-│  [▓▓▓▓▓▓▓▓▓░] spare: 888
+├─ AND constraints: 31,780 used (97.0% of 2^15)
+│  [▓▓▓▓▓▓▓▓▓░] spare: 988
 ├─ MUL constraints: 0 used (0.0% of 2^0)
 │  [░░░░░░░░░░] spare: 1
-└─ Distinct value indices: 73,026
-   ├─ Distinct shifted value indices: 41,322
-   └─ Distinct unshifted value indices: 31,704
+└─ Distinct value indices: 72,825
+   ├─ Distinct shifted value indices: 41,222
+   └─ Distinct unshifted value indices: 31,603
 
 Value Vector
 ├─ Public Section: 111 used (86.7% of 2^7)
 │  [▓▓▓▓▓▓▓▓░░] spare: 17
 │  ├─ Constants: 111
 │  └─ Inout: 0
-├─ Private Section: 31,594 used (96.8%)
-│  [▓▓▓▓▓▓▓▓▓░] spare: 1,046
+├─ Private Section: 31,494 used (96.5%)
+│  [▓▓▓▓▓▓▓▓▓░] spare: 1,146
 │  ├─ Witness: 804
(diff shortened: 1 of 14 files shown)


jimpo (approved):
Good idea!

PR #1654, opened 2026-07-01
BINIUS-179: automatically determine oracle specs via OracleSetupChannel

benji-potamus:
Closes [BINIUS-179](https://linear.app/jimpo/issue/BINIUS-179).

## What

`IOPVerifier::oracle_specs` (in both Binius64 `binius-verifier` and IronSpartan `binius-spartan-verifier`) previously returned a **hardcoded** oracle-spec sequence that had to be kept in sync by hand with the actual `recv_oracle` calls in `verify`. This derives it automatically instead.

## How

- **`IOPVerifierChannel::recv_oracle`** now takes the oracle's description:
  ```rust
  fn recv_oracle(&mut self, log_msg_len: usize, is_witness_dependent: bool) -> Result<Self::Oracle, Error>;
  ```
- **New `OracleSetupChannel`** (`crates/iop/src/oracle_setup_channel.rs`) — an `IOPVerifierChannel` that performs no real verification (all `recv_*` return zero, `assert_zero` is a no-op — the same dry-run shape as the existing `SizeTrackingChannel`) and simply records an `OracleSpec { log_msg_len, is_zk: self.is_zk && is_witness_dependent }` on each `recv_oracle`.
- **`oracle_specs()`** is reimplemented to run `verify` against an `OracleSetupChannel` and read back the recorded specs. Because the oracle sequence is data-independent, the dry-run recovers it exactly, and it now stays in lockstep with `verify` automatically. (Binius64: one witness oracle. IronSpartan: the precommit oracle is received by the outer `Verifier::verify` before delegating, so `oracle_specs()` records it first, then dry-runs `verify` for the private + mask oracles.)
- The concrete channels take the new signature: `NaiveVerifierChannel` uses `log_msg_len` for its transcript read; `BaseFold`/`SizeTracking`/wrapper channels read fixed-size data and keep their pre-supplied specs (which are now the auto-derived ones), so they ignore `log_msg_len`.

The `SizeTrackingChannel` precedent (used in `proof_size_test`) already proves `verify` runs cleanly on an all-zero dry-run channel, so this reuses a known-safe pattern.

## Notes / judgment calls

- **IronSpartan mask oracle** (`lib.rs`): it's a fresh random mask rather than witness data, but it's currently committed with ZK hiding (`new_zk`). I pass `is_witness_dependent = true` to **preserve that behavior** (keeping it ZK) — flagging in case you'd prefer the mask be treated as *not* witness-dependent (which would drop its masking and change the proof). Easy to flip.
- **`binius-m4-verifier`** has its own separate `oracle_specs` (a single non-ZK oracle); the issue scopes this to Binius64 + IronSpartan, so I only updated M4's `recv_oracle` **call** to the new signature and left its `oracle_specs` hardcoded. Happy to convert it too as a follow-up if you'd like the consistency.

## Test

Build + full test suite green across `binius-iop`, `binius-iop-prover`, `binius-verifier`, `binius-spartan-verifier`, `binius-spartan-prover`, `binius-m4-verifier` — the prove/verify roundtrip tests are the real gate here (they exercise `oracle_specs()` during setup and would fail if the derived specs didn't match, since the BaseFold compiler indexes FRI params by spec). `fmt --check`, `clippy --all-targets -D warnings`, and `rustdoc -D warnings` all clean.

### DIFF
--- a/crates/iop/src/channel.rs
+++ b/crates/iop/src/channel.rs
@@ -106,10 +106,18 @@ pub trait IOPVerifierChannel<'r, F: Field>: IPVerifierChannel<F> {
 
 	/// Receives an oracle commitment from the prover.
 	///
-	/// # Preconditions
-	///
-	/// `remaining_oracle_specs()` must be non-empty.
-	fn recv_oracle(&mut self) -> Result<Self::Oracle, Error>;
+	/// The caller describes the oracle being received: `log_msg_len` is the log2 of the message
+	/// length, and `is_witness_dependent` is whether the oracle's contents depend on the witness.
+	/// These let a channel record the oracle's [`OracleSpec`] rather than requiring the specs to be
+	/// supplied up front. The resulting oracle is zero-knowledge iff the channel is configured for
+	/// ZK *and* the oracle is witness-dependent — a non-witness-dependent oracle (e.g. a
+	/// pre-indexed commitment to the wiring matrix for succinctness, a planned feature) is never
+	/// masked.
+	fn recv_oracle(
+		&mut self,
+		log_msg_len: usize,
+		is_witness_dependent: bool,
+	) -> Result<Self::Oracle, Error>;
 
 	/// Queues oracle linear relations to be opened.
 	///
--- a/crates/iop/src/oracle_setup_channel.rs
+++ b/crates/iop/src/oracle_setup_channel.rs
@@ -1,0 +1,234 @@
+// Copyright 2026 The Binius Developers
+
+//! An [`IOPVerifierChannel`] that records the sequence of oracle specifications an IOP uses,
+//! without performing any actual verification.
+//!
+//! Running an IOP verifier against an `OracleSetupChannel` (a dry run: all `recv_*` methods return
+//! dummy values and `assert_zero` is a no-op, like [`SizeTrackingChannel`]) discovers the
+//! [`OracleSpec`] sequence directly from the `recv_oracle` calls, so the specs need not be
+//! hardcoded and kept in sync with the verification logic by hand.
+//!
+//! [`SizeTrackingChannel`]: crate::size_tracking_channel::SizeTrackingChannel
+
+use std::{
+	iter::{Product, Sum},
+	ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
+};
+
+use binius_field::{
+	BinaryField128bGhash, ExtensionField, Field, FieldOps,
+	arithmetic_traits::{InvertOrZero, Square},
+	util::FieldFn,
+};
+use binius_ip::channel::IPVerifierChannel;
+
+use crate::channel::{Error, IOPVerifierChannel, OracleLinearRelation, OracleSpec};
+
+/// A zero-sized dummy field element for [`OracleSetupChannel`].
+///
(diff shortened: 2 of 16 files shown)


crates/iop/src/channel.rs
```diff
@@ -106,10 +106,17 @@ pub trait IOPVerifierChannel<'r, F: Field>: IPVerifierChannel<F> {
 
 	/// Receives an oracle commitment from the prover.
 	///
-	/// # Preconditions
-	///
-	/// `remaining_oracle_specs()` must be non-empty.
-	fn recv_oracle(&mut self) -> Result<Self::Oracle, Error>;
+	/// The caller describes the oracle being received: `log_msg_len` is the log2 of the message
+	/// length, and `is_witness_dependent` is whether the oracle's contents depend on the witness
+	/// (as opposed to being, say, a fresh random mask). These let a channel record the oracle's
```
jimpo: The specific case where the oracle isn't witness dependent is when it's a pre-indexed commitment to the wiring matrix for succinctness. This isn't supported in the codebase yet, but is a planned feature.

crates/iop/src/oracle_setup_channel.rs
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	/// Consumes the channel and returns the recorded oracle specs, in the order received.
+	pub fn into_oracle_specs(self) -> Vec<OracleSpec> {
+		self.oracle_specs
+	}
+}
+
+impl<F: BinaryField> IPVerifierChannel<F> for OracleSetupChannel<F> {
+	type Elem = F;
```
jimpo: Perhaps `Elem = DummyElem` works, where DummyElem is a unit struct that impls `FieldOps`?

jimpo (changes requested):

crates/iop/src/oracle_setup_channel.rs
benji-potamus: Done. Added a `DummyElem` zero-sized unit struct that impls `FieldOps` (all arithmetic is no-ops, `Scalar = BinaryField128bGhash` to satisfy the `FieldOps<Scalar = B128>` bound `binius_verifier`'s verify requires). `OracleSetupChannel` is now non-generic (dropped the `<F>`), with `Elem = DummyElem` and blanket `impl<F: Field> IPVerifierChannel<F>` / `IOPVerifierChannel<'r, F>`. `compute_public_value` skips `f` and returns `DummyElem` (permitted by the trait contract). The dry run now does no real field arithmetic. Re-validated green (full test suite incl. roundtrips, clippy/fmt/rustdoc).

crates/iop/src/channel.rs
benji-potamus: Thanks — that clears up the semantics. Updated the doc to give that as the example (a pre-indexed commitment to the wiring matrix for succinctness, a planned feature) rather than the mask. So the IronSpartan mask stays `is_witness_dependent = true` (it's witness-independent in the "random" sense but not the "indexed/public" sense, and must remain masked) — the genuinely non-witness-dependent case is the future indexed commitment.

jimpo (approved):

benji-potamus:
Thanks for the review! Approved, all 8 CI checks green, `mergeable_state: clean`, based on current `main`. Ready to merge / add to the queue whenever you are.

PR #1655, opened 2026-07-01
BINIUS-178: use Word exponents in intmul Witness; borrow a_root

benji-potamus:
Closes [BINIUS-178](https://linear.app/jimpo/issue/BINIUS-178/reducing-cloning-in-intmul-witness-construction).

Two things:

**1. Use `Word` exponents in the intmul `Witness` (per review).** The `Witness` and its leaf-building helpers were generic over `B: Bitwise` / `S: AsRef<[B]>`, which forced callers to convert their `Vec<Word>` to `Vec<u64>` (the `convert_to_u64` dance in `prove.rs`, with a comment noting the fix was to implement `Bitwise` for `Word`). Now:
- `Witness<P, S: AsRef<[Word]>>` — the `B` generic and `PhantomData<B>` marker are gone; `two_valued_field_buffer` / `compute_b_leaves` select bits directly on `Word`.
- `Word` gains `Shr<usize>` + `From<u8>` so it satisfies `Bitwise` (phase 3's `SelectorMlecheckProver` still requires it), plus `Shr<i32>` (via a small macro) so shift-by-literal stays ergonomic — bare `word >> 63` literals default to `i32` and existing `u32` shift callers are unaffected, so no call sites outside intmul change.
- The `convert_to_u64` conversion at the prover boundary is deleted; exponents are `Word` end to end.

**2. Borrow `a_root` instead of cloning it.** `compute_b_leaves` only reads `bases`, so it takes `&FieldBuffer<P>` — drops the `a_root.clone()` in `Witness::new`.

The remaining big copy, `b_leaves.clone()`, is load-bearing and can't be pulled from the prodcheck's final sumcheck (index bits fold first); tracked as follow-up [BINIUS-180](https://linear.app/jimpo/issue/BINIUS-180) (needs a custom base-layer sumcheck).

## Verification

`cargo test` for `binius-core` / `binius-prover` / `binius-verifier` / `binius-ip` / `binius-ip-prover` — all pass, including the intmul roundtrip `prove_and_verify`. Full workspace builds (tests + benches); clippy `-D warnings` + fmt clean.

### DIFF
--- a/crates/core/src/word.rs
+++ b/crates/core/src/word.rs
@@ -94,6 +94,13 @@ impl Word {
 		Word(value)
 	}
 
+	/// Returns the bit at position `i`, counting from the least significant bit.
+	///
+	/// `i` must be in `0..64`.
+	pub const fn extract_bit(self, i: usize) -> bool {
+		(self.0 >> i) & 1 == 1
+	}
+
 	/// Performs parallel 32-bit additions on the upper and lower halves with carry-in.
 	///
 	/// Each 32-bit half is added independently, like [`sll32`](Word::sll32) operates on
--- a/crates/prover/src/protocols/intmul/prove.rs
+++ b/crates/prover/src/protocols/intmul/prove.rs
@@ -42,30 +42,27 @@ use crate::fold_word::{fold_across_words, fold_words};
 
 /// A helper structure that encapsulates switchover settings and the prover channel for
 /// the integer multiplication protocol.
-pub struct IntMulProver<'a, P, S, Channel> {
+pub struct IntMulProver<'a, P, Channel> {
 	_p_marker: PhantomData<P>,
-	_s_marker: PhantomData<S>,
 
 	switchover: usize,
 	channel: &'a mut Channel,
 }
 
-impl<'a, P, S, Channel> IntMulProver<'a, P, S, Channel> {
+impl<'a, P, Channel> IntMulProver<'a, P, Channel> {
 	pub const fn new(switchover: usize, channel: &'a mut Channel) -> Self {
 		Self {
 			_p_marker: PhantomData,
-			_s_marker: PhantomData,
 			switchover,
 			channel,
 		}
 	}
 }
 
-impl<'a, F, P, S, Channel> IntMulProver<'a, P, S, Channel>
+impl<F, P, Channel> IntMulProver<'_, P, Channel>
 where
 	F: BinaryField,
 	P: PackedField<Scalar = F>,
-	S: AsRef<[u64]> + Sync,
 	Channel: IPProverChannel<F>,
 {
 	/// Prove an integer multiplication statement.
@@ -91,7 +88,7 @@ where
 	/// The output of this protocol is a set of evaluation claims on the `b` selectors representing
 	/// all of `a`, `b`, `c_lo` and `c_hi` as column-major bit matrices, at a common evaluation
 	/// point.
-	pub fn prove(&mut self, witness: Witness<P, u64, S>) -> Result<IntMulOutput<F>, Error> {
+	pub fn prove(&mut self, witness: Witness<'_, P>) -> Result<IntMulOutput<F>, Error> {
 		let Witness {
 			log_bits,
 			a_exponents,
@@ -107,7 +104,6 @@ where
 			c_hi_prodcheck,
 			c_hi_root,
 			c_root,
-			_b_marker,
 		} = witness;
 
 		let n_vars = c_root.log_len();
@@ -144,7 +140,7 @@ where
 			&twisted_eval_points,
 			&twisted_evals,
 			a_root,
-			b_exponents.as_ref(),
+			b_exponents,
 			[c_lo_root, c_hi_root],
 			&initial_eval_point,
 			exp_eval,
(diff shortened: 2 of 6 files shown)


crates/prover/src/protocols/intmul/witness.rs
```diff
@@ -181,7 +181,7 @@ where
 /// The leaves are concatenated: `[L_0, L_1, ..., L_{2^k-1}]`
 fn compute_b_leaves<F, P, B, S>(
 	log_bits: usize,
-	bases: FieldBuffer<P>,
+	bases: &FieldBuffer<P>,
 	exponents: &S,
```
jimpo: This Bitwise abstraction thing isn't helpful. Just make the exponents Words.

jimpo (changes requested):

crates/prover/src/protocols/intmul/witness.rs
benji-potamus: Done (force-pushed `e68bfc0`). The intmul `Witness` is now concrete over `Word` exponents (`S: AsRef<[Word]>`) — the `B: Bitwise` / `S: AsRef<[B]>` generics and the `PhantomData<B>` marker are gone, and `two_valued_field_buffer` / `compute_b_leaves` do the bit selection directly on `Word`.

To get there I implemented what the old comment in `prove.rs` anticipated: `Word` now has `Shr<usize>` and `From<u8>`, so it satisfies `Bitwise` (which `SelectorMlecheckProver` in phase 3 still requires). That let me delete the `convert_to_u64` dance at the prover boundary — the exponents flow through as `Word` end to end.

One wrinkle: adding `Shr<usize>` alone made bare integer-literal shifts on `Word` (`word >> 63`, etc.) ambiguous, since they'd previously resolved to the sole `Shr<u32>`. I gave `Word` a small macro'd `Shr` over `i32`/`u32`/`usize` (like a primitive int) so literals default to `i32` and existing `u32` callers keep working — no call sites outside intmul needed changing. `a_root` is still borrowed rather than cloned. Tests + clippy + fmt green.

crates/prover/src/protocols/intmul/prove.rs
```diff
@@ -71,7 +71,7 @@ impl<'a, F, P, S, Channel> IntMulProver<'a, P, S, Channel>
 where
 	F: BinaryField,
 	P: PackedField<Scalar = F>,
-	S: AsRef<[u64]> + Sync,
+	S: AsRef<[Word]> + Sync,
```
jimpo: Is this generic param necessary? &'a [Word] should work just fine, I think.

crates/prover/src/protocols/intmul/witness.rs
```diff
@@ -37,7 +36,7 @@ use crate::protocols::prodcheck::ProdcheckProver;
 /// separately.
 #[derive(Clone, Getters)]
 #[getset(get = "pub")]
-pub struct Witness<P: PackedField, B: Bitwise, S: AsRef<[B]> + Sync> {
+pub struct Witness<P: PackedField, S: AsRef<[Word]> + Sync> {
```
jimpo: Can we just use &'a [Word] in place of S?

jimpo (changes requested):

crates/prover/src/protocols/intmul/witness.rs
benji-potamus: Done (force-pushed `9279595`). `Witness` is now `Witness<'a, P>` with `a_exponents`/`b_exponents`/`c_lo_exponents: &'a [Word]` — the `S: AsRef<[Word]>` generic is gone, and the leaf helpers (`constant_base_leaves`, `compute_b_leaves`, `two_valued_field_buffer`) take `&[Word]` directly.

crates/prover/src/protocols/intmul/prove.rs
benji-potamus: Right — dropped it. `IntMulProver` is now `IntMulProver<'a, P, Channel>` (no `S`), and `prove` takes `Witness<'_, P>`. All the exponent generics are gone; exponents flow as `&[Word]`. Tests + clippy + fmt green; rebased on latest main.

crates/core/src/word.rs
```diff
@@ -72,11 +72,28 @@ impl Shl<u32> for Word {
 	}
 }
 
-impl Shr<u32> for Word {
-	type Output = Self;
+// Right-shift by the common integer types so `Word` can act as a `Bitwise` bitmask (which requires
+// `Shr<usize>`) while keeping shift-by-literal ergonomic (integer literals default to `i32`) and
+// preserving existing `u32` shift callers.
+macro_rules! impl_word_shr {
```
jimpo: Revert this. The `Shr<u32>` impl was fine.

Instead add a method `extract_bit(&self, i: usize) -> bool`, which you can use in all the places in the prover where it appears you need right shift behavior.

jimpo (changes requested):

crates/core/src/word.rs
benji-potamus: Done (pushed `aa0f6a8`, rebased on latest main).

- Reverted `word.rs` to the single `impl Shr<u32>` and dropped the `From<u8>` impl.
- Added `Word::extract_bit(self, i: usize) -> bool` and used it for the bit selection in `constant_base_leaves`, `compute_b_leaves` (both paths), and `two_valued_field_buffer`.

One spot still needs a bitmask that satisfies `Bitwise`: the phase-3 `SelectorMlecheckProver`, which reads the `b` exponent bits through its compact `BitSelector`/switchover representation (shared ip-prover code, generic over a primitive-integer bitmask). Rather than reintroduce a `Bitwise` impl on `Word`, I pass a zero-copy `bytemuck::cast_slice::<Word, u64>(b_exponents)` there — `Word` is `repr(transparent)` over `u64`, which its own docs sanction. Happy to change if you'd prefer a different adapter.

Tests (incl. `prove_and_verify`) + clippy + fmt + rust-doc green.

jimpo (approved):

PR #1656, opened 2026-07-01
BINIUS-182: batch prove/verify for fracaddcheck

benji-potamus:
Implements [BINIUS-182](https://linear.app/jimpo/issue/BINIUS-182/batch-prove-and-verify-for-fracaddcheck).

## Summary

Adds batched proving for `fracaddcheck`, handling `m ≥ 1` instances of equal tree depth, modeled on `prodcheck::batch_prove`. New public API in `crates/ip-prover/src/fracaddcheck.rs`:

- **`batch_prove_until_final_layer(provers, claimed_fractions, selector_point, content_point, channel)`** — runs `n_layers - 1` reductions, then hands back, per input prover, its reduced `(num, den)` fraction paired with the `MleCheckProver` for its final (widest) layer, along with the shared evaluation point. A caller finishes the final layer (e.g. the logUp* final layer splicing it into another reduction). It returns a named struct generic over the prover type:
  ```
  pub struct BatchProveUntilFinalLayerOutput<F, MP> {
      pub eval_point: Vec<F>,
      pub final_layer: Vec<((F, F), MP)>,
  }
  ```
  where the function's `MP` is `impl MleCheckProver<F>`.
- **`batch_prove(...) -> BatchProveOutput { eval_point, fractions }`** — calls `batch_prove_until_final_layer`, runs the final layer's MLE-check, and returns each input prover's reduced `(num, den)` fraction at the reduced `eval_point` (one per input prover, no provers).

The `m` instances are combined by multilinear interpolation over `k = ⌈log₂ m⌉` selector variables (eq-weighting).

## Verify side: no new code

Mirroring prodcheck, the batched proof is checked by the **existing** `binius_ip::fracaddcheck::verify(k + n_layers, claim)` recursion, where the batched claim is the eq(selector)-weighted combination of the returned per-prover fractions, with the selector coordinates prepended to the claim point.

## The fracaddcheck-specific wrinkle

Each fracaddcheck layer is a 4-multilinear (`num_0, num_1, den_0, den_1`) MLE-check with **two** output claims combined via a `batch_coeff` that `batch_verify_mle` samples once per layer. So a layer reduction (`reduce_layer`) samples one `batch_coeff`, Horner-folds each instance's `[num, den]` round polynomials with it, eq(selector)-weights across instances for the content rounds, and runs the selector rounds as a `frac_add_mle` over the padded (2ᵏ) reduced halves with the *same* `batch_coeff`.

## Testing

Prover → (existing) verifier roundtrips, asserting the eq(selector)-weighted combination of `batch_prove`'s returned fractions equals the interpolation of the input witnesses:
`test_batch_prove_power_of_two_provers`, `test_batch_prove_non_power_of_two_provers` (padding), `test_batch_prove_single_prover`, `test_batch_prove_single_layer`, `test_batch_prove_with_content` (non-scalar fractions). All pass; `cargo fmt`, clippy (`-D warnings`), and `cargo doc -D warnings` are green.

Not wired into a consumer (e.g. logup*) in this PR — it adds the batching capability, as prodcheck's `batch_prove` was.

### DIFF
--- a/crates/ip-prover/src/fracaddcheck.rs
+++ b/crates/ip-prover/src/fracaddcheck.rs
@@ -1,15 +1,20 @@
 // Copyright 2025-2026 The Binius Developers
 
+use std::iter;
+
 use binius_field::{Field, PackedField};
-use binius_ip::prodcheck::MultilinearEvalClaim;
-use binius_math::{FieldBuffer, line::extrapolate_line_packed};
+use binius_ip::{mlecheck, prodcheck::MultilinearEvalClaim, sumcheck::RoundCoeffs};
+use binius_math::{
+	FieldBuffer, line::extrapolate_line_packed, multilinear::eq::eq_ind_partial_eval,
+};
 use binius_utils::rayon::iter::{IntoParallelIterator, ParallelIterator};
+use itertools::izip;
 
 use crate::{
 	channel::IPProverChannel,
 	sumcheck::{
 		batch::batch_prove_mle_and_write_evals,
-		common::MleCheckProver,
+		common::{MleCheckProver, SumcheckProver},
 		frac_add_mle::{self, FractionalBuffer},
 	},
 };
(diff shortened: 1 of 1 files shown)


jimpo (dismissed):
batch_prove should not return provers -- it should output the eval point and the fractional evaluations for each input prover as a Vec.

However, make another function called batch_prove_until_final_layer, which doesn't do the final MLE-check. Instead of returning FracAddProver though, it should get the final layer provers and return the `impl MleCheckProver<F>` for each input prover. So the output shape is the evaluation point for the final MLE check, then a vec of fractional evals along with MLE-check provers for the final layer.

`batch_prove` simply calls `batch_prove_until_final_layer`, then does the final mle-check (and returning the new eval point, of course).

benji-potamus:
Addressed the review + rebased onto `main` (`98d5a91`).

**Review change** — split the API as requested:
- `batch_prove` no longer returns provers. It returns `BatchProveOutput { eval_point, fractions }` — the eval point and each input prover's reduced `(num, den)` fraction as a `Vec` — by calling `batch_prove_until_final_layer` and then running the final layer's MLE-check.
- New `pub fn batch_prove_until_final_layer(...) -> (eval_point, Vec<((num, den), impl MleCheckProver<F>)>)` — runs `n_layers - 1` reductions and, for each input prover, pops its final layer as an `impl MleCheckProver<F>` (via `layer_prover`) paired with its fraction, so a caller (e.g. the logUp* final layer) can finish it.

It's a pure restructure of the existing (passing) reduction — the transcript is unchanged, so `binius_ip::fracaddcheck::verify` still checks it with no changes.

**Rebase note:** `main` (`#1693`) made the `MleCheckProver` sumcheck API infallible since this branch was opened, so I also dropped the now-dead `Result`/`?`/`SumcheckError` plumbing from the batch code. One nuance worth flagging: `batch_prove_until_final_layer` returns `impl MleCheckProver<F>`, and under Rust 2024's impl-Trait capture rules I had to name the channel generic and add `+ use<F, P, Channel>` so the returned provers don't borrow the channel (letting `batch_prove` reuse it for the final layer).

Full `binius-ip-prover` suite green (incl. all 5 batch tests); clippy `-D warnings`, `cargo doc -D warnings`, nightly fmt, typos clean.

crates/ip-prover/src/fracaddcheck.rs
```diff
@@ -223,15 +228,357 @@ where
...
+///
+/// Arguments and preconditions are as for [`batch_prove`].
+#[allow(clippy::type_complexity)]
+pub fn batch_prove_until_final_layer<F, P, Channel>(
+	provers: Vec<FracAddCheckProver<P>>,
+	claimed_fractions: Vec<(F, F)>,
+	selector_point: Vec<F>,
+	content_point: Vec<F>,
+	channel: &mut Channel,
+) -> (Vec<F>, Vec<((F, F), impl MleCheckProver<F> + use<F, P, Channel>)>)
```
jimpo: Make the output type a struct with a Box<dyn MleCheckProver<F>>.

jimpo (changes requested):

jimpo:
Also rebase onto main

crates/ip-prover/src/fracaddcheck.rs
benji-potamus: Rebased onto main (`7e3e40a`).

On the struct: `MleCheckProver`/`SumcheckProver` is explicitly **not object-safe** (the doc says so, and `fn finish(self) -> Vec<F>` takes `self` by value), so `Box<dyn MleCheckProver<F>>` doesn't compile directly — a trait object can't have a by-value-`self` method, and there's no `Box<dyn …Prover>` anywhere in the tree yet. Which way do you want to go?

1. **Object-safe wrapper (my default):** add a small `BoxedMleCheckProver<F>` trait (`finish(self: Box<Self>)`, everything else delegating) with a blanket impl for `T: MleCheckProver<F>`, so the output can be a struct holding `Box<dyn BoxedMleCheckProver<F>>`. Localized, ~30 lines; the output becomes a clean non-generic struct as you asked.
2. **Make the base trait object-safe:** change `SumcheckProver::finish(self)` → `finish(self: Box<Self>)` (touches every prover impl). Cleaner long-term but a wider change than this PR.
3. **Named struct, generic over the prover** (`…Output<F, MP: MleCheckProver<F>>`) — a struct but keeps the type parameter instead of erasing it.

Happy to do (1) unless you'd prefer (2)/(3).

benji-potamus: Went with option (1) (`3153323`), since it delivers the struct-with-boxed-prover you asked for and is the most localized. `batch_prove_until_final_layer` now returns a struct:

```rust
pub struct BatchProveUntilFinalLayerOutput<F: Field> {
    pub eval_point: Vec<F>,
    pub final_layer: Vec<((F, F), Box<dyn BoxedMleCheckProver<F>>)>,
}
```

`BoxedMleCheckProver<F>` is a small object-safe mirror of `MleCheckProver` (`finish(self: Box<Self>)`, everything else delegating) with a blanket impl for every `T: MleCheckProver<F>`, plus `SumcheckProver`/`MleCheckProver` impls for `Box<dyn BoxedMleCheckProver<F>>` so the boxed final-layer provers flow through the same `reduce_layer` reduction as concrete ones. The `impl Trait + use<F, P, Channel>` from the last revision is gone.

One wrinkle: because the blanket impl gives every `MleCheckProver` an identically named `execute`/`fold`/`finish`, the reduction's method calls needed `SumcheckProver::`-qualification to disambiguate (noted in a comment). If you'd rather avoid that and just make the base trait object-safe — option (2), `SumcheckProver::finish(self: Box<Self>)` — I'm happy to switch; it's cleaner at the trait but touches every prover impl. Full suite green (55 tests), clippy `-D`, doc `-D`, fmt, typos clean.

jimpo: Forget the boxing, and revert that part. Go with option (3), the struct generic over the MleCheckProver.

benji-potamus: Done (`93ac0c5`) — reverted the boxing (removed `BoxedMleCheckProver` and the `SumcheckProver::` qualifications) and went with option (3): the output is a named struct generic over the prover.

```rust
pub struct BatchProveUntilFinalLayerOutput<F, MP> {
    pub eval_point: Vec<F>,
    pub final_layer: Vec<((F, F), MP)>,
}
```

`batch_prove_until_final_layer` returns `BatchProveUntilFinalLayerOutput<F, impl MleCheckProver<F> + use<F, P, Channel>>`. Also rebased onto latest main (`bd3deef6`). Full suite green (55 tests), clippy `-D`, doc `-D`, fmt, typos clean.

jimpo (approved):

PR #1657, opened 2026-07-01
BINIUS-183: deduplicate structurally-identical gates before emitting constraints

tcoratger:
Closes [BINIUS-183](https://linear.app/jimpo/issue/BINIUS-183/deduplicate-structurally-identical-gates-before-emitting-constraints).

Collapses gates that compute the same thing, so a repeated subexpression costs one constraint instead of several. Bit-exact — no protocol change, no new soundness assumption.

Rebased on top of the merged dead-code-elimination pass (#1653); the two passes are complementary and coexist.

### The problem

The circuit compiler interns constants, but nothing else.

A circuit that builds the same non-constant subexpression twice — same opcode, same input wires, same immediates — emits a separate gate each time.

- Each such gate becomes its own AND/MUL constraint and its own committed wire.
- AND constraints are the prover's baseline cost unit, so the duplicates are pure waste.
- Hand-written hashes have almost none; generated and crypto circuits (secp256k1, HD-wallet derivation, zk-login) have many.

### The idea

Two gates with the same opcode, constant and input wires, immediates, and dimensions compute identical outputs.

- Keep the first occurrence as the canonical gate.
- Collapse every later duplicate onto it, pointing the duplicate's readers at the canonical wires.
- A collapsed duplicate then emits no constraint and claims no witness slot.

### The change

One forward sweep in emission (topological) order:

```
for gate in emission_order:
    rewrite gate's input wires through the remap    // canonicalize inputs
    h = structural_hash(gate)                        // opcode, const+input wires, imm, dims
    if canonical[h] is a true structural match and gate has no observable output:
        remap[gate.outputs[i]] = canonical_outputs[i]   // collapse
        mark gate dead
    else:
        canonical[h] = gate

# the constrain loop skips dead gates
```

A reader is always visited after the gate that feeds it, so rewriting each gate's own inputs redirects every reader of a collapsed duplicate. The pass therefore needs no use-def chains and allocates almost nothing.

The pass runs before dead-code elimination, so the dead-code pass sees the canonicalized graph.

### Soundness

- A collapsed duplicate is bit-identical to its canonical gate: same structure means same value.
- Its readers are redirected to the canonical wire, which holds exactly that value.
- Public outputs and force-committed wires are never collapsed, since their wires must stay in place.
- Hash collisions are rejected by an explicit structural-equality check, so a collision never causes a false collapse.

### Scope

- **new:** `cse.rs` — the `dedup_gates` sweep plus unit tests (duplicate collapsed / distinct kept / transitive collapse / observable kept).
- **changed:** `build()` runs the pass after constant propagation and before dead-code elimination, skipping collapsed duplicates; a compiler option (default on) gates it, with a `BINIUS_DISABLE_CSE` env hatch for A/B measurement.
- **left untouched:** the eval form still evaluates every gate, so witness generation is unchanged; a collapsed duplicate's output simply becomes a scratch value.
- the gate-fusion snapshot tests opt out of the pass to stay focused on fusion.

### Verification

- `cargo test -p binius-frontend` — 154 pass (CSE and dead-code unit tests together); `cargo test -p binius-circuits` — 334 pass (build gadgets, generate witnesses, run `verify_constraints`).
- Built the keccak and ethsign circuits and both proved **and** verified with the pass on.

### Benchmark

AND-constraint count, this pass off vs on (dead-code elimination on in both, so this isolates CSE's marginal effect):

| circuit | CSE off | CSE on | reduction |
|---|---|---|---|
| sha256 | 16,938 | 16,908 | 30 (0.18%) |
| keccak | 6,675 | 6,652 | 23 (0.34%) |
| hashsign (XMSS) | 301,536 | 298,066 | 3,470 (1.15%) |
| bip32 | 905,911 | 861,937 | 43,974 (4.86%) |
| ethsign (secp256k1) | 209,134 | 197,938 | 11,196 (5.35%) |
| ec_msm | 207,239 | 196,006 | 11,233 (5.42%) |
| zklogin | 430,714 | 370,749 | 59,965 (13.92%) |

The win is circuit-dependent: negligible on hand-written hashes, but a real cut on the secp256k1 and zk-login circuits, where repeated select/comparison/AND subexpressions abound. The two passes target disjoint gates (unreachable vs duplicate), so CSE's reduction is essentially the same with or without dead-code elimination.

Build-time cost is negligible: on bip32 (~1M gates) the circuit build is unchanged with the pass on, since it is a single allocation-light sweep with no use-def rebuild.

Note: the committed `CircuitStat` snapshots under `crates/examples/snapshots/` are re-blessed in this PR, since those AND counts legitimately drop.


jimpo (approved):
Interesting that this is so common. Since it does seem to produce savings, lets do it.

tcoratger:
> Interesting that this is so common. Since it does seem to produce savings, lets do it.

Just fixed conflicts :)

tcoratger:
@jimpo Just noticed new conflicts, solved

jimpo:
More conflicts :-( But lots of code improvements :grin:

tcoratger:
> More conflicts :-( But lots of code improvements 😁

No problem, just fixed conflicts here

PR #1658, opened 2026-07-01
BINIUS-185: benchmark intmul prover phases and costly components

benji-potamus:
Closes [BINIUS-185](https://linear.app/jimpo/issue/BINIUS-185).

## What

Extends `crates/prover/benches/intmul.rs` to benchmark the intmul prover broken down by **phase** and by **costly component**, as requested:

- **Per-phase benches** (`intmul/phases` group): `phase1`, `phase2_frobenius_twist`, `phase3`, `phase4`, `phase5` — each measured in isolation.
- **Component benches** (`intmul/components` group), the three the issue names:
  - `b_leaves` — computing the variable-base exponentiation tree leaves (`a` root as base, `b` as exponents) via `compute_b_leaves`.
  - `product_tree` — building the GKR product tree over those leaves (`ProdcheckProver::new`).
  - `selector_sumcheck` — constructing + proving the phase-3 `SelectorMlecheckProver`.

The existing coarse benches (`witness` / `prove` / `combined`) are unchanged in behavior; their group was renamed from the (misleading) `intmul_phases` to `intmul/prove`, since the real per-phase breakdown now lives in `intmul/phases`.

## How

The proving phases are **sequential and stateful** — each consumes the previous one's outputs plus the transcript. To measure a phase in isolation, its inputs are reconstructed in the (unmeasured) `iter_batched` **setup** by running the earlier phases against a throwaway transcript; the phase itself is then measured against a fresh transcript. The specific transcript challenges don't affect the work a phase performs, so a fresh transcript (and random initial evaluation point) is sufficient, and this avoids threading a `&mut Channel` across the setup/measure boundary. Three small helpers (`run_to_phase2/3/4`) build the prefixes.

## Design call to flag

Benches are separate crates, so they can only call `pub` items. To let the bench drive each phase, this exposes `IntMulProver::phase1/phase3/phase4/phase5` and `compute_b_leaves` as **`#[doc(hidden)] pub`** — the "public for benchmarking, not a stable API" idiom (keeps them out of the rendered docs). The `PhaseNOutput` structs were already `pub` in `binius-verifier`. This is isolated in the first commit (`expose intmul prover phases…`). If you'd prefer a different seam (e.g. keeping them private and only benching the three components, or a dedicated bench-only feature gate), easy to change.

## Notes

- Benches run **single-threaded** (`ThreadPoolBuilder::num_threads(1)`), matching the pre-existing intmul bench — measures total work rather than parallel wall-clock. Easy to lift globally if you'd rather see the parallel numbers.
- Sizes match the existing bench (`log_num = 14`, `log_bits = 6`), `sample_size(10)`. The `phase4`/`phase5` benches have heavy setup (they re-run the earlier phases each iteration), so those groups take a while — correct, just not fast.

## Test

`cargo bench -p binius-prover --bench intmul -- --test` runs all 11 benchmarks once: all `Success` (the reconstructed phase inputs satisfy every internal assertion). `cargo test -p binius-prover intmul` green (incl. the prove/verify roundtrip — the visibility change is behavior-neutral). `fmt` and `clippy --tests --benches -D warnings` clean.

### DIFF
--- a/crates/prover/benches/intmul.rs
+++ b/crates/prover/benches/intmul.rs
@@ -1,15 +1,33 @@
-// Copyright 2025 Irreducible Inc.
+// Copyright 2025-2026 The Binius Developers
 use binius_core::word::Word;
-use binius_field::PackedBinaryGhash1x128b;
-use binius_prover::protocols::intmul::{prove::IntMulProver, witness::Witness};
+use binius_field::{BinaryField128bGhash, Field, PackedBinaryGhash1x128b};
+use binius_ip_prover::{
+	prodcheck::ProdcheckProver,
+	sumcheck::{
+		batch::batch_prove,
+		selector_mle::{Claim, SelectorMlecheckProver},
+	},
+};
+use binius_math::{
+	multilinear::{eq::eq_ind_partial_eval_scalars, evaluate::evaluate},
+	test_utils::random_scalars,
+};
+use binius_prover::protocols::intmul::{
+	prove::IntMulProver,
+	witness::{Witness, compute_b_leaves, constant_base_leaves},
+};
 use binius_transcript::ProverTranscript;
 use binius_utils::rayon::ThreadPoolBuilder;
-use binius_verifier::config::StdChallenger;
-use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+use binius_verifier::{config::StdChallenger, protocols::intmul::common::frobenius_twist};
+use criterion::{BatchSize, BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
 use rand::prelude::*;
 
 type P = PackedBinaryGhash1x128b;
+type F = BinaryField128bGhash;
 
+/// Number of exponents is `2^LOG_NUM`.
+const LOG_NUM: usize = 14;
+/// Log of the integer bit width; the exponentiation tree has `2^LOG_BITS` leaves per node.
 const LOG_BITS: usize = 6;
 
 fn generate_test_data(log_num: usize) -> (Vec<Word>, Vec<Word>, Vec<Word>, Vec<Word>) {
@@ -42,13 +60,12 @@ fn generate_test_data(log_num: usize) -> (Vec<Word>, Vec<Word>, Vec<Word>, Vec<W
 fn bench_intmul_prove(c: &mut Criterion) {
 	ThreadPoolBuilder::new().num_threads(1).build_global().ok();
 
-	let mut group = c.benchmark_group("intmul_phases");
+	let mut group = c.benchmark_group("intmul/prove");
 	group.sample_size(10);
 	group.throughput(Throughput::Elements(1));
 
-	let log_num = 14;
-	let num_exponents = 1 << log_num;
-	let (a, b, c_lo, c_hi) = generate_test_data(log_num);
+	let num_exponents = 1 << LOG_NUM;
+	let (a, b, c_lo, c_hi) = generate_test_data(LOG_NUM);
 
 	group.bench_with_input(
 		BenchmarkId::new("witness", num_exponents),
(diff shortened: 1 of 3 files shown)


crates/prover/benches/intmul.rs
```diff
@@ -95,5 +120,278 @@ fn bench_intmul_prove(c: &mut Criterion) {
...
+					.unwrap()
+			},
+			BatchSize::SmallInput,
+		);
+	});
+
+	group.bench_function("phase2_frobenius_twist", |bencher| {
+		let mut rng = rand::rng();
+		bencher.iter_batched(
+			|| run_to_phase2(&witness, &mut rng),
```
jimpo: You don't need to rerun all preceding steps in the setup. Between each group, just advance the state of the protocol, then clone the datastructures that get consumed by the phase in this setup closure.

crates/prover/benches/intmul.rs
```diff
@@ -95,5 +120,278 @@ fn bench_intmul_prove(c: &mut Criterion) {
...
+			},
+			BatchSize::SmallInput,
+		);
+	});
+
+	group.finish();
+}
+
+fn bench_intmul_components(c: &mut Criterion) {
+	ThreadPoolBuilder::new().num_threads(1).build_global().ok();
```
jimpo: Get rid of this.

jimpo (changes requested):
Also bench the component `constant_base_leaves`.

crates/prover/benches/intmul.rs
benji-potamus: Done. The phases group now advances the protocol **once** (phase1→phase4 on a throwaway transcript) at the top of `bench_intmul_phases`, capturing each phase's inputs. Each phase's per-iteration setup only clones the datastructures that phase consumes by value — `phase1` clones `b_prodcheck`, `phase3` clones the roots, `phase4` clones the three prodcheck provers, `phase5` clones the retained leaf layers. `phase2_frobenius_twist` consumes nothing by value so it's a plain `iter()` with no clone. The `run_to_phase2/3/4` helpers are gone. Same treatment applied to `selector_sumcheck` in the components group (phase2 derived once, setup just rebuilds the cheap claims/eq-weights).

crates/prover/benches/intmul.rs
benji-potamus: Removed. `build_global()` only takes effect on its first call per process, so the copies in `bench_intmul_phases` and `bench_intmul_components` were dead — `bench_intmul_prove` (first in the group) already pins the pool to one thread. Dropped both duplicates; kept the single one in `bench_intmul_prove`.

benji-potamus:
Pushed `43976ced` addressing the review:

1. **`constant_base_leaves` component bench** added to `intmul/components` (exposed `#[doc(hidden)] pub` alongside `compute_b_leaves` in commit 1). Benches the constant-base exponentiation-tree leaves (fixed generator base, `a` exponents).
2. **Setup no longer re-runs preceding phases per iteration** — the protocol is advanced once, and each phase's `iter_batched` setup just clones what it consumes by value (`run_to_phase2/3/4` helpers deleted). Same for `selector_sumcheck`.
3. **Removed the redundant `build_global()`** calls (dead after the first).

Also rebased onto latest `main`, which had moved `prodcheck`/`sumcheck` into `binius-ip-prover` — updated the bench imports accordingly.

Validation (from the worktree, `-Ctarget-cpu=native`): `cargo bench --bench intmul -- --test` → all **12** benches `Success`; `cargo test -p binius-prover intmul` → 4 passed; `clippy -D warnings --tests --benches` + nightly `fmt --check` clean.

benji-potamus:
Rebased onto latest `main` (`e2c6b916`), which now includes #1655 (Word exponents in the intmul `Witness` + borrowed `a_root`). Adapted the bench + exposure to the new API: `Witness<'a, P>` with `&[Word]` exponents, `IntMulProver<'a, P, Channel>` (no `S` param), `compute_b_leaves(log_bits, &a_root, &[Word])` / `constant_base_leaves(log_bits, g, &[Word])`, and the selector prover fed `bytemuck::cast_slice(b_exponents)` bitmasks (mirroring `phase3`). Since `compute_b_leaves` now borrows its base, the `b_leaves` component bench no longer needs a per-iteration clone. All review points from the previous round are still in place. Re-validated: 12 benches `Success`, intmul tests + clippy(`-D warnings`) + nightly fmt green.

jimpo (approved):

PR #1659, opened 2026-07-01
BINIUS-186: fold idempotent bitwise identities on equal operands

tcoratger:
Closes [BINIUS-186](https://linear.app/jimpo/issue/BINIUS-186/fold-idempotent-bitwise-identities-on-equal-operands).

Folds bitwise ops whose two operands are the same wire, so a trivial identity costs no gate. Bit-exact — no protocol change, no new soundness assumption.

### The problem

The builder emits a gate for `band`, `bor`, `bxor`, and `select` even when both operands are the same wire.

- `band(x, x)` and `bor(x, x)` each become an AND constraint.
- `select(_, t, t)` becomes an AND constraint; `bxor(x, x)` a linear gate.
- Each computes a trivial identity, yet costs a gate (and, for AND, a committed wire).

Constant operands are already foldable, but equal non-constant operands are not.

### The idea

On equal operands the result is a known identity that needs no gate:

```
band(x, x)      -> x
bor(x, x)       -> x
bxor(x, x)      -> 0
select(_, t, t) -> t
```

Compare the two operand wires at the builder site and return the simplified wire.

### The change

```
band / bor / select : if the operands are the same wire -> return it, emit no gate
bxor                : if the operands are the same wire -> return the interned zero constant
```

### Soundness

- Each identity is bit-exact for all 64 bits: `x AND x = x`, `x OR x = x`, `x XOR x = 0`.
- `select(_, t, t) = t` reads no bit of the condition, so it is independent of the MSB-boolean convention.
- No new assumption; a folded op emits nothing.

- Complementary to constant folding (which folds a *constant* operand) and to common-subexpression elimination (which collapses two *duplicate gates*).
- This folds a *single* gate whose two operands are equal, which neither of those catches.
- The absorption identities (`band(x, bor(x, y)) -> x`, `bxor(x, bxor(x, y)) -> y`) are out of scope: they need per-operand def-tracking and are partly subsumed by common-subexpression elimination.

### Scope

- **changed:** `band` / `bor` / `bxor` / `select` in `mod.rs`; a compiler option (default on) with a `BINIUS_DISABLE_ALGEBRAIC_FOLDING` env hatch for A/B measurement.
- **tests:** a structural fold check plus a witness / `verify_constraints` check that `x ^ x` carries 0.
- **left untouched:** the gate-fusion snapshot tests opt out of the pass to stay focused on fusion.

### Verification

- `cargo test -p binius-frontend` — 152 pass (incl. the two new fold tests); `cargo test -p binius-circuits` — 334 pass (build gadgets, generate witnesses, run `verify_constraints`).

### Benchmark

AND-constraint count, folding off vs on (dead-code elimination on in both):

| circuit | folding off | folding on | reduction |
|---|---|---|---|
| hashsign (XMSS) | 301,536 | 301,536 | 0 |
| sha256 | 16,938 | 16,938 | 0 |
| keccak | 6,675 | 6,674 | 1 |
| zklogin | 430,714 | 430,708 | 6 |
| ethsign (secp256k1) | 209,134 | 207,281 | 1,853 (0.89%) |

Honest read: the win is narrow. Hand-written hashes never write `x & x` or `x ^ x`, so the folds are inert there. The only material hit is the secp256k1 circuit (~0.9%), whose constant-time field code produces equal-operand selects. It is a cheap, safe simplification that mostly pays off on structured field arithmetic.

Note: the committed `CircuitStat` snapshots under `crates/examples/snapshots/` are re-blessed in this PR, since a few AND counts legitimately drop.


jimpo (approved):

jimpo:
Sorry, needs a re-bless

tcoratger:
> Sorry, needs a re-bless

Just fixed the remaining conflicts

PR #1660, opened 2026-07-01
BINIUS-181: experimental portable auto-vectorized Blake3 leaf hasher

tcoratger:
Experiment for [BINIUS-181](https://linear.app/jimpo/issue/BINIUS-181/blake3-parallel-impl-using-auto-vectorization).

A pure-Rust Blake3 multi-lane leaf hasher that leans on LLVM auto-vectorization instead of the `blake3` crate's hand-written SIMD kernel. Output is bit-identical to `blake3::hash`.

> **Standalone alternative to [#1651](https://github.com/binius-zk/binius64/pull/1651) (BINIUS-168), not stacked on it.** This branch is cut from `main` and adds only new files, so it touches none of #1651's code. A reviewer can merge either PR without the other. All three columns below were measured together on the same machine in one session, each arm run on its own so thermal throttling hits them equally. #1651's kernel is not in this branch; its column was measured by temporarily overlaying that PR's code, then discarded.

### The idea

The `blake3` crate hits SIMD through hand-written intrinsics, one backend per target (NEON, AVX2, AVX-512).

The bet from the BINIUS-168 review: write the compression in *scalar-looking* Rust and let LLVM find the SIMD.

- Hold each of the 16 compression-state words as `[u32; N]`, one lane per message.
- Write every step as a fixed-width `0..N` loop of plain `u32` add / xor / rotate.
- The lanes are independent, so the vectorizer maps the loop onto whatever the target has.

```
state word v[i]:  [ lane_0 | lane_1 | ... | lane_{N-1} ]   // one [u32; N]
one G-mix line :  for lane in 0..N { v[a][lane] += v[b][lane] + m[lane] }   // vectorizes across lanes
```

Each lane keeps a single 64-byte block buffer and runs the standard Blake3 chunk-state stream: fill a block, compress it, and defer the trailing block to finalization where it alone carries `CHUNK_END | ROOT`. No intrinsics, no `unsafe`, no per-target code, and no compile-time length.

### Why keep this over the crate's SIMD kernel

**It hashes messages shorter than one block.** The crate's `hash_many` only accepts whole-block-multiple inputs, so sub-block and partial-block leaves fall back to the scalar path. This kernel handles any length up to one chunk in the vectorized path — the sub-block leaves exercised by the ZK leaf-hashing path included.

### The finding

Apple M-series (NEON), fixed 1 MiB of field elements folded into leaves, single-threaded per batch. Throughput, higher is better. Portable column is `N = 16` lanes.

| leaf size | main (scalar adapter) | #1651 (crate SIMD) | portable (this PR) |
|---|---|---|---|
| 16 B  | 144 MiB/s | 144 MiB/s * | **233 MiB/s** |
| 32 B  | 277 MiB/s | 277 MiB/s * | **459 MiB/s** |
| 48 B  | 393 MiB/s | 392 MiB/s * | **679 MiB/s** |
| 64 B  | 457 MiB/s | 1.03 GiB/s | 964 MiB/s |
| 128 B | 495 MiB/s | 1.13 GiB/s | 1.07 GiB/s |
| 256 B | 500 MiB/s | 1.18 GiB/s | 1.15 GiB/s |
| 512 B | 521 MiB/s | 1.21 GiB/s | 1.21 GiB/s |
| 1024 B| 533 MiB/s | 1.21 GiB/s | 1.23 GiB/s |

`*` The crate kernel cannot hash sub-block messages, so at 16 / 32 / 48 B it falls back to the scalar adapter (the `main` column).

Two regimes:

- **Sub-block (16–48 B):** portable is **1.6×–1.7×** over both other paths, which drop to scalar here. This is the new capability and the reason to keep this approach.
- **Whole-block (64–1024 B):** portable tracks the crate kernel within a few percent — behind at 64–128 B (0.92×–0.95×), level at 256–512 B, ahead at 1024 B (1.02×). Both are ~2.1×–2.4× over the scalar baseline.

### Lane width is the whole story

The batch width `N` handed to the vectorizer is not monotonic — the naive choice is a trap. Throughput at a 512 B leaf:

| kernel | throughput |
|---|---|
| portable, 4 lanes  | 563 MiB/s |
| portable, 8 lanes  | 478 MiB/s |
| portable, 16 lanes | 1.17 GiB/s |

- **4 lanes** (one NEON register per state word): barely above scalar — four independent lanes don't give the pipeline enough parallelism to hide the per-lane dependency chain.
- **8 lanes**: *slower than 4* — 16 state words $\times$ 2 registers each spill the 32-register NEON file.
- **16 lanes**: the sweet spot — enough independent work to saturate the units.

### The real prize is elsewhere

The NEON parity is fine, but not the point. On **AVX-512 / SVE2**, 16 lanes map to one register per word — untestable on this Apple box, and where the portable approach could pull clearly ahead. That, plus the sub-block support, is why this stays worth pursuing.

### Correctness

- Output is pinned bit-identical to `blake3::hash` across lengths `0..=1024` and lane widths (4 / 8 / 16).
- A split-update test checks that a message fed in two `update` calls hashes the same as one.
- Routing is checked end-to-end: sub-block, partial, whole-block, and multi-chunk leaf sizes all reproduce the reference.

### Scope

- **new:** `crates/hash/src/blake3_portable.rs` (the portable kernel + `PortableBlake3ParallelDigest<LANES>`) and `crates/hash/benches/blake3_portable.rs`.
- **new:** one `pub mod` line and one `[[bench]]` entry.
- **left to future work:** multi-chunk messages (`> 1024` B) still need the tree / CV-stack, so they route to the scalar adapter; the ideal `LANES` is ISA-specific (16 on NEON).
- the scalar `blake3::hash` stays as the test reference; the in-repo bench compares the scalar baseline against the portable kernel at 4 / 8 / 16 lanes.

### Verification

- `cargo clippy -p binius-hash --all-targets`, nightly `cargo +nightly fmt` — clean.
- `cargo test -p binius-hash` — 12 tests pass, including the portable length-sweep, split-update, and routing tests.
- Note: `cargo check --workspace` has a pre-existing unrelated breakage in `binius-prover` on this base; this change touches `crates/hash` only.

### DIFF
--- a/crates/hash/src/blake3_portable.rs
+++ b/crates/hash/src/blake3_portable.rs
@@ -15,7 +15,7 @@
 //! - SVE2 on capable ARM64 -> width-agnostic vectors.
 //!
 //! Output is bit-identical to `blake3::hash`, pinned to the reference in tests.
-//! Scope: a single chunk of whole 64-byte blocks (`M <= 1024`), like the SIMD leaf digest.
+//! Scope: any message up to one 1024-byte chunk, including sub-block and partial-block leaves.
 
 use std::{array, mem::MaybeUninit};
 
@@ -101,6 +101,24 @@ fn permute<const N: usize>(m: &mut [[u32; N]; 16]) {
 	*m = permuted;
 }
 
+/// Loads one 64-byte block per lane into 16 little-endian message words.
+#[inline(always)]
+fn load_block_words<const N: usize>(block: &[[u8; BLOCK_LEN]; N]) -> [[u32; N]; 16] {
+	let mut m = [[0u32; N]; 16];
+	for lane in 0..N {
+		for (w, slot) in m.iter_mut().enumerate() {
+			let off = w * 4;
+			slot[lane] = u32::from_le_bytes([
+				block[lane][off],
+				block[lane][off + 1],
+				block[lane][off + 2],
+				block[lane][off + 3],
+			]);
+		}
+	}
+	m
+}
+
 /// Compresses one 64-byte block across all `N` lanes, updating the chaining value in place.
 ///
 /// The counter, block length, and flags are shared by every lane, so they broadcast.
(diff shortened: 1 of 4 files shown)


crates/hash/src/blake3_portable.rs
```diff
@@ -0,0 +1,432 @@
...
+			_ => unreachable!("a single chunk has at most CHUNK_LEN / BLOCK_LEN = 16 full blocks"),
+		}
+	}};
+}
+
+/// Portable multi-lane Blake3 digest over `N` messages of `M` bytes each.
+///
+/// The compression runs in pure Rust `[u32; N]` lane loops, left for LLVM to auto-vectorize.
+#[derive(Clone)]
+pub struct PortableBlake3MultiDigest<const M: usize, const N: usize> {
```
jimpo: Don't make `M` const len here, not necessary anymore. The buffers should be [[u8; BLOCK_LEN]; N]. And it gets filled up to the block len once per compression call. You can get rid of the dispatch macro now too.

jimpo (commented):
Nice results. Kind of annoying that we can't force vectorization with a smaller lane width, but throughput with 16 lanes is great.

I'm in favor of keeping this approach instead of the Platform SIMD. The main concrete reason is that it supports messages less than block size, which gets exercised in the ZK leaf hashing path.

crates/hash/src/blake3_portable.rs
tcoratger: Fixed, the numbers are a bit better now :) I'm updating the PR description right now.

tcoratger:
> Nice results. Kind of annoying that we can't force vectorization with a smaller lane width, but throughput with 16 lanes is great.
> 
> I'm in favor of keeping this approach instead of the Platform SIMD. The main concrete reason is that it supports messages less than block size, which gets exercised in the ZK leaf hashing path.

Great, just fixed the comment, let me know if we need to do anything before merging this one :)

tcoratger:
A note on the benchmark numbers, since the whole-block throughput is a little lower than an earlier version of this branch and I wanted to explain why.

**Short version:** on whole-block leaves this kernel is now on par with the crate's SIMD (roughly 0.9–1.0×) instead of a few percent ahead. That's a deliberate trade for being able to hash *any* leaf size, including ones smaller than a block — which is the case the ZK leaf-hashing path actually hits, and where this kernel is 1.6–1.7× faster.

**Why the whole-block numbers slipped a bit:**

We hash a leaf 64 bytes at a time — a 128-byte leaf is 2 chunks of work, a 1024-byte leaf is 16.

The earlier version baked the leaf length into the type, so the compiler knew the exact number of chunks up front and could lay them out as flat, straight-line code with nothing else around it. This version takes the length as a plain runtime value, so instead the compiler emits a normal loop: do one chunk, check if there's more, repeat.

Same work either way, but the loop version pays a small fixed cost per leaf for the loop itself.

That fixed cost is tiny, and it's the same no matter the leaf size. What changes is how much real work it's spread over:

- a 128-byte leaf is only 2 chunks of work, so the small overhead is a noticeable slice of the total → ~0.94× the crate
- a 1024-byte leaf is 16 chunks, so the same overhead basically disappears → ~1.03× the crate

So the dip is biggest on the smallest whole-block leaves and fades to nothing on the large ones. Nothing algorithmic changed — it's just that a fixed-length loop the compiler can pre-plan is a hair faster than a variable-length one it can't.

I did try to win it back by trimming the per-chunk bookkeeping, and it made no measurable difference — which confirms the cost is the variable-length loop itself, not anything we can tidy up. Getting it back would mean baking the length in again, which is exactly what we dropped here.

**Why it's worth it:** the fixed-length version could only handle leaves that were a whole number of 64-byte blocks; anything shorter fell back to the slow scalar path. The variable-length loop is what lets us take sub-block leaves (16–48 bytes) through the fast path, and there it's 1.6–1.7× ahead. That's the trade: give up a couple of percent on large whole-block leaves to cover every leaf size in one path, including the small ones the ZK path produces.

jimpo (approved):

PR #1661, opened 2026-07-02
perf(circuits): use the swap_bytes gadget in rs256 byte swapping

wstran:
## What

`fixedbytevec_le_to_biguint` converts a little-endian byte vector to a big-endian `BigUint`. It byte-swapped each 64-bit limb by hand: mask out one byte per lane (`band(_, 0xFF)`, 8 AND constraints per limb), shift each byte into place, and combine the lanes with `bor` (one AND constraint and one committed word each).

## Fix

Reuse the existing `binius_circuits::bytes::swap_bytes` gadget, which reverses the byte order Hacker's Delight style: 4 AND constraints per word, fewer committed intermediate words, and the lane combines are already the linear `bxor`.

## Numbers

Measured on the full RS256-2048 circuit (`test_real_rsa_signature_verification_with_message`):

| version | AND constraints | committed words |
| --- | --- | --- |
| main (manual masks + bor) | 175523 | 180115 |
| this PR (swap_bytes) | 174947 | 179539 |

576 fewer AND constraints and 576 fewer committed words per verification. The byte-swap runs once per signature limb and once per modulus limb (32 limbs each for 2048-bit RSA). The zklogin snapshot drops the same amounts.

repro: print `cs.constraint_system().and_constraints.len()` and `cs.constraint_system().value_vec_layout.committed_total_len` after `builder.build()` in the rs256 test.

## Correctness

`swap_bytes` computes the same byte reversal (property-tested in `bytes.rs` against `u64::swap_bytes`). The `rs256` tests pass, including the negative cases (a wrong message and an invalid PKCS#1 prefix are still rejected). All other example snapshots are unchanged.

## Scope

Only `fixedbytevec_le_to_biguint` and the zklogin snapshot it feeds.

### DIFF
--- a/crates/circuits/src/rs256.rs
+++ b/crates/circuits/src/rs256.rs
@@ -1,3 +1,4 @@
+// Copyright 2026 The Binius Developers
 // Copyright 2025 Irreducible Inc.
 use binius_frontend::{CircuitBuilder, Wire, WitnessFiller};
 use num_integer::Integer;
@@ -5,6 +6,7 @@ use num_integer::Integer;
 use super::fixed_byte_vec::ByteVec;
 use crate::{
 	bignum::{BigUint, ModReduce, assert_eq, optimal_mul, optimal_sqr},
+	bytes::swap_bytes,
 	sha256::Sha256,
 };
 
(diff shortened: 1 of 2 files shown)


jimpo (changes requested):
Nice find, but there's an even more efficient `binius_circuits::bytes::swap_bytes` gadget. Use that instead. It does have some AND gates for masking, but it commits fewer words.

wstran:
Done, switched to `swap_bytes`. Measured on the full RS256-2048 circuit it beats the bxor version on both axes: 177153 -> 176577 AND constraints and 181505 -> 180929 committed words vs main. zklogin snapshot re-blessed, all other snapshots unchanged.

jimpo (approved):

jimpo:
Nice! Needs rebase onto main

PR #1662, opened 2026-07-02
Optimize shift key accumulation

latifkasuli:
## Summary
Fuses the lambda-power weighting into `Key::accumulate`'s consecutive-operand scan in the
shift protocol (phase 1 `build_g_parts` hot loop). The new private `accumulate_weighted`
computes the same weighted sum as `accumulate_by_operand(..).map(|(i, acc)| acc *
lambda_powers[i]).sum()` in a single pass, without the lazy iterator state machine and
per-group tuples.

Phase 2 (`build_monster_multilinear`) is intentionally unchanged: it weights each operand
group by `scalars[key.id + operand_index * SHIFT_VARIANT_COUNT * WORD_SIZE_BITS]` rather
than `lambda_powers`, so it still needs `accumulate_by_operand`.

Prover output is unchanged — the fused scan produces identical field elements, so proof
bytes are byte-for-byte the same (confirmed by A/B runs: identical proof sizes,
verification passes).

## Benchmarks (N=4096, LOG_INV_RATE=1, Apple M4 Pro, single-threaded)
Measured on current main (`75dc9506`), which includes the independent-hash benchmark fix
from #1663.

| Primitive | Proof gen | Witness + prove |
|---|---:|---:|
| SHA-256 compressions | 699.25 ms → 551.75 ms (−21.1%) | 740.96 ms → 594.08 ms (−19.8%) |
| BLAKE3 compressions | 556.79 ms → 443.36 ms (−20.4%) | 580.63 ms → 459.53 ms (−20.9%) |
| Keccak-f[1600] permutations | 696.18 ms → 524.17 ms (−24.7%) | 751.39 ms → 586.89 ms (−21.9%) |

Proof size is byte-identical with and without this change (360,928 B for all three
primitives at these settings).

## Test plan
- Rebased onto main (`75dc9506`)
- `cargo test -p binius-prover --lib` (19/19; includes a differential equivalence test
  covering grouped, non-contiguous, and empty ranges)
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-prover --document-private-items --no-deps`
- A/B benchmark at N=4096 on fixed circuits: proof bytes identical, timings above

### DIFF
--- a/crates/prover/src/protocols/shift/key_collection.rs
+++ b/crates/prover/src/protocols/shift/key_collection.rs
@@ -120,6 +120,40 @@ impl Key {
 		})
 	}
 
+	/// Accumulates the same weighted sum as [`Self::accumulate_by_operand`], but fuses the
+	/// `lambda_powers[operand_index]` weighting into the consecutive-operand scan.
+	#[inline]
+	fn accumulate_weighted<F: Field>(
+		&self,
+		constraint_indices: &[ConstraintIndex],
+		operator_data: &PreparedOperatorData<F>,
+	) -> F {
+		let Range { start, end } = self.range;
+		let mut constraint_indices = constraint_indices[start as usize..end as usize].iter();
+
+		let Some(first) = constraint_indices.next() else {
+			return F::ZERO;
+		};
+
+		let mut operand_index = first.operand_index as usize;
+		let mut acc = F::ZERO;
+		let mut result = F::ZERO;
+		let tensor = operator_data.r_x_prime_tensor.as_ref();
+		acc += tensor[first.constraint_index as usize];
+
+		for current in constraint_indices {
+			let current_operand_index = current.operand_index as usize;
+			if current_operand_index != operand_index {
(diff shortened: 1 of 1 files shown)


crates/prover/src/protocols/shift/key_collection.rs
```diff
@@ -130,9 +164,7 @@ impl Key {
 		constraint_indices: &[ConstraintIndex],
 		operator_data: &PreparedOperatorData<F>,
 	) -> F {
-		self.accumulate_by_operand(constraint_indices, operator_data)
-			.map(|(operand_index, acc)| acc * operator_data.lambda_powers[operand_index])
-			.sum()
+		self.accumulate_weighted(constraint_indices, operator_data)
```
jimpo, line 167: accumulate_weighted is redundant -- just move the code to this function body

jimpo (approved):
Nice! I'm disappointed this didn't inline and compile to the same thing, but I guess it's quite complicated.

PR #1664, opened 2026-07-02
BINIUS-195: replace AES strategy wrappers with a PackedAesArithmetic trait

tcoratger:
@jimpo Just a proposal to limit a bit the macro stuffs, looks like I obtained as well performance improvements locally, feel free to close if you don't like it :) 

Closes [BINIUS-195](https://linear.app/jimpo/issue/BINIUS-195).

Replaces the AES packed-field strategy wrappers and their token-tree macro wiring with one underlier-keyed trait plus blanket operator impls. Pure refactor — outputs are bit-identical, no protocol or behaviour change.

### The problem

The AES packings (`PackedAESBinaryField{1x,16x,32x,64x}8b`) get their arithmetic through a stack of indirection.

Each of `Mul`/`Square`/`InvertOrZero`/`WideMul` is wired by a token-tree macro (`define_packed_binary_field!`, `impl_strategy!`, `impl_mul_with!`).
Each macro forwards to a `TransparentWrapper` "strategy" newtype (`MulFromWideMul`, `Divide`, `ReuseMultiply`, `Gfni`, `VmullWideMul`).
Following one multiply means expanding four macro layers plus a wrap/peel dance.

The strategy is passed as raw token-trees only because a matched `:ty` fragment cannot have `<$name>` appended — the macro contorts around a macro limitation.
IDE "go to definition" and compiler errors both die inside the expansion.

### The idea

Key the arithmetic on the underlier with one ordinary trait, and forward the operators to it with blanket impls.

```
trait PackedAesArithmetic: UnderlierType {
    type WideProduct;
    fn wide_mul(a, b) -> WideProduct;
    fn reduce(w) -> Self;
    fn square(a) -> Self;
    fn invert_or_zero(a) -> Self;
    fn mul(a, b) -> Self { reduce(wide_mul(a, b)) }   // default
}
```

`PackedPrimitiveType` is a local type, so `impl<U: PackedAesArithmetic> Mul for PackedPrimitiveType<U, AESTowerField8b>` is coherent.
Four such blanket impls replace the macro-generated ones, and each backend implements the trait once — no wrappers, no token-trees.

### The change

```diff
-define_packed_binary_field!(
-    PackedAESBinaryField16x8b, AESTowerField8b, M128,
-    (MulFromWideMul), (AesSquare16x), (AesInvert16x), (AesWideMul16x)
-);
+pub type PackedAESBinaryField16x8b = PackedPrimitiveType<M128, AESTowerField8b>;
+// arithmetic comes from the blanket impls on PackedPrimitiveType<U, AESTowerField8b>
```

One concrete `PackedAesArithmetic` impl per underlier, selected per architecture:

- `u8` (byte tables) and a generic `ScaledUnderlier<U, N>` impl — new `arch/aes_arithmetic.rs`
- NEON `vmull` — `aarch64/packed_aes_128.rs`
- GFNI-or-byte-divide — `x86_64/packed_aes_{128,256,512}.rs`
- portable byte-divide — `portable/packed_aes_128.rs`

### Soundness

Pure refactor.
The functions called at the leaves are unchanged (`aes_mul_8b`, the NEON/GFNI intrinsics, the byte tables).
Only the NEON `vmull` path defers a real reduction; every other backend already reduces, so its `WideProduct` is the packing and `reduce` is a no-op — the same behaviour the wrappers encoded.
Outputs are bit-identical, pinned by the existing `PackedAESBinaryField*` multiply/square/invert proptests.

### Scope

- **new:** `arch/aes_arithmetic.rs` (trait, blanket impls, `u8` + `ScaledUnderlier` impls, byte-divide helpers, tables)
- **changed:** the four arch `packed_aes_128/256/512` files, `packed_aes.rs`, `gfni_arithmetics.rs` (wrappers → free fns), `arch/mod.rs` + module lists
- **deleted:** `portable/{packed_aes_8, packed_aes_256, packed_aes_512, pairwise_table_arithmetic, reuse_multiply_arithmetic}.rs`, and the `Gfni`/`GfniWideMul`/`VmullWideMul` wrappers
- **left untouched:** `Divide`/`Scaled`/`MulFromWideMul` — GHASH still uses them; converting GHASH is the follow-up

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-field --all-targets`, nightly `fmt` — clean
- `cargo test -p binius-field` — 193 pass (AES mul/square/invert proptests cover the `u8`, NEON-128, and `ScaledUnderlier`-256/512 paths)
- x86_64 cross-`check` clean in four feature sets: baseline sse2, gfni+avx2+avx512f, gfni-only, avx2-without-gfni (x86 was compile-verified, not run, on this aarch64 host — CI should run its test suite)

### Benchmark

Before vs after on aarch64 (`git stash` A/B), ns per op:

| op | 16x8b | 32x8b | 64x8b |
|---|---|---|---|
| multiply | 55.2 → 55.5 (=) | 141 → 142 (=) | 287 → 266 (−8%) |
| square | 25.1 → 25.9 (=) | 52.6 → 52.5 (=) | 204 → 111 (−45%) |
| invert | 77.4 → 77.5 (=) | 266 → 176 (−34%) | 379 → 347 (−9%) |

Single-register 128b and scalar 8b are unchanged (identical leaf code); the 256/512b packings are faster because the direct `ScaledUnderlier` impl inlines better than the old nested wrapper chain.


jimpo:
Sorry, I think the arithmetic strategy approach is better. The bitvector types are shared across target features, but depending what target features are available changes what the optimal arithmetic strategy is.

If stuff isn't getting inlined properly through the strategy recursion, we should fix that directly.

PR #1668, opened 2026-07-02
BINIUS-198: remove redundant c_root from intmul Witness

benji-potamus:
Closes [BINIUS-198](https://linear.app/jimpo/issue/BINIUS-198/remove-c-root-from-intmul-witness).

The intmul `Witness` stored a `c_root` field (the full product `c` root, computed as `c_lo_root * c_hi_root`). It was only used to open the protocol as the MLE root — but that value equals `b_root`, the root of the variable-base `b`-exponent tree (the protocol proves exactly `(G^a)^b = G^{c_lo} · (G^{2^{2^m}})^{c_hi}`, so the two roots coincide). So `c_root` is redundant.

This removes the field and its `buffer_bivariate_product(c_lo_root, c_hi_root)` computation in `Witness::new`, and switches `prove` (and the phase benchmarks) to use `b_root` — which was already destructured-and-discarded (`b_root: _`) — as the MLE root. Transcript-identical: `exp_eval` is computed from a buffer equal to the old `c_root`.

The `check_consistency` test previously asserted `b_root == c_root` by reading the field; it now recomputes `c_root` from `c_lo_root`/`c_hi_root` and asserts the same equality, so the invariant that justifies the removal is still checked.

Verified: `cargo test -p binius-core -p binius-prover -p binius-verifier` (incl. `prove_and_verify`), `cargo build --workspace --tests --benches`, clippy `-D warnings`, nightly fmt, and rust-doc all green.

### DIFF
--- a/crates/prover/src/protocols/intmul/prove.rs
+++ b/crates/prover/src/protocols/intmul/prove.rs
@@ -23,7 +23,7 @@ use binius_math::{
 	inner_product::inner_product_buffers,
 	multilinear::{
 		eq::{eq_ind_partial_eval, eq_ind_partial_eval_scalars},
-		evaluate::evaluate,
+		evaluate::{evaluate, evaluate_inplace},
 	},
 };
 use binius_utils::{checked_arithmetics::log2_ceil_usize, rayon::prelude::*};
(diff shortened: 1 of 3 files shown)


crates/prover/src/protocols/intmul/prove.rs
```diff
@@ -97,21 +97,22 @@ where
...
-		let n_vars = c_root.log_len();
+		// `b_root` (the variable-base `b`-exponent tree root) equals the full product `c` root, so
+		// it serves as the MLE root that opens the protocol.
+		let n_vars = b_root.log_len();
 		assert!(log_bits >= 1);
 
 		let initial_eval_point = self.channel.sample_many(n_vars);
 
-		let exp_eval = evaluate(&c_root, &initial_eval_point);
+		let exp_eval = evaluate(&b_root, &initial_eval_point);
```
jimpo: Do a consuming in-place evaluation.

jimpo (changes requested):

crates/prover/src/protocols/intmul/prove.rs
benji-potamus: Done (pushed `aacea49`). Switched to `evaluate_inplace(b_root, &initial_eval_point)`, which folds `b_root` in place and consumes it — `b_root` isn't used after this point. `prove_and_verify` still passes (transcript-identical).

jimpo (approved):

PR #1669, opened 2026-07-02
BINIUS-200: commit the logUp* pushforward via an IOP channel

tcoratger:
Closes [BINIUS-200](https://linear.app/jimpo/issue/BINIUS-200).

Makes the logUp* protocol commit its pushforward `Y` and prove it opens correctly, instead of returning an unbound evaluation claim. The reduction math is unchanged — this only adds the commitment and its opening.

### The problem

logUp* proves an indexed lookup by committing a small pushforward `Y = I_* eq_r` instead of the full looked-up vector.
The reduction ends with a claim: "`Y` at point `r'` equals `v`".
But `Y` was never committed — nothing bound it.
So a cheating prover could invent a fake `Y` that fits a false lookup and just assert its evaluation.
The verifier had no fingerprint of `Y` to check against, so the claim floated unanchored.

### The idea

Commit `Y` before the challenge is drawn, then open that commitment at the reduced point.
- Committing first (Fiat-Shamir) locks `Y` in, so it cannot be chosen after seeing the challenge.
- Opening at the end proves the claimed value is the real committed `Y`'s value.

Sending a commitment needs the IOP channel, which carries oracle commitments.
The bare IP channel only carries field elements, so it cannot express this.

### A concrete example

Table `T = [5, 9, 2, 7]`, indices `[2, 0, 2, 3]`, weights `[3, 1, 4, 2]`.

```
lookup   e = 3*T[2] + 1*T[0] + 4*T[2] + 2*T[3] = 33
scatter  Y = [1, 0, 7, 2]              (each weight added onto the slot its row hits)
duality  <T, Y> = 5*1 + 2*7 + 7*2 = 33 = e
```

The proof runs on the 4-entry `Y`, not the looked-up list — and in the real regime the table is tiny while the index list is huge, so `Y` is the whole saving.
Before this change the "`Y(r') = v`" claim was unanchored; now `Y` is committed and opened.

### The change

The crate layering is `ip` (commitment-free reductions) then `iop` (adds oracle commitments) on top.
So the reduction stays low and a thin committing wrapper sits above it.

- **`ip-prover`**: split the reduction body into `prove_reduction`, which takes a pre-built `eq_r` and `Y`. The self-contained `prove` builds both and delegates, so the committing layer builds `Y` once, commits it, and reuses it with no second scatter-add.
- **`iop-prover`**: `logup_star::prove` builds `eq_r` and `Y`, commits `Y` before the reduction, runs the reduction, and returns the table and index claims plus the pushforward oracle relation `<Y, eq_r'> = Y(r')`.
- **`iop`**: `logup_star::verify` receives the `Y` commitment and returns the matching relation.

`Y` is the only oracle this protocol introduces (Soukhanov25, Section 3).
The table `T` and index `I` stay the caller's oracles, so their claims are returned for the caller to open.
The caller batches all the openings into one channel open.

### Soundness

- The reduction transcript is unchanged; only the `Y` commitment is added, and it precedes the logUp challenge.
- The opening binds the claimed `Y(r')` to the committed `Y`, closing the gap where the claim was unchecked.

### Scope

- **new:** `crates/iop-prover/src/logup_star.rs`, `crates/iop/src/logup_star.rs`
- **changed:** `crates/ip-prover/src/logup_star/{prove.rs, mod.rs}` — extract `prove_reduction`, expose the witness builders
- **untouched:** the reduction transcript and the bare `binius-ip` verifier

### Verification

- `cargo check --workspace --all-targets`, `cargo clippy -p binius-ip-prover -p binius-iop -p binius-iop-prover --all-targets -- -D warnings`, nightly `cargo fmt --all` — all clean
- `RUSTDOCFLAGS=-D warnings cargo doc --no-deps` — intra-doc links resolve
- round-trip tests over the naive channel (shapes with $m \ll n$, $m = n$, $m = 1$) and over the real BaseFold/FRI channel, which commits `Y` and verifies its opening end-to-end
- a negative test: a wrong evaluation claim is rejected

### DIFF
--- a/crates/iop-prover/src/lib.rs
+++ b/crates/iop-prover/src/lib.rs
@@ -30,5 +30,6 @@ pub mod basefold_channel;
 pub mod basefold_compiler;
 pub mod channel;
 pub mod fri;
+pub mod logup_star;
 pub mod merkle_tree;
 pub mod naive_channel;
--- a/crates/iop-prover/src/logup_star.rs
+++ b/crates/iop-prover/src/logup_star.rs
@@ -1,0 +1,402 @@
+// Copyright 2026 The Binius Developers
+
+//! logUp* proving with the pushforward committed as an oracle.
+//!
+//! The bare reduction returns a claimed evaluation of the pushforward `Y = I_* eq_r`.
+//! It commits nothing, so that claim cannot be checked on its own.
+//! This layer commits `Y` over the IOP channel and returns the relation that opens it.
+//!
+//! The commit precedes the reduction, so the logUp challenge binds the committed `Y`.
+//! The table `T` and index `I` stay the caller's oracles, so their claims are returned unchanged.
+//! `Y` is the only oracle this protocol introduces, per [Soukhanov25, Section 3].
+//!
+//! [Soukhanov25]: <https://eprint.iacr.org/2025/946>
+
+use binius_field::{BinaryField1b, ExtensionField, Field, PackedField};
+use binius_ip_prover::logup_star::{self as reduction, witness};
+use binius_math::{FieldBuffer, multilinear::eq::eq_ind_partial_eval};
+
+use crate::channel::IOPProverChannel;
+
+/// An error raised while proving a committed logUp* reduction.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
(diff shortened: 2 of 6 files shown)


jimpo (approved):
Nice, very clean!

PR #1670, opened 2026-07-02
BINIUS-201: unify the round-state enums in the sumcheck provers

tcoratger:
Closes [BINIUS-201](https://linear.app/jimpo/issue/BINIUS-201).

Collapses five copy-pasted per-prover state enums into one shared type. Pure refactor — no behavior change.

### The problem

Every sumcheck-style prover in `crates/ip-prover/src/sumcheck` is the same two-phase state machine.
After it emits a round polynomial, it waits for a challenge to fold.
After it folds, it holds the running claim for the next round.

Eight provers each declared their **own private enum** for that one idea.
Five names for the same two states: `RoundCoeffsOrSum`, `RoundCoeffsOrEval`, `RoundCoeffsOrSums`, `RoundCoeffsOrEvals<M>`, `RoundCoeffsOrClaim`.

The call-order contract was copy-pasted too:
- the execute-entry guard (`... else return ExpectedFold`),
- the fold-entry guard (`... else return ExpectedExecute`),
- the finish-entry guard that maps the current phase to the right error.

That is ~7 hand-written copies of the same match, one edit away from disagreeing on the protocol contract.

### The idea

One generic enum, parameterised by the two containers so each prover keeps its exact shape and allocation.

```rust
pub enum RoundState<Coeffs, Claim> {
    Coeffs(Coeffs),  // round polynomial(s) produced, awaiting the fold
    Claim(Claim),    // claim(s) awaiting the next round polynomial
}
```

- single-claim prover uses `RoundState<RoundCoeffs<F>, F>`
- dynamic multi-claim prover uses `RoundState<Vec<RoundCoeffs<F>>, Vec<F>>`
- fixed-count prover uses `RoundState<[RoundCoeffs<F>; M], [F; M]>`

Three small `const fn` accessors put the contract in one place:
- one returns the carried claim, or `ExpectedFold`
- one returns this round's polynomial(s), or `ExpectedExecute`
- one returns the right error when the protocol is told to finish too early

### The change

```
- let RoundCoeffsOrSum::Sum(last_sum) = &self.state else { return Err(Error::ExpectedFold) };
+ let last_sum = self.state.claim()?;
```

And every hand-written `finish` match collapses to `self.state.unfinished_err()`.

### Soundness

Pure refactor — no protocol, transcript, or arithmetic change.
The state transitions and the error each guard returns are identical to before; only their definition is now shared.

### Scope

- **new:** `sumcheck/round_state.rs` — the shared enum and its `const fn` accessors.
- **changed:** 8 provers migrated; the five private enums deleted.
- **untouched:** all round-polynomial math, folding, and the transcript.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-ip-prover --all-targets`, nightly `fmt` — clean
- `binius-ip-prover` — 50 tests pass (incl. the batch-vs-single and multi-vs-single conformance tests)
- direct consumers `binius-prover` and `binius-iop-prover` — tests pass


jimpo (commented):
Seems reasonable, needs rebase onto main

tcoratger:
> Seems reasonable, needs rebase onto main

@jimpo Just fixed the conflicts

jimpo (approved):

PR #1671, opened 2026-07-02
BINIUS-203: collapse the single-claim quadratic MLE-check prover onto the batch prover

tcoratger:
Closes [BINIUS-203](https://linear.app/jimpo/issue/BINIUS-203).

> **Stacked on #1670** (BINIUS-201). Until that merges, the diff here also shows its commit — review the top commit (`BINIUS-203`) only, and merge after #1670.

Makes the single-claim quadratic MLE-check prover a thin wrapper over the batch prover, deleting the duplicated round loop. Pure refactor — no behavior change.

### The problem

Two provers implement the same degree-2 MLE-check algorithm.
`QuadraticMleCheckProver` proves one composition; `BatchQuadraticMleCheckProver` proves `M` at once.

The batch prover's per-round loop is chunked, so the equality-indicator chunk stays in L1, and it accumulates in unreduced (wide) form.
The single-claim prover carried its own copy of that loop, and it was the worse copy:
- it was not chunked,
- its hot loop did bounds-checked random indexing into a `Vec` of split halves, one index per input per row.

So there were two round loops for one algorithm, and the hot single-claim path was the slower one.

### The idea

The single-claim prover is just the batch prover with `M = 1`.

A small trait captures "evaluate the `M` compositions on one packed row":

```rust
pub trait QuadraticComposition<P, const N: usize, const M: usize> {
    fn evaluate(&self, inputs: [P; N], outputs: &mut [P; M]);
}
```

- Every `Fn([P; N], &mut [P; M])` closure implements it via a blanket impl, so existing batch call sites are unaffected.
- A single-output `Fn([P; N]) -> P` is adapted to `M = 1` by a wrapper that writes its one value into the length-1 output.

`BatchQuadraticMleCheckProver` switches its bound to this trait.
`QuadraticMleCheckProver` becomes a newtype over `BatchQuadraticMleCheckProver<..., 1>` that adapts the composition and delegates.

### The change

```
- for j in 0..N {
-     evals_1[j]   = splits_1[j].as_ref()[i];        // bounds-checked, per input per row
-     evals_inf[j] = splits_0[j].as_ref()[i] + splits_1[j].as_ref()[i];
- }
- // ...its own non-chunked round loop...
+ // QuadraticMleCheckProver::new wraps each Fn([P; N]) -> P and defers to the batch prover,
+ // which runs the chunked, wide-accumulated loop.
```

### Soundness

Pure refactor — no protocol, transcript, or arithmetic change.
The chunked loop only reorders a sum of field elements, and field addition is commutative, so the round polynomials are bit-identical.
The existing conformance test (a batch of two vs. two single-claim provers) passes unchanged and guards this.

### Scope

- **new:** the `QuadraticComposition` trait, its blanket impl, and a single-output adapter.
- **removed:** the single-claim prover's hand-rolled round loop and state machine (~150 lines).
- **unchanged:** the public type `QuadraticMleCheckProver` and its `new` — every call site (`spartan-prover`, `and_reduction`, `intmul`, benches) is untouched.
- the fractional-addition prover annotates its closure arguments, since the blanket impl does not back-propagate the expected closure signature to untyped parameters.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-ip-prover -p binius-prover -p binius-spartan-prover --all-targets`, nightly `fmt` — clean
- `binius-ip-prover` — 50 tests pass, incl. `test_batch_quadratic_mlecheck_conforms_to_single_quadratic_mlecheck`
- `binius-prover` and `binius-spartan-prover` — tests pass

### DIFF
--- a/crates/ip-prover/src/sumcheck/batch_quadratic_mle.rs
+++ b/crates/ip-prover/src/sumcheck/batch_quadratic_mle.rs
@@ -18,6 +18,26 @@ use crate::sumcheck::{
 	round_state::RoundState,
 };
 
+/// Produces the composite value of each of the `M` claims for one packed row of `N` inputs.
+///
+/// Evaluating all claims together lets those that share sub-expressions reuse the work.
+/// A closure of type `Fn([P; N], &mut [P; M])` implements this directly.
+pub trait QuadraticComposition<P: PackedField, const N: usize, const M: usize> {
+	/// Writes the `M` composite values for the given inputs into `outputs`.
+	fn evaluate(&self, inputs: [P; N], outputs: &mut [P; M]);
+}
+
+impl<P, T, const N: usize, const M: usize> QuadraticComposition<P, N, M> for T
+where
+	P: PackedField,
+	T: Fn([P; N], &mut [P; M]),
+{
+	fn evaluate(&self, inputs: [P; N], outputs: &mut [P; M]) {
+		// A closure that already has the right shape is its own evaluator.
+		self(inputs, outputs)
+	}
+}
+
 /// Batch MLE-check prover for M quadratic compositions over N multilinears.
 ///
 /// This prover runs a single sumcheck instance that amortizes the work of M independent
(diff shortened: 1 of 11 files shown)


jimpo:
Let's hold off on this, I'm hoping for a better solution via https://linear.app/jimpo/issue/BINIUS-202/. I kind of hate the Batch sumcheck provers.

PR #1672, opened 2026-07-02
BINIUS-184: bit-twiddling SIMD Rijndael multiply (SSE2/AVX2/AVX-512)

benji-potamus:
Closes [BINIUS-184](https://linear.app/jimpo/issue/BINIUS-184/arith-bench-benchmark-simd-rijndael-mult-algorithm).

Adds a **non-GFNI** software path for packed Rijndael GF(2⁸) multiplication to `binius-arith-bench` — the x86_64 counterpart to the existing aarch64 `vmull::mul`, for CPUs without GFNI (and, with the same kernel, Wasm) — with a criterion bench across SSE2 / AVX2 / AVX-512BW register widths.

## Algorithm

The per-byte kernel is the branch-free, constant-time Russian-peasant multiply from Vault's Shamir secret-sharing implementation:

```text
r = 0
for i in (0..8).rev():
    r = (-(b >> i & 1) & a) ^ (-(r >> 7) & 0x1b) ^ (r + r)
```

Every op is lane-parallel, so it vectorizes directly. The lane primitives it needs — 8-bit add, a per-byte "sign mask", and a 16-bit lane shift — are abstracted behind a small **`RijndaelBytewise`** trait (`src/rijndael/russian_peasant.rs`), so one generic `mul<U: RijndaelBytewise>` runs over `__m128i`, `__m256i`, and `__m512i`. The two's-complement masks `-(x&1)` become signed byte compares on SSE/AVX and AVX-512 mask registers (`_mm512_movm_epi8(_mm512_movepi8_mask(·))`) on 512; `r+r` is an 8-bit add; bit-`i` selection is a 16-bit lane shift into the sign bit.

Adding the AVX-512 width required a minimal `__m512i` `Underlier` + `PackedUnderlier<u8>` impl in `src/arch/x86_64.rs` (it wasn't an underlier before).

## Benchmark

x86_64, `-C target-cpu=native`, AVX-512 host **without GFNI** (so `gfni::mul` is unavailable — a realistic no-GFNI stand-in):

| bench | throughput |
|---|---|
| `rijndael/russian_peasant::mul::<__m128i>` | ≈ 1.65 Gelem/s |
| `rijndael/russian_peasant::mul::<__m256i>` | ≈ 3.18 Gelem/s |
| `rijndael/russian_peasant::mul::<__m512i>` | ≈ 4.10 Gelem/s |

AVX2 scales ~1.9× over SSE; AVX-512 adds ~1.3× more (sub-linear, as expected from AVX-512 downclocking + the mask-register sign path). GFNI's hardware `gf2p8mul` is a single instruction (typically another ~5–10×), so this quantifies the software fallback's cost.

## Correctness

`RijndaelBytewise` proptests pack random lanes, run the SIMD `mul` per width, and compare every lane against an independent **LSB-first** scalar oracle (a different algorithm than the MSB-first SIMD kernel). All three widths pass.

## Validation

- Proptest (128/256/512) green; `clippy -D warnings --all-targets` + nightly `fmt --check` + `typos` clean.
- Cross-arch: `wasm32-unknown-unknown` build and `aarch64-unknown-linux-gnu` check both green (the new code is x86_64-gated; AVX-512 paths gated `avx512bw`, 128b gated `sse4.1` to match `PackedUnderlier<u8>`).

### DIFF
--- a/crates/arith-bench/src/rijndael/mod.rs
+++ b/crates/arith-bench/src/rijndael/mod.rs
@@ -9,5 +9,6 @@
 //! transformation (S-Box) to provide non-linearity in the cipher.
 
 pub mod gfni;
+pub mod russian_peasant;
 #[cfg(target_arch = "aarch64")]
 pub mod vmull;
--- a/crates/arith-bench/src/rijndael/russian_peasant.rs
+++ b/crates/arith-bench/src/rijndael/russian_peasant.rs
@@ -1,0 +1,193 @@
+// Copyright 2026 The Binius Developers
+//! GF(2^8) Rijndael field multiplication via a branch-free bit-twiddling algorithm, vectorized
+//! per-byte across a SIMD underlier.
+//!
+//! Unlike [`gfni::mul`](super::gfni::mul), which lowers to a single hardware `gf2p8mul`
+//! instruction, this path uses no GFNI — making it a candidate software fallback for x86_64 CPUs
+//! without GFNI (and, with the same idea, for Wasm). The per-byte kernel is the constant-time
+//! Russian-peasant multiply from Vault's Shamir secret-sharing implementation:
+//!
+//! ```text
+//! r = 0
+//! for i in (0..8).rev():
+//!     r = (-(b >> i & 1) & a) ^ (-(r >> 7) & 0x1b) ^ (r + r)
+//! ```
+//!
+//! Every operation (byte doubling, sign-bit mask, AND, XOR) is lane-parallel, so the scalar
+//! kernel vectorizes directly. The lane primitives it needs — 8-bit add, a per-byte "sign mask",
+//! and a 16-bit lane shift — are abstracted behind [`RijndaelBytewise`] so the same [`mul`] runs
+//! over `__m128i` (SSE2), `__m256i` (AVX2), and `__m512i` (AVX-512BW). The two's-complement masks
+//! `-(x & 1)` become signed byte compares (or AVX-512 mask registers), `r + r` becomes an 8-bit
+//! add, and testing bit `i` of each byte is a left shift that moves it into the sign bit.
+
+use seq_macro::seq;
+
+use crate::underlier::{PackedUnderlier, Underlier};
+
+/// Byte-lane primitives for the Russian-peasant GF(2^8) multiply, specialized per SIMD underlier.
+pub trait RijndaelBytewise: Underlier + PackedUnderlier<u8> {
+	/// Lane-wise wrapping 8-bit addition. Adding a value to itself is a byte-lane `<< 1` (with the
+	/// carry out of bit 7 dropped — the reduction step accounts for it).
+	fn add_epi8(a: Self, b: Self) -> Self;
+
+	/// Per-byte mask: `0xff` in lanes whose bit 7 is set, `0x00` otherwise.
+	fn sign_mask_epi8(a: Self) -> Self;
+
+	/// Logical left shift of each 16-bit lane by `count` (a small runtime amount, 0..8). Used to
+	/// move bit `i` of each byte into the byte's sign bit so
+	/// [`sign_mask_epi8`](Self::sign_mask_epi8) can broadcast it. A runtime count keeps one
+	/// primitive across widths, since the AVX-512 immediate-shift intrinsic uses a different
+	/// const-generic type than the SSE/AVX ones.
+	fn sll_epi16(a: Self, count: u32) -> Self;
+}
+
+/// Multiply packed GF(2^8) Rijndael elements (the byte lanes of `U`) using the branch-free
+/// Russian-peasant algorithm, with the AES reduction polynomial x^8 + x^4 + x^3 + x + 1.
+#[inline]
+pub fn mul<U: RijndaelBytewise>(a: U, b: U) -> U {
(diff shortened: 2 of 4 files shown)


jimpo:
Rename to russian_peasant. Implement for AVX2/AVX512 via a specialized underlier trait.

benji-potamus:
Done, pushed `17f23b5c` (marked ready for review):
1. **Renamed** `vault` → `russian_peasant`.
2. **AVX2 + AVX-512** implemented, generic over a new `RijndaelBytewise` trait carrying the three byte primitives (8-bit add, per-byte sign mask, 16-bit lane shift) — one `mul<U>` runs over `__m128i`/`__m256i`/`__m512i`. AVX-512's sign mask uses the opmask path (`_mm512_movm_epi8(_mm512_movepi8_mask(·))`). Adding the 512 width needed a minimal `__m512i` `Underlier` + `PackedUnderlier<u8>` in `arch/x86_64.rs`.

Throughput (no-GFNI AVX-512 host): 128b ≈ 1.65, 256b ≈ 3.18, 512b ≈ 4.10 Gelem/s. Proptest correct for all three widths; clippy/fmt/typos/wasm/aarch64 green.

One impl note: the lane shift is a *runtime* count (`_mm{,256,512}_sll_epi16`) rather than an immediate, because AVX-512's `_mm512_slli_epi16` uses a `const IMM8: u32` while the SSE/AVX ones use `i32` — bridging that in one trait method needs `generic_const_exprs`. Immediate shifts would need a per-width const-generic method; happy to switch if you'd prefer.

benji-potamus:
Fixed a CI failure on the `rust-checks-x86_64-portable` leg (pushed `4e122d55`): the `mul_m128i` proptest was gated `sse2` but the `__m128i` impl (like `PackedUnderlier<u8>`) is gated `sse4.1`, so on the baseline-x86_64 leg the test referenced a missing impl. Re-gated the test to `sse4.1`. Reproduced the portable leg locally (`RUSTFLAGS="" cargo test --all-targets --no-run`) — now green, and native tests/clippy/fmt/typos still pass.

crates/arith-bench/src/rijndael/russian_peasant.rs
```diff
@@ -0,0 +1,195 @@
...
+	/// [`sign_mask_epi8`](Self::sign_mask_epi8) can broadcast it. A runtime count keeps one
+	/// primitive across widths, since the AVX-512 immediate-shift intrinsic uses a different
+	/// const-generic type than the SSE/AVX ones.
+	fn sll_epi16(a: Self, count: u32) -> Self;
+}
+
+/// Multiply packed GF(2^8) Rijndael elements (the byte lanes of `U`) using the branch-free
+/// Russian-peasant algorithm, with the AES reduction polynomial x^8 + x^4 + x^3 + x + 1.
+#[inline]
+#[cfg(target_arch = "x86_64")]
```
jimpo: Does this need the cfg gate?

crates/arith-bench/src/rijndael/russian_peasant.rs
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	#[inline]
+	fn sll_epi16(a: Self, count: u32) -> Self {
+		use std::arch::x86_64::{_mm_cvtsi32_si128, _mm512_sll_epi16};
+		unsafe { _mm512_sll_epi16(a, _mm_cvtsi32_si128(count as i32)) }
+	}
+}
+
+#[cfg(all(test, target_arch = "x86_64"))]
```
jimpo: get rid of the target_arch cfg gate. Only on the functions that need it.

jimpo (changes requested):

crates/arith-bench/src/rijndael/mod.rs
```diff
@@ -9,5 +9,7 @@
 //! transformation (S-Box) to provide non-linearity in the cipher.
 
 pub mod gfni;
+#[cfg(target_arch = "x86_64")]
```
jimpo: Why is this necessary?

crates/arith-bench/src/rijndael/russian_peasant.rs
benji-potamus: No — removed. The generic `mul` and the `RijndaelBytewise` trait are arch-independent (only the impls touch `std::arch::x86_64`), so nothing here needs the gate. Fixed in `74a3387d`.

crates/arith-bench/src/rijndael/russian_peasant.rs
benji-potamus: Done — `mod tests` is now just `#[cfg(test)]`. The scalar oracle and the generic `check` are arch-independent, and each proptest is gated on its x86_64-only `target_feature` (sse4.1 / avx2 / avx512bw), so they're simply inactive off x86_64. Verified it still compiles on aarch64 + wasm.

crates/arith-bench/src/rijndael/mod.rs
benji-potamus: It wasn't — removed. `russian_peasant` is now an unconditional module like `gfni`; the only arch-specific parts are the trait impls, which self-gate on `target_arch`/`target_feature`. `74a3387d`.

jimpo (approved):

PR #1675, opened 2026-07-03
BINIUS-126: optimize SHA-256 circuit gadget with SWAR

tcoratger:
Closes [BINIUS-126](https://linear.app/jimpo/issue/BINIUS-126/optimize-sha2-circuit-gadget-with-swar).

Runs SHA-256's block chain two blocks at a time, packed into the two 32-bit lanes of one 64-bit word. Output is unchanged; every consumer gets ~2x fewer AND constraints for free.

### The problem

In binius64 **every gate costs one AND constraint regardless of word width**.
`iadd_32`, `rotr32`, `srl32` operate on *both* 32-bit halves of a 64-bit word for the price of one.

SHA-256's word is 32 bits, but the gadget chained **single-lane** `sha256_compress` — one block per compression, high 32 bits of every word left empty.
So it did half the useful work per gate and came out **more expensive than SHA-512**, which fills all 64 bits:

| bytes | SHA-256 (before) | SHA-512 |
|---|---|---|
| 1024 | 16,938 | 11,946 |

A 32-bit hash costing more than a 64-bit hash is backwards.

### The idea

Pack two SHA-256 lanes into each 64-bit word and run both through one compression core (SWAR).
Within a single message the blocks chain — block $b{+}1$'s input state is block $b$'s output — so the two lanes are not independent.

Break the dependency with a hint, exactly as `blake3_compress_2x_seq` does:

```
high lane [32:64]:  S1 = compress(input state, block b)
low  lane [0:32] :  S2 = compress(S1,          block b+1)
```

A `Sha256CompressHint` precomputes `S1` off-circuit and seeds the low lane's input.
`S1` is then constrained two ways, so the hint cannot lie:

- its high half must equal the real input state (the first compression's input);
- its low half must equal the first compression's in-circuit output.

### Per block pair

- **before:** 2 single-lane compressions
- **after:** 1 two-lane compression + a hint bound by a handful of asserts

→ the compression cost of a message roughly halves.

### The change

```
Sha256::new / sha256_fixed:
- for each block: state = sha256_compress(state, block)          // one block per core
+ for each pair : out   = sha256_compress_2x_seq(state, [b0, b1]) // two blocks per core
+ trailing odd block compressed single-lane
```

- **new:** `sha256_compress_2x_seq` + `Sha256CompressHint` + a module-level `ref_compress`.
- **changed:** the compression chains in `Sha256::new` and `sha256_fixed`.
- **untouched:** `sha256_compress` (single-lane) stays — the trailing-odd-block path and the scalar reference both use it.

This is the SHA-256 analog of the BLAKE3 double-compression track (`blake3_compress_2x_seq`, then its use in `blake3_fixed`).
It is **not** a protocol or padding change — the padding, length field, and digest selection are identical; only the compression chain is repacked.
SHA-512 cannot be SWAR'd this way: its words already fill the full 64-bit lane.

### Soundness

- The hint is untrusted: both halves of its output are re-derived in-circuit and asserted, so a wrong `S1` fails a constraint.
- The result is bit-identical to the single-lane chain — validated against RFC 6234 / NIST known-answer digests (`"abc"`, the two-block 56-byte `TEST2_1`, and lengths 55–512), plus random-size cross-checks against the `sha2` crate.

### Scope

- Public API of `Sha256::new` / `sha256_fixed` unchanged, so zklogin, double-SHA256, rs256, bip32, p2pkh, hmac all benefit transparently.
- Single-block hashes (e.g. p2pkh) are unchanged — there is nothing to pair.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-circuits --all-targets`, nightly `fmt --all` — clean.
- `binius-circuits` full suite (336 tests) and `binius-prover` SHA-256 prove/verify pass; the `sha256` example proves end-to-end (exercises the hint).
- New primitive tests: sequential-fusion vs scalar reference, plus a direct RFC 6234 two-block known-answer test.
- Snapshots re-blessed for `sha256`, `zklogin`, `bitcoin_headers`.

### Benchmark

AND constraints, `<example> stat --max-len-bytes N`:

| bytes | SHA-256 before | SHA-256 after | reduction | SHA-512 |
|---|---|---|---|---|
| 64 | 1,988 | 1,174 | -41% | 1,300 |
| 128 | 2,989 | 2,175 | -27% | 2,586 |
| 256 | 4,986 | 3,356 | -33% | 3,927 |
| 1024 | 16,938 | 10,414 | -38.5% | 11,946 |

SHA-256 is now cheaper than SHA-512 at every size, and the 1024-byte example drops from the $2^{15}$ into the $2^{14}$ AND-constraint domain.
Whole-circuit knock-on: zklogin 430,714 → 421,746, bitcoin header chain 31,780 → 23,640.

### DIFF
--- a/crates/circuits/src/sha256/compress.rs
+++ b/crates/circuits/src/sha256/compress.rs
@@ -1,6 +1,10 @@
 // Copyright 2025 Irreducible Inc.
+use std::{array, iter};
+
 use binius_core::word::Word;
-use binius_frontend::{CircuitBuilder, Wire, WitnessFiller};
+use binius_frontend::{CircuitBuilder, Hint, Wire, WitnessFiller};
+
+use crate::util::clear_high_bits;
 
 const IV: [u32; 8] = [
 	0x6a09e667, 0xbb67ae85, 0x3c6ef372, 0xa54ff53a, 0x510e527f, 0x9b05688c, 0x1f83d9ab, 0x5be0cd19,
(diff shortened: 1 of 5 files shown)


jimpo:
Nice! I wonder how much the difference is for the `sha256_fixed` function (fixed-len payload). I bet it's more.

jimpo (approved):

PR #1676, opened 2026-07-03
Remove the binius-math fold module

tcoratger:
Removes the top-level `fold` module (`fold_cols` / `fold_rows`) from `binius-math`. Pure deletion — no behavior change.

Scoped down from an earlier, broader dead-code sweep: per review, the other unused utilities (`Matrix`, `BatchInversion`, `EvaluationDomain`, and assorted unused methods) are worth keeping even if currently uncalled, so this PR now touches **only** `fold`.

### The problem

The top-level `fold` module is a naive, single-threaded, scalar matrix-vector fold:

- `fold_cols(mat, vec)` — combine columns (a matrix-vector product $Mv$).
- `fold_rows(mat, vec)` — combine rows ($v^\top M^\top$).

It has been superseded by the more specialized linear transformations the provers actually use — `multilinear::fold::fold_highest_var_inplace` in the sumcheck provers, and the FRI codeword fold. Nothing in production calls the generic version anymore.

### The change

- **`fold_rows`** — no callers anywhere in the workspace. Deleted.
- **`fold_cols`** — no production caller. Its one use was as a reference oracle in the iop-prover FRI fold test (`crates/iop-prover/src/fri/fold.rs`, under `#[cfg(test)]`). That test now inlines the equivalent computation directly:

```
// fold each row against the eq tensor of the challenges
let folded_vals: Box<[B128]> = msg
    .chunks(arity)
    .map(|row| inner_product_buffers(&row, &query))
    .collect();
let folded_msg = FieldBuffer::new(log_dim, folded_vals);
```

- Deletes `crates/math/src/fold.rs` and its `pub mod fold;` line.

### Scope

- **removed:** `crates/math/src/fold.rs` (321 lines) + the module declaration.
- **changed:** one iop-prover FRI test, which keeps its correctness check via an inlined reference fold.
- **left untouched:** `multilinear::fold` and every other math utility.

### Verification

- `cargo check --workspace --all-targets --all-features` — clean
- `cargo clippy -p binius-math -p binius-iop-prover --all-targets --all-features` — no warnings
- `cargo test -p binius-math --all-features` — 112 passed, 0 failed (+ 2 doctests)
- `cargo test -p binius-iop-prover fri::fold` — FRI fold proptest passes against the inlined reference


jimpo (commented):
Most of these utilities are good still, even if no longer used. Especially batch_invert and matrix.

fold is the only one that I think we should remove (it's basically replaced by different more specialized linear transformations).

tcoratger:
> Most of these utilities are good still, even if no longer used. Especially batch_invert and matrix.
> 
> fold is the only one that I think we should remove (it's basically replaced by different more specialized linear transformations).

Fixed

jimpo (approved):

PR #1678, opened 2026-07-03
BINIUS-205: rework the intmul phase 4/5 handoff

benji-potamus:
Closes [BINIUS-205](https://linear.app/jimpo/issue/BINIUS-205/rework-the-intmul-phase-4phase-5-handoff).

Reworks the phase 4 → phase 5 handoff of the integer-multiplication protocol so phase 4 stops binding the all-but-last-layer evaluations and instead passes the reduced prodcheck claims straight to phase 5, and phase 5 finishes each tree with a regular prodcheck-layer MLE-check rather than a single `3·2^k`-claim multi-MLE.

## Phase 4
The batched prodcheck already stops one layer short and returns each tree's reduced claim (eval + retained final layer). Phase 4 now just hands those three claims and the reduced content-and-node point straight to phase 5 — it no longer recomputes or sends the `3·2^{k-1}` all-but-last-layer evals, and the verifier no longer receives/binds them.

## Phase 5
- **Before the sumcheck**, the prover sends one reduced eval per tree (`a`, `c_lo`, `c_hi`); the verifier checks they recombine, weighted by `eq(selector)`, to the phase-4 prodcheck output claim.
- The final sumcheck **batches five** provers: the final (widest) GKR layer of each of the three trees — each a regular prodcheck-layer bivariate-product MLE-check (`ProdcheckProver::layer_prover`) seeded by its reduced eval and wrapped in `MleToSumCheck` — plus the overflow parity zerocheck and the `b` re-randomization. The latter two span only the `n_vars` content variables, so they're padded up to the trees' `n_vars + (k-1)` variables via the (previously unused) `PaddedSumcheckDecorator`.
- **After the sumcheck**, the prover sends the raw per-bit output evals (computed directly from the exponents, as `b` already did); the verifier reconstructs the selected leaf values and recombines them via `eq(r_bit)` to bind the three final-layer sumchecks, and applies the `eq(0; r_bit)` padding factor to the overflow / `b` terms.

The protocol output (per-bit evals at a common `n_vars` point) is unchanged, so the downstream shift reduction is unaffected.

## Notes
- Adds `Witness::c_hi_exponents` (already passed to `Witness::new`, now retained) so phase 5 can compute the raw `c_hi` output evals from the exponents.
- Removes the now-dead multi-MLE path and its `split_pairs` / `deinterleave_pairs` / `split_leaf_layer` helpers, plus the orphaned `normalize_a_c_exponent_evals` / `recover_selectors` (the verifier's forward `reconstruct_selecteds` remains). Net −128 lines.

## Validation
- `prove_and_verify` (full 5-phase intmul) and `test_shift_prove_and_verify` (downstream consumer) green; full `binius-prover` + `binius-verifier` suites pass.
- `clippy --all-targets -D warnings`, nightly `fmt --check`, and `typos` clean; portable-leg (`RUSTFLAGS=""`) build green.

### DIFF
--- a/crates/verifier/src/protocols/intmul/verify.rs
+++ b/crates/verifier/src/protocols/intmul/verify.rs
@@ -1,6 +1,6 @@
 // Copyright 2025 Irreducible Inc.
 
-use std::{iter, slice};
+use std::iter;
 
 use binius_field::{BinaryField, Field, field::FieldOps};
 use binius_ip::{
@@ -9,11 +9,13 @@ use binius_ip::{
 	sumcheck::{BatchSumcheckOutput, batch_verify},
 };
 use binius_math::{
-	multilinear::{eq::eq_ind, evaluate::evaluate_inplace_scalars},
+	multilinear::{
+		eq::{eq_ind, eq_ind_zero},
+		evaluate::evaluate_inplace_scalars,
+	},
 	univariate::evaluate_univariate,
 };
 use binius_utils::checked_arithmetics::log2_ceil_usize;
-use itertools::chain;
 
 use super::{
 	common::{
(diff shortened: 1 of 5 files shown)


crates/verifier/src/protocols/intmul/verify.rs
```diff
@@ -298,25 +282,31 @@ where
...
-		strided_products(&c_hi_selected_evals),
-	)
-	.collect::<Vec<_>>();
+	let expected_a = tree_product(&a_selected_evals);
+	let expected_c_lo = tree_product(&c_lo_selected_evals);
+	let expected_c_hi = tree_product(&c_hi_selected_evals);
+
+	// The overflow and b checks span only the content variables, so their padded contribution
+	// carries the factor eq(0^{n_extra} ; r_bit) = Π_j (1 - r_bit_j).
+	let eq_pad = r_bit
```
jimpo: use eq_ind_zero

jimpo (changes requested):

crates/verifier/src/protocols/intmul/verify.rs
benji-potamus: Done — replaced the manual `Π(1 - r_bit_j)` fold with `eq_ind_zero(r_bit)` in `ad1216d`. Tests + clippy + fmt still green.

jimpo (approved):

PR #1679, opened 2026-07-03
BINIUS-206: wire the portable Blake3 leaf hasher into a selectable HashSuite

tcoratger:
Closes [BINIUS-206](https://linear.app/jimpo/issue/BINIUS-206).

Makes the portable Blake3 leaf hasher from #1660 (BINIUS-181) reachable from the proving stack, by exposing it as a `HashSuite`. Pure plumbing — no protocol change, no new soundness assumption.

### The problem

BINIUS-181 landed a portable, auto-vectorized Blake3 leaf hasher — but nothing can reach it.

- Its public types are referenced only by its own benchmark and unit tests.
- The Merkle tree picks its leaf hasher through a `HashSuite`, and no suite points at the kernel.
- So the fast path exists, was measured in isolation, and then sat unused.

Underneath there is a smaller blocker.

- A `HashSuite`'s parallel leaf hasher is built with no arguments — the trait bound is `+ Default`.
- The portable parallel digest implemented only a `new` constructor, not `Default`.
- So even a hand-written suite pointing at it would not have compiled.

### The idea

Give the kernel a front door: a `HashSuite` whose leaf path *is* the portable kernel.

- Every leaf digest is a standard Blake3 hash — bit-identical to any other Blake3 hasher.
- Only the parallel compute path changes.
- The two-to-one inner-node compression is the existing Blake3 one, untouched.

Then the prover and verifier can be instantiated with it directly:

```
Prover::<_, PortableBlake3HashSuite>::setup(...)
Verifier::<PortableBlake3HashSuite>::setup(...)
```

### The change

```diff
  // crates/hash/src/blake3_portable.rs — the digest a suite must construct via Default
- pub struct PortableBlake3ParallelDigest<const LANES: usize>;
+ #[derive(Debug, Clone, Default)]
+ pub struct PortableBlake3ParallelDigest<const LANES: usize>;

  // crates/hash/src/blake3.rs — the new suite, next to Blake3HashSuite
+ pub struct PortableBlake3HashSuite;
+ impl HashSuite for PortableBlake3HashSuite {
+     type LeafHash       = blake3::Hasher;
+     type Compression    = Blake3Compression;                       // unchanged
+     type ParLeafHash    = PortableBlake3ParallelDigest<16>;        // the only difference
+     type ParCompression = ParallelCompressionAdaptor<Blake3Compression>;
+ }
```

The batch width is fixed at 16 lanes: the throughput sweet spot on NEON in the kernel's own benchmark, and the width the AVX2 / AVX-512 vectorizer fills (4 and 8 lanes measure slower).

### Soundness

Nothing to argue — plumbing, not protocol.

- Every leaf digest is a standard Blake3 hash, pinned bit-identical to the reference.
- The inner-node compression is unchanged.
- The suite is opt-in; the SHA-256 default (`StdHashSuite`) is untouched everywhere.

### The one caveat

The kernel takes the vectorized route only for leaves within one 1024-byte chunk.

- Leaf $\le$ one chunk: hashed by the batch kernel.
- Larger leaf: falls back to the scalar adapter that walks the tree.

So the speedup is confined to the sub-chunk regime the ZK leaf-hashing path exercises. This is documented on the type, not hidden.

### Scope

- **new:** `PortableBlake3HashSuite` and its crate-root re-export; a suite-parity test.
- **changed:** one `derive` on `PortableBlake3ParallelDigest`.
- **untouched:** the kernel itself, the SHA-256 default, and every existing suite.

### Verification

- `cargo clippy -p binius-hash --all-targets` — clean.
- `cargo +nightly fmt --all` — no changes.
- `cargo test -p binius-hash` — 13 passed, including the new suite-parity test, which runs both Blake3 suites' leaf hashers through the exact fixed-length call the Merkle tree makes and pins them equal to `blake3::hash` across the 1024-byte routing boundary (leaf sizes 0, 1, 63, 100, 1000, 1024 on the batch route; 1025, 4096 on the fallback route).

### DIFF
--- a/crates/hash/src/blake3.rs
+++ b/crates/hash/src/blake3.rs
@@ -7,7 +7,6 @@ use digest::Output;
 use super::{
 	binary_merkle_tree::HashSuite, blake3_portable::PortableBlake3ParallelDigest,
 	compress::CompressionFunction, parallel_compression::ParallelCompressionAdaptor,
-	parallel_digest::ParallelDigestAdapter,
 };
 
 /// A two-to-one compression function that hashes the concatenation of its inputs with Blake3.
@@ -24,22 +23,9 @@ impl CompressionFunction<Output<blake3::Hasher>, 2> for Blake3Compression {
 }
 
 /// Blake3 [`HashSuite`]: Blake3 leaves and a Blake3 compression function for inner nodes.
-#[derive(Debug, Clone, Default)]
-pub struct Blake3HashSuite;
-
-impl HashSuite for Blake3HashSuite {
-	type LeafHash = blake3::Hasher;
-	type Compression = Blake3Compression;
-	type ParLeafHash = ParallelDigestAdapter<blake3::Hasher>;
-	type ParCompression = ParallelCompressionAdaptor<Blake3Compression>;
-}
-
-/// Blake3 [`HashSuite`] whose leaves run the portable auto-vectorized kernel.
 ///
-/// Every leaf digest is a standard Blake3 hash.
-///
-/// Only the parallel compute path differs from the other Blake3 suite:
-/// - Leaves within one 1024-byte chunk are hashed by the vectorized batch kernel.
+/// Every leaf digest is a standard Blake3 hash; only the parallel compute path is specialized.
+/// - Leaves within one 1024-byte chunk are hashed by the portable auto-vectorized batch kernel.
 /// - Larger leaves fall back to the scalar adapter that walks the tree.
 ///
 /// The speedup is confined to the sub-chunk regime the ZK leaf-hashing path exercises.
@@ -49,9 +35,9 @@ impl HashSuite for Blake3HashSuite {
 /// - The width the AVX2 / AVX-512 vectorizer fills.
 /// - 4 and 8 lanes both measure slower.
 #[derive(Debug, Clone, Default)]
-pub struct PortableBlake3HashSuite;
+pub struct Blake3HashSuite;
 
-impl HashSuite for PortableBlake3HashSuite {
+impl HashSuite for Blake3HashSuite {
 	type LeafHash = blake3::Hasher;
 	type Compression = Blake3Compression;
 	type ParLeafHash = PortableBlake3ParallelDigest<16>;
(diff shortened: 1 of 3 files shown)


crates/hash/src/blake3.rs
```diff
@@ -33,6 +34,30 @@ impl HashSuite for Blake3HashSuite {
...
+/// - Larger leaves fall back to the scalar adapter that walks the tree.
+///
+/// The speedup is confined to the sub-chunk regime the ZK leaf-hashing path exercises.
+///
+/// The batch width is fixed at 16 lanes:
+/// - The throughput sweet spot on NEON in the portable-kernel benchmark.
+/// - The width the AVX2 / AVX-512 vectorizer fills.
+/// - 4 and 8 lanes both measure slower.
+#[derive(Debug, Clone, Default)]
+pub struct PortableBlake3HashSuite;
```
jimpo: Just name it the Blake3HashSuite.

crates/hash/src/blake3.rs
```diff
@@ -33,6 +34,30 @@ impl HashSuite for Blake3HashSuite {
...
+
+impl HashSuite for PortableBlake3HashSuite {
+	type LeafHash = blake3::Hasher;
+	type Compression = Blake3Compression;
+	type ParLeafHash = PortableBlake3ParallelDigest<16>;
+	type ParCompression = ParallelCompressionAdaptor<Blake3Compression>;
```
jimpo, line 44: Follow-up: can you make a parallel compression too?

jimpo (commented):
Another follow-up would be to be able to select this HashSuite with the example CLI. There's already a CLI flag (though I believe it's misnamed simply "compression").

crates/hash/src/blake3.rs
tcoratger: Fixed, I wasn't sure if we wanted to keep both but this is much better to have a single one, I agree :)

crates/hash/src/blake3.rs
tcoratger: Just created a Linear ticket: https://linear.app/jimpo/issue/BINIUS-213/add-a-parallel-blake3-compression-for-batched-merkle-inner

tcoratger:
> Another follow-up would be to be able to select this HashSuite with the example CLI. There's already a CLI flag (though I believe it's misnamed simply "compression").

Just fixed review comments and created 2 Linear tickets for the 2 followup PRs:
- [BINIUS-212](https://linear.app/jimpo/issue/BINIUS-212/select-the-blake3-hash-suite-from-the-example-cli-and-rename-the)
- [BINIUS-213](https://linear.app/jimpo/issue/BINIUS-213/add-a-parallel-blake3-compression-for-batched-merkle-inner-node)

jimpo (approved):

PR #1680, opened 2026-07-03
Split constraint_system.rs into a folder module

tcoratger:
@jimpo Just a proposal that I designed with Claude to limit a bit the size of the big file, as usual, if you don't like, feel free to close :) 

Pure refactor — no behavior change. The public API is byte-for-byte unchanged.

`crates/core/src/constraint_system.rs` had grown to ~2600 lines. This turns it into a folder module, one file per concept, with tests colocated next to the code they exercise.

### The problem

One file held a dozen unrelated types — `ValueIndex`, `ShiftVariant`, the constraints, `ConstraintSystem`, `ValueVecLayout`, `ValueVec`, `ValuesData`, `Proof` — plus ~1350 lines of tests.

Over half the file was tests, mixed together regardless of what they covered.

Finding anything meant scrolling through ~2600 lines.

### The idea

Split into a folder module whose `mod.rs` re-exports every submodule:

```
constraint_system/
  mod.rs          module doc + `pub use` of every submodule
  value_index.rs  ValueIndex
  shift.rs        ShiftVariant, ShiftedValueIndex
  constraint.rs   Operand, AndConstraint, MulConstraint
  system.rs       ConstraintSystem
  layout.rs       ValueVecLayout
  value_vec.rs    WordPair, AlignedWords (private), ValueVec
  values_data.rs  ValuesData
  proof.rs        Proof
```

Because `mod.rs` does `pub use <submodule>::*`, the path `binius_core::constraint_system::Foo` resolves exactly as before. This is how `std` lays out modules like `std::collections` — a directory of private submodules re-exported from `mod.rs`.

The module keeps the name `constraint_system`: it is the crate's central concept, and 53 external import sites spell that path out. The problem was file size, not the name.

### The change

Same code, moved. Each type keeps its `SerializeBytes` / `DeserializeBytes` impls next to it, and each file has its own `#[cfg(test)] mod tests` holding exactly the tests for that type. The set of 46 test functions is identical before and after (verified by diff).

The only non-mechanical touches, all forced by the file boundaries:

- `ValueVecLayout::is_padding` / `is_committed_oob` go from module-private to `pub(super)` — `system.rs::validate` now calls them across a file boundary.
- The three `include_bytes!` reference-binary paths gain one `../` (the tests moved one directory deeper).
- Four doc-only intra-doc links that now cross a module boundary get qualified targets (e.g. `` [`ValueVec`] `` → `` [`ValueVec`](super::ValueVec) ``) so they still resolve. No prose reworded.
- `test_word_serialization_round_trip` moves to `word.rs` — it only tests `Word`, so that is its true home.

### Scope

- **changed:** `crates/core/src/constraint_system.rs` → `crates/core/src/constraint_system/` (9 files); one test moved into `crates/core/src/word.rs`.
- **unchanged:** every public path, every type, every impl, all serialized formats and version constants.
- no downstream crate needed an edit.

### Verification

- `cargo clippy -p binius-core --all-targets` — clean
- `cargo +nightly fmt` — clean (matches the repo's nightly `rustfmt.toml`)
- `cargo test -p binius-core` — 70 passed, 3 ignored (the reference-binary generators, as before)
- `cargo doc -p binius-core` with `-D rustdoc::broken_intra_doc_links` — clean
- built every downstream crate (`frontend`, `verifier`, `prover`, `m4-*`, `spartan-*`, `examples`) against the unchanged module path — all green


jimpo (approved):

jimpo:
Nice, needs rebase

tcoratger:
> Nice, needs rebase

Just fixed the conflicts

PR #1681, opened 2026-07-03
BINIUS-209: allow 0 mul constraints and skip the intmul protocol

benji-potamus:
Closes [BINIUS-209](https://linear.app/jimpo/issue/BINIUS-209/allow-0-mul-constraints-and-disable-intmul-protocol).

Lets a constraint system have **zero MUL constraints**, in which case both the prover and verifier skip the IntMul reduction entirely (no transcript messages on either side). This enables [BINIUS-208](https://linear.app/jimpo/issue/BINIUS-208) without imposing the IntMul cost on circuits that have no integer multiplication (e.g. pure-AND circuits like SHA-256).

## Changes
- **`ConstraintSystem::validate_and_prepare`**: an empty MUL constraint set is kept at zero rather than padded up to a single dummy constraint (so the "0 mul" state becomes reachable). The AND path is untouched — still padded to the next power of two.
- **`IOPProver::prove` / `IOPVerifier::verify`**: guard the IntMul reduction on `n_mul_constraints() > 0`. Both sides apply the **identical** guard, so the Fiat-Shamir transcript stays in lockstep.
- When skipped, each side synthesizes a **zero IntMul `OperatorData`** (four zero evals at an empty `r_x_prime`, sharing BitAnd's `r_zhat_prime`). The shift reduction's monster evaluation iterates the (now empty) MUL constraints and this claim contributes zero, so the shift/PCS phases are unchanged and still receive a (zero) claim rather than an `Option`.
- **Circuit-statistics display** (`stat.rs`): with zero MUL constraints nothing is allocated, so there's no power-of-two exponent to report — print `0 used (0.0% of 0)` instead of the wraparound `2^64` (`trailing_zeros(0)`). Regenerated the affected snapshots (the pure-AND hash circuits: sha256, sha512, keccak, blake2s, blake2b, blake3_compress, hashsign).

## Soundness
The existing "IntMul before BitAnd" ordering (IntMul evals must be bound before `r_zhat_prime` is drawn) only matters when IntMul runs; when it's skipped there are no evals to bind, and `r_zhat_prime` is drawn solely by BitAnd — consistent between prover and verifier because both skip identically.

## Validation
- `test_prove_verify_zero_mul_constraints` (new): asserts the SHA-256 circuit has `n_mul_constraints() == 0` and runs both plain and ZK prove→verify through the skip path.
- The existing `test_prove_verify_sha256_preimage` / `test_zk_prove_verify_sha256_preimage` now exercise the skip path too (SHA-256 has no muls, previously masked by the pad-to-1).
- Updated `tests/shift.rs::test_shift_prove_and_verify` to handle a zero-MUL constraint system (empty point + zero claim), matching the production skip.
- Full `binius-core` / `binius-prover` / `binius-verifier` / `binius-frontend` / `binius-spartan-*` / `binius-m4-prover` suites pass; all 9 circuit-snapshot examples pass `check-snapshot`; `clippy --all-targets -D warnings`, nightly `fmt --check`, and `typos` clean; portable-leg (`RUSTFLAGS=""`) build green.

### DIFF
--- a/crates/core/src/constraint_system.rs
+++ b/crates/core/src/constraint_system.rs
@@ -603,9 +603,15 @@ impl ConstraintSystem {
 	pub fn validate_and_prepare(&mut self) -> Result<(), ConstraintSystemError> {
 		self.validate()?;
 
-		// Require all constraint types to have a power-of-two count.
+		// Require all constraint types to have a power-of-two count. An empty MUL constraint set is
+		// kept at zero (rather than padded to a single dummy constraint) so the prover and verifier
+		// can skip the IntMul reduction entirely — see `IOPProver::prove` / `IOPVerifier::verify`.
 		let and_target_size = self.and_constraints.len().next_power_of_two();
-		let mul_target_size = self.mul_constraints.len().next_power_of_two();
+		let mul_target_size = if self.mul_constraints.is_empty() {
+			0
+		} else {
+			self.mul_constraints.len().next_power_of_two()
+		};
 
 		self.and_constraints
 			.resize_with(and_target_size, AndConstraint::default);
--- a/crates/examples/snapshots/blake2b.snap
+++ b/crates/examples/snapshots/blake2b.snap
@@ -7,8 +7,8 @@ Gates & Instructions
 Constraints
 ├─ AND constraints: 8,163 used (99.6% of 2^13)
 │  [▓▓▓▓▓▓▓▓▓░] spare: 29
-├─ MUL constraints: 0 used (0.0% of 2^0)
-│  [░░░░░░░░░░] spare: 1
+├─ MUL constraints: 0 used (0.0% of 0)
+│  [░░░░░░░░░░] spare: 0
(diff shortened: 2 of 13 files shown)


jimpo (approved):

benji-potamus:
Heads up @jimpo — since your approval I pushed two follow-ups, so this may need a fresh look before enqueueing:
- Fixed the `circuits-snapshot` CI failure: with 0 mul constraints the stats display rendered `2^64` (`trailing_zeros(0)`); now prints `0 used (0.0% of 0)`, and regenerated the 7 affected (pure-AND) snapshots.
- Rebased onto latest `main` (`#1661`).

All CI was green on the pre-rebase commit; re-validated locally after the rebase (full core/prover/verifier/frontend suites, all 9 circuit snapshots, clippy `-D warnings`, nightly fmt). Now at `fc702d10`.

jimpo (approved):

PR #1682, opened 2026-07-04
BINIUS-208: require IOP channels in the intmul protocol

benji-potamus:
Part 1 of the [BINIUS-208](https://linear.app/jimpo/issue/BINIUS-208/use-logup-for-fixed-base-exponentiation-in-intmul) stack (logup*-based fixed-base exponentiation; design clarified in a comment on the issue).

Mechanical preparation for the logup*-based phase 5, which commits the pushforward oracle mid-protocol: the top-level intmul prove/verify entry points now take IOP channels. Both call sites already hold IOP channels, so no behavior changes and the transcript is identical.

- `intmul::verify` bound goes `IPVerifierChannel` → `IOPVerifierChannel`; the phase verifier functions keep their IP bounds (they never receive oracles).
- `IntMulProver::prove` moves to its own impl block bound on `IOPProverChannel`; the `#[doc(hidden)]` phase methods stay on the IP-bound block, so the phase benches keep using bare transcripts.
- The prove/combined benches run over a real BaseFold prover channel (production parameters, one ZK oracle spec declared for the upcoming pushforward), so oracle transmission costs are measured once the lookup lands. The round-trip test wraps its transcripts in the naive channels.

Stacked on top (pushed, PRs opened in sequence as each merges): the logup* Elem generalization ([`benji/binius-208-logup-star-elem-generic`](https://github.com/binius-zk/binius64/tree/benji/binius-208-logup-star-elem-generic)), the BINIUS-210 multi-looker logup* generalization ([BINIUS-210](https://linear.app/jimpo/issue/BINIUS-210/generalize-logup-for-multiple-looker-tables), [`benji/binius-210-generalize-logup-for-multiple-looker-tables`](https://github.com/binius-zk/binius64/tree/benji/binius-210-generalize-logup-for-multiple-looker-tables)), then the core phase 4/5 replacement ([`benji/binius-208-intmul-logup-lookup`](https://github.com/binius-zk/binius64/tree/benji/binius-208-intmul-logup-lookup)).

### DIFF
--- a/crates/prover/benches/intmul.rs
+++ b/crates/prover/benches/intmul.rs
@@ -1,6 +1,13 @@
 // Copyright 2025-2026 The Binius Developers
 use binius_core::word::Word;
 use binius_field::{BinaryField128bGhash, Field, PackedBinaryGhash1x128b};
+use binius_hash::StdHashSuite;
+use binius_iop::{
+	basefold_compiler::BaseFoldVerifierCompiler, channel::OracleSpec, fri::MinProofSizeStrategy,
+};
+use binius_iop_prover::{
+	basefold_compiler::BaseFoldProverCompiler, merkle_tree::prover::BinaryMerkleTreeProver,
+};
 use binius_ip_prover::{
 	prodcheck::ProdcheckProver,
 	sumcheck::{
@@ -10,6 +17,7 @@ use binius_ip_prover::{
 };
 use binius_math::{
 	multilinear::{eq::eq_ind_partial_eval_scalars, evaluate::evaluate},
+	ntt::{NeighborsLastSingleThread, domain_context::GenericPreExpanded},
 	test_utils::random_scalars,
 };
 use binius_prover::protocols::intmul::{
@@ -29,6 +37,41 @@ type F = BinaryField128bGhash;
 const LOG_NUM: usize = 14;
 /// Log of the integer bit width; the exponentiation tree has `2^LOG_BITS` leaves per node.
 const LOG_BITS: usize = 6;
+
+/// The Reed-Solomon rate of the standard prover configuration.
+const LOG_INV_RATE: usize = 1;
+/// One FRI test query keeps the parameters cheap; the benchmarks measure commitment cost, not
+/// opening soundness.
+const N_TEST_QUERIES: usize = 1;
+/// Log-length of the oracle budget the channel declares. intmul commits no oracle yet; this
+/// matches the 2^16-entry exponentiation-table pushforward its logup*-based phase 5 commits,
+/// so the harness measures that cost as soon as it lands.
+const LOG_ORACLE_LEN: usize = 16;
+
+/// Builds a BaseFold prover compiler for the prove benchmarks.
+///
+/// The prove benchmarks run over a real BaseFold channel so that oracle transmission (NTT
+/// encoding and Merkle tree construction) is measured; a naive channel would serialize
+/// oracles for free. The final batched opening in `finish` is not measured, since the full
+/// system amortizes it into the single opening shared with the witness trace.
+fn basefold_compiler() -> BaseFoldProverCompiler<
+	P,
+	NeighborsLastSingleThread<GenericPreExpanded<F>>,
+	BinaryMerkleTreeProver<F, StdHashSuite>,
+> {
+	let merkle_prover = BinaryMerkleTreeProver::<F, StdHashSuite>::new();
+	let verifier_compiler = BaseFoldVerifierCompiler::new(
+		merkle_prover.scheme().clone(),
+		vec![OracleSpec::new(LOG_ORACLE_LEN)],
+		LOG_INV_RATE,
+		N_TEST_QUERIES,
+		&MinProofSizeStrategy,
+	);
+	let domain_context =
+		GenericPreExpanded::generate_from_subspace(verifier_compiler.max_subspace());
+	let ntt = NeighborsLastSingleThread::new(domain_context);
+	BaseFoldProverCompiler::from_verifier_compiler(&verifier_compiler, ntt, merkle_prover)
+}
 
 fn generate_test_data(log_num: usize) -> (Vec<Word>, Vec<Word>, Vec<Word>, Vec<Word>) {
 	let num_exponents = 1 << log_num;
@@ -75,20 +118,26 @@ fn bench_intmul_prove(c: &mut Criterion) {
 
 	// prove
 	let witness = Witness::<P>::new(LOG_BITS, &a, &b, &c_lo, &c_hi).unwrap();
+	let compiler = basefold_compiler();
 
+	// The channel is created in the setup closures. It borrows its transcript, and a closure
+	// cannot return a borrow of captured state, so each setup leaks one small transcript; the
+	// leak is bounded by the iteration count.
 	group.bench_with_input(
 		BenchmarkId::new("prove", num_exponents),
 		&witness,
 		|bencher, witness| {
-			bencher.iter_with_setup(
+			bencher.iter_batched_ref(
 				|| {
-					let prover_transcript = ProverTranscript::<StdChallenger>::default();
-					(witness.clone(), prover_transcript)
+					let transcript =
+						Box::leak(Box::new(ProverTranscript::<StdChallenger>::default()));
+					(Some(witness.clone()), compiler.create_channel_without_zk(transcript))
 				},
-				|(witness, mut prover_transcript)| {
-					let mut intmul_prover = IntMulProver::new(0, &mut prover_transcript);
-					intmul_prover.prove(witness);
+				|(witness, channel)| {
+					let mut intmul_prover = IntMulProver::new(0, channel);
+					intmul_prover.prove(witness.take().expect("set in setup"));
 				},
+				BatchSize::SmallInput,
 			)
 		},
 	);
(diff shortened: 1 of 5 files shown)


crates/prover/benches/intmul.rs
```diff
@@ -86,7 +87,9 @@ fn bench_intmul_prove(c: &mut Criterion) {
 					(witness.clone(), prover_transcript)
 				},
 				|(witness, mut prover_transcript)| {
-					let mut intmul_prover = IntMulProver::new(0, &mut prover_transcript);
+					let mut channel =
+						NaiveProverChannel::<F, _>::new(&mut prover_transcript, vec![]);
```
jimpo: For benchmarking, it's unfortunately important to use a real `BaseFold` channel, because the cost of sending an oracle is non-trivial.

jimpo (changes requested):

crates/prover/benches/intmul.rs
benji-potamus: Done — in the stacked core-lookup commit, where the pushforward oracle actually exists: the prove/combined/phase5 benches now run over a real BaseFold channel (production parameters: `OracleSpec::new_zk(LIMB_BITS)`, `SECURITY_BITS = 96`, `LOG_INV_RATE = 1`). The channel is created outside the timed window (`iter_custom` — the channel borrows the transcript, so it can't be returned from an `iter_batched` setup closure); the timed region includes `send_oracle` (mask + NTT encode + Merkle commit) but not the final batched opening in `finish`, which the full system amortizes into the opening shared with the witness trace.

It stays naive in this PR only because nothing is sent here yet: `BaseFoldProverCompiler::new` asserts a non-empty oracle spec list and `finish` asserts every spec was consumed, so a zero-oracle BaseFold channel can't be constructed — and with no oracle the naive wrapper is transcript-identical.

You were right that it's non-trivial: committing the 2^16-entry ZK pushforward costs ~100 ms single-threaded, dominated by Merkle hashing. Corrected numbers (LOG_NUM=14, 1 thread, vs main): prove 129.9 → 242.5 ms, combined witness+prove 262.7 → 275.8 ms (+5%), where the naive-channel numbers had shown −26.5%. The commit cost is fixed (table-sized, independent of the trace), so it amortizes: at LOG_NUM=17 prove is 1.17 s, with the commit at ~9%. I'll put the full comparison in the core PR description.

benji-potamus: Pushed the rest of the stack so the bench change is visible: the BaseFold-channel harness is in [`9e9cd76` on `benji/binius-208-intmul-logup-lookup`](https://github.com/binius-zk/binius64/commit/9e9cd764d360) (see `pushforward_compiler` and the `iter_custom` loops in `crates/prover/benches/intmul.rs`), stacked as `benji/binius-208-logup-star-elem-generic` → `benji/binius-210-generalize-logup-for-multiple-looker-tables` → `benji/binius-208-intmul-logup-lookup`; PRs for those follow as each parent merges.

benji-potamus: Moved into this PR as requested — the prove/combined benches here now run over a real BaseFold channel (`basefold_compiler()`: `OracleSpec::new_zk(16)` for the upcoming pushforward, `SECURITY_BITS = 96`, `LOG_INV_RATE = 1`; the compiler is hoisted and the channel is created in the setup path of `iter_with_setup`, and since intmul commits nothing yet the channel is dropped without `finish`). The stacked core commit keeps the same harness and just swaps the spec constant for `LIMB_BITS` and adds phase5. Also dropped the `iter_custom` detour — channel creation from the hoisted compiler is microseconds, so plain `iter_with_setup`/`iter_batched` it is. Verified the benches run at this commit standalone and at the stack tip; all branches re-pushed (head here is now `1ce1694c`).

crates/prover/benches/intmul.rs
```diff
@@ -30,6 +42,38 @@ const LOG_NUM: usize = 14;
...
+	let n_test_queries = calculate_n_test_queries(SECURITY_BITS, LOG_INV_RATE);
+	let merkle_prover = BinaryMerkleTreeProver::<F, StdHashSuite>::new();
+	let verifier_compiler = BaseFoldVerifierCompiler::new(
+		merkle_prover.scheme().clone(),
+		vec![OracleSpec::new_zk(LOG_ORACLE_LEN)],
+		LOG_INV_RATE,
+		n_test_queries,
+		&MinProofSizeStrategy,
+	);
+	let domain_context = GenericOnTheFly::generate_from_subspace(verifier_compiler.max_subspace());
```
jimpo: Use the expanded twiddles, not on the fly.

crates/prover/benches/intmul.rs
```diff
@@ -30,6 +42,38 @@ const LOG_NUM: usize = 14;
...
+fn basefold_compiler() -> BaseFoldProverCompiler<
+	P,
+	NeighborsLastSingleThread<GenericOnTheFly<F>>,
+	BinaryMerkleTreeProver<F, StdHashSuite>,
+> {
+	let n_test_queries = calculate_n_test_queries(SECURITY_BITS, LOG_INV_RATE);
+	let merkle_prover = BinaryMerkleTreeProver::<F, StdHashSuite>::new();
+	let verifier_compiler = BaseFoldVerifierCompiler::new(
+		merkle_prover.scheme().clone(),
+		vec![OracleSpec::new_zk(LOG_ORACLE_LEN)],
```
jimpo: shouldn't be ZK

crates/prover/benches/intmul.rs
```diff
@@ -30,6 +42,38 @@ const LOG_NUM: usize = 14;
...
+/// The prove benchmarks run over a real BaseFold channel so that oracle transmission (masking,
+/// NTT encoding, and Merkle tree construction) is measured; a naive channel would serialize
+/// oracles for free. The final batched opening in `finish` is not measured, since the full
+/// system amortizes it into the single opening shared with the witness trace.
+fn basefold_compiler() -> BaseFoldProverCompiler<
+	P,
+	NeighborsLastSingleThread<GenericOnTheFly<F>>,
+	BinaryMerkleTreeProver<F, StdHashSuite>,
+> {
+	let n_test_queries = calculate_n_test_queries(SECURITY_BITS, LOG_INV_RATE);
```
jimpo: Fine to just hardcode `n_test_queries` to 1.

crates/prover/benches/intmul.rs
```diff
@@ -98,13 +145,16 @@ fn bench_intmul_prove(c: &mut Criterion) {
...
 		&num_exponents,
 		|bencher, _| {
-			bencher.iter(|| {
-				let witness = Witness::<P>::new(LOG_BITS, &a, &b, &c_lo, &c_hi).unwrap();
-
-				let mut prover_transcript = ProverTranscript::<StdChallenger>::default();
-				let mut intmul_prover = IntMulProver::new(0, &mut prover_transcript);
-				intmul_prover.prove(witness);
-			})
+			bencher.iter_with_setup(
```
jimpo: iter_with_setup is hidden in docs. Use `iter_batched_ref` instead, which is recommended.

crates/prover/benches/intmul.rs
```diff
@@ -98,13 +145,16 @@ fn bench_intmul_prove(c: &mut Criterion) {
...
+				ProverTranscript::<StdChallenger>::default,
+				|mut prover_transcript| {
+					let mut channel =
+						compiler.create_channel(&mut prover_transcript, StdRng::seed_from_u64(0));
```
jimpo: Channel creating should happen in the setup closure. You seem to have a problem doing that, presumably because of the lifetime issue of borrowing the newly created transcript. hack around it but declaring a mut transcript above the benchmark call, then in the setup closure, assign it to a new one and make the wrapping channel

jimpo (changes requested):

crates/prover/benches/intmul.rs
benji-potamus: Done — `GenericPreExpanded` (matching the production `ProverNTT` in `crates/prover/src/prove.rs`).

crates/prover/benches/intmul.rs
benji-potamus: Done — `OracleSpec::new(LOG_ORACLE_LEN)`, and the channels are now built with `create_channel_without_zk` (no rng). Same in the stacked core commit (`OracleSpec::new(LIMB_BITS)`).

crates/prover/benches/intmul.rs
benji-potamus: Done — `const N_TEST_QUERIES: usize = 1`, dropped the `calculate_n_test_queries`/`SECURITY_BITS` imports.

crates/prover/benches/intmul.rs
benji-potamus: Done — both prove benches (and phase5 in the core commit) use `iter_batched_ref` now.

crates/prover/benches/intmul.rs
benji-potamus: Channel creation is in the setup closure now. The outer-`mut`-transcript variant doesn't get past the borrow checker — the setup closure is `FnMut`, and returning a channel that borrows the captured slot is a lending closure ("captured variable cannot escape `FnMut` closure body"). Closest safe equivalent: the setup `Box::leak`s a fresh transcript (a few KB each, bounded by iteration count in a short-lived bench process) and builds the channel from the `'static` borrow, so the routine is exactly `IntMulProver::new` + `prove`. Pushed as `ef61b87f`; happy to switch to a raw-pointer reuse of one outer transcript instead if you'd rather avoid the leak.

jimpo: OK, we'll fixed this later with BINIUS-214.

jimpo (approved):

PR #1683, opened 2026-07-04
BINIUS-211: remove the lifetime parameter from TransparentEvalFn

tcoratger:
Closes [BINIUS-211](https://linear.app/jimpo/issue/BINIUS-211/refactor-transparentevalfn-to-remove-lifetime-parameter).

Removes the lifetime parameter from the transparent-evaluation closure so it is `'static`.
Pure refactor — no protocol or behavior change.

### The problem

The closure that evaluates a transparent polynomial was tied to a caller lifetime:

```
pub type TransparentEvalFn<'a, Elem> = Box<dyn Fn(&[Elem]) -> Elem + 'a>;
```

That `'a` was contagious.
It sat on `OracleLinearRelation<'a, ...>`, on the whole `IOPVerifierChannel<'r, F>` trait, on every channel impl, and on every generic `verify` that named the trait.
A channel could not be `'static` while it carried these relations, which is what the issue asks to fix.

### The idea

Make the closure own everything it reads and drop the parameter:

```
pub type TransparentEvalFn<Elem> = Box<dyn Fn(&[Elem]) -> Elem + 'static>;
```

`OracleLinearRelation`, the trait, and the `verify` signatures all shed their lifetime with it.
`'static` on a boxed closure forces `Elem: 'static`, so the trait now reads `IPVerifierChannel<F, Elem: 'static>`.

### Where the `'static` bites, and why it is fine

Almost every transparent closure already owns its captures, so it was `'static` in all but name.

Two things were not:

- `eval_transparent` borrowed the constraint system.
  It now owns the multiplication constraints through a shared `Rc` — mirroring the `Rc` already used for the equality-indicator tensor in the same function.
- The two symbolic wrapper channels have an `Elem` of `CircuitElem<F, InstanceGenerator<'a>>` / `...<WitnessGenerator<'a>>`, which borrowed a `&'a WitnessLayout`.
  Those generators now **own** their layout.

An F-level transparent evaluator (the direction the code's own `HACK` note gestures at) is not viable here: the generic verifier only holds `Elem`-typed challenges in the symbolic flow, so it cannot build an `F`-valued closure.
Owning the layout is the self-contained fix.

### Soundness

- The transparent closures compute the same values; the boxed function body is unchanged.
- Nothing in the transcript, challenge sampling, or opening changes.
- The generator layout clone is a copy of read-only data, so the reconstructed public vector is identical.

### Scope

- **changed:** `TransparentEvalFn`, `OracleLinearRelation`, the `IOPVerifierChannel` trait, and all 8 channel impls (iop, spartan-verifier, spartan-prover) plus the generic `verify` fns in `spartan-verifier` and `verifier`.
- **changed:** `InstanceGenerator` and `WitnessGenerator` now own a cloned `WitnessLayout` (their `'a` type parameter is gone); `ReplayChannel` loses its lifetime as a result.
- **left untouched:** protocol logic, the prover's `LogupProof`, and every channel's runtime behavior.

### Performance note

The generators clone their layout once per proof.
That clone is `O(circuit)`, but the wrapper prover/verifier already replays the entire circuit (`O(circuit)`), so it is a negligible add.
If preferred, threading `Rc<WitnessLayout>` through the config structs is a trivial follow-up that removes even that.

### Verification

- `cargo clippy` on all touched crates (`--all-targets`) — clean
- tests pass across `binius-iop`, `binius-spartan-verifier`, `binius-spartan-prover`, `binius-iop-prover`, `binius-m4-verifier`, `binius-spartan-frontend`, `binius-verifier`, including the end-to-end `test_zk_wrapped_prove_verify` (ZK wrapper + replay round-trip) and the logUp* IOP verifier tests.


jimpo:
Ah, shoot. The clone of the witness layout is actually pretty bad. I forgot about that. Maybe we could make `WitnessGenerator` generic over `LayoutRef: Deref<Target=WitnessLayout>` so that it could take either a ref or an Rc?

tcoratger:
> Ah, shoot. The clone of the witness layout is actually pretty bad. I forgot about that. Maybe we could make `WitnessGenerator` generic over `LayoutRef: Deref<Target=WitnessLayout>` so that it could take either a ref or an Rc?

Just used Arc, looks cleaner now

jimpo (approved):

tcoratger:
@jimpo Just fixed the conflicts!

jimpo (approved):

PR #1688, opened 2026-07-04
Enforce the 32-bit half-word shift-amount bound in validation

tcoratger:
Tightens shift-amount validation so the 32-bit half-word variants reject amounts they cannot actually represent. No behavior change for any valid circuit — the frontend already produces only in-range amounts; this closes the gap on the paths that accept externally-supplied constraint systems.

### The problem

A shifted operand term carries a shift variant and an amount.
The four half-word variants — `Sll32`, `Srl32`, `Sra32`, `Rotr32` — act on each 32-bit half independently and read only the **lower 5 bits** of the amount.
`Word::sll32` and its siblings literally mask `n & 0x1F`, so the meaningful range is `0..32`.

The frontend constructors enforce `amount < 32` for those variants.
But the two paths that admit an **externally-supplied** term only checked `amount >= 64`:

- witness / constraint-system **deserialization**
- `ConstraintSystem::validate`

So a crafted or corrupted system could carry a `Sll32` with `amount = 40`, and it would pass validation.

### Why that is a real gap

The two consumers then disagree about what that term means:

```
apply(Sll32, 40)   ->  Word::sll32 masks 40 & 31 = 8   ->  shift by 8
shift protocol      ->  indexes shift-amount row 40      ->  shift by 40
```

The word-level semantics say "shift by 8"; the shift argument encodes row 40.
Validation exists precisely to reject terms whose meaning is not well-defined across prover and verifier, and this one slipped through.

### The idea

Put the per-variant bound in one place on the enum, and have every path that accepts a term consult it.

- `ShiftVariant::is_half_word()` — true for the `*32` family.
- `ShiftVariant::max_amount()` — `32` for half-word variants, `64` otherwise.

Construction, validation, and deserialization now enforce the same bound, so a term that survives any of them denotes the same shift everywhere.

### The change

```
ShiftedValueIndex::deserialize:
- if amount >= 64 { .. }
+ if amount >= shift_variant.max_amount() { .. }

ConstraintSystem::validate (validate_operand):
- if term.amount >= 64 { ShiftAmountTooLarge { .., shift_amount } }
+ let max_amount = term.shift_variant.max_amount();
+ if term.amount >= max_amount { ShiftAmountTooLarge { .., shift_amount, max_amount } }
```

The `ShiftAmountTooLarge` error now carries `max_amount`, so its message reports the real limit (`n=32>=32`) instead of a misleading `>=64`.

### Scope

- **new:** `ShiftVariant::is_half_word`, `ShiftVariant::max_amount` in `crates/core/src/constraint_system/shift.rs`
- **changed:** the amount check in `ShiftedValueIndex::deserialize` and in `ConstraintSystem::validate`; the `ShiftAmountTooLarge` variant gains a `max_amount` field
- **left untouched:** the frontend constructors (already correct) and `apply` (masking behavior is unchanged)

### Soundness

- No valid circuit is affected: the frontend never emits an out-of-range half-word amount, so accepted systems are unchanged.
- The only new behavior is **rejecting** previously-accepted-but-ill-defined terms, which strengthens the prover/verifier consistency validation is meant to guarantee.
- Serialization wire format is unchanged; full-width variants still accept `0..64`.

### Verification

- `cargo +nightly fmt --all`, `cargo check --workspace --all-targets`, `cargo clippy -p binius-core --all-targets` — clean
- `cargo test -p binius-core` — 73 passed
- new tests: `max_amount` / `is_half_word` per variant; deserialization accepts `Sll32` amount 31 and rejects 32 (full-width still accepts 32–63); `validate` rejects a `Sll32` amount-32 term with `ShiftAmountTooLarge { max_amount: 32 }`


jimpo (approved):

tcoratger:
@jimpo Just fixed conflicts

PR #1689, opened 2026-07-04
BINIUS-215: add Merkle IP channel traits over transcript

benji-potamus:
Introduces the Merkle IP channel abstraction from [BINIUS-215](https://linear.app/jimpo/issue/BINIUS-215/abstract-merkle-ip-channels-over-transcript). This is the first PR of the stack; the follow-up converts FRI and BaseFold to use it.

## Commits

1. **Fix salt index in hiding Merkle opening proofs** — `BinaryMerkleTreeProver::prove_opening` wrote `get_salt(index >> layer_depth)`, i.e. the salt of the wrong leaf whenever `layer_depth > 0`. Invisible in-tree because production trees are non-hiding and the existing hiding tests only exercise `layer_depth = 0`; caught by the new channel round-trip test. Includes a regression test opening hiding commitments against internal layers.
2. **Add the Merkle IP channel traits + transcript-backed impls** — `MerkleIPVerifierChannel<F>: IPVerifierChannel<F>` in `binius_iop::merkle_channel` and `MerkleIPProverChannel<F>: IPProverChannel<F>` (with `P: PackedField<Scalar = F>` on the data-carrying methods) in `binius_iop_prover::merkle_channel`, implemented by `VerifierMerkleTranscriptChannel<T, Challenger_, F, H: HashSuite>` / `ProverMerkleTranscriptChannel<T, Challenger_, F, H>` wrapping a transcript (owned or `&mut`, via `BorrowMut`) together with a `BinaryMerkleTreeScheme<F, H>` / `BinaryMerkleTreeProver<F, H>`. Roots travel as observed messages; openings as unobserved decommitment advice, using the optimal-layer batching FRI uses today. Also adds `MerkleTreeProver::prove_vector`, the prover mirror of `MerkleTreeScheme::verify_vector` (writes the per-leaf salts for hiding trees; today that side is open-coded saltless in the FRI fold prover). Round-trip, hiding, borrowed-transcript, and negative tests included.

## Design decisions beyond the issue sketch

- **`Result`-returning receive methods.** `recv_merkle_commitment`/`recv_openings` return `Result<_, merkle_channel::Error>` rather than bare values, matching `IPVerifierChannel::recv_one` — reads can hit an empty tape and openings can fail verification.
- **`recv_committed_vector` / `send_committed_vector`.** FRI's terminal codeword is sent in full and checked with `MerkleTreeScheme::verify_vector`; that receive is Merkle-bound, so it belongs on the channel. Without it the FRI conversion would still need a raw transcript escape hatch.
- **`sample_bits`.** FRI samples query indices via `CanSampleBits`, which the IP channel traits don't expose. Putting index sampling on the Merkle channel traits lets the query phase go fully behind the abstraction.
- **Prover `send_openings(commitment, data, indices)` re-takes the data.** The commitment handle owns the Merkle tree (`Committed`) but borrows the committed values at open time, so FRI keeps owning its codewords across fold rounds unchanged.

### DIFF
--- a/crates/iop-prover/src/merkle_channel.rs
+++ b/crates/iop-prover/src/merkle_channel.rs
@@ -1,0 +1,242 @@
+// Copyright 2026 The Binius Developers
+
+//! Channel abstraction for provers of protocols using Merkle commitments.
+//!
+//! This module provides the [`MerkleIPProverChannel`] trait, the prover-side counterpart of
+//! `binius_iop::merkle_channel::MerkleIPVerifierChannel`. It extends [`IPProverChannel`] with the
+//! ability to send Merkle commitments and openings of the committed leaves.
+//!
+//! The [`ProverMerkleTranscriptChannel`] implementation wraps a [`ProverTranscript`] and commits
+//! with a [`BinaryMerkleTreeProver`]: commitment roots are written as observed messages, while
+//! opening proofs are written as unobserved decommitment advice bound to the already-observed
+//! roots.
+
+use std::{borrow::BorrowMut, marker::PhantomData};
+
+use binius_field::{Field, PackedField};
+use binius_hash::binary_merkle_tree::{BinaryMerkleTree, HashSuite};
+use binius_iop::merkle_tree::MerkleTreeScheme;
+use binius_ip_prover::channel::IPProverChannel;
+use binius_math::FieldSlice;
+use binius_transcript::{
+	ProverTranscript,
+	fiat_shamir::{CanSampleBits, Challenger},
+};
+use binius_utils::{SerializeBytes, checked_arithmetics::log2_strict_usize};
+use digest::Output;
+use rand::CryptoRng;
+
+use crate::merkle_tree::{MerkleTreeProver, commit_field_buffer, prover::BinaryMerkleTreeProver};
+
+/// An extension of [`IPProverChannel`] that can send and open Merkle commitments.
+pub trait MerkleIPProverChannel<F: Field>: IPProverChannel<F> {
+	/// A Merkle commitment, carrying the data required to open it later.
+	type Commitment;
+
+	/// Commits `data` as a Merkle tree with leaves of exactly `leaf_size` scalars each and sends
+	/// the commitment.
+	///
+	/// The tree depth is `log2(data.len() / leaf_size)`.
+	///
+	/// ## Preconditions
+	///
+	/// * `data.len()` must be a multiple of `leaf_size`, and the resulting leaf count must be a
+	///   power of two.
+	fn send_merkle_commitment<P: PackedField<Scalar = F>>(
+		&mut self,
+		data: FieldSlice<P>,
+		leaf_size: usize,
+	) -> Self::Commitment;
+
+	/// Sends a multi-opening of committed leaves, bound by a Merkle commitment.
+	///
+	/// All indices must be less than `2^depth` for the commitment's tree depth. The verifier
+	/// receives `indices.len() * leaf_size` field elements via its matching `recv_openings` call.
+	///
+	/// ## Preconditions
+	///
+	/// * `data` must be the buffer passed to [`Self::send_merkle_commitment`] for this commitment.
+	fn send_openings<P: PackedField<Scalar = F>>(
+		&mut self,
+		commitment: &Self::Commitment,
+		data: FieldSlice<P>,
+		indices: &[usize],
+	);
+
+	/// Sends the full committed vector, bound by a Merkle commitment.
+	///
+	/// ## Preconditions
+	///
+	/// * `data` must be the buffer passed to [`Self::send_merkle_commitment`] for this commitment.
+	fn send_committed_vector<P: PackedField<Scalar = F>>(
+		&mut self,
+		commitment: &Self::Commitment,
+		data: FieldSlice<P>,
+	);
+
+	/// Samples a uniform integer with the given number of bits.
+	///
+	/// Protocols use this to sample query indices for [`Self::send_openings`], matching the
+	/// verifier's samples.
+	fn sample_bits(&mut self, bits: usize) -> usize;
+}
+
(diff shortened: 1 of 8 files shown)


crates/iop/src/merkle_channel.rs
```diff
@@ -0,0 +1,249 @@
...
+	fn sample_bits(&mut self, bits: usize) -> usize {
+		CanSampleBits::sample_bits(self.transcript.transcript_mut(), bits) as usize
+	}
+}
+
+/// Access to a mutably borrowed [`VerifierTranscript`].
+///
+/// This is implemented for both owned and mutably borrowed transcripts, so that channel wrappers
+/// generic over it can take either.
+pub trait AsVerifierTranscriptMut {
```
jimpo: Remove this custom trait. You can use `BorrowMut<VerifierTranscript<Challenger_>>` in its place

crates/iop-prover/src/merkle_channel.rs
```diff
@@ -0,0 +1,249 @@
...
+	fn sample_bits(&mut self, bits: usize) -> usize {
+		CanSampleBits::sample_bits(self.transcript.transcript_mut(), bits) as usize
+	}
+}
+
+/// Access to a mutably borrowed [`ProverTranscript`].
+///
+/// This is implemented for both owned and mutably borrowed transcripts, so that channel wrappers
+/// generic over it can take either.
+pub trait AsProverTranscriptMut {
```
jimpo: Like the verifier side, use BorrowMut instead of a custom trait

crates/iop-prover/src/merkle_channel.rs
```diff
@@ -0,0 +1,249 @@
...
+	fiat_shamir::{CanSampleBits, Challenger},
+};
+use binius_utils::{SerializeBytes, checked_arithmetics::log2_strict_usize};
+use digest::Output;
+use rand::CryptoRng;
+
+use crate::merkle_tree::{MerkleTreeProver, commit_field_buffer, prover::BinaryMerkleTreeProver};
+
+/// An extension of [`IPProverChannel`] that can send and open Merkle commitments.
+pub trait MerkleIPProverChannel<P: PackedField>: IPProverChannel<P::Scalar> {
```
jimpo: I think `MerkleIPProverChannel` should just be generic over `F` and the methods be generic over `P: PackedField<Scalar=F>`. That will integrate better into FRI, which uses packed elements for the top level codeword, but F for the folded codewords.

jimpo (changes requested):

crates/iop/src/merkle_channel.rs
benji-potamus: Done — the helper trait is gone; the wrapper is now `VerifierMerkleTranscriptChannel<T, Challenger_, S>` with `T: BorrowMut<VerifierTranscript<Challenger_>>`. One note: the challenger had to become a struct parameter (held as `PhantomData`) rather than living only in the where clause, since a `BorrowMut` bound alone leaves it unconstrained (E0207). Call sites that construct the channel annotate it (`::<_, Challenger_, _>`), which inference occasionally needs anyway because `&mut VerifierTranscript<C>` implements both `BorrowMut<VerifierTranscript<C>>` and the reflexive `BorrowMut<&mut VerifierTranscript<C>>`.

crates/iop-prover/src/merkle_channel.rs
benji-potamus: Done — same shape on the prover: `ProverMerkleTranscriptChannel<T, Challenger_, MP>` with `T: BorrowMut<ProverTranscript<Challenger_>>` and the challenger as a `PhantomData` parameter (same E0207 constraint as the verifier side).

crates/iop-prover/src/merkle_channel.rs
benji-potamus: Done — `MerkleIPProverChannel<F: Field>` with `P: PackedField<Scalar = F>` on `send_merkle_commitment`/`send_openings`/`send_committed_vector`. You called it: in the FRI conversion I have staged on top of this PR, `FRIQueryProver::prove_queries` needs only a single `MerkleIPProverChannel<F, Commitment = ...>` bound while opening the packed top-level codeword and the scalar folded codewords through the same channel. (It also fixed a type-inference wart where `sample_bits` alone couldn't pin `P`.)

crates/iop-prover/src/merkle_channel.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/// A [`MerkleIPProverChannel`] over a [`ProverTranscript`], committing with a
+/// [`MerkleTreeProver`].
+///
+/// The transcript is held through a [`BorrowMut`] bound, so the channel can own the transcript or
+/// mutably borrow one. Likewise the Merkle tree prover may be held by value or by reference.
+pub struct ProverMerkleTranscriptChannel<T, Challenger_, MP> {
+	transcript: T,
+	merkle_prover: MP,
```
jimpo: fix the merkle_prover to BinaryMerkleTreeProver and BinaryMerkleTreeScheme on the verifier side. The channel structs can be generic over the F and HashSuite instead.

jimpo (changes requested):

crates/iop-prover/src/merkle_channel.rs
benji-potamus: Done — both wrappers are concrete again: `VerifierMerkleTranscriptChannel<T, Challenger_, F, H: HashSuite>` holds `BinaryMerkleTreeScheme<F, H>` by value and `ProverMerkleTranscriptChannel<T, Challenger_, F, H>` holds `BinaryMerkleTreeProver<F, H>` (with `new()` / `hiding(transcript, rng, salt_len)` / `with_scheme`/`with_merkle_prover` ctors). Downstream in the stacked conversion, the protocol layers (FRI, BaseFold) are generic over the channel traits only, and these concrete wrappers get constructed once in the BaseFold compilers.

jimpo (approved):

PR #1690, opened 2026-07-04
Shrink ShiftedValueIndex from 16 bytes to 8

tcoratger:
Halves the size of the value type the prover reads while evaluating operands. Pure in-memory change — the serialized wire format is unchanged, so there is no protocol change and no soundness assumption.

### The problem

A shifted operand term is a value index, a shift variant, and a shift amount:

```
value_index: u32 | shift_variant | amount: usize
```

The amount is always `< 64`, so it needs 6 bits.
But it was stored as a `usize` — 8 bytes — and with alignment padding that made the whole struct **16 bytes**.

Constraint systems hold **millions** of these terms.
The prover walks them on its hot path, evaluating each operand as a XOR of shifted words.
So the struct's size directly sets how much memory that traversal touches and how many terms fit in cache.

### The idea

The amount fits in a byte, so store it in one:

```
before:  value_index: u32 | shift_variant | amount: usize   -> 16 bytes
after:   value_index: u32 | shift_variant | amount: u8       ->  8 bytes
```

A single 8-byte word instead of two.
Half the memory for the same data, and twice as many terms per cache line.

### Keeping the API and the wire format stable

Two things stay exactly as they were, so the change stays contained:

- **Constructors.** `sll`, `srl`, `sar`, `rotr`, and the `*32` variants still take `amount: usize` and still range-check it.
  They narrow to `u8` internally, so **no call site changes**.
- **Wire format.** A `usize` serializes as a `u32` in this repo.
  Serialization still writes the amount as a `u32` and narrows it on read after the range check.
  So serialized systems stay byte-for-byte identical — **no `SERIALIZATION_VERSION` bump, no reference-binary regeneration**.

Field reads that feed a `usize`-typed sink (the shift kernel, the verifier's per-amount table index, the error's reported amount) gain a single `as usize`.

### Scope

- **changed:** `amount` field type in `crates/core/src/constraint_system/shift.rs`; constructors narrow to `u8`; serialize/deserialize keep the `u32` wire encoding
- **cast sites:** one `as usize` each in `system.rs` (error field), `verifier/.../shift/monster.rs` (table index), `m4-prover/src/bitand.rs` (shift kernel)
- **left untouched:** constructor signatures, the serialized format, and every construction call site across `frontend` / `circuits` / `examples`

### Soundness

- No protocol change: the serialized transcript and constraint-system bytes are identical to before.
- No behavior change: the amount range (`< 64`) is unchanged and still enforced on construction and deserialization.
- Purely a storage-width change on an in-memory struct.

### Verification

- `cargo check --workspace --all-targets`, `cargo +nightly fmt --all`, `cargo clippy -p binius-core -p binius-verifier -p binius-m4-prover --all-targets` — clean
- core lib **71 passed** (incl. a new `size_of::<ShiftedValueIndex>() == 8` guard so the win cannot silently regress)
- verifier shift **6**, m4-prover **19**, prover `shift` **8** + `prove_verify` **1** — the full prove/verify pipeline exercises the amount encoding through the shift argument end-to-end
- all serialization round-trip and reference-binary tests pass unchanged, confirming the wire format is stable

### DIFF
--- a/crates/core/src/constraint_system/shift.rs
+++ b/crates/core/src/constraint_system/shift.rs
@@ -130,10 +130,11 @@ pub struct ShiftedValueIndex {
 	pub value_index: ValueIndex,
 	/// The flavour of the shift that the value must be shifted by.
 	pub shift_variant: ShiftVariant,
-	/// The number of bits by which the value must be shifted by.
+	/// The number of bits to shift by.
 	///
+	/// Stored as a byte to keep the struct small: constraint systems hold millions of these.
 	/// Must be less than 64.
-	pub amount: usize,
+	pub amount: u8,
 }
 
 impl ShiftedValueIndex {
@@ -156,7 +157,7 @@ impl ShiftedValueIndex {
 		Self {
 			value_index,
 			shift_variant: ShiftVariant::Sll,
-			amount,
+			amount: amount as u8,
 		}
 	}
 
@@ -169,7 +170,7 @@ impl ShiftedValueIndex {
 		Self {
 			value_index,
 			shift_variant: ShiftVariant::Slr,
-			amount,
+			amount: amount as u8,
 		}
 	}
 
@@ -185,7 +186,7 @@ impl ShiftedValueIndex {
 		Self {
 			value_index,
 			shift_variant: ShiftVariant::Sar,
-			amount,
+			amount: amount as u8,
 		}
 	}
 
@@ -200,7 +201,7 @@ impl ShiftedValueIndex {
 		Self {
 			value_index,
 			shift_variant: ShiftVariant::Rotr,
-			amount,
+			amount: amount as u8,
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


crates/core/src/constraint_system/shift.rs
```diff
@@ -130,10 +130,11 @@ pub struct ShiftedValueIndex {
 	pub value_index: ValueIndex,
 	/// The flavour of the shift that the value must be shifted by.
 	pub shift_variant: ShiftVariant,
```
jimpo, line 132: Maybe this struct should be `repr(u8)` as well

jimpo (approved):
Good catch!

crates/core/src/constraint_system/shift.rs
tcoratger: Done in #1710 — added `#[repr(u8)]` with explicit 0..=7 discriminants so the one-byte layout is guaranteed rather than incidental, and the serialize path collapses to `*self as u8`. Wire format is unchanged.

PR #1691, opened 2026-07-04
BINIUS-210: generalize logup* to multiple lookers sharing one table and pushforward

benji-potamus:
Implements [BINIUS-210](https://linear.app/jimpo/issue/BINIUS-210/generalize-logup-for-multiple-looker-tables): several looker columns can now share one table and one committed pushforward, instead of running a full logup* instance per looker. Part 3 of the [BINIUS-208](https://linear.app/jimpo/issue/BINIUS-208/use-logup-for-fixed-base-exponentiation-in-intmul) stack (follows #1685) — the intmul phase 5 rework consumes exactly this (12 twisted limb claims against one shared exponentiation table).

Construction: the looker-batching challenge γ is sampled before the pushforward commitment; looker j's numerator is γʲ·eq_{r_j}, the combined pushforward is Y = Σⱼ γʲ (I_j)_* eq_{r_j}, and ⟨T, Y⟩ = Σⱼ γʲ e_j. The looker side runs as one GKR circuit over `n + log_lookers` variables: the top `log_lookers` layers fractionally add the per-looker circuit roots (interpolated as real multilinears, numerators 0-padded, denominators 1-padded), and the remaining `n` layers run the per-looker circuits through `fracaddcheck::batch_prove`, seeded at the selector point the top layers bind. The verifier is the single-looker verifier shape: one root fraction per side, cross-multiplied identity, and the existing `fracaddcheck::verify` recursion.

- `binius_ip::logup_star`: `verify_reduction(gamma, ...)` is the single entry point (the IOP layer samples γ). It evaluates the batched leaf numerator transparently (Σⱼ eq(sel, j)·γʲ·eq(r_j, ·)), receives the per-looker index evaluations, checks they combine with the selector coordinates to the leaf denominator (padding denominators are 1), and returns one index claim per looker at the shared content point (`LogupOutput::index_eval_claims`).
- `binius_ip_prover::logup_star`: `prove` takes `Looker`s; `witness::combined_lookers` builds the γ-scaled numerators and the combined pushforward; `prove_reduction` runs the two-phase looker side.
- `binius_ip_prover::fracaddcheck::batch_prove`: padding slots now default to the constant zero fraction `0/1` — each batched content round adds the padding's constant prime polynomial contribution, and the reduced denominator halves pad with 1.
- `binius_iop{,_prover}::logup_star`: committed wrappers sample γ before the `Y` commitment so it binds the batching, then run the reduction.
- Single-looker callers are the 1-element special case (no top layers); round-trip tests cover 1 and 3 lookers at both the bare and committed layers.

Next (last) in the stack: the intmul core phase 4/5 replacement ([`benji/binius-208-intmul-logup-lookup`](https://github.com/binius-zk/binius64/tree/benji/binius-208-intmul-logup-lookup)), which consumes the per-looker index claims directly.

### DIFF
--- a/crates/iop-prover/src/logup_star.rs
+++ b/crates/iop-prover/src/logup_star.rs
@@ -13,8 +13,11 @@
 //! [Soukhanov25]: <https://eprint.iacr.org/2025/946>
 
 use binius_field::{BinaryField, Divisible, PackedField};
+pub use binius_ip_prover::logup_star::Looker;
 use binius_ip_prover::logup_star::{self as reduction, witness};
-use binius_math::{FieldBuffer, multilinear::eq::eq_ind_partial_eval};
+use binius_math::{
+	FieldBuffer, multilinear::eq::eq_ind_partial_eval, univariate::evaluate_univariate,
+};
 
 use crate::channel::IOPProverChannel;
 
@@ -42,10 +45,10 @@ pub struct LogupProof<P: PackedField, Oracle> {
 	pub table_eval_point: Vec<P::Scalar>,
 	/// The claimed evaluation of the table `T` at the point.
 	pub table_eval_claim: P::Scalar,
-	/// The `n`-coordinate point of the index claim.
+	/// The `n`-coordinate point shared by the index claims.
 	pub index_eval_point: Vec<P::Scalar>,
-	/// The claimed evaluation of the embedded index column at its point.
-	pub index_eval_claim: P::Scalar,
+	/// The claimed evaluations of the per-looker embedded index columns at the point.
+	pub index_eval_claims: Vec<P::Scalar>,
 	/// The oracle relation that opens the pushforward at the reduced point.
 	pub pushforward: PushforwardRelation<P, Oracle>,
 }
@@ -62,9 +65,8 @@ pub struct LogupProof<P: PackedField, Oracle> {
 /// # Arguments
 ///
 /// * `table` - The table multilinear `T` over `m` variables (`2^m` entries).
-/// * `index` - The index column, one table position per looker row, of length `2^n`.
-/// * `eval_point` - The `n`-coordinate evaluation point `r`.
-/// * `eval_claim` - The claimed evaluation `e = (I^* T)(eval_point)`.
+/// * `lookers` - The looker columns and claims; every evaluation point must have the same length
+///   `n` and every index column `2^n` entries.
 /// * `channel` - The IOP prover channel, whose next oracle has message length `2^m`.
 ///
 /// # Preconditions
@@ -73,9 +75,7 @@ pub struct LogupProof<P: PackedField, Oracle> {
 /// - Every index entry is less than `2^m`.
 pub fn prove<F, P, Channel>(
 	table: &FieldBuffer<P>,
-	index: &[usize],
-	eval_point: &[F],
-	eval_claim: F,
+	lookers: &[Looker<'_, F>],
 	channel: &mut Channel,
 ) -> LogupProof<P, Channel::Oracle>
 where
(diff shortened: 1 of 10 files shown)


crates/iop-prover/src/logup_star.rs
```diff
@@ -85,18 +85,34 @@ where
...
+	// the logUp challenge c.
 	//
-	//     eq_r = eq(eval_point, .)     the looker numerator
-	//     Y    = I_* eq_r              the pushforward, scattered onto table positions
-	let eq_r = witness::equality_indicator::<F, P>(eval_point);
-	let pushforward = witness::pushforward::<F, P>(&eq_r, index, m);
+	//     N = the stacked gamma-scaled equality indicators    the looker numerator
+	//     Y = I_* N                                           the pushforward
+	let gamma = channel.sample();
+	let (stacked_index, numerator) = witness::stacked_lookers::<F, P>(lookers, gamma);
```
jimpo: Don't create stacked columns. Use the new `fracaddcheck::batch_prove`.

crates/ip/src/logup_star/verify.rs
```diff
@@ -66,8 +83,29 @@ use crate::{
 /// - the batched final layer is inconsistent.
 pub fn verify<F, C>(
```
jimpo: this should take `gamma` and be `verify_reduction` (move its code here). The method that does the sampling is at the IOP layer.

crates/ip/src/logup_star/verify.rs
```diff
@@ -118,12 +163,25 @@ where
...
-	// The looker numerator is the eq_r multilinear, which the verifier evaluates by itself.
+	// The stacked numerator is transparent, so the verifier evaluates it by itself: block j holds
+	// gamma^j · eq(r_j, ·) (padding blocks are zero), so its MLE at the leaf point is the eq-fold
+	// of the per-looker scaled equality evaluations over the looker coordinates.
 	//
-	//     eq_r(point_l) = eq(eval_point, point_l)
-	let expected_eq = eq_ind::<C::Elem>(eval_point, &index_eval_point);
+	//     N(point_l) = sum_j eq(looker_coords, j) · gamma^j · eq(r_j, content_coords)
+	let (content_point, looker_point) = index_eval_point.split_at(n);
+	let mut power = C::Elem::one();
```
jimpo: Don't update mut var. Zip with powers(gamma)

crates/ip/src/logup_star/verify.rs
jimpo: Return one eval claim per index, as a vec. In this routine, receive the split out claimed evals and check that they combine with the looker point to the `looker_den`.

jimpo (changes requested):

crates/iop-prover/src/logup_star.rs
benji-potamus: Done in 447e922: the stacked columns are gone. The prover now builds one `FracAddCheckProver` per looker (unscaled `eq_{r_j}` numerators — γ only weights the root identity and the combined pushforward) and runs them through `fracaddcheck::batch_prove` with the selector challenges sampled after the root fractions. The verifier checks the batched claim with the existing `fracaddcheck::verify` recursion over the eq(selector)-weighted combination of the per-looker roots.

crates/ip/src/logup_star/verify.rs
benji-potamus: Done: `verify` is gone; `verify_reduction(gamma, ...)` is now the single entry point with the code moved in, and the IOP layer samples gamma before receiving the pushforward commitment. Bare-layer tests sample gamma from the transcript themselves.

crates/ip/src/logup_star/verify.rs
benji-potamus: Done — the mut accumulator is gone. The root identity now folds the per-looker fractions zipped with `powers(gamma)`, and the stacked-numerator evaluation this comment was on no longer exists (the batched leaf numerator is the eq(selector)-weighted combination of plain `eq(r_j, ·)` evaluations).

crates/ip/src/logup_star/verify.rs
benji-potamus: Done: `LogupOutput` now carries `index_eval_claims: Vec` (one per looker) at the shared content point. The verifier receives the split-out claimed evals from the prover and checks that the eq(selector)-weighted combination of `c − eval_j` equals the batched `looker_den` before returning them. This also let the intmul phase 5 drop its own split-and-recombine round — it consumes the per-column claims directly.

crates/ip/src/logup_star/verify.rs
```diff
@@ -77,57 +105,115 @@ where
...
+		lookers.iter().all(|looker| looker.eval_point.len() == n),
+		"every looker evaluation point must have the same length"
+	);
+	let log_lookers = lookers.len().next_power_of_two().ilog2() as usize;
 
 	// Sample the logUp challenge c that randomizes the logarithmic-derivative denominators.
 	let c = channel.sample();
 
-	// Read the root fractions of both fractional-addition GKR circuits.
+	// Read the root fractions of every fractional-addition GKR circuit: one per looker, then the
```
jimpo: Change the looker-side verifier logic back. So it should receive just one fractional claim for the root of the tree on the looker side, and run a `fracaddcheck_verify` with `n + log_lookers` layers.

On the prover side, it should prove in two phases. Make `n_lookers` FracAddProvers, one per looker. Then, take the root claims from those trees and interpolate a multilinear over them (one for num, one for den). These are real FieldBuffer<F>. The numerator gets padded with 0s, the denominator padded with 1s. Make a FracAddProver for this and run it to completion, for log_lookers layers. Then do the batch prove over the original FracAddProvers.

One thing to be aware of is that `batch_prove` may need to be modified for the 1-padded denominator (which should be the default behavior for that function). So in each round, when it computes the batched round polynomials, the numerator is padded with a constant 0 function and the denominator with a constant 1 function.

jimpo (changes requested):

crates/ip/src/logup_star/verify.rs
benji-potamus: Done in e9d3566. The verifier is back to receiving one looker-side root fraction and running `fracaddcheck::verify(n + log_lookers, ...)`. The prover runs the two phases as described: per-looker `FracAddCheckProver`s (γʲ-scaled numerators), then the roots interpolated into real `FieldBuffer`s (numerators 0-padded, denominators 1-padded) with a `FracAddCheckProver` run to completion over the `log_lookers` layers, then `batch_prove` over the original provers seeded at the resulting selector point.

`batch_prove` now treats the padding slots as the constant fraction 0/1 by default: the constant composition has the constant prime round polynomial (0 for the numerator, 1 for the denominator) and its claims stay (0, 1) through every fold, so each batched content round adds `pad_eq_sum · batch_coeff` to the constant coefficient, and the reduced halves pad the denominators with 1. The existing batch tests (including non-power-of-two padding) pass against the unmodified verify recursion with the 0/1-padded combined claims.

jimpo (approved):

PR #1693, opened 2026-07-04
BINIUS-215: convert FRI and BaseFold to the Merkle IP channels

benji-potamus:
Second and final PR for [BINIUS-215](https://linear.app/jimpo/issue/BINIUS-215/abstract-merkle-ip-channels-over-transcript), on top of #1689. FRI and BaseFold now interact with the transcript and Merkle trees exclusively through `MerkleIPProverChannel` / `MerkleIPVerifierChannel` — no protocol code touches a transcript or a Merkle scheme/prover directly, and the "must take the concrete transcript" escape hatches in the BaseFold channels are gone.

**Transcripts are byte-identical.** All round-trip tests and the three `test_proof_size_*` byte-count tests pass with unmodified assertions.

## What moved where

- **Verifier**: `verify_mlecheck_basefold` / `verify_batch_zk_basefold` / `BaseFoldVerifierChannel` take `&mut impl MerkleIPVerifierChannel<F>`; roots are received as commitment handles via `recv_merkle_commitment(leaf_size, depth)` (same observed digest read as before), query openings via `recv_openings` (same optimal-layer batched Merkle verification), the terminal codeword via `recv_committed_vector`. The FRI oracles (`BrakedownOracle`/`FRIOracle`/`BatchBrakedownOracle`) are generic over an opaque commitment handle and no longer hold a scheme reference.
- **Prover**: the FRI fold/query stack and `prove_mlecheck_basefold` / `BaseFoldProverChannel` are generic over `MerkleIPProverChannel<F>`. `fri::commit_interleaved`/`commit_masked` split into `encode_*` (pure encoding, `fri/encode.rs`) + one `channel.send_merkle_commitment(...)` at the call site. `FRIFoldProver::execute_fold_round(channel)` commits through the channel; the BaseFold round loop is reordered to coeffs-write → fold+commit (root write) → sample. The *computation* of the commit moves after the coeffs write, but since folding/committing does no transcript I/O the byte stream is unchanged (coeffs, then observed root, then sample — exactly as before).
- **Construction sites**: the concrete `{Verifier,Prover}MerkleTranscriptChannel` wrappers are built only in the BaseFold compilers (`create_channel`); `crates/verifier`, `crates/prover`, spartan, and m4 all go through them. All `merkle_prover` plumbing in prover-side setup paths is deleted, along with `ProverMerkleCommitment::new` (handles are only minted by `send_merkle_commitment`) and the now-unneeded `#[auto_impl(&)]` on the Merkle tree traits.

## Size note

This is one large PR rather than separate FRI/BaseFold steps: routing the fold-phase roots through the channels (BaseFold) required restructuring the same fold-prover state machine the query-phase conversion (FRI) introduces, so an FRI-only intermediate would not have survived review intact. The change is mechanical — signatures and plumbing — with the proof-size tests pinning behavior; net −415 lines.

### DIFF
--- a/crates/iop-prover/src/basefold_compiler.rs
+++ b/crates/iop-prover/src/basefold_compiler.rs
@@ -8,43 +8,46 @@
 use std::marker::PhantomData;
 
 use binius_field::{BinaryField, PackedField};
+use binius_hash::binary_merkle_tree::HashSuite;
 use binius_iop::{
 	basefold_compiler::BaseFoldVerifierCompiler, channel::OracleSpec, fri::FRIParams,
-	merkle_tree::MerkleTreeScheme,
+	merkle_tree::BinaryMerkleTreeScheme,
 };
 use binius_math::ntt::AdditiveNTT;
 use binius_transcript::{ProverTranscript, fiat_shamir::Challenger};
 use binius_utils::SerializeBytes;
+use digest::Output;
 use rand::{Rng, SeedableRng, rngs::StdRng};
 
-use crate::{basefold_channel::BaseFoldProverChannel, merkle_tree::MerkleTreeProver};
+use crate::{
+	basefold_channel::BaseFoldProverChannel, merkle_channel::ProverMerkleTranscriptChannel,
+};
 
 /// A compiler that creates BaseFold ZK prover channels with precomputed parameters.
 ///
 /// This compiler builds a single combined FRI over all oracles, with ZK oracles configured for
 /// zero-knowledge mode.
 #[derive(Debug)]
-pub struct BaseFoldProverCompiler<P, NTT, MerkleProver_>
+pub struct BaseFoldProverCompiler<P, NTT, H>
 where
 	P: PackedField<Scalar: BinaryField>,
 	NTT: AdditiveNTT<Field = P::Scalar> + Sync,
-	MerkleProver_: MerkleTreeProver<P::Scalar>,
+	H: HashSuite,
 {
 	ntt: NTT,
-	merkle_prover: MerkleProver_,
 	oracle_specs: Vec<OracleSpec>,
 	/// The combined FRI parameters over **all** oracles.
 	fri_params: FRIParams<P::Scalar>,
-	_p_marker: PhantomData<P>,
+	_marker: PhantomData<(P, H)>,
 }
 
-impl<F, P, NTT, MerkleScheme, MerkleProver_> BaseFoldProverCompiler<P, NTT, MerkleProver_>
+impl<F, P, NTT, H> BaseFoldProverCompiler<P, NTT, H>
 where
 	F: BinaryField,
 	P: PackedField<Scalar = F>,
 	NTT: AdditiveNTT<Field = F> + Sync,
-	MerkleScheme: MerkleTreeScheme<F, Digest: SerializeBytes>,
-	MerkleProver_: MerkleTreeProver<F, Scheme = MerkleScheme>,
+	H: HashSuite,
+	Output<H::LeafHash>: SerializeBytes,
 {
 	/// Creates a new compiler with precomputed combined FRI parameters.
 	///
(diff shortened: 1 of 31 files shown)


crates/iop/src/basefold_compiler.rs
```diff
@@ -119,19 +124,32 @@ where
 	}
 
 	/// Creates a ZK verifier channel from this compiler and a transcript.
+	///
+	/// The returned channel drives all prover interaction through a
+	/// [`VerifierMerkleTranscriptChannel`] over the transcript, constructed here with a scheme
+	/// matching this compiler's Merkle scheme.
 	pub fn create_channel<'a, Challenger_>(
 		&'a self,
 		transcript: &'a mut VerifierTranscript<Challenger_>,
```
jimpo, line 133: This should instead take in an `impl MerkleIPVerifierChannel<F>` by value.

Then remove the `merkle_scheme` field from the struct.

crates/iop/src/basefold_compiler.rs
```diff
@@ -110,7 +113,9 @@ where
 	}
 
 	/// Creates a [`SizeTrackingChannel`] from this compiler's oracle specs.
-	pub fn create_size_tracking_channel(&self) -> SizeTrackingChannel<'_, F, MerkleScheme_> {
+	pub fn create_size_tracking_channel(
```
jimpo, line 116: Remove this method, it's unused.

crates/iop/src/basefold_channel.rs
```diff
@@ -358,10 +312,15 @@ where
 
 		let index = self.next_oracle_index;
 
+		// Receive the commitment with its Merkle tree shape, matching the prover-side commit: the
+		// oracle's codeword has dimension `log_dim - log_lift` and one interleaved coset of
+		// `2^log_batch_size` scalars per leaf.
+		let fri_oracle = &self.fri_params.input_oracles()[index];
+		let depth = (self.fri_params.rs_code().log_dim() - fri_oracle.log_lift)
```
jimpo, line 319: assert consistency with `log_msg_len`

crates/iop-prover/src/basefold_compiler.rs
```diff
@@ -121,13 +115,22 @@ where
 
 	/// Creates a ZK prover channel from this compiler, a transcript, and an RNG.
 	///
-	/// The `rng` is used to seed an internal `StdRng` for mask generation.
+	/// The returned channel drives all prover interaction through a
+	/// [`ProverMerkleTranscriptChannel`] over the transcript, constructed here with a non-hiding
+	/// Merkle tree prover. The `rng` is used to seed an internal `StdRng` for mask generation.
 	pub fn create_channel<'a, Challenger_: Challenger>(
 		&'a self,
 		transcript: &'a mut ProverTranscript<Challenger_>,
```
jimpo, line 123: Similar to verifier side: instead of taking a transcript by ref, take a channel by value.

jimpo (approved):
Address comments in a follow-up PR

crates/iop/src/basefold_compiler.rs
benji-potamus: Done in #1696 — `create_channel` takes an `impl MerkleIPVerifierChannel<F>` by value and the `merkle_scheme` field is gone (the compiler collapses to `BaseFoldVerifierCompiler<F>`; `new` still takes a scheme transiently for proof-size estimation, with `H` demoted to a method generic).

crates/iop/src/basefold_compiler.rs
benji-potamus: Removed in #1696 — it had one caller (the spartan proof-size test), which now constructs the `SizeTrackingChannel` directly; the hardcoded size assertion is unmodified.

crates/iop/src/basefold_channel.rs
benji-potamus: Added in #1696 — `recv_oracle` now asserts that the committed message length implied by `(leaf_size, depth)` equals the spec's `log_msg_len`, plus one variable for a ZK oracle's interleaved mask.

crates/iop-prover/src/basefold_compiler.rs
benji-potamus: Done in #1696 — both `create_channel` and `create_channel_without_zk` take the channel by value, and the prover compiler collapses to `BaseFoldProverCompiler<P, NTT>`.

PR #1694, opened 2026-07-04
BINIUS-216: sumcheck shared-multilinear store and round evaluators

benji-potamus:
Implements [BINIUS-216](https://linear.app/jimpo/issue/BINIUS-216/sumcheck-shared-multilinear-store-and-round-evaluators-eager-fold), step 1 of the BINIUS-202 sumcheck rearchitecture (Option A): a shared-multilinear store + round-evaluator architecture with eager folding, behind the unchanged `SumcheckProver`/`MleCheckProver` traits.

Rebased onto current `main`, which since this branch's base gained a batched fracaddcheck (BINIUS-182/236) and a multi-looker + pushforward logup\* (BINIUS-210/220). The full store rewiring is ported onto those, structured as a **commit series** so the whole change is reviewable commit-by-commit; **every commit builds + tests + clippy green on its own**.

## Commit series

1. **`add sumcheck shared-multilinear store and round evaluators`** — the infrastructure, added *additively* (old provers untouched): `MleStore` (borrowed/owned columns behind `ColId`, owns all folding, dedups `Gruen32` eq trackers; module doc states the invariant *the store folds, evaluators only read*), `RoundEvaluator` + `SharedSumcheckProver`/`SharedMleCheckProver` (generic over the evaluator type, `#[auto_impl(Box)]`), `QuadraticMleEvaluator`, `BivariateProductEvaluator`, `MleToSumCheckEvaluator`, `frac_add_mle::evaluators`, and the **differential-equivalence tests** (old provers vs. new store+evaluator path — byte-identical round polys / challenges / emitted evals), landing before any call site is rewired, as the issue prescribes.
2. **`port the batch_quadratic_mle bench to the shared evaluator`**.
3. **`port the single-layer fracaddcheck reduction to the store`** — `prove_layers` folds each layer in place through a store (`pop_last_layer` + a borrowing `layer_mlecheck_prover`), zero up-front copy.
4. **`port the batched fracaddcheck reduction to the store`** — `batch_prove` / `reduce_layer` / `batch_prove_layer`: the per-instance layer provers borrow their columns from owned layer buffers the driver keeps alive (`batch_prove_until_final_layer` returns the owned final-layer buffers rather than built provers, sidestepping the self-referential owner+borrower), and the selector-merge prover runs through the store. Public `batch_prove` signature unchanged, so the multi-looker logup\* caller is untouched.
5. **`port the logup* final layer to the store and remove the pre-store fracaddcheck path`** — `prove_final_layer` builds its fractional prover on a store; then the now-unused consuming `layer_prover` and `frac_add_mle::new` are removed and **`batch_quadratic_mle` is deleted**.

Standalone provers that don't fit the store (selector/rerand, zk mask, `PaddedSumcheckDecorator`, inout) are untouched and batch alongside via `Either`, as today. The `finish()` transcript contract (each shared column evaluated once, emitted in store order) keeps written evals byte-identical.


crates/ip-prover/src/sumcheck/bivariate_product_evaluator.rs
```diff
@@ -0,0 +1,156 @@
...
+	evals: WideRoundEvals2<P::Output>,
+}
+
+impl<P: WideMul> Default for ProductAccum<P> {
+	fn default() -> Self {
+		Self {
+			evals: WideRoundEvals2::default(),
+		}
+	}
+}
```
tcoratger: For stuffs like this we should just derive Default directly on the struct no?

crates/ip-prover/src/sumcheck/bivariate_product_evaluator.rs
```diff
@@ -0,0 +1,156 @@
...
+
+#[derive(Debug, Clone)]
+enum RoundCoeffsOrSum<F: Field> {
+	Coeffs(RoundCoeffs<F>),
+	Sum(F),
+}
```
tcoratger, line 129: To avoid having this, we should maybe merge https://github.com/binius-zk/binius64/pull/1670 first?

crates/ip-prover/src/sumcheck/quadratic_mle_evaluator.rs
```diff
@@ -0,0 +1,283 @@
...
+	pub fn new(
+		store: &mut MleStore<'_, P>,
+		cols: [ColId; N],
+		composition: Composition,
+		infinity_composition: InfinityComposition,
+		eval_point: Vec<F>,
+		eval_claims: [F; M],
+	) -> Self {
+		// precondition
+		assert!(N > 0 && M > 0);
```
tcoratger: Just so that it is known at compile time :)

```suggestion
		const { assert!(N > 0 && M > 0); }
```

crates/ip-prover/src/sumcheck/quadratic_mle_evaluator.rs
```diff
@@ -0,0 +1,283 @@
...
+			y_inf: array::from_fn(|_| Default::default()),
+		}
+	}
+}
+
+#[derive(Debug, Clone)]
+enum RoundCoeffsOrEvals<F: Field, const M: usize> {
+	Coeffs([RoundCoeffs<F>; M]),
+	Evals([F; M]),
+}
```
tcoratger, line 218: Same we can probably avoid having this

crates/ip-prover/src/sumcheck/bivariate_product_evaluator.rs
benji-potamus: Outdated now: the per-worker accumulators were reworked from `dyn Any` boxes to flat `WideMul::Output` slices, which deleted the `ProductAccum` struct entirely — there's no longer a struct here to `derive(Default)` on (the slots are `Default`-initialized generically in `SharedSumcheckProver::execute`).

crates/ip-prover/src/sumcheck/quadratic_mle_evaluator.rs
benji-potamus: Outdated now: each evaluator proves a single claim (the const `M` was removed), so there's no `M > 0` to assert; `N > 0` is guaranteed by the composition's array arity. The compile-time-assert suggestion no longer has anything to guard.

crates/ip-prover/src/sumcheck/bivariate_product_evaluator.rs
benji-potamus: The manual accumulator machinery this pointed at was removed by the `dyn Any` → flat wide-slice rework, so this should be moot now. @jimpo — I'll leave the #1670 sequencing call to you if it still matters for other reasons.

crates/ip-prover/src/sumcheck/quadratic_mle_evaluator.rs
```diff
@@ -0,0 +1,218 @@
...
+		// precondition
+		assert!(N > 0);
+		// precondition
+		assert_eq!(
+			eval_point.len(),
+			store.n_vars(),
+			"evaluation point length must equal the store's number of variables"
+		);
+
+		let eq_tracker = store.register_eq_tracker(&eval_point);
```
jimpo, line 79: I don't like how this is responsible for registering the eval point. The constructor should take the EqId as an arg and a const ref to the MleStore. Add whatever methods to MleStore necessary to get a reference to the full point from the EqId

crates/ip-prover/src/fracaddcheck.rs
```diff
@@ -92,40 +132,15 @@ where
 		self.layers.len()
 	}
 
-	/// Pops the last layer and returns a sumcheck prover for it.
```
jimpo, line 95: What's wrong with this interface? It seems fine, it can simply return a SharedMleCheckProver (and the colIds potentially, as a tuple). Then logup* will just add an evaluator to it.

jimpo (changes requested):

PR #1616, opened 2026-06-28
BINIUS-161: fold trivial gate identities in the circuit builder

tcoratger:
Closes [BINIUS-161](https://linear.app/jimpo/issue/BINIUS-161).

Folds trivial bitwise/select identities in the circuit builder so they emit no gate.
Bit-exact rewrites — no protocol change, no new soundness assumption.

### The problem

Circuit authors write bitwise ops whose operands are sometimes constants.

A masking step might be `band(x, all-1)`, padding might be `bxor(x, 0)`, a branch might `select` on a constant flag.

Each still emits a gate, and every `band`/`bor` gate becomes an AND constraint — the baseline unit of prover cost.

So these redundant gates inflate the proof for nothing.

### The idea

These operations have bit-exact identities that need no constraint at all:

```
c & d / c ^ d / c | d  (both constant)  -> fold to a constant
0 & y -> 0        all-1 & y -> y
x ^ 0 -> x
0 | y -> y        all-1 | y -> all-1
select(const cond) -> one arm           select(_, t, t) -> t
```

The builder already interns constants, so it returns the simplified wire directly and skips the gate.

### The change

```
CircuitBuilder::band / bxor / bor / select:
+ match on the constant value of each operand (a new const_of helper)
+ return the simplified wire for the identities above
  ... otherwise emit the gate as before
```

### Soundness

Every folded case is a bit-exact rewrite of the operation it replaces.

- The returned wire holds exactly the value the gate would have produced.
- No information-carrying constraint is dropped; the removed constraints were tautologies.
- `select` reads only the most significant bit, matching the gate's MSB-bool convention.

### Scope

- **changed:** `CircuitBuilder::band/bxor/bor/select` and a small `const_of` helper in `crates/frontend/src/compiler/mod.rs`.
- **tests:** two gate-fusion snapshot tests used `band(x, all-1)` only as scaffolding to materialize an XOR cone; they now use a witness mask so they still exercise the same fusion path.
- left untouched: the fusion pass, the eval form, every other gate.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-frontend --all-targets`, nightly `fmt --all` — clean.
- `cargo test -p binius-frontend` — 148 pass; `cargo test -p binius-examples` — 15 pass.
- End-to-end soundness: built the XMSS circuit, generated a real witness, and `verify_constraints` passed on all 1,731,476 constraints.

### Benchmark

[...]

PR #1695, opened 2026-07-04
perf(circuits): use XOR for jwt_claims mutually exclusive terminator flags

wstran:
## What

`JwtClaims`'s value-terminator scan combines flags with OR twice per scanned position:

```rust
let is_terminator = b.bor(b.bor(is_quote, is_comma), is_close_brace);   // line 165
...
found_end = b.bor(found_end, found_here);                               // line 173
```

Root cause: each `bor` costs one AND constraint, but in both cases the operands are never set at the same time.

- `is_quote`, `is_comma`, `is_close_brace` are equality tests of the same byte against three distinct constants (`"`, `,`, `}`), so at most one MSB is set.
- `found_here` is gated by `should_check`, which ANDs in `not_found_yet = bnot(found_end)`. Once `found_end` is set no later position contributes, so the accumulator and the new term never share the MSB.

Fix: replace both ORs with XOR, which is linear and free.

```rust
let is_terminator = b.bxor(b.bxor(is_quote, is_comma), is_close_brace);
...
found_end = b.bxor(found_end, found_here);
```

This is the same constraint-enforced-disjointness reasoning as the merged `biguint_lt` swap (#1686).

## Numbers

AND constraints from the zklogin snapshot (the only circuit that uses `JwtClaims`; every other snapshot is unchanged):

| circuit | before | this PR | reduction |
|---------|--------|---------|-----------|
| zklogin | 421,170 | 413,896 | -1.73% (-7,274) |

The terminator loop runs once per byte position per attribute, so the three ANDs saved per position (two from `is_terminator`, one from the `found_end` accumulator) compound across the scan.

```
repro: cargo run --release --example zklogin -- check-snapshot
```

## Correctness

The change preserves the MSB of both rewritten wires for every witness, and only the MSB is read downstream:

- `is_quote` / `is_comma` / `is_close_brace` come from `icmp_eq`, whose non-MSB bits are undefined and whose MSB is pinned to `(byte == const)`. A single byte cannot equal two distinct constants, so their MSBs are mutually exclusive and `bor` equals `bxor` there. `is_terminator` is consumed only by `band(should_check, is_terminator)`, an MSB read.
- For `found_end`, `MSB(found_here) = 1` forces `MSB(found_end) = 0`, because `found_here = band(should_check, is_terminator)` and `should_check` ANDs in `bnot(found_end)`. So the accumulator and the new term never share the MSB, and `bxor` reproduces `bor`'s MSB at every step. `found_end` is read only by `bnot` (feeding the gate above) and `assert_true`, both MSB reads, so the diverging low bits never reach an observation.

The pattern-search accumulator `found_start` at line 122 is deliberately left as `bor`: a pattern can match at several positions, so those flags are not mutually exclusive.

All 336 `binius-circuits` tests pass (including the JWT extraction roundtrips), the zklogin snapshot is re-blessed, and no other snapshot changes.

## Scope

Only the two mutually-exclusive ORs in the terminator scan are changed. The `band`s that AND independent predicates (`matches_here`, `should_check`) and the `found_start` OR are genuine and left as is.

### DIFF
--- a/crates/circuits/src/jwt_claims.rs
+++ b/crates/circuits/src/jwt_claims.rs
@@ -1,3 +1,4 @@
+// Copyright 2026 The Binius Developers
 // Copyright 2025 Irreducible Inc.
 use binius_core::word::Word;
 use binius_frontend::{CircuitBuilder, Wire, WitnessFiller, util::pack_bytes_into_wires_le};
@@ -161,8 +162,10 @@ impl JwtClaims {
 				let is_comma = b.icmp_eq(byte_at_pos, comma);
 				let is_close_brace = b.icmp_eq(byte_at_pos, close_brace);
 
-				// It's a terminator if it's any of the three
-				let is_terminator = b.bor(b.bor(is_quote, is_comma), is_close_brace);
+				// It's a terminator if it's any of the three. The three equality flags test
+				// the same byte against distinct constants, so at most one is set and OR
+				// agrees with the linear XOR on the only bit (the MSB) they define.
+				let is_terminator = b.bxor(b.bxor(is_quote, is_comma), is_close_brace);
 
 				let found_here = b.band(should_check, is_terminator);
 
@@ -170,7 +173,10 @@ impl JwtClaims {
 				// When found_here is all-1s, include pos_wire in value_end
 				// When found_here is all-0s, masked_pos is 0 and OR leaves value_end unchanged
 				value_end = b.select(found_here, pos_wire, value_end);
-				found_end = b.bor(found_end, found_here);
+				// `should_check` gates `found_here` on `not_found_yet = bnot(found_end)`, so once
+				// `found_end` is set no later position contributes again. The accumulator and the
+				// new term never share the MSB, so XOR matches OR there and is linear.
+				found_end = b.bxor(found_end, found_here);
 			}
 
 			// Assert that we found a terminator (found_end should be all-1s)
--- a/crates/examples/snapshots/zklogin.snap
+++ b/crates/examples/snapshots/zklogin.snap
@@ -5,24 +5,24 @@ Gates & Instructions
 └─ Number of evaluation instructions: 549,064
 
 Constraints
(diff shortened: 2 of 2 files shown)


benji-potamus (approved):
Nice one — same constraint-enforced-disjointness pattern as the merged #1686, applied cleanly to the terminator scan. I verified the MSB-preservation argument independently against the surrounding code and it holds:

- **`is_terminator` OR→XOR** — `is_quote`/`is_comma`/`is_close_brace` are `icmp_eq` of the *same* byte against three distinct constants, so at most one MSB is set and OR agrees with XOR on the MSB. `is_terminator` is only ever consumed via `band(should_check, …)` and then MSB-broadcast (`select`) or MSB-accumulated, so the low-bit divergence between OR and XOR is never observed.
- **`found_end` OR→XOR** — `found_here`'s MSB is gated by `not_found_yet = bnot(found_end)` inside `should_check`, so `MSB(found_end)=1 ⇒ MSB(found_here)=0`; the accumulator and the new term never share the MSB, XOR reproduces OR's MSB at every step, and the recursion is MSB-only (it only feeds `bnot`→`band`→`should_check` and the final `assert_true`). Worth noting the "all-1s" wording in the existing comments was already loose — under the old `bor`, `found_here` was never genuinely all-1s either (its operands are MSB-boolean with undefined low bits), which is exactly why `assert_true` is MSB-only and why nothing here regresses.
- `found_start` is correctly left as `bor` — a pattern can match at multiple positions, so those flags aren't mutually exclusive.

Copyright header addition matches CONTRIBUTING.md, CI is green, and the zklogin snapshot is the only one that moves (−1.73%), consistent with `JwtClaims` being the sole consumer. Well-reasoned writeup as usual. Approving — I'll leave the actual merge to a maintainer.

jimpo:
@wstran Please rebase onto main and re-bless snapshot

wstran:
Done, rebased and re-blessed. zklogin holds at 421,110 -> 413,842 (-1.73%), orthogonal to #1616 so unchanged.

jimpo (approved):

PR #1702, opened 2026-07-05
BINIUS-208: replace intmul phase 4/5 with a logup* fixed-base exponentiation lookup

benji-potamus:
Implements [BINIUS-208](https://linear.app/jimpo/issue/BINIUS-208/use-logup-for-fixed-base-exponentiation-in-intmul): intmul phases 4 and 5 now read the twisted per-limb exponentiation claims from a shared table of generator powers `i ↦ Gⁱ` via the committed multi-looker logup* reduction (#1691), instead of materializing the looked-up columns and batching them with an MLE-eval sumcheck. Final PR of the BINIUS-208 stack (follows #1682, #1685, #1691).

- Phase 4 is a single full-depth batched prodcheck: `prodcheck::batch_prove` now runs all layers (with the all-but-last body split out as `prodcheck::batch_prove_until_final_layer`), so the verifier is one `prodcheck::verify(LOG_N_LIMBS)` call plus one recombination check of the received per-limb evals.
- Phase 5 builds the `N_LIMB_COLUMNS = 12` per-limb index columns (limb slices of the `a`, `c_lo`, `c_hi` exponent words), Frobenius-twists the phase-4 claims onto the shared table, and reads them from the table via the committed logup* reduction. The pushforward oracle is committed inside and its opening relation is queued on the channel there (deferred to `finish()` for BaseFold).
- The reduction returns one embedded-index claim per column at a shared content point; phase 5 folds them by a sampled ρ and runs a separate rerandomization sumcheck bringing the folded index claim, the `b̃` rerandomization, and the parity zerocheck to one shared output point.
- The table is succinct: the verifier checks the reduced table claim against the power-table MLE directly. The pre-existing phase-4/5 machinery for materialized lookup columns is removed.
- intmul is now fixed to `WORD_SIZE_BITS`/`LOG_WORD_SIZE_BITS` constants: the `log_bits` runtime parameter is gone from the witness, prover, verifier, and bench surfaces.

Benchmark numbers (BaseFold channel, production params, main vs branch at LOG_NUM 14 and 17) are in the comments below.

### DIFF
--- a/crates/verifier/src/protocols/intmul/verify.rs
+++ b/crates/verifier/src/protocols/intmul/verify.rs
@@ -3,36 +3,35 @@
 
 use std::iter;
 
-use binius_field::{BinaryField, Field, field::FieldOps};
-use binius_iop::channel::IOPVerifierChannel;
+use binius_field::{BinaryField, BinaryField1b, ExtensionField, Field, field::FieldOps};
+use binius_iop::{channel::IOPVerifierChannel, logup_star};
 use binius_ip::{
 	channel::IPVerifierChannel,
+	logup_star::LookerClaim,
 	prodcheck::{self, MultilinearEvalClaim},
 	sumcheck::{BatchSumcheckOutput, batch_verify},
 };
 use binius_math::{
-	multilinear::{
-		eq::{eq_ind, eq_ind_zero},
-		evaluate::evaluate_inplace_scalars,
-	},
+	multilinear::{eq::eq_ind, evaluate::evaluate_inplace_scalars},
 	univariate::evaluate_univariate,
 };
 use binius_utils::checked_arithmetics::log2_ceil_usize;
 
 use super::{
 	common::{
-		IntMulOutput, Phase1Output, Phase2Output, Phase3Output, Phase4Output, frobenius_twist,
-		reconstruct_selecteds,
+		IntMulOutput, LIMB_BITS, LOG_N_LIMBS, N_LIMB_COLUMNS, N_LIMBS, Phase1Output, Phase2Output,
+		Phase3Output, Phase4Output, eval_power_table_mle, frobenius_twist, limb_column_twists,
+		twist_limb_claim,
 	},
 	error::Error,
 };
+use crate::config::{LOG_WORD_SIZE_BITS, WORD_SIZE_BITS};
 
 /// Verify Phase 1: GKR step on the exponentiation product tree.
 ///
 /// Runs prodcheck verification to reduce the root claim on $\widetilde{Q}$ to $2^k$ leaf
 /// evaluation claims, then verifies the leaf evaluations against the prover's claimed values.
 fn verify_phase_1<F, C>(
-	log_bits: usize,
 	initial_eval_point: &[C::Elem],
 	initial_b_eval: C::Elem,
 	channel: &mut C,
@@ -48,16 +47,16 @@ where
 		eval: initial_b_eval,
 		point: initial_eval_point.to_vec(),
 	};
-	let output_claim = prodcheck::verify(log_bits, claim, channel)?;
+	let output_claim = prodcheck::verify(LOG_WORD_SIZE_BITS, claim, channel)?;
 
 	// Split output point: first n are x-point, last k are z-challenges
 	let (eval_point, z_suffix) = output_claim.point.split_at(n_vars);
 
 	// Read 2^k leaf evaluations from channel
-	let b_leaves_evals = channel.recv_many(1 << log_bits)?;
+	let b_leaves_evals = channel.recv_many(WORD_SIZE_BITS)?;
 
 	// Verify: output_claim.eval = multilinear_eval(b_leaves_evals, z_suffix)
-	// The leaf evals form a multilinear over log_bits variables; evaluate at z_suffix
+	// The leaf evals form a multilinear over LOG_WORD_SIZE_BITS variables; evaluate at z_suffix
 	let expected_eval = evaluate_inplace_scalars(b_leaves_evals.clone(), z_suffix);
 
 	channel.assert_zero(expected_eval - output_claim.eval)?;
@@ -74,7 +73,6 @@ where
 /// claims to exponent evaluations on $\widetilde{b}$ and a selector on $\widetilde{P}$, and
 /// (b) the product claim $\widetilde{\textsf{LO}} \cdot \widetilde{\textsf{HI}}$.
 fn verify_phase_3<F, C>(
-	log_bits: usize,
 	twisted_eval_points: Vec<Vec<C::Elem>>,
 	twisted_evals: Vec<C::Elem>,
 	c_eval_point: &[C::Elem],
@@ -87,7 +85,7 @@ where
 {
 	let n_vars = c_eval_point.len();
 
-	assert_eq!(twisted_eval_points.len(), 1 << log_bits);
+	assert_eq!(twisted_eval_points.len(), WORD_SIZE_BITS);
 
 	for twisted_eval_point in &twisted_eval_points {
 		assert_eq!(twisted_eval_point.len(), c_eval_point.len());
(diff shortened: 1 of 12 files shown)


benji-potamus:
Benchmark results, main (`6bd9085e`) vs this branch, criterion `intmul/prove` group at `LOG_NUM = 14` (2^14 exponent triples), single thread (`RAYON_NUM_THREADS=1`), `RUSTFLAGS="-C target-cpu=native"`, AWS Linux box, 10 samples per bench (median of the reported interval):

| bench | main | branch | delta |
|---|---|---|---|
| witness | 172.9 ms | 59.0 ms | **−65.9%** |
| prove | 138.0 ms | 113.7 ms | **−17.6%** |
| combined (witness + prove) | 274.9 ms | 144.1 ms | **−47.6%** |

The prover uses the real BaseFold channel with production parameters (`SECURITY_BITS` 96, `LOG_INV_RATE` 1), so the prove number includes the `send_oracle` commitment of the 2^16-entry ZK pushforward — a fixed table-sized cost that amortizes as the trace grows. Witness generation drops the most because the looked-up column materialization is gone entirely.

LOG_NUM = 17 numbers coming in a follow-up comment.

benji-potamus:
Follow-up: `LOG_NUM = 17` (2^17 exponent triples), same setup (single thread, native target, real BaseFold channel, medians):

| bench | main | branch | delta |
|---|---|---|---|
| witness | 1652 ms | 415 ms | **−74.9%** |
| prove | 1181 ms | 750 ms | **−36.5%** |
| combined | 2673 ms | 1120 ms | **−58.1%** |

The gains widen with trace size, as expected: the fixed table-sized pushforward commitment (2^16 entries) amortizes, while the removed looked-up-column materialization and the per-limb GKR work scale with the trace.

crates/verifier/src/protocols/intmul/verify.rs
```diff
@@ -200,55 +198,17 @@ where
 		point: [selector, eval_point.to_vec()].concat(),
 	};
 
-	// Run the batched prodcheck verification (n_layers = log_bits - 1 reduction layers).
-	let output_claim = prodcheck::verify(n_layers, claim, channel)?;
+	// Run the batched prodcheck verification (all but the trees' final layer).
+	let output_claim = prodcheck::verify(LOG_N_LIMBS - 1, claim, channel)?;
```
jimpo: This whole function seems like it would be simpler if it just ran the prodcheck::verify for all LOG_N_LIMBS. On the prover side, `prodcheck::batch_prove` should do all layers, and `prodcheck::batch_prove_until_final_layer` should do all but the last (and batch_prove calls batch_prove_until_final_layer).

crates/verifier/src/protocols/intmul/verify.rs
```diff
@@ -344,13 +409,14 @@ where
...
-		eval_point: r_content.to_vec(),
+	let output = IntMulOutput {
+		eval_point: r_out.to_vec(),
 		a_evals,
 		b_evals,
 		c_lo_evals,
 		c_hi_evals,
-	})
+	};
+	Ok((output, logup_proof.pushforward))
```
jimpo: change logup_star::verify so that instead of returning the pushforward relation, it just calls verify_linear_relation on the channel. This gets deferred to finish() on the channel anyway (for BaseFoldChannel, which is what we care about).

crates/verifier/src/protocols/intmul/verify.rs
```diff
@@ -258,58 +218,165 @@ where
...
+	let index_content_point = logup_proof.index_eval_point.as_slice();
+	let mut padded_column_evals = logup_proof.index_eval_claims.clone();
+	padded_column_evals.resize(1 << log_cols, C::Elem::zero());
+
+	// Collapse the per-column claims into a single claim on the eq(ρ)-folded column V by sampling
+	// ρ, so the final unification runs over the content variables only.
+	let rho = channel.sample_many(log_cols);
+	let folded_index_claim = evaluate_inplace_scalars(padded_column_evals, &rho);
+
+	// Final unification: one batched sumcheck brings the folded index claim, the recombined b
```
jimpo, line 303: These should really get batched along with the final fracaddcheck layer on the looker columns. We can use batch_prove_until_final_layer for that. Do that in this PR.

jimpo (changes requested):

crates/verifier/src/protocols/intmul/verify.rs
benji-potamus: Done in 61f7ed2c. `prodcheck::batch_prove` now runs all layers (returning `BatchProveOutput { eval_point, evals }`, one reduced eval per input prover); the old all-but-last body is `batch_prove_until_final_layer`, which `batch_prove` delegates to before finishing the retained layer as a regular reduction layer. Phase 4 is now a single `prodcheck::verify(LOG_N_LIMBS)` call: the mid-protocol receive of the three per-tree claims and the hand-rolled batched final layer are gone, replaced by one recombination check of the per-limb evals (eq(r_limb)-fold per tree, eq(selector)-weighted across trees) against the prodcheck output claim.

crates/verifier/src/protocols/intmul/verify.rs
benji-potamus: Done in 61f7ed2c. `logup_star::verify` (and the prover symmetrically) now calls `verify_oracle_relations` / `prove_oracle_relations` on the channel internally — deferred to `finish()` for BaseFold, immediate-but-symmetric for the naive channel. `LogupProof` no longer carries the relation, and intmul plus the top-level `prove.rs`/`verify.rs` stopped threading it (the final open is back to just the trace relation).

crates/verifier/src/protocols/intmul/verify.rs
benji-potamus: Done in 61f7ed2c. logup* gained `prove/verify_until_looker_final_layer` variants (bare + committed): the looker batch runs through `fracaddcheck::batch_prove_until_final_layer`, the per-looker reduced fractions are sent and checked against the batched layer claim, and the retained per-column final-layer MLE-check provers come back to intmul. Phase 5's final unification batches those 12 layer provers with the zerocheck and the b rerandomization in one sumcheck over `n_vars - 1` variables plus a shared line-fold (the two n_vars-variable claims split on the highest variable of r_2, mirroring the table-side final layer). The ρ-fold and the materialized folded index column are gone: the verifier evaluates the transparent leaf-numerator halves (γ^j eq_{r_j}) itself, folds the received denominator halves to get each column's index claim at r_out, and binds it directly to the basis-weighted per-bit evals — one equality per column instead of the eq-weighted detour. The earlier bench tables predate this transcript; I'll re-measure and post updated numbers.

benji-potamus:
Re-measured the branch after the review-round restructuring (61f7ed2c) — same setup as the tables above (criterion `intmul_prove`, `RAYON_NUM_THREADS=1`, `-C target-cpu=native`, BaseFold channel, production params). Main baselines are unchanged from the earlier comments (no intmul changes landed on main since).

| bench | LOG_NUM=14 (prev branch → now) | LOG_NUM=17 (prev branch → now) |
|---|---|---|
| witness | 59.0 → 58.8 ms | 415 → 414.8 ms |
| prove | 113.7 → 112.2 ms | 750 → 756.0 ms |
| combined | 144.1 → 144.0 ms | 2673 baseline; 1120 → 1118.0 ms |

The restructuring is performance-neutral (all deltas within noise, ±1%), so the vs-main improvements from the earlier tables stand: combined −47.6% at LOG_NUM=14 and −58.1% at LOG_NUM=17.

jimpo:
Needs rebase onto main

jimpo (changes requested):
Sorry, I was wrong. Phase 5 is too complicated now with the batching of the last layer with the overflow & b re-randomization. Run logup_star to completion, reducing to evaluation claims on the table indices. Then do a separate re-randomization sumcheck for the table lookup limbs, the overflow, and the b exponent claims.

benji-potamus:
Reverted the final-layer fusion in c3c05524: logup* runs to completion again (the `*_until_looker_final_layer` variants are deleted), returning the per-column index claims at the shared content point, and phase 5 runs the separate rerandomization sumcheck over `n_vars` variables — the eq(ρ)-folded index column MLE-check batched with the overflow zerocheck and the b exponent rerandomization, exactly the pre-review shape. The phase-4 full-depth prodcheck and the channel-opened pushforward relation from the earlier round are kept. All gates green locally (intmul + e2e BaseFold tests, clippy, fmt, doc). This restores the configuration measured in the first bench tables (LOG_NUM=14: 59.0/113.7/144.1 ms, LOG_NUM=17: 415/750/1120 ms), and the re-measurement showed the two phase-5 shapes are within noise of each other, so those numbers stand.

jimpo (approved):

PR #1703, opened 2026-07-05
Optimize build_g_parts scatter for small words

latifkasuli:
## Summary
In `build_g_parts` (shift protocol phase 1), each key's accumulated contribution is
scattered into the g multilinears byte-by-byte. The old loop materialized
`word.0.to_le_bytes()` and processed all 8 bytes, even when the remaining high bytes were
zero. On the optimized SHA256 `shift_reduction` benchmark, many witness words are small
values, so those high-byte iterations only add zero masks.

This keeps the existing byte-level masked-select scatter, but consumes the word directly:

- take the low byte with `remaining_word as u8`
- process that byte exactly as before
- shift with `remaining_word >>= 8`
- stop once the remaining high bytes are zero

This is a single code path for every word. It removes the previous popcount threshold and
the separate set-bit scatter path, so there is no microarchitecture-tuned popcount
constant and less semantic duplication to maintain.

Prover output is unchanged: skipped high zero bytes would only have selected zero masks in
the old loop, and nonzero low bytes are processed in the same low-to-high byte order with
the same lane masks.

## Benchmarks
Fresh local 3-way Criterion run on Apple M4 Pro against current `main` (`947e1377`, after
#1739). The previous sparse-popcount branch is included only as a comparison point; this
PR now uses the early-exit byte loop.

| Benchmark | main | previous sparse-popcount | this PR | Δ vs main |
|---|---:|---:|---:|---:|
| `shift_reduction_phases/phase1_build_g_parts` | 6.0089 ms | 5.7712 ms | 5.4169 ms | -9.85% |
| `shift_reduction_log2_12_bytes_4096/prove` | 2.6846 ms | 2.5875 ms | 2.5040 ms | -6.73% |

Relative to the previous sparse-popcount implementation, the early-exit byte loop is also
faster:

| Benchmark | Δ vs previous sparse-popcount |
|---|---:|
| `phase1_build_g_parts` | -6.14% |
| 4096-byte `prove` | -3.23% |

## Test plan
- `cargo test -p binius-prover --lib` (22/22)
- `cargo clippy -p binius-prover --lib --tests -- -D warnings`
- Fresh `shift_reduction` Criterion comparison above

### DIFF
--- a/crates/prover/src/protocols/shift/phase_1.rs
+++ b/crates/prover/src/protocols/shift/phase_1.rs
@@ -229,11 +229,14 @@ pub fn build_g_parts<F: BinaryField, P: PackedField<Scalar = F>>(
 					//     }
 					// }
 					let start = key.id as usize * (WORD_SIZE_BITS >> P::LOG_WIDTH);
-					let word_bytes = word.0.to_le_bytes();
-					for (&byte, values) in word_bytes.iter().zip(
-						multilinears[start..start + (WORD_SIZE_BITS >> P::LOG_WIDTH)]
-							.chunks_exact_mut(WORD_SIZE_BYTES >> P::LOG_WIDTH),
-					) {
+					let values = &mut multilinears[start..start + (WORD_SIZE_BITS >> P::LOG_WIDTH)];
+					let values_per_byte = WORD_SIZE_BYTES >> P::LOG_WIDTH;
+					let mut remaining_word = word.0;
+					let mut byte_index = 0;
+					while remaining_word != 0 {
+						let byte = remaining_word as u8;
+						let byte_values =
+							&mut values[byte_index * values_per_byte..][..values_per_byte];
 						for value_index in 0..(8 >> P::LOG_WIDTH) {
 							unsafe {
 								let packed_mask_index =
@@ -249,10 +252,12 @@ pub fn build_g_parts<F: BinaryField, P: PackedField<Scalar = F>>(
 								//   due to the chunking
 								// - `value_index` is always in bounds because we iterate over 0..(8
 								//   >> P::LOG_WIDTH)
-								*values.get_unchecked_mut(value_index) +=
+								*byte_values.get_unchecked_mut(value_index) +=
 									acc_packed.select(packed_mask);
 							}
 						}
+						remaining_word >>= 8;
+						byte_index += 1;
 					}
 				}
 


benji-potamus (commented):
Thanks for another carefully-measured optimization — the writeup (bimodal density table, the two failed dense-path attempts, the threshold sweep, and the byte-identical A/B verification) makes this easy to reason about.

I went through the correctness independently and it holds:

- **Dense path:** for byte `b` and `value_index v`, the target element is index `b*8 + v`, which is exactly word bit `b*8 + v` (little-endian byte, bit within byte). `packed_mask_index = (byte >> v) & 1`, and `acc_packed.select(mask)` adds `acc_packed` when the bit is set, `0` when clear.
- **Sparse path:** walks set bits `i` and adds `acc_packed` to `values[i]`.
- Both add `acc_packed` to `values[i]` iff bit `i` is set. Skipping a clear bit vs. adding a zero-select is identical in field addition; key order is unchanged; and element indices within a key are distinct, so intra-key ordering doesn't matter. So the output is provably identical, matching your empirical byte-for-byte result.
- The `unsafe` is sound: `P::LOG_WIDTH == 0 ⇒ values.len() == 64`, and `i = bits.trailing_zeros()` on a `u64` is in `0..64`. The `use_sparse_scatter` flag is loop-invariant per word, and since it's gated on the `P::LOG_WIDTH == 0` const the sparse branch fully const-folds away for wide packings — nice.

CI is green across the board (format, clippy, arm64 + x86_64 + portable checks, doc, tests, wasm), so the mechanical pass is clean too.

**One thing I want to leave to @jimpo rather than decide myself:** this isn't a read-only accessor — it adds a *second* scatter implementation, platform-tuned (M4/NEON) with a magic `SPARSE_SCATTER_MAX_ONES = 24`, to a load-bearing prover hot loop that the core team owns and maintains. The two paths now have to stay semantically in sync forever, and the threshold is calibrated to one microarchitecture. Correctness doesn't gate this at all — it's purely a taste/maintenance-burden call about whether that duplication is worth a real ~8% win on the SHA-2/BLAKE-shaped (hashsign) workloads. That's why I'm leaving this as a comment rather than an approval: the merge decision hinges on that preference, which is the core team's to make.

Two minor, non-blocking notes if it does go in:
- The new comment says "add `acc` directly" but the code adds `acc_packed` (`= P::broadcast(acc)`); harmless since `WIDTH == 1` here, but "add `acc_packed`" would be marginally clearer.
- The constant's doc is M4-specific by nature; a one-line "re-tune per target if this ever matters on other narrow-packing arches" would flag it for the next person.

No changes requested from me — correctness and CI are solid; over to @jimpo on the maintenance tradeoff.

jimpo:
I'm skeptical of the premise that we should optimize for sparse words. The blake3 and sha2 benchmark don't appear to use the optimized SWAR gadgets (blake3_compress_2x) and the sha256_fixed one.

I'm going to change the default behavior of the examples to have less overhead, then we can re-assess.

jimpo:
Does this still produce a speedup on the new shift_reduction benchmark? (It's been changed to use the newer, optimized SHA256 gadget)

latifkasuli:
Reran this on current `main` (`73aef610`) with this patch rebased locally onto it. It still speeds up the directly affected `build_g_parts` phase, but the effect is smaller on the new optimized-SHA256 `shift_reduction` benchmark.

| Benchmark | main | this PR | delta |
|---|---:|---:|---:|
| `shift_reduction_phases/phase1_build_g_parts` | 6.1333 ms | 5.7815 ms | -5.74% |
| `shift_reduction_log2_12_bytes_4096/prove` | 2.7456 ms | 2.7116 ms | -1.24% |

So the sparse-word path still helps the scatter itself, but whole-prover impact is much smaller with the newer SHA256 gadget. I updated the PR body to make `shift_reduction` the primary benchmark and keep the older independent-hash numbers as secondary context.

jimpo:
Interesting... I'm on the fence.

Could you try one more idea for comparison? I suspect that these "sparse" wires are actually just wires with small values in them (so dense in the bottom N bits, 0 at the top).

The loop currently uses `word.0.to_le_bytes()` to get the bytes, then bit selects with bits in each byte. Instead, you could write a loop that gets the bytes with `while word != 0`, `word as u8` to get the byte, and `word >> 8` to get to the next iteration. Basically, the same loop, but it early exits when the remaining top bits are 0. This would be the code path that every word goes through -- no conditional based on popcount or anything.

latifkasuli:
Tested this, and that hypothesis matches the data. The early-exit byte loop is better than both current `main` and the previous popcount/sparse-path version.

I force-pushed this PR to use that implementation instead (`d09c4eca`). It keeps the same byte-level masked-select scatter, but consumes the word with `remaining_word as u8` / `remaining_word >>= 8` and stops once the remaining high bytes are zero. So there is no popcount threshold and no second set-bit scatter implementation anymore.

Fresh 3-way Criterion run on Apple M4 Pro against current `main` (`947e1377`):

| Benchmark | main | previous sparse-popcount | this PR | delta vs main |
|---|---:|---:|---:|---:|
| `shift_reduction_phases/phase1_build_g_parts` | 6.0089 ms | 5.7712 ms | 5.4169 ms | -9.85% |
| `shift_reduction_log2_12_bytes_4096/prove` | 2.6846 ms | 2.5875 ms | 2.5040 ms | -6.73% |

Checks run on the updated branch:

- `cargo test -p binius-prover --lib` (22/22)
- `cargo clippy -p binius-prover --lib --tests -- -D warnings`

jimpo (approved):

PR #1704, opened 2026-07-05
BINIUS-224: fine-grained shift_reduction phase benchmarks

benji-potamus:
Implements [BINIUS-224](https://linear.app/jimpo/issue/BINIUS-224/fine-grained-benchmarks-for-shift-reduction-proving).

Breaks the `shift_reduction` benchmark down into per-phase benchmarks, mirroring the `intmul/phases` breakdown, timing each of the five shift-reduction prover phase functions individually:

- **phase 1:** `build_g_parts`, `build_h_parts`, `run_phase_1_sumcheck`
- **phase 2:** `build_monster_multilinear`, `run_sumcheck`

## Benchmark structure

New `bench_shift_phases` in `benches/shift_reduction.rs` (added to the `criterion_group!`). The five phases are sequential and stateful, so — as `bench_intmul_phases` does — it builds the expensive circuit / witness / key-collection **once**, runs phase 1 once (throwaway transcript) to capture each phase's inputs (`g_parts`/`h_parts`, and phase 1's `r_j`/`r_s`/`gamma` → `r_j_witness`/`monster_multilinear`), then times each phase in its own `bench_function`, using `iter_batched` to clone only the by-value inputs (`run_phase_1_sumcheck`, `run_sumcheck`) and plain `iter` for the by-reference builders. One fixed size (4096-byte SHA256 message), `sample_size(10)`.

Sample run (`RUSTFLAGS="-C target-cpu=native" cargo bench -p binius-prover --bench shift_reduction`), one machine, indicative only:

| phase | time |
|---|---|
| phase1_build_g_parts | ~5.1 ms |
| phase1_build_h_parts | ~64 µs |
| phase1_run_sumcheck | ~450 µs |
| phase2_build_monster_multilinear | ~1.6 ms |
| phase2_run_sumcheck | ~0.5 ms |

(`build_g_parts` dominates phase 1; `build_monster_multilinear` dominates phase 2.)

## Visibility

To reach the phase functions from an integration bench (which sees only the public API), the `phase_1`/`phase_2`/`monster` modules are made `pub mod` and the exposed functions marked `#[doc(hidden)] pub` — **mirroring exactly the `#[doc(hidden)] pub` mechanism `intmul` already uses** for its phase methods (`protocols/intmul/prove.rs`). `#[doc(hidden)]` keeps them out of the docs and signals "not a stable API."

## Drive-by fix: the existing bench was broken

`bench_prove_and_verify` panicked on run: it computed the IntMul operator's `r_x_prime` via `strict_log_2(cs.mul_constraints.len()).unwrap()`, but SHA256 has **zero** MUL constraints, so `strict_log_2(0)` is `None`. (Benches aren't in CI, so this went unnoticed.) Fixed by synthesizing the IntMul operator as the **zero claim at an empty point** — four zero evals, empty `r_x_prime`, sharing BitAnd's `r_zhat_prime` — exactly as the real prover/verifier do in their `None` branch (`prove.rs` / `verify.rs`). The complete-run `prove`/`verify` benchmarks now run.

## Validation

- `cargo bench -p binius-prover --bench shift_reduction` — all benchmarks run and produce numbers (verified with a short criterion config); the previously-panicking `prove`/`verify` benches now work.
- `cargo clippy -p binius-prover --all-targets -- -D warnings`, nightly `fmt --check`, `typos` — clean.

### DIFF
--- a/crates/prover/src/protocols/shift/mod.rs
+++ b/crates/prover/src/protocols/shift/mod.rs
@@ -1,11 +1,18 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
 use binius_verifier::protocols::shift::{BITAND_ARITY, INTMUL_ARITY, SHIFT_VARIANT_COUNT};
 
 mod key_collection;
-mod monster;
-mod phase_1;
-mod phase_2;
+// `monster`, `phase_1`, and `phase_2` are internal implementation, exposed (via `#[doc(hidden)]`
+// `pub mod`) only so the `shift_reduction` benchmark can time individual phase functions (see
+// `benches/shift_reduction.rs`). Not a stable API.
+#[doc(hidden)]
+pub mod monster;
+#[doc(hidden)]
+pub mod phase_1;
+#[doc(hidden)]
+pub mod phase_2;
 mod prove;
 
 pub use key_collection::{KeyCollection, build_key_collection};
--- a/crates/prover/benches/shift_reduction.rs
+++ b/crates/prover/benches/shift_reduction.rs
@@ -1,4 +1,5 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
 use binius_circuits::sha256::Sha256;
 use binius_core::{ValueVec, constraint_system::ConstraintSystem};
@@ -56,7 +57,7 @@ pub fn create_sha256_cs_with_witness(
 }
 
 fn bench_prove_and_verify(c: &mut Criterion) {
-	use binius_field::{BinaryField128bGhash, PackedBinaryGhash1x128b, Random};
+	use binius_field::{BinaryField128bGhash, Field, PackedBinaryGhash1x128b, Random};
 	type F = BinaryField128bGhash;
 	type P = PackedBinaryGhash1x128b;
 	let mut rng = rand::rng();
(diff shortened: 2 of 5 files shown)


crates/prover/src/protocols/shift/mod.rs
```diff
@@ -1,11 +1,16 @@
...
 mod key_collection;
-mod monster;
-mod phase_1;
-mod phase_2;
+// `monster`, `phase_1`, and `phase_2` are internal implementation, exposed only so the
+// `shift_reduction` benchmark can time individual phase functions (see
+// `benches/shift_reduction.rs`). The bench-exposed functions are marked `#[doc(hidden)]`; they are
+// not a stable API.
+pub mod monster;
+pub mod phase_1;
```
jimpo, line 13: put doc(hidden) on the modules themselves, not on the functions inside them

jimpo (changes requested):
The circuit has no

crates/prover/src/protocols/shift/mod.rs
benji-potamus: Done in 2c82be6 — moved `#[doc(hidden)]` onto the `phase_1`/`phase_2`/`monster` module declarations and removed it from the individual functions (the module-level attribute hides the whole module from docs). The functions stay `pub` so the bench can still reach them; clippy `--all-targets`, fmt, and the bench build are green.

(Your review summary looked truncated — it just says "The circuit has no ". If there's more you meant re: the SHA256 circuit having no MUL constraints / the IntMul zero-claim, let me know and I'll address it.)

jimpo (approved):

PR #1707, opened 2026-07-05
BINIUS-145: split KeyCollection into public and non-public segments

benji-potamus:
Mechanical prep for splitting the public segment out of the committed trace ([BINIUS-145](https://linear.app/jimpo/issue/BINIUS-145/handling-constants), PR 1 of the sequence in the [plan](https://linear.app/jimpo/issue/BINIUS-145/handling-constants#comment-c759bc33)). Proofs are byte-identical: verified by comparing serialized transcripts (plain and ZK, fixed seeds) for the SHA-256 test circuit before and after the change.

**Commit 1 — segment word-count accessors on `ValueVecLayout`.** Adds `n_public_words` / `log_public_words` / `n_non_public_words` and replaces the hand-rolled `log2_ceil(offset_witness)` derivations in `Verifier::setup` and `ZKVerifier::setup`. (The virtual-address index translation lands with PR 2, where it is first used.)

**Commit 2 — `KeyCollection` becomes two `KeySegment`s.** One segment for the public words, one for the non-public committed words, each self-contained (`keys` / `key_ranges` / `constraint_indices`, segment-relative indices). The phases still iterate both segments in absolute value-vector order: `build_g_parts` chains the two per-segment zips into one fold, and `build_monster_multilinear` maps chunk indices over the concatenated word range so the packing is unchanged at every position. `KeyCollection` serialization version bumps to 2. Naming follows `ValueVec::public()` / `non_public()` rather than "committed", which would collide with `committed_total_len` still including the public prefix until the PR 3 layout reshape.

Both commits build and pass tests and clippy independently.

### DIFF
--- a/crates/core/src/constraint_system/layout.rs
+++ b/crates/core/src/constraint_system/layout.rs
@@ -1,4 +1,5 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 use binius_utils::serialization::{DeserializeBytes, SerializationError, SerializeBytes};
 use bytes::{Buf, BufMut};
 
@@ -51,6 +52,25 @@ impl ValueVecLayout {
 		}
 
 		Ok(())
+	}
+
+	/// Returns the number of words in the public segment: the constants and inout values,
+	/// including padding up to the power-of-two segment length.
+	pub const fn n_public_words(&self) -> usize {
+		self.offset_witness
+	}
+
+	/// Returns the base-2 logarithm of the public segment length in words.
+	///
+	/// [`Self::validate`] guarantees that the public segment length is a power of two.
+	pub const fn log_public_words(&self) -> usize {
+		self.offset_witness.trailing_zeros() as usize
+	}
+
+	/// Returns the number of non-public words: the witness and internal values, including
+	/// padding up to `committed_total_len`.
+	pub const fn n_non_public_words(&self) -> usize {
+		self.committed_total_len - self.offset_witness
 	}
 
 	/// Returns true if the given index points to an area that is considered to be padding.
--- a/crates/prover/src/protocols/shift/key_collection.rs
+++ b/crates/prover/src/protocols/shift/key_collection.rs
@@ -1,4 +1,5 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
 use std::{iter, mem, ops::Range};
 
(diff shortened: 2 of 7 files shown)


crates/core/src/constraint_system/layout.rs
```diff
@@ -53,6 +54,25 @@ impl ValueVecLayout {
...
+	/// Returns the base-2 logarithm of the public segment length in words.
+	///
+	/// [`Self::validate`] guarantees that the public segment length is a power of two.
+	pub const fn log_public_words(&self) -> usize {
+		self.offset_witness.trailing_zeros() as usize
+	}
+
+	/// Returns the number of non-public words: the witness and internal values, including
+	/// padding up to `committed_total_len`.
+	pub const fn n_non_public_words(&self) -> usize {
```
jimpo, line 72: n_witness_words

crates/prover/src/protocols/shift/key_collection.rs
```diff
@@ -168,32 +169,66 @@ impl Key {
...
+///
+/// The keys are split by value-vector segment: one [`KeySegment`] for the public words
+/// (value-vector indices `[0, n_public_words)`) and one for the non-public committed words
+/// (indices `[n_public_words, committed_total_len)`). Word indices within each segment are
+/// segment-relative. The split prepares for committing the two segments separately
+/// (BINIUS-145); the phases still iterate both segments in absolute value-vector order.
+#[derive(Debug, Clone)]
+pub struct KeyCollection {
+	pub public: KeySegment,
+	pub non_public: KeySegment,
```
jimpo, line 222: non_public -> witness

jimpo (approved):
Just a naming change. To do as a follow-up.

crates/core/src/constraint_system/layout.rs
benji-potamus: Landing in PR 3 (the layout reshape, stacked on #1708): the accessor is gone entirely there — `committed_total_len` becomes a stored `n_witness_words` field that counts the witness+internal words without the public segment, so the quantity carries this name everywhere.

crates/prover/src/protocols/shift/key_collection.rs
benji-potamus: Renamed to `KeyCollection { public, witness }` — lands in PR 3 (stacked on #1708) together with the `n_witness_words` layout rename, so the segment naming is consistent across layout, key collection, and the phase code.

PR #1708, opened 2026-07-05
BINIUS-145: commit only the hidden segment, splitting the witness claim

benji-potamus:
The protocol change for [BINIUS-145](https://linear.app/jimpo/issue/BINIUS-145/handling-constants) (PR 2 of the [plan](https://linear.app/jimpo/issue/BINIUS-145/handling-constants#comment-c759bc33), stacked on #1707): the trace oracle commits only the witness's **hidden segment**; the **public segment** (constants and inout values) enters the protocol as a verifier-known polynomial. `IOPProver::prove` / `IOPVerifier::verify` signatures are unchanged.

**Protocol.** The witness spans `log_witness_words + 1` word-index variables: the public segment at the base of the low half-cube, the hidden segment at the base of the high half-cube, selected by the top variable (`r_segment`) — which the existing sumcheck binds in round 1, so the planned sparse special first round (PR 4) needs no transcript change. The layout now guarantees the hidden segment is at least as long as the public segment (the frontend allocator pads), so `layout.log_witness_words() >= log_public_words()` with no `max`. As discussed on the issue, the claim splits **before** ring-switch: the prover derives and sends the hidden-segment (`witness_eval`) value, and the verifier reconstructs the full witness evaluation via `extrapolate_line(eq_ind_zero(..) * public_eval, witness_eval, r_segment)`, computing `public_eval` from its **own** public words via the same `compute_public_value` mechanism as the monster evaluation. Ring-switch and BaseFold then run unchanged on the witness claim.

**The standalone public input check is deleted, not moved** — `eval_pubcheck_eq`, the batching in `verify.rs`, and `compute_batched_transparent` on the prover side. Its binding role is subsumed: a prover using different public values fails the reconstructed final check w.h.p. ZK masks now cover only the committed oracle (specs track automatically via `OracleSetupChannel`).

**Commit 1** adds `ValueVecLayout::log_witness_words` with the validate/allocator guarantee and the bit-level public-words MLE evaluation (`evaluate_words_mle`) with unit tests. **Commit 2** is the protocol change across prover and verifier, plus bench and test updates.

**Proof sizes** (SHA-256 compress test circuit, `log_inv_rate=1`, verification round-tripped): plain 43,408 → 35,936 bytes (−17.2%); ZK 362,544 → 355,120 bytes (−2.0%). The phase-2 prover domain temporarily grows to `2^(log_witness_words + 1)` (dense buffers); PR 4 removes that cost transcript-identically.

### DIFF
--- a/crates/prover/src/protocols/shift/key_collection.rs
+++ b/crates/prover/src/protocols/shift/key_collection.rs
@@ -11,7 +11,10 @@ use binius_core::{
 	consts::LOG_WORD_SIZE_BITS,
 };
 use binius_field::Field;
-use binius_utils::serialization::{DeserializeBytes, SerializationError, SerializeBytes};
+use binius_utils::{
+	checked_arithmetics::log2_ceil_usize,
+	serialization::{DeserializeBytes, SerializationError, SerializeBytes},
+};
 use bytes::{Buf, BufMut};
 
 use super::{BITAND_ARITY, INTMUL_ARITY, PreparedOperatorData};
@@ -212,20 +215,30 @@ impl KeySegment {
 /// A collection of keys that organizes the prover's view of the constraint system.
 ///
 /// The keys are split by value-vector segment: one [`KeySegment`] for the public words
-/// (value-vector indices `[0, n_public_words)`) and one for the non-public committed words
-/// (indices `[n_public_words, committed_total_len)`). Word indices within each segment are
-/// segment-relative. The split prepares for committing the two segments separately
-/// (BINIUS-145); the phases still iterate both segments in absolute value-vector order.
+/// (value-vector indices `[0, n_public_words)`) and one for the hidden words (indices
+/// `[n_public_words, committed_total_len)`). Word indices within each segment are
+/// segment-relative. The phases iterate both segments in absolute value-vector order.
 #[derive(Debug, Clone)]
 pub struct KeyCollection {
 	pub public: KeySegment,
-	pub non_public: KeySegment,
+	pub hidden: KeySegment,
 }
 
 impl KeyCollection {
 	/// The total number of words covered by both segments.
 	pub const fn n_words(&self) -> usize {
-		self.public.n_words() + self.non_public.n_words()
+		self.public.n_words() + self.hidden.n_words()
+	}
+
+	/// The base-2 logarithm of the hidden segment length in words, rounded up to a power of
+	/// two.
+	///
+	/// Matches [`ValueVecLayout::log_witness_words`] for the layout the collection was built
+	/// from; the layout guarantees this is at least the public segment's logarithm.
+	///
+	/// [`ValueVecLayout::log_witness_words`]: binius_core::constraint_system::ValueVecLayout::log_witness_words
+	pub const fn log_witness_words(&self) -> usize {
+		log2_ceil_usize(self.hidden.n_words())
 	}
 }
 
@@ -340,10 +353,10 @@ pub fn build_key_collection(cs: &ConstraintSystem) -> KeyCollection {
 
 	// Split the builder keys lists at the public segment boundary and build one `KeySegment`
 	// per half.
-	let non_public_lists = builder_key_lists.split_off(cs.value_vec_layout.n_public_words());
+	let hidden_lists = builder_key_lists.split_off(cs.value_vec_layout.n_public_words());
 	KeyCollection {
 		public: build_key_segment(builder_key_lists),
-		non_public: build_key_segment(non_public_lists),
+		hidden: build_key_segment(hidden_lists),
 	}
 }
 
@@ -500,7 +513,7 @@ impl SerializeBytes for KeyCollection {
 		VERSION.serialize(&mut write_buf)?;
 
 		self.public.serialize(&mut write_buf)?;
-		self.non_public.serialize(write_buf)
+		self.hidden.serialize(write_buf)
 	}
 }
 
@@ -515,9 +528,9 @@ impl DeserializeBytes for KeyCollection {
 		}
 
 		let public = KeySegment::deserialize(&mut read_buf)?;
-		let non_public = KeySegment::deserialize(read_buf)?;
+		let hidden = KeySegment::deserialize(read_buf)?;
 
-		Ok(KeyCollection { public, non_public })
+		Ok(KeyCollection { public, hidden })
 	}
 }
 
--- a/crates/prover/src/protocols/shift/phase_2.rs
+++ b/crates/prover/src/protocols/shift/phase_2.rs
@@ -10,8 +10,12 @@ use binius_ip_prover::{
 		ProveSingleOutput, bivariate_product::BivariateProductSumcheckProver, prove_single,
 	},
 };
-use binius_math::{FieldBuffer, multilinear::eq::eq_ind_partial_eval};
-use binius_verifier::config::LOG_WORD_SIZE_BITS;
+use binius_math::{
+	FieldBuffer,
+	multilinear::eq::{eq_ind_partial_eval, eq_ind_zero},
+};
+use binius_utils::checked_arithmetics::checked_log_2;
+use binius_verifier::{config::LOG_WORD_SIZE_BITS, protocols::shift::evaluate_words_mle};
 use tracing::instrument;
 
 use super::{
@@ -27,7 +31,7 @@ use crate::fold_word::fold_words;
 ///
 /// # Protocol Steps
 /// 1. **Challenge Splitting**: Splits phase 1 challenges into `r_j` and `r_s` components
-/// 2. **Witness Folding**: Folds the witness words using the `r_j` challenges
+/// 2. **Segment Folding**: Folds the public and hidden words using the `r_j` challenges
 /// 3. **Monster Multilinear Construction**: Builds the monster multilinear from key collection and
 ///    operator data
 /// 4. **Sumcheck Execution**: Runs bivariate product sumcheck to prove witness ×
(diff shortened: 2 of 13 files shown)


crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -71,13 +120,14 @@ pub struct VerifyOutput<F> {
...
 	/// Challenge point for the shift variables (length `LOG_WORD_SIZE_BITS`).
 	pub r_s: Vec<F>,
-	/// Challenge point for the word index variables (length `log_word_count`).
+	/// Challenge point for the virtual trace word index variables (length
+	/// `log_virtual_half_words + 1`; the top variable selects the segment).
 	pub r_y: Vec<F>,
 	/// Final evaluation claim from the second sumcheck.
 	eval: F,
-	/// The claimed witness evaluation at the challenge point.
+	/// The claimed evaluation of the committed-half multilinear at `(r_j, r_y[..m])`.
```
jimpo: witness eval is a perfectly good name. keep the name and comment. "witness" refers to the committed segment now.

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -21,7 +25,52 @@ use super::{
...
+///
+/// The virtual trace splits the value vector across two half-cubes selected by its top
+/// word-index variable: the public segment occupies the low half and the non-public committed
+/// words the high half, each zero-padded to `2^m` words where `m` is the returned value. The
+/// virtual trace thus has `m + 1` word-index variables.
+///
+/// The public segment length is a power of two of at least one full element
+/// ([`ValueVecLayout::validate`]), so the `max` also guarantees element alignment of both
+/// halves.
+pub fn log_virtual_half_words(layout: &ValueVecLayout) -> usize {
```
jimpo: I don't like this method name. It really should be `layout.log_witness_words`, which should be guaranteed to be >= log_public_words by the layout.

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -71,13 +120,14 @@ pub struct VerifyOutput<F> {
 	pub r_j: Vec<F>,
 	/// Challenge point for the shift variables (length `LOG_WORD_SIZE_BITS`).
 	pub r_s: Vec<F>,
-	/// Challenge point for the word index variables (length `log_word_count`).
+	/// Challenge point for the virtual trace word index variables (length
```
jimpo: Maybe it'll be cleaner to simply split that one challenge off onto a different field. Like `r_segment: F`.

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -266,17 +324,55 @@ where
...
+		channel.compute_public_value(&inputs, PublicWordsEvalFn { public })
+	};
+
+	// Reconstruct the virtual trace evaluation from its two halves. The public half is
+	// zero-padded from `2^k` to `2^m` words, contributing the `(1 - r_y_i)` padding factors.
+	let one = C::Elem::one();
+	let log_half = log_virtual_half_words(&constraint_system.value_vec_layout);
+	let (r_y_pad, r_top) = r_y[log_public_words..].split_at(log_half - log_public_words);
+	let r_top = r_top[0].clone();
+	let padded_public_eval = r_y_pad
```
jimpo: eq_ind_zero helper

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -266,17 +324,55 @@ where
...
+		.iter()
+		.fold(public_eval, |acc, r| acc * (one.clone() - r));
+	let trace_eval = (one - &r_top) * padded_public_eval + r_top * committed_eval.clone();
```
jimpo: extrapolate_line helper

crates/prover/src/protocols/shift/phase_2.rs
```diff
@@ -112,25 +152,32 @@ pub fn run_sumcheck<F: Field, P: PackedField<Scalar = F>, Channel: IPProverChann
...
+	#[cfg(not(debug_assertions))]
+	let _ = multilinear_evals;
+
+	// Send the committed-half evaluation at `r_y[..log_half]`. The committed half is zero-padded
+	// from the folded length up to the half size, contributing the `(1 - r)` padding factors.
+	let log_half = r_y.len() - 1;
+	let (r_y_low, r_y_pad) = r_y[..log_half].split_at(committed_folded.log_len());
+	let committed_eval = r_y_pad
+		.iter()
+		.fold(evaluate(committed_folded, r_y_low), |acc, &r_y_i| acc * (F::ONE - r_y_i));
```
jimpo: Don't do this re-evaluation. Evaluate the public segment instead (like the verifier does), which is cheaper. Then subtract it off and scale to get the witness eval.

jimpo (changes requested):

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Restored — the field is `witness_eval` again with the original doc comment, and the same naming now runs through the prover side.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done: `ValueVecLayout::log_witness_words()` (just `log2_ceil(n_non_public_words)`), with `validate()` now guaranteeing the witness segment is at least as long as the public segment and the frontend allocator padding to match — so no `max` anywhere. `log_virtual_half_words` is gone; `IOPVerifier::log_witness_elems` and the shift sumcheck size both derive from the layout method.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done — `VerifyOutput` now carries `r_y` (length `log_witness_words`) plus a separate `r_segment: F`, which also removed the `[..len - 1]` slicing at the ring-switch call site.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Using `eq_ind_zero` now (and on the prover side of the derivation too).

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Using `extrapolate_line` now.

crates/prover/src/protocols/shift/phase_2.rs
benji-potamus: Done — the prover now evaluates the public segment with the same `evaluate_words_mle` the verifier uses, subtracts its padded contribution off the sumcheck's trace evaluation, and scales by `r_segment⁻¹` to get `witness_eval`. The O(2^m) re-evaluation is gone (this also drops the buffer that was kept alive for it). Transcript value is unchanged; noted in a comment that completeness now fails with negligible probability when the segment challenge samples to zero.

crates/verifier/src/verify.rs
```diff
@@ -264,30 +261,19 @@ impl IOPVerifier {
 		)
 		.entered();
 
-		// Ring-switching verification
+		// Ring-switching verification of the witness claim; the segment selector challenge was
+		// already consumed by `shift::check_eval`.
```
jimpo:
```suggestion
		// Ring-switching verification of the witness claim.
```

crates/prover/src/protocols/shift/phase_2.rs
```diff
@@ -68,41 +72,80 @@ where
...
-	run_sumcheck(r_j_witness, monster_multilinear, r_j, gamma, channel)
+	run_sumcheck(virtual_folded, public_words, monster_multilinear, r_j, gamma, channel)
+}
+
+/// Assembles the folded virtual trace from its two folded segments.
+///
+/// Each segment sits at the base of its half-cube, zero-padded up to `2^log_half` entries. The
+/// interim dense representation materializes the mostly-zero low half; a sparse special first
+/// sumcheck round can remove this cost without changing the transcript.
+pub fn assemble_virtual_trace<F: Field, P: PackedField<Scalar = F>>(
```
jimpo: implement this better, using memcpy s and as_ref/as_mut on FieldBuffer. You can use split_half_mut on the output buffer

crates/prover/src/protocols/shift/key_collection.rs
```diff
@@ -227,6 +230,17 @@ impl KeyCollection {
 	pub const fn n_words(&self) -> usize {
 		self.public.n_words() + self.non_public.n_words()
```
jimpo: rename self.non_public to self.hidden

jimpo (changes requested):
We need to get more consistent with the naming. We can refer to the "witness" as consisting of a "public" segment (itself having constants and inout values) and a "hidden" segment.

So all this "virtual trace" stuff -- it's just the "witness".

crates/verifier/src/verify.rs
benji-potamus: Applied verbatim.

crates/prover/src/protocols/shift/phase_2.rs
benji-potamus: Rewritten with `split_half_mut` + `copy_from_slice` on the packed slices (the trailing partial packed element of each segment carries zero high lanes, which is exactly the padding the destination wants, so whole-element memcpys are correct).

crates/prover/src/protocols/shift/key_collection.rs
benji-potamus: Renamed to `hidden`, and swept the witness/public/hidden glossary through the rest of the PR: all "virtual trace" wording is now just the witness (with public and hidden segments), the layout error is `HiddenSegmentTooShort`, and variables follow (`hidden_folded`, `hidden_words`, …). Kept the names you'd blessed explicitly: `witness_eval`, `layout.log_witness_words`, `r_segment`.

jimpo (approved):

benji-potamus:
@jimpo the REST merge endpoint refuses with 405 ("Changes must be made through the merge queue") and nothing lands in the queue — I can't enqueue from this side. Could you hit "Merge when ready"? Everything is green: your approval on the current head, CI 8/8, and the stacked PR 3 (layout reshape) + PR 4 (sparse first round) are verified and ready to go up as soon as this lands.

PR #1713, opened 2026-07-06
Replace Underlier::zero() with a ZERO associated constant

tcoratger:
Turns `Underlier::zero()` in `binius-arith-bench` into a `const ZERO: Self`.
Pure API cleanup — no behavior change, codegen is identical.

### The problem

The `Underlier` trait in `binius-arith-bench` exposed zero as a method:

```rust
fn zero() -> Self;
```

But the core `UnderlierType` trait in `binius-field` already exposes it as a constant (`const ZERO: Self`).
So arith-bench was the odd one out, and zero — which is a compile-time constant for every underlier — was hiding behind a function call.

### The idea

Make it a `const`, matching `binius-field` and `std` (`i32::MAX`, `f32::INFINITY`):

```rust
const ZERO: Self;
```

This reads better, sits right next to the existing `const BITS`, and — unlike a method — is usable in const contexts like `[T::ZERO; N]` and `static` initializers.

### The change

The scalar and generic impls are trivial constants:

```diff
- fn zero() -> Self { 0 }                          // native uints
+ const ZERO: Self = 0;

- fn zero() -> Self { std::array::from_fn(|_| U::zero()) }   // [U; N]
+ const ZERO: Self = [U::ZERO; N];
```

The SIMD impls needed a little care.
`_mm*_setzero_*` and `vdupq_n_u64` are `unsafe` and **not** `const fn`, so they cannot seed a constant.
Following the pattern already used in `binius-field`, the SIMD constants are built by const-transmuting zeroed bytes, which *is* const-stable:

```diff
- fn zero() -> Self { unsafe { _mm256_setzero_si256() } }
+ const ZERO: Self = unsafe { std::mem::transmute([0u128; 2]) };
```

All-zero bytes are a valid bit pattern for every SIMD register, and lane order is irrelevant for zero, so this is correct for `__m128i` / `__m256i` / `__m512i` (x86_64) and `uint64x2_t` / `poly64x2_t` (NEON).

### Scope

- **changed:** the `Underlier` trait and all its impls — native uints, `[U; N]`, x86_64 SIMD, NEON SIMD.
- **updated call sites:** `U::zero()` → `U::ZERO` in `test_utils.rs`, `ghash/{clmul,bit_sliced}.rs`, `rijndael/russian_peasant.rs`, `benches/field_mul.rs`, and the aarch64 test.
- 8 files, +27 / −50.

### Verification

- `cargo clippy -p binius-arith-bench --all-targets` — clean.
- `cargo +nightly fmt --all` — clean.
- `cargo test -p binius-arith-bench` (native aarch64 / NEON) — 63 tests + 1 doctest pass.
- The x86_64 SIMD const-transmutes were cross-checked with `--target x86_64-apple-darwin` under `+avx2,+avx512f,+avx512bw` (those impls are feature-gated off on the host) — `check` and `clippy` both clean.

### DIFF
--- a/crates/arith-bench/src/arch/x86_64.rs
+++ b/crates/arith-bench/src/arch/x86_64.rs
@@ -9,6 +9,8 @@ use crate::underlier::{PackedUnderlier, Underlier};
 #[cfg(target_feature = "sse2")]
 impl Underlier for __m128i {
 	const BITS: usize = 128;
+	// Safety: all-zero bytes are a valid bit pattern for every 128-bit SIMD register.
+	const ZERO: Self = unsafe { std::mem::transmute(0u128) };
 
 	#[inline]
 	fn and(a: Self, b: Self) -> Self {
@@ -18,11 +20,6 @@ impl Underlier for __m128i {
 	#[inline]
 	fn xor(a: Self, b: Self) -> Self {
 		unsafe { _mm_xor_si128(a, b) }
-	}
-
-	#[inline]
-	fn zero() -> Self {
-		unsafe { _mm_setzero_si128() }
 	}
 
 	#[inline]
@@ -216,6 +213,8 @@ impl crate::underlier::OpsClmul for __m128i {
 #[cfg(target_feature = "avx2")]
 impl Underlier for __m256i {
 	const BITS: usize = 256;
+	// Safety: all-zero bytes are a valid bit pattern for every 256-bit SIMD register.
+	const ZERO: Self = unsafe { std::mem::transmute([0u128; 2]) };
 
 	#[inline]
 	fn and(a: Self, b: Self) -> Self {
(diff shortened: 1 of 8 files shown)


crates/arith-bench/src/arch/x86_64.rs
```diff
@@ -216,6 +213,8 @@ impl crate::underlier::OpsClmul for __m128i {
 #[cfg(target_feature = "avx2")]
 impl Underlier for __m256i {
 	const BITS: usize = 256;
+	// Safety: all-zero bytes are a valid bit pattern for every 256-bit SIMD register.
+	const ZERO: Self = unsafe { std::mem::transmute([0u128; 2]) };
```
jimpo, line 217: Are you sure there's no alignment issue?

jimpo (commented):
Seems OK, but does this actually make any perf difference?

tcoratger:
> Seems OK, but does this actually make any perf difference?

No perf difference. They should compile to exactly same machine code but the ZERO const stuff is a way for this to be usable under const or static expressions, just this, the rest is unchanged :) This is something we do exactly the same in Plonky3, feel free to close if you prefer the function path

crates/arith-bench/src/arch/x86_64.rs
tcoratger: Yes for me there is no alignment issue and since this is compile time known, this part runs at compilation time and so an unsound transmute would fail const-eval no?

jimpo (approved):

PR #1716, opened 2026-07-06
Add parallel witness evaluation for independent components

latifkasuli:
## Summary

Adds an explicit `Circuit::populate_wire_witness_parallel` path that evaluates independent eval-form bytecode components in parallel. The default `populate_wire_witness` path remains sequential and unchanged at the API level.

The parallel path builds connected components over eval bytecode, checks that component writes are disjoint, and then runs those components with a shared raw `ValueStore`. The `unsafe` boundary is intentionally contained to the interpreter store. Soundness depends on two invariants:

- write-disjointness: checked at eval-form build time over outputs, aux wires, and scratch wires
- read-closure: established by a local def-map inside `gate_component_roots`, not by pass-populated `GateGraph::wire_def`, so this remains correct even when optional passes such as DCE are disabled

This also adds a Criterion benchmark group for independent hash circuits that measures `populate_witness + populate_wire_witness_parallel + prove`.

## Benchmarks

Fresh Criterion sweep on Apple M4 Pro, `RAYON_NUM_THREADS=8`, `N_PRIMITIVES=4096`, `LOG_INV_RATE=1`, `--features rayon`, generic `aarch64-apple-darwin` target.

| Primitive | Witness + prove | Parallel witness + prove | Delta |
|---|---:|---:|---:|
| SHA-256 compressions | 218.43 ms | 192.27 ms | -12.0% |
| BLAKE3 compressions | 140.99 ms | 134.25 ms | -4.8% |
| Keccak-f[1600] permutations | 264.15 ms | 228.14 ms | -13.6% |

BLAKE3 was rerun because the first pass was noisy; two follow-up runs clustered around a smaller ~5% win. Proof size at these settings is unchanged by this PR: 352992 bytes (344.72 KiB) for all three independent hash benches.

## Safety / Correctness Notes

- The default sequential witness path is still available and used unless callers opt into `populate_wire_witness_parallel`.
- If the eval form has <=1 component or component writes are not disjoint, the parallel path falls back to sequential evaluation.
- The local def-map avoids relying on `GateGraph::wire_def`, which is only populated by optional compiler passes.
- Tests compare sequential and parallel full `ValueVec` backing storage, including scratch space.
- A DCE-disabled regression covers the original failure mode under `BINIUS_DISABLE_DCE=1`.

## Test Plan

- `cargo test -p binius-frontend`
- `cargo test -p binius-frontend --features rayon`
- `BINIUS_DISABLE_DCE=1 RAYON_NUM_THREADS=8 cargo test -p binius-frontend --features rayon populate_wire_witness_parallel_matches_sequential_when_dce_is_disabled -- --nocapture`
- `cargo check --workspace`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-frontend --features rayon --document-private-items --no-deps`
- `cargo test -p binius-examples --features rayon independent_hashes --lib`
- `cargo bench -p binius-examples --features rayon --bench sha256_compressions --no-run`
- `cargo bench -p binius-examples --features rayon --bench blake3_compressions --no-run`
- `cargo bench -p binius-examples --features rayon --bench keccak_permutations --no-run`
- Scale smoke under `BINIUS_DISABLE_DCE=1 RAYON_NUM_THREADS=8 N_PRIMITIVES=4096 RUNS=10`: SHA-256 and Keccak both passed 10/10 sequential-vs-parallel full-store equality plus constraint verification.

### DIFF
--- a/crates/core/src/constraint_system/value_vec.rs
+++ b/crates/core/src/constraint_system/value_vec.rs
@@ -143,6 +143,11 @@ impl ValueVec {
 		&self.data[self.layout.offset_witness..self.layout.committed_total_len]
 	}
 
+	/// Returns the full mutable backing store, including scratch space.
+	pub fn as_mut_slice(&mut self) -> &mut [Word] {
+		&mut self.data
+	}
+
 	/// Returns the witness portion of the values vector.
 	pub fn witness(&self) -> &[Word] {
 		let start = self.layout.offset_witness;
--- a/crates/examples/Cargo.toml
+++ b/crates/examples/Cargo.toml
@@ -86,7 +86,7 @@ harness = false
 [features]
 default = []
 perfetto = ["tracing-profile/perfetto"]
-rayon = ["binius-utils/rayon"]
+rayon = ["binius-frontend/rayon", "binius-utils/rayon"]
 
 # Tutorial examples
 [[example]]
--- a/crates/examples/benches/utils/independent_hashes.rs
+++ b/crates/examples/benches/utils/independent_hashes.rs
@@ -133,13 +133,24 @@ pub fn run_independent_hash_benchmark<E>(
 		benchmark,
 		primitive.group_prefix(),
 		peak_alloc,
-		bench_witness_generation_and_proving,
+		bench_independent_hash_extra_groups,
 	);
+}
+
+fn bench_independent_hash_extra_groups<B>(
+	c: &mut Criterion,
+	ctx: ConstraintSystemBenchmarkContext<'_, B>,
+) where
+	B: ExampleBenchmark,
+{
+	bench_witness_generation_and_proving(c, &ctx);
+	#[cfg(feature = "rayon")]
+	bench_parallel_witness_generation_and_proving(c, &ctx);
 }
 
 fn bench_witness_generation_and_proving<B>(
 	c: &mut Criterion,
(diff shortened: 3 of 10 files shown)


jimpo:
Thanks for the PR, but we're going to handle witness generation parallelization in a different way, based on a clearer 2D witness abstraction. See https://github.com/binius-zk/binius64/pull/1739.

If you would like to take it, one follow-on issue from that is to parallelize witness gen over the ValueTable2 (it'll become the canonical ValueTable soon) using the StridedArray2D abstraction. We basically want to slice the table up into vertical slices, where the width of each one is some sweet spot and run the batch interpreter over these slices in parallel. My thinking is that the sweet spot is something like 256 words. 256 words * 8 bytes per word * 4 operand gate = 8 KiB, which fits easily in L1. Could even be 512 or 1024 words, but there's probably diminishing returns to longer strides at that point.

@latifkasuli Do you want to take this? I could write it up as an issue if helpful.

latifkasuli:
I'll take the `ValueTable2` follow-up.

This PR's eval-form component parallelization is superseded by the `ValueTable2` / `StridedArray2D` direction from #1739, so I'm closing it rather than continuing on this path. I'll move the work to a new PR that parallelizes the batch witness-generation path by slicing `ValueTable2` into vertical stripes and benchmarking stripe widths around the suggested 256-word point (with 512/1024 as comparison points).

PR #1717, opened 2026-07-06
Experiment: alternate build_g_parts over a WiringInfo sparse-matrix layout

benji-potamus:
Implements the BINIUS-228 experiment: an alternate `build_g_parts` for shift-reduction phase 1, built on a new `WiringInfo` sparse-matrix layout, kept alongside the existing `KeyCollection` path and benchmarked head-to-head.

Linear: https://linear.app/jimpo/issue/BINIUS-228

## Result (benchmark decides)

Per-phase bench (`shift_reduction_phases`, 4096-byte SHA256 message, `PackedBinaryGhash1x128b`), two runs:

| variant | time |
|---|---|
| `phase1_build_g_parts` (KeyCollection, current) | ~5.2 ms |
| `phase1_build_g_parts_wiring` (this PR) | ~15.4 ms |

**The wiring path is ~2.9× slower here.** The predicted cost — a bit-scatter per *wiring entry* (per constraint reference), versus one aggregated scatter per `(word, variant, amount)` in the KeyCollection path — dominates the smaller-working-set win. See the Linear thread for the mechanism and a load-balance hypothesis.

Opening as **draft** per the experiment lane: the code is complete, proptested, and benchmarked, but since it loses on this workload the adoption / retirement call is yours. Happy to cancel, keep it as a reference sibling, or iterate (e.g. on the parallel load-balance angle).

## Commits

1. **`WiringInfo` sparse-matrix layout + `build_wiring_info`** — one `WiringInfo` per value-vector segment (public/hidden), indexed `[operand][shift_variant][shift_amount]`, mirroring the `KeyCollection` split.
2. **`build_g_parts_wiring` alternate algorithm** — parallelizes over the 512 `(variant, amount)` output rows with no shared state, plus a proptest asserting byte-identical output to `build_g_parts` across three `P` widths.
3. **Benchmark** — `phase1_build_g_parts_wiring` alongside `phase1_build_g_parts`.

## Note on the premise

The issue references the `KeyCollection { public, hidden }` split from BINIUS-145; that has since merged to `main`, so `WiringInfo` mirrors it exactly (one per segment, segment-relative `word_index`) as the issue describes.

### DIFF
--- a/crates/prover/src/protocols/shift/phase_1.rs
+++ b/crates/prover/src/protocols/shift/phase_1.rs
@@ -4,7 +4,9 @@
 use std::{iter, ops::Range};
 
 use binius_core::word::Word;
-use binius_field::{AESTowerField8b, BinaryField, Field, PackedField};
+use binius_field::{
+	AESTowerField8b, BinaryField, Field, PackedField, UnderlierType, WithUnderlier,
+};
 use binius_ip::sumcheck::{SumcheckOutput, common::RoundCoeffs};
 use binius_ip_prover::{
 	channel::IPProverChannel,
@@ -24,6 +26,7 @@ use super::{
 	key_collection::{KeyCollection, Operation},
 	monster::build_h_parts,
 	prove::PreparedOperatorData,
+	wiring::{WiringCollection, WiringEntry, WiringInfo},
 };
 
 // This is the number of variables in the g (and h) multilinears of phase 1.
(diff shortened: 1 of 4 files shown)


benji-potamus:
## Benchmark: `build_g_parts` head-to-head

**What was measured:** `cargo bench -p binius-prover --bench shift_reduction -- phase1_build_g_parts`, the `shift_reduction_phases` group. Single circuit: SHA256 over a 4096-byte message (`LOG_MESSAGE_LEN_BYTES = 12`), `F = BinaryField128bGhash`, `P = PackedBinaryGhash1x128b`. `RUSTFLAGS="-C target-cpu=native"`, criterion `sample_size(10)`. Two independent runs.

| variant | run 1 (median) | run 2 (median) |
|---|---|---|
| `phase1_build_g_parts` (KeyCollection, current) | 5.26 ms | 5.18 ms |
| `phase1_build_g_parts_wiring` (this PR) | 15.11 ms | 15.39 ms |

**Delta: the wiring path is ≈ +10 ms, ≈ 2.9× slower** on this circuit. Output is verified byte-identical to the current path by the new proptest (three `P` widths), so this is a pure performance comparison.

### Why it loses

The dominant factor is exactly the "Cost" the ticket predicted. The KeyCollection path aggregates *all* operand/constraint contributions for a `(word, variant, amount)` into a single scalar (`Key::accumulate`) and then does **one** bit-scatter over that word's 64 bits. The wiring path scatters **once per wiring entry** — i.e. once per constraint reference. For a word reused across *K* constraints under the same `(variant, amount)`, KeyCollection does ~one 64-bit scatter while the wiring path does ~*K*, and SHA256 reuses words heavily. The register-resident 64-scalar working set doesn't recover that.

A secondary suspect (not yet measured): parallelizing over the 512 `(variant, amount)` rows load-balances poorly on real circuits, which only populate a sparse subset of those rows (SHA256 uses a handful of variants × a couple dozen distinct amounts). Work concentrates in few rows → stragglers, whereas the KeyCollection path parallelizes over thousands of words. If the transpose is worth pursuing, re-parallelizing over non-empty rows (or a finer axis) is the lever to try.

### Recommendation

As implemented per the ticket spec, the wiring transpose does not win on SHA256, so I would **not** adopt it / retire the KeyCollection phase-1 path. The sibling + proptest + bench are in place if you want to keep them as a reference or iterate on the load-balance angle. Your call on merge vs. cancel.

crates/prover/src/protocols/shift/phase_1.rs
```diff
@@ -284,3 +285,289 @@ fn build_multilinear_parts<P: PackedField>(
...
+			let mut acc = [F::ZERO; WORD_SIZE_BITS];
+			for &WiringEntry {
+				constraint_index,
+				word_index,
+			} in matrix
+			{
+				let t = tensor[constraint_index as usize];
+				// Scatter `t` into every set bit of the word (field add is XOR).
+				let mut remaining = segment_words[word_index as usize].0;
+				while remaining != 0 {
```
jimpo: Fixed length iterate over WORD_SIZE_BITS. Use masking for constant time accumulation.

jimpo (changes requested):

jimpo:
Bump LOG_MESSAGE_LEN_BYTES to 14 in the benchmarks.

crates/prover/src/protocols/shift/phase_1.rs
benji-potamus: Done — pushed (71ae4ca0). The scatter now iterates the full `0..WORD_SIZE_BITS` and accumulates via an underlier mask: `acc[bit] += F::from_underlier(t.to_underlier() & F::Underlier::fill_with_bit((word >> bit) & 1))`, so the work is constant-time (independent of the word's set-bit pattern), no branch. Added `F: WithUnderlier` to the bound. The identity proptest still passes (byte-identical to `build_g_parts`, 3 `P` widths).

Heads up on perf: the constant-time version does 64 masked adds per wiring entry regardless of popcount, so on the SHA256 bench it's now ~24 ms (was ~15 ms with the data-dependent set-bit loop; KeyCollection is ~5 ms). So this makes the correctness/side-channel behavior clean but widens the gap vs. the KeyCollection path on this workload — flagging in case that changes your read on whether the transpose is worth pursuing.

benji-potamus:
Bumped `LOG_MESSAGE_LEN_BYTES` to 14 (16384-byte SHA256 message, ~4× the constraints of the previous size-12 run) and re-benched. Pushed in `5386d222`.

`shift_reduction_phases`, `PackedBinaryGhash1x128b`, `-C target-cpu=native`, constant-time masking scatter:

| variant | size 12 (4 KiB msg) | size 14 (16 KiB msg) |
|---|---|---|
| `phase1_build_g_parts` (KeyCollection) | ~5.2 ms | ~19.5 ms |
| `phase1_build_g_parts_wiring` (this PR) | ~24 ms | ~91.6 ms |
| ratio | ~4.6× | ~4.7× |

So the gap is essentially flat across sizes — scaling the circuit up doesn't help the transpose; both grow ~linearly in constraint count and the wiring path stays ~4.7× behind. (Numbers here are the constant-time masking version you asked for; the earlier data-dependent set-bit loop was ~2.9× at size 12.)

The load-balance hypothesis (parallelizing over the sparse set of populated `(variant, amount)` rows) is still the untested lever if you want the transpose to be competitive — that's option (c) from the Linear thread. Let me know which way you'd like to go.

jimpo:
OK, this is a failure. Closing.

benji-potamus:
Agreed — the transpose loses decisively (~4.7×) and doesn't close the gap at scale. Closing the PR and marking BINIUS-228 cancelled. The finding (why the per-reference scatter can't beat the aggregated one) is written up on the Linear issue for the record.

PR #1719, opened 2026-07-06
BINIUS-218: build the logUp* looker denominator in one packed pass

tcoratger:
Part of [BINIUS-218](https://linear.app/jimpo/issue/BINIUS-218). This is one of several witness inefficiencies tracked there, so it should **not** close the ticket on merge.

Builds the logUp* looker denominator in one packed pass instead of a scalar `Vec` plus a repack. Output is bit-identical — no protocol change.

### The problem

The looker denominator is $c - \iota(\text{index}[i])$, one entry per looker row over the $2^n$ cube.
Building it took two full-size steps on the dominant $n$-axis:

- compute every denominator scalar into a `Vec`,
- repack that `Vec` into a `FieldBuffer`, packing scalar-by-scalar.

So two `$2^n$` allocations and two passes, and the scalar `Vec` exists only to be immediately repacked and dropped.

### The idea

Unlike the numerator (a scaled equality indicator), this **cannot** be a broadcast: $\iota(\text{index}[i])$ is data-dependent per row.
But the intermediate scalar `Vec` is still avoidable.

Chunk the index into `P::WIDTH`-sized groups and pack each word as it is computed:

```
for each chunk of WIDTH rows:
    word = from_scalars( c - iota(index[i]) for i in chunk )
```

The buffer is filled directly, so compute and pack fuse into one pass with one allocation.

### Per looker denominator

- **before:** scalar `Vec` ($2^n$), then repack into a buffer ($2^n$) — 2 allocations, 2 passes
- **after:** one fused compute-and-pack pass — 1 allocation, 1 pass

→ one allocation and one pass removed.

### The change

```diff
- let values = index.iter().map(|&i| c - embed_position::<F>(i)).collect::<Vec<_>>();
- FieldBuffer::from_values(&values)
+ let log_len = index.len().ilog2() as usize;
+ let packed_len = 1 << log_len.saturating_sub(P::LOG_WIDTH);
+ let mut packed = Vec::with_capacity(packed_len);
+ packed.extend(index.chunks(P::WIDTH).map(|chunk| {
+     P::from_scalars(chunk.iter().map(|&i| c - embed_position::<F>(i)))
+ }));
+ FieldBuffer::new(log_len, packed.into_boxed_slice())
```

### Soundness

- The denominator is bit-identical to before, value for value.
- `from_scalars` fills up to `P::WIDTH` lanes from the iterator, leaving the rest zero — exactly what `from_values`' internal packing did.
- Same scalars, same chunk order, same padding, so the fractional-addition leaf and the transcript are unchanged.

### Scope

- **changed:** `looker_denominator` in `crates/ip-prover/src/logup_star/witness.rs`
- **left untouched:** the numerator, the pushforward, the table denominator, the verifier
- **not addressed here:** the fractional-addition layer footprint noted on BINIUS-218 — left for a follow-up under the same ticket

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-ip-prover --all-targets`, nightly `fmt --all --check` — clean
- `binius-ip-prover` + `binius-iop-prover` logUp* round-trips (5 + 12) — pass, so the end-to-end proof still verifies
- new proptest pins the packed build to a scalar reference over `n` below, at, and above the packing width, plus a deterministic single-row and four-row case

### Benchmark

Building the `$2^n$` denominator column:

| n_vars | baseline (Vec + repack) | direct packed build | speedup |
|---|---|---|---|
| 16 (65K) | 112.1 µs | 84.6 µs | 1.33× |
| 20 (1.0M) | 1.758 ms | 1.315 ms | 1.34× |
| 24 (16.8M) | 27.70 ms | 20.91 ms | 1.32× |

Throughput rises from ~585 Melem/s to ~790 Melem/s — a steady ~1.33×, consistent with dropping one of the two passes. Smaller than the numerator win, which removed two of three.

The bench is a throwaway used only to produce these numbers and is **not part of this PR**. To reproduce: drop the file below into `crates/math/benches/looker_denominator.rs`, add a matching `[[bench]]` entry (`name = "looker_denominator"`, `harness = false`) to `crates/math/Cargo.toml`, then run `cargo bench -p binius-math --bench looker_denominator`.

<details>

```rust

//! Before/after benchmark for the logUp* looker denominator `c - iota(index[i])`.
//!
//! Each row's denominator is the challenge shifted by the row's embedded index value.
//! Two ways to build the `2^n`-entry buffer are compared:
//!
//!     baseline: compute every scalar into a Vec, then repack it into a FieldBuffer
//!     direct  : compute and pack P::WIDTH scalars per word in one pass, no scalar Vec

use binius_field::{
    BinaryField, Divisible, PackedField,
    arch::{OptimalB128, OptimalPackedB128},
};
use binius_math::{FieldBuffer, test_utils::random_scalars};
use rand::prelude::*;

/// Embed a table position into the field, matching the protocol's `GF(2)`-linear embedding.
fn embed<F: BinaryField<Underlier: Divisible<u64>>>(j: usize) -> F {
    F::from_underlier(F::Underlier::from_iter(std::iter::once(j as u64)))
}

fn bench_looker_denominator(c: &mut Criterion) {
    type F = OptimalB128;
    type P = OptimalPackedB128;

    let mut group = c.benchmark_group("looker_denominator");

    // The logUp challenge, a general field element.
    let challenge = random_scalars::<F>(&mut rng, 1)[0];

    for n_vars in [16, 20, 24] {
        // Throughput is the number of denominators produced, one per looker row.
        let n_output_elems = 1u64 << n_vars;
        group.throughput(Throughput::Elements(n_output_elems));

        // One random index column, reused by both strategies.
        let index = (0..(1usize << n_vars))
            .map(|_| rng.random_range(0..(1usize << 12)))
            .collect::<Vec<_>>();

        // Baseline: scalar Vec, then repack.
        group.bench_function(BenchmarkId::new("baseline_vec_then_repack", n_vars), |b| {
            b.iter(|| {
                let values = index
                    .iter()
                    .map(|&i| challenge - embed::<F>(i))
                    .collect::<Vec<_>>();
                FieldBuffer::<P>::from_values(&values)
            });
        });

        // Direct: compute and pack in one pass.
        group.bench_function(BenchmarkId::new("direct_packed_build", n_vars), |b| {
            b.iter(|| {
                let log_len = index.len().ilog2() as usize;
                let packed_len = 1 << log_len.saturating_sub(P::LOG_WIDTH);
                let mut packed = Vec::with_capacity(packed_len);
                packed.extend(index.chunks(P::WIDTH).map(|chunk| {
                    P::from_scalars(chunk.iter().map(|&i| challenge - embed::<F>(i)))
                }));
                FieldBuffer::<P>::new(log_len, packed.into_boxed_slice())
            });
        });
    }

    group.finish();
}

criterion_group!(benches, bench_looker_denominator);
criterion_main!(benches);
```

</details>

### DIFF
--- a/crates/ip-prover/src/logup_star/witness.rs
+++ b/crates/ip-prover/src/logup_star/witness.rs
@@ -115,16 +115,13 @@ where
 	// n, the number of looker variables, from the 2^n rows.
 	let log_len = index.len().ilog2() as usize;
 
-	// One denominator per row: shift the challenge by the row's embedded index value.
-	// Pack P::WIDTH consecutive denominators into each word as they are computed.
-	// So the packed buffer is built in one pass, with no scalar buffer materialized first.
-	let packed_len = 1 << log_len.saturating_sub(P::LOG_WIDTH);
-	let mut packed = Vec::with_capacity(packed_len);
-	packed.extend(
-		index
-			.chunks(P::WIDTH)
-			.map(|chunk| P::from_scalars(chunk.iter().map(|&i| c - embed_position::<F>(i)))),
-	);
+	// One denominator per row: c minus the row's embedded index value.
+	// Subtract a full word at a time: one packed subtraction per word, built in parallel.
+	let c_packed = P::broadcast(c);
+	let packed = index
+		.par_chunks(P::WIDTH)
+		.map(|chunk| c_packed - P::from_scalars(chunk.iter().copied().map(embed_position::<F>)))
+		.collect::<Vec<_>>();
 
 	FieldBuffer::new(log_len, packed.into_boxed_slice())
 }
--- a/crates/ip-prover/src/logup_star/witness.rs
+++ b/crates/ip-prover/src/logup_star/witness.rs
@@ -103,17 +103,30 @@ where
 /// Build the looker denominator `c - I` over the `n`-variable looker cube.
 ///
 /// Entry `i` is `c - iota(index[i])`, the logUp denominator for looker row `i`.
+///
+/// # Preconditions
+///
+/// * `index.len()` is a power of two.
 pub fn looker_denominator<F, P>(c: F, index: &[usize]) -> FieldBuffer<P>
 where
 	F: BinaryField<Underlier: Divisible<u64>>,
 	P: PackedField<Scalar = F>,
 {
-	// One denominator per looker row: shift the challenge by the row's embedded index value.
-	let values = index
-		.iter()
-		.map(|&i| c - embed_position::<F>(i))
-		.collect::<Vec<_>>();
-	FieldBuffer::from_values(&values)
+	// n, the number of looker variables, from the 2^n rows.
+	let log_len = index.len().ilog2() as usize;
+
+	// One denominator per row: shift the challenge by the row's embedded index value.
+	// Pack P::WIDTH consecutive denominators into each word as they are computed.
+	// So the packed buffer is built in one pass, with no scalar buffer materialized first.
+	let packed_len = 1 << log_len.saturating_sub(P::LOG_WIDTH);
+	let mut packed = Vec::with_capacity(packed_len);
+	packed.extend(
+		index
+			.chunks(P::WIDTH)
+			.map(|chunk| P::from_scalars(chunk.iter().map(|&i| c - embed_position::<F>(i)))),
+	);
+
+	FieldBuffer::new(log_len, packed.into_boxed_slice())
 }
 
 /// Build the table denominator `c - J` over the `m`-variable table cube.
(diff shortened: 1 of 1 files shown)


crates/ip-prover/src/logup_star/witness.rs
```diff
@@ -103,17 +103,30 @@ where
...
+	// One denominator per row: shift the challenge by the row's embedded index value.
+	// Pack P::WIDTH consecutive denominators into each word as they are computed.
+	// So the packed buffer is built in one pass, with no scalar buffer materialized first.
+	let packed_len = 1 << log_len.saturating_sub(P::LOG_WIDTH);
+	let mut packed = Vec::with_capacity(packed_len);
+	packed.extend(
+		index
+			.chunks(P::WIDTH)
+			.map(|chunk| P::from_scalars(chunk.iter().map(|&i| c - embed_position::<F>(i)))),
+	);
```
jimpo: minor improvements: rayon, packed subtraction

```rust
let c_packed = P::broadcast(c);
let packed = index.par_chunks(P::WIDTH)
    .map(|chunk| c_packed - P::from_scalars(chunk.iter().copied().map(embed_position::<F>)))
   .collect();
```

jimpo (changes requested):

crates/ip-prover/src/logup_star/witness.rs
tcoratger: Fixed

jimpo (approved):

PR #1720, opened 2026-07-06
BINIUS-230: use Maskable to compute intmul b leaves without branching

benji-potamus:
Rewrites `compute_b_leaves` and `compute_b_leaves_parallel` to select each b-leaf with the branchless `Maskable` API (`make_mask` + `select`) instead of the per-bit `if bit { base } else { F::ONE }`. Resolves the existing `// TODO: Optimize with bit-masking for selection` in the parallel path.

Linear: https://linear.app/jimpo/issue/BINIUS-230

## Change

The selection `if bit { base } else { 1 }` becomes, per lane, `ones + (base - ones).select(&mask)`:
- on **selected** lanes (bit set) the mask is all-ones, so `select` keeps `base - 1` and the `+ 1` restores `base`;
- on **unselected** lanes `select` yields `0`, so `+ 1` gives `F::ONE`.

`Maskable` is a parent trait of both `Field` and `PackedField`, so this needs **no new bound** at the call site. The parallel path drops its scalar decompose→branch→`from_scalars` recompose round-trip in favour of one packed `sub`/`select`/`add`.

## Correctness

Behavior-preserving — output is byte-identical. Added `compute_b_leaves_matches_spec`, a direct oracle test checking every leaf against the spec (`bases[i]^(2^z)` if bit `z` set, else `1`) over **both** the parallel path (`n_vars ≥ P::LOG_WIDTH`) and the scalar fallback (`n_vars < P::LOG_WIDTH`). Existing `test_forwards*` consistency tests still pass.

## Performance

Benchmarked `intmul/components/b_leaves` (criterion, `-C target-cpu=native`) before/after: **no change in performance detected** (change `[-0.87%, +3.26%]`, p = 0.38). The per-column squaring dominates, so the selection cost is negligible either way — this lands as a branchless cleanup / TODO resolution rather than a measurable speedup. Numbers in a comment below.

### DIFF
--- a/crates/prover/src/protocols/intmul/witness.rs
+++ b/crates/prover/src/protocols/intmul/witness.rs
@@ -233,9 +233,11 @@ where
 
 	for (i, (mut base, &exp)) in iter::zip(bases.iter_scalars(), exponents).enumerate() {
 		for z in 0..WORD_SIZE_BITS {
-			let bit = exp.extract_bit(z);
-
-			out.set(z * n_elems + i, if bit { base } else { F::ONE });
+			// Branchless select of `base` when bit `z` is set, else `F::ONE`: on the selected lane
+			// `mask` is all-ones so `select` keeps `base - 1` and the `+ 1` restores `base`; on the
+			// unselected lane `select` yields `0` and the `+ 1` gives `F::ONE`.
+			let mask = F::make_mask(iter::once(exp.extract_bit(z)));
+			out.set(z * n_elems + i, F::ONE + (base - F::ONE).select(&mask));
 
 			base = base.square();
 		}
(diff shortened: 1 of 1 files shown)


benji-potamus:
## Benchmark: `intmul/components/b_leaves`

**What was measured:** `cargo bench -p binius-prover --bench intmul -- intmul/components/b_leaves`, criterion `sample_size(10)`, `RUSTFLAGS="-C target-cpu=native"`, `P = PackedBinaryGhash4x128b`. Baseline = `main` (`061a4110`) via a throwaway sibling commit, compared against this branch with criterion `--save-baseline`/`--baseline`.

| | median |
|---|---|
| before (main, branching) | ~9.32 ms |
| after (this PR, Maskable) | ~9.51 ms |
| criterion change | `[-0.87%, +1.08%, +3.26%]`, p = 0.38 |

**No change in performance detected.** The two are within run-to-run noise (the after median landed slightly high on this run; earlier clean runs were ~9.23 ms before vs ~9.26 ms after). The `b_leaves` cost is dominated by the 64 sequential packed squarings per column, so the per-bit selection — whether branch or masked `select` — is a small fraction of the total.

So I'd frame this as a **cleanup** (branchless `Maskable` selection, resolving the parallel-path TODO) that's perf-neutral, not a speedup. No regression.

jimpo (approved):

PR #1721, opened 2026-07-06
BINIUS-229: optimize power_table with strided packed field muls

benji-potamus:
Rewrites `power_table` (intmul witness) to fill the table with strided packed multiplies instead of a serial scalar power chain, exposing instruction-level parallelism.

Linear: https://linear.app/jimpo/issue/BINIUS-229

## Approach

The old version walked `base^i` one scalar at a time (`row *= base`) — a fully serial dependency chain through the multiplier. The new version:
1. builds the first **stride block** of `2^(P::LOG_WIDTH + LOG_STRIDE)` sequential powers, packed into `2^LOG_STRIDE` packed elements;
2. fills each subsequent block by multiplying the previous block, element-wise, by `incr = base^(2^(P::LOG_WIDTH + LOG_STRIDE))` broadcast.

The `2^LOG_STRIDE` multiplies **within** a block are independent, so the CPU pipelines them; only the block-to-block step carries a dependency. `LOG_STRIDE = 4` (16 packed elements per block). Small tables (≤ one block, incl. `log_size < P::LOG_WIDTH`) keep the simple sequential build. A small `sequential_powers` helper dedups the two sequential loops.

## Correctness

Behavior-preserving (row `i` is still exactly `base^i`). Added `power_table_matches_sequential`, a direct oracle test checking every row against a sequential reference across both the fallback and strided paths (sizes 0/3/6 fallback, 7/10 strided). Existing `test_forwards*` pass.

## Performance

`intmul/components/power_table` (criterion, `-C target-cpu=native`, `P = PackedBinaryGhash4x128b`, `log_size = LIMB_BITS = 16`), baseline = `main` via a throwaway sibling commit:

| | median |
|---|---|
| before (serial scalar chain) | ~766 µs |
| after (strided packed) | ~196 µs |
| criterion change | **−74.4%** (p = 0.00), ≈ 3.9× |

Numbers in a comment below.

### DIFF
--- a/crates/prover/src/protocols/intmul/witness.rs
+++ b/crates/prover/src/protocols/intmul/witness.rs
@@ -163,19 +163,70 @@ where
 	}
 }
 
+/// Number of packed columns filled as one independent group before chaining to the next block.
+///
+/// The power chain `base^i` is inherently serial; striding the fill into blocks of
+/// `2^LOG_STRIDE` independent packed multiplies exposes instruction-level parallelism so the
+/// multiplier pipeline stays busy. `4` keeps a block (16 packed elements) comfortably in registers.
+const LOG_STRIDE: usize = 4;
+
+/// The first `count` powers of `base` (`base^0 .. base^(count-1)`) plus the next power
+/// `base^count`.
+fn sequential_powers<F: Field>(count: usize, base: F) -> (Vec<F>, F) {
+	let mut scalars = Vec::with_capacity(count);
+	let mut row = F::ONE;
+	for _ in 0..count {
+		scalars.push(row);
+		row *= base;
+	}
+	(scalars, row)
+}
+
 /// Build the power table of `base` with `2^log_size` rows: row `i` holds `base^i`.
 pub fn power_table<F, P>(log_size: usize, base: F) -> FieldBuffer<P>
 where
 	F: Field,
 	P: PackedField<Scalar = F>,
 {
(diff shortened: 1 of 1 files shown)


benji-potamus:
## Benchmark: `intmul/components/power_table`

**What was measured:** `cargo bench -p binius-prover --bench intmul -- intmul/components/power_table`, criterion `sample_size(10)`, `RUSTFLAGS="-C target-cpu=native"`, `P = PackedBinaryGhash4x128b`, `log_size = LIMB_BITS = 16` (65536-row table). Baseline = `main` via a throwaway sibling commit, compared with criterion `--save-baseline`/`--baseline`.

| | median |
|---|---|
| before (serial scalar chain) | ~766 µs |
| after (strided packed, `LOG_STRIDE=4`) | ~196 µs |
| criterion change | **−74.4%** `[-74.57%, -74.27%]`, p = 0.00 |

**≈ 3.9× faster.** The win is instruction-level parallelism: the old `row *= base` chain serialized every multiply on the previous result, whereas the strided fill issues 16 independent packed multiplies per block, keeping the multiplier pipeline full. Only the block-to-block step carries a dependency.

This feeds `Witness::new` (2·N_LIMBS = 8 tables built in parallel), so it shaves the witness-setup cost of the intmul protocol.

jimpo (approved):
My suggestion for the strided filling was premature optimization.

PR #1722, opened 2026-07-06
BINIUS-119: M4 batched BitAnd shift reduction (prover + verifier)

tcoratger:
Part of [BINIUS-119](https://linear.app/jimpo/issue/BINIUS-119).

Implements the M4 batched shift reduction for the BitAnd operands, prover and verifier.
The core reuses the single-instance reduction unchanged; the only new work is folding the batch.

### The problem

The shift reduction is the step that rewrites the BitAnd operand claims into one evaluation of the *committed* witness, which the commitment can then open.

M4 commits $K = 2^k$ copies of one circuit at once, stacked instance-major.
So the BitAnd reduction now runs over `K * n_and` rows, and its evaluation point gains a chunk of high coordinates selecting *which copy*.
The single-instance shift reduction has no notion of a copy, so it cannot consume that point.

### The idea

The copies share their wiring and differ only in data.
So the copy index `kappa` enters the operand claim only through the committed witness bit, never through the shift structure.
It therefore factors out through one equality weight — fold the batch onto a single virtual copy whose words are field elements:

```
W_tilde(j, y) = sum_kappa eq(r_kappa, kappa) * bit_j(w_kappa[y])
```

Each 0/1 bit becomes one $\mathbb{F}_{2^{128}}$ element, folded across the batch rows.
The existing two-phase reduction then runs on this virtual copy verbatim, producing a claim on the committed batch witness at $(r_j, r_y, r_\kappa)$, with the copy index on the high coordinates.

The `g` parts and the phase-2 fold are linear in the witness, so building them from the folded witness is exact — no per-copy approximation.

### Where the copy index comes from

The BitAnd reduction reverses its sumcheck challenges, so its point is ordered `row = kappa * n_and + x` with `x` in the low bits.
The split is then purely positional:

```
eval_point = [ x .......... | kappa ...... ]
               low c_and       high k coords
```

### The change

- `crates/prover/.../shift/batch.rs` — `prove_batch`: fold + field-valued `g`-build, reusing the single-instance shift kernels, monster, and both sumcheck runners.
- `crates/verifier/.../shift/batch.rs` — `verify_batch`: BitAnd-only mirror; checks `eval == witness_eval * monster_eval`.
- `crates/m4-prover/src/shift.rs`, `crates/m4-verifier/src/shift.rs` — thin glue on `ShiftReductionOutput` that splits the point and assembles $(r_j, r_y, r_\kappa)$.

The single-instance shift code is untouched apart from marking two sumcheck runners `pub(crate)` so the batched prover can reuse them.

- It is Binius64's *bit-shift operand* reduction (`SLL/SRL/SRA/ROTR`), not Flock §4.6's hash-chain successor shift — same word, different thing.
- It does **not** try to share `r_kappa` across BitAnd and IntMul.
  Per the ticket discussion those reductions use different sumcheck challenges and align only at the bit-index level, so a shared copy point is not available.
- IntMul is out of scope here: it would land at a different copy point, so batching both cleanly is deferred.

### Soundness

- The fold is an exact algebraic identity, not a probabilistic step — it adds no error.
- The two phases and the closing monster identity are the single-instance reduction, run on the folded witness.
- `verify_batch` computes the monster evaluation directly over the channel element type, which assumes a non-zero-knowledge channel; M4 commits without zero-knowledge, so this holds.

### Scope

- **new:** the two `batch.rs` files (prover + verifier) and the two M4 glue modules.
- **changed:** two `pub(crate)` markers and the module/lib export lines.
- **left untouched:** the single-instance shift prover and verifier.

### Verification

- `cargo +nightly fmt --all`, `cargo check --workspace --all-targets`, `cargo clippy` on the four crates — all clean.
- Tests (`binius-m4-prover`, `binius-m4-verifier`, single-instance shift): pass.
- The round-trip test runs the real chain — build K copies, real BitAnd reduction, shift reduction, verify — and checks the claimed value equals the witness evaluated honestly at $(r_j, r_y, r_\kappa)$.
  Plus: `K = 1` reproduces the single-instance path, and a tampered proof is rejected.


crates/m4-prover/src/shift.rs
```diff
@@ -0,0 +1,355 @@
...
+	//
+	//     inputs : x, y, w, z   (all inout)
+	//     gate   : and = x & y
+	//     assert : and ^ w == z
+	struct AndCircuit {
+		circuit: Circuit,
+		x: Wire,
+		y: Wire,
+		w: Wire,
+		z: Wire,
```
benji-potamus, line 125: **Test coverage gap (my main ask).** Every operand in this circuit is an *unshifted* committed word, so the batched reduction only ever exercises the shift-amount-0 / identity-variant key. But `build_batch_g_parts` and `build_flat_monster` are new code that iterates all 8 shift variants (SLL/SRL/SRA/ROTR × amounts) — none of those paths are hit here.

The round-trip test is a genuine prover↔verifier cross-check (a disagreement between the prover's `build_flat_monster` and the verifier's `evaluate_monster_multilinear_for_operation` would trip `assert_zero`), which is exactly why it's worth making it cover shifted operands. Suggest a circuit with e.g. `band(shl(x, k), rotr(y, m))` so multiple variants and non-zero amounts flow through the batched g-build and flat monster. As-is, the most subtle new code is validated only on the trivial variant.

crates/m4-prover/src/shift.rs
```diff
@@ -0,0 +1,355 @@
...
+		let c = and_circuit();
+		let table = populate_table(&c, inputs);
+		let cs = prepared_cs(&c);
+		let key_collection = build_key_collection(&cs);
+
+		// Prover: batched BitAnd reduction feeds the shift reduction on one transcript.
+		let mut prover_transcript = ProverTranscript::new(StdChallenger::default());
+		let and_witness = BatchAndCheckWitness::build(&table, &cs.and_constraints);
+		let log_total = checked_log_2(and_witness.a().len());
+		let and_output = and_witness.prove::<P, _>(&mut prover_transcript);
```
benji-potamus, line 253: Minor: the literal `6` here (and `[6..]` below) is `LOG_WORD_SIZE_BITS`. Prefer the constant so the split doesn't silently break if the word size ever changes.

crates/m4-prover/src/shift.rs
```diff
@@ -0,0 +1,355 @@
...
+		let (r_x, r_kappa) = eval_point.split_at(c_and);
+
+		let bitand_data = OperatorData {
+			evals: vec![a_eval, b_eval, c_eval],
+			r_zhat_prime: z_challenge,
+			r_x_prime: r_x.to_vec(),
+		};
+
+		// One committed-word slice per instance, in instance order.
+		let instances: Vec<&[Word]> = (0..table.n_instances())
```
benji-potamus: Nit (CONTRIBUTING → "Turbofish over type annotations"): prefer `.collect::<Vec<_>>()` over the local type annotation `let instances: Vec<&[Word]> = …`.

crates/prover/src/protocols/shift/batch.rs
benji-potamus, line 1: **Taste / design note (not a blocker).** This file reimplements `build_g_parts` → `build_batch_g_parts`, the phase-2 monster → `build_flat_monster`, and the phase-2 sumcheck → `run_flat_sumcheck` rather than reusing `prove_phase_2`. The merge commit explains why: main moved the single-instance path onto a segmented "skip-c" half-cube layout, while M4's `ValueTable` commits the flat full witness — so the batch legitimately can't reuse the new phase-2.

That's defensible (some duplication beats fusing two witness layouts into one confusing path). The concern is that the duplicated code is subtle monster construction where *drift = a soundness bug*, and `build_flat_monster` (here) vs the verifier's `evaluate_monster_multilinear_for_operation` are two independent implementations validated only by the round-trip test — which currently runs shift-free (see my comment on the test circuit). Broadening that test is the cheap mitigation.

Longer term, worth a tracked note on either unifying once M4 moves to the segmented commitment, or factoring the shared per-word monster scalar so the flat and segmented builders can't drift apart.

benji-potamus (commented):
Nicely documented, and algebraically sound where I could check it. No hard blockers. The one substantive ask is broadening test coverage of the shift-variant machinery; the rest is a taste discussion and a few nits (inline).

**Verified correct**
- Challenge ordering is consistent: `run_phase_1_sumcheck` reverses internally, so the batch prover splits `r_j`(low)/`r_s`(high) with no further reverse while the verifier reverses after `verify_sumcheck` and splits identically; both reverse `r_y`.
- Phase-2 variable counts agree: prover `log2_ceil(key_collection.n_words())` vs verifier `log2_ceil(committed_total_len)`, both `.max(LOG_WORDS_PER_ELEM)`, and `n_words() == committed_total_len`.
- Lambda batching matches: prover `Σ evals·λ^{i+1}` = verifier `λ · evaluate_univariate(evals, λ)`.
- Prover/verifier separation is clean (packed `P` + Rayon on the prover; scalar-only, no Rayon on the verifier). Error boundaries are right: prover uses documented `assert!` preconditions with a `# Panics` section; verifier returns `Result` at the proof boundary.
- No `unwrap` in library code; `expect`s carry justifications; the two `#[allow(clippy::useless_vec)]` are justified by comment. Copyright headers match the repo's actual convention.
- Good defensive guards + matching `#[should_panic]` length-guard test, tamper-rejection test, and proptest.

**Substantive (inline on the test circuit):** the test circuit is shift-free (`band(x, y)`), so the new `build_batch_g_parts` / `build_flat_monster` are only exercised on the trivial identity variant. Since the round-trip test is a real prover↔verifier monster cross-check, extending it to shifted operands is the high-value addition here.

**Taste (inline on `batch.rs`):** the batch path introduces a second, "flat" phase-2 + monster construction parallel to the single-instance "segmented skip-c" one, driven by M4's flat commitment. Defensible, but it's soundness-sensitive duplication validated only by the (currently shift-free) round-trip test — worth a tracked note toward eventual unification.

**Housekeeping**
- History carries a `Merge upstream/main …` commit — please rebase onto main to drop it (jj/rebase workflow) rather than keeping a merge in the stack.
- The PR description's "changed: two `pub(crate)` markers" is stale — the diff touches no `phase_*.rs`; `run_phase_1_sumcheck` / `build_h_parts` are already `pub` in their `pub mod` (cleaner than described). Worth updating the description.

crates/m4-prover/src/shift.rs
tcoratger: Fixed

crates/m4-prover/src/shift.rs
tcoratger: Fixed

crates/m4-prover/src/shift.rs
tcoratger: Fixed

crates/prover/src/protocols/shift/batch.rs
tcoratger: Fixed

crates/verifier/src/protocols/shift/batch.rs
```diff
@@ -0,0 +1,154 @@
...
+/// # Errors
+///
+/// Returns an error if either sumcheck fails or the final monster identity does not hold.
+///
+/// # Panics
+///
+/// Assumes a non-zero-knowledge channel.
+/// The monster evaluation is computed directly over the channel element type.
+/// M4 commits without zero-knowledge, so this holds.
+pub fn verify_batch<F, C>(
```
jimpo, line 75: It should be possible to use the existing shift reduction verify and check_eval without modification. And only change the prover side. Could you try that?

jimpo (changes requested):
Can you split this and put up a PR for just the batch shift prover, and see if you can make it work with the existing verifier? Basically the only different in the batched setting is that the witness eval (which is a field on the shift::verify output) is partially evaluated along the instance axis.

crates/verifier/src/protocols/shift/batch.rs
tcoratger: Done in #1732

tcoratger:
> Can you split this and put up a PR for just the batch shift prover, and see if you can make it work with the existing verifier? Basically the only different in the batched setting is that the witness eval (which is a field on the shift::verify output) is partially evaluated along the instance axis.

I've just done this in #1732 to split this PR. My proposal is that:
- We merge #1732
- I rebase this one to remove the batch verify stuff so that the scope is much more limited

tcoratger:
Closing with https://github.com/binius-zk/binius64/pull/1735

PR #1726, opened 2026-07-07
BINIUS-145: prove the shift phase-2 first round sparsely over the segment selector

benji-potamus:
Final PR of [BINIUS-145](https://linear.app/jimpo/issue/BINIUS-145/handling-constants) (the [plan](https://linear.app/jimpo/issue/BINIUS-145/handling-constants#comment-c759bc33); follows #1707, #1708, #1724). Prover-only optimization — the proof transcript is **byte-identical** to a build without it (verified plain + ZK on the SHA-256 test circuit), so no verifier or proof-format change.

**Context.** #1708 committed only the hidden segment and moved the public segment out of the trace oracle, but its phase-2 sumcheck materialized the full witness as a dense buffer over `log_witness_words + 1` variables — the low half (the zero-padded public segment) is mostly zeros. This PR removes that interim cost.

**The sparse first round.** The witness's top word-index variable is the segment selector, bound in round 1. Instead of building the combined `2^(log_witness_words + 1)` buffer, the prover keeps the two segments separate (`build_monster_segments` returns the public and hidden monster pieces; the folded witness stays split too) and computes round 1 directly:

* the dense `Σ hidden · monster_hidden` pass — the same `O(2^log_witness_words)` dominant term as before;
* two `O(2^log_public_words)` public-prefix corrections for `y_inf`, exploiting that the public/`M_p` buffers vanish past the public length so the stray hidden-lane terms cancel between the two sums.

After the selector challenge `α`, the segment pairs fold into single dense buffers (`fold_segments`) and the remaining rounds run through the standard `BivariateProductSumcheckProver` — so every round message is identical to the dense prover's. The witness evaluation is derived from the final combined evaluation by the same public-segment subtract-and-scale the verifier uses.

The first commit exposes `binius-ip-prover`'s `round_evals` as a `#[doc(hidden)] pub` module so the shift prover can reuse the exact `RoundEvals2` interpolation the in-crate provers use.

### DIFF
--- a/crates/prover/src/protocols/shift/monster.rs
+++ b/crates/prover/src/protocols/shift/monster.rs
@@ -7,7 +7,7 @@ use binius_field::{AESTowerField8b, BinaryField, Field, PackedField, WideMul};
 use binius_math::{
 	BinarySubspace, FieldBuffer, multilinear::eq::eq_ind_partial_eval, univariate::lagrange_evals,
 };
-use binius_utils::rayon::prelude::*;
+use binius_utils::{checked_arithmetics::checked_log_2, rayon::prelude::*};
 use binius_verifier::{
 	config::{LOG_WORD_SIZE_BITS, WORD_SIZE_BITS},
 	protocols::shift::{BITAND_ARITY, INTMUL_ARITY, evaluate_h_op},
@@ -131,16 +131,18 @@ where
 ///
 /// # Usage
 ///
-/// Used in phase 2 of the shift protocol where the prover needs a single multilinear combining
-/// all shift-related constraints for efficient sumcheck computation.
-#[instrument(skip_all, name = "build_monster_multilinear")]
-pub fn build_monster_multilinear<F, P: PackedField<Scalar = F>>(
+/// Used in phase 2 of the shift protocol. The two returned buffers are the segments of the
+/// witness's monster multilinear: the public piece over `log_public_words` variables and the
+/// hidden piece over `log_witness_words` variables (the hidden words at the base, zeros above).
+/// The sparse first sumcheck round consumes them without materializing the combined buffer.
+#[instrument(skip_all, name = "build_monster_segments")]
+pub fn build_monster_segments<F, P: PackedField<Scalar = F>>(
 	key_collection: &KeyCollection,
 	bitand_operator_data: &PreparedOperatorData<F>,
 	intmul_operator_data: &PreparedOperatorData<F>,
 	r_j: &[F],
 	r_s: &[F],
-) -> FieldBuffer<P>
+) -> (FieldBuffer<P>, FieldBuffer<P>)
 where
 	F: BinaryField + From<AESTowerField8b>,
 {
@@ -189,10 +191,6 @@ where
 		&intmul_h_ops,
 	);
 
-	let log_half = key_collection.log_witness_words();
-	let log_len = log_half + 1;
-	let capacity = 1 << log_len.saturating_sub(P::LOG_WIDTH);
-
 	// The scalar for one word of a segment: the accumulated contribution of all its keys. The
 	// per-key wide accumulations are summed unreduced and reduced once at the end.
 	let word_scalar = |segment: &KeySegment, index: usize| {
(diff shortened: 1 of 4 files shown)


crates/prover/src/protocols/shift/monster.rs
```diff
@@ -215,31 +213,31 @@ where
...
-		})
-		.collect_into_vec(&mut monster_multilinear);
-	monster_multilinear.resize(capacity, P::default());
+	// Each segment sits at the base of its buffer: the public piece fills its power-of-two
+	// length exactly, the hidden piece is zero-padded up to the hidden segment length.
+	let build_segment = |segment: &KeySegment, log_len: usize| {
+		let capacity = 1 << log_len.saturating_sub(P::LOG_WIDTH);
+		let n_words = segment.n_words();
+		let mut values = Vec::<P>::with_capacity(capacity);
+		(0..n_words.div_ceil(P::WIDTH))
```
jimpo: Structure the loop so that the chunk indices are (0..n_words / P::WIDTH) (rounding down). In the loop, iterate over exactly P::WIDTH elements to optimize from_scalars. Then handle the remaining elements conditionally at the end of the loop.

crates/prover/src/protocols/shift/phase_2.rs
```diff
@@ -73,102 +76,171 @@ where
...
-	// the low half-cube, the hidden words at the base of the high half-cube.
+	// Fold each segment separately; the combined witness is never materialized. The public fold
+	// fills its power-of-two length exactly, and the hidden fold is zero-extended to the hidden
+	// segment length.
 	let (public_words, hidden_words) = words.split_at(key_collection.public.n_words());
 	let public_folded = fold_words::<_, P>(public_words, r_j_tensor.as_ref());
-	let hidden_folded = fold_words::<_, P>(hidden_words, r_j_tensor.as_ref());
-	let witness_folded =
-		assemble_witness(&public_folded, &hidden_folded, key_collection.log_witness_words());
+	let hidden_folded = zero_extend(
```
jimpo: fold_words already pads to the next power of two with zeros. Why is zero_extend necessary?

crates/prover/src/protocols/shift/phase_2.rs
```diff
@@ -73,102 +76,171 @@ where
...
 ) -> FieldBuffer<P> {
-	let mut witness_folded = FieldBuffer::zeros(log_half + 1);
-	{
-		let mut split = witness_folded.split_half_mut();
-		let (mut lo_half, mut hi_half) = split.halves();
-		lo_half.as_mut()[..public_folded.as_ref().len()].copy_from_slice(public_folded.as_ref());
-		hi_half.as_mut()[..hidden_folded.as_ref().len()].copy_from_slice(hidden_folded.as_ref());
+	let alpha_broadcast = P::broadcast(alpha);
+	let one_minus_alpha = P::broadcast(F::ONE - alpha);
+	let mut values = hidden
```
jimpo: par_iter for scaling hidden. sequential iter for operations on the public segment.

crates/prover/src/protocols/shift/phase_2.rs
```diff
@@ -73,102 +76,171 @@ where
...
-/// Each segment sits at the base of its half-cube, zero-padded up to `2^log_half` entries. The
-/// interim dense representation materializes the mostly-zero low half; a sparse special first
-/// sumcheck round can remove this cost without changing the transcript.
-pub fn assemble_witness<F: Field, P: PackedField<Scalar = F>>(
-	public_folded: &FieldBuffer<P>,
-	hidden_folded: &FieldBuffer<P>,
-	log_half: usize,
+/// Returns the dense buffer `(1 - alpha) * public_padded + alpha * hidden`, exactly the result
+/// of `fold_highest_var_inplace` on the materialized witness buffer.
+fn fold_segments<F: Field, P: PackedField<Scalar = F>>(
```
jimpo, line 161: fold_segments can consume and mutate arguments for memory efficiency

crates/prover/src/protocols/shift/phase_2.rs
```diff
@@ -73,102 +76,171 @@ where
...
 	F: BinaryField + From<AESTowerField8b>,
 {
-	#[cfg(debug_assertions)]
-	let cloned_witness_folded_for_debugging = witness_folded.clone();
+	let log_hidden = hidden_folded.log_len();
+	assert_eq!(hidden_monster.log_len(), log_hidden);
+	assert_eq!(public_monster.log_len(), public_folded.log_len());
+	assert!(public_folded.log_len() <= log_hidden);
+
+	// Round 1: the dense hidden-segment pass.
```
jimpo: Factor round 1 out into a helper function.

crates/prover/src/protocols/shift/monster.rs
benji-potamus: Done — the parallel loop now runs over `0..n_words / P::WIDTH` and each iteration builds `from_scalars` from exactly `P::WIDTH` words (statically sized). The trailing partial element (when `n_words` isn't a multiple of `P::WIDTH`) is pushed once after the loop.

crates/prover/src/protocols/shift/phase_2.rs
benji-potamus: You're right — it isn't. `hidden_words.len()` is exactly the hidden segment length (`n_hidden_words`), so `fold_words` already pads to `log2_ceil(n_hidden_words) = log_witness_words` variables. Dropped `zero_extend` entirely; the folds are used directly, and the `run_sumcheck` asserts (`hidden_monster.log_len() == hidden_folded.log_len()`) still guard the invariant.

crates/prover/src/protocols/shift/phase_2.rs
benji-potamus: Done — `fold_segments` now scales the hidden segment with `par_iter_mut` (the dominant `O(2^log_witness_words)` work) and applies the public-prefix combination with a sequential loop.

crates/prover/src/protocols/shift/phase_2.rs
benji-potamus: Done — `fold_segments` now takes `hidden` by value and scales it in place (`par_iter_mut`), then adds the public prefix into it, returning the same buffer. No new allocation for the dominant buffer. The callers pass `hidden_folded` / `hidden_monster` by value (they're dead after round 1); `public` stays borrowed.

crates/prover/src/protocols/shift/phase_2.rs
benji-potamus: Done — round 1 is now `first_round_coeffs(...)`, which computes the dense hidden pass plus the public-prefix corrections and returns the interpolated `RoundCoeffs`. `run_sumcheck` just calls it, sends the message, samples `α`, and proceeds to the fold + standard prover.

jimpo (approved):

PR #1727, opened 2026-07-07
BINIUS-231: FieldFn::call_native with a WideMul + rayon MonsterEvalFn native path

benji-potamus:
Adds a native `call_native` fast path to `FieldFn` and accelerates the verifier's `MonsterEvalFn` monster evaluation, rebased on the #1728 rewrite (single loop over constraints, one accumulator).

Linear: https://linear.app/jimpo/issue/BINIUS-231

**Redone per review steer:** dropped the tensor-expansion-hoisting (moot after #1728), and split the acceleration into the two commits requested:

**1. `FieldFn::call_native` + wide-mul native path.**
- New `FieldFn::call_native(&self, inputs: &[F]) -> F`, default `self.call::<F>(inputs)` — every impl gets it free, no bound propagation.
- `MonsterEvalFn::call`/`call_native` share an `evaluate<E, _>` helper (input splitting + tensor expansions) that takes a closure for the per-operation evaluation; `call` passes the generic `evaluate_monster_multilinear_for_operation`, `call_native` passes the new `evaluate_monster_multilinear_for_operation_native`.
- The native variant accumulates the per-constraint contributions into a **single unreduced wide element**, reduced exactly once at the end (reduction is `F`-linear ⇒ identical result). The generic path can't do this because `E: FieldOps` doesn't imply `WideMul`. Extracted `operand_shift_scalar_table` so both paths share the table build.
- The 6 native `compute_public_value` sites now call `call_native`.

**2. Parallelize the native loop with rayon.** Now that #1728 made it one loop over all constraints with a single wide accumulator element, the constraints partition cleanly: `par_iter().map(|c| wide_mul(...)).sum::<Wide>()`. No large per-task accumulator (the earlier naive attempt on the old `[[E; 64]; 8]` structure oversplit ~90× slower — that problem is gone).

## Correctness

`evaluate_monster_native_matches_generic` checks the native result equals the generic `E=F` result over random inputs, and the full shift prove→verify round-trip passes — real verification now runs `call_native`. Both commits are independently green.

## Performance

On this **4-core** box, `evaluate_matrices`-equivalent monster eval at ~256 K constraints × ~6 terms: scalar ~46 ms vs native (wide + rayon) ~45 ms — a modest ~2–7%. The loop is largely memory-bandwidth-bound (random `r_y_tensor` access), so 4 cores don't reach 4×; it should scale further on many-core machines. Numbers in a comment.

### DIFF
--- a/crates/verifier/src/protocols/shift/monster.rs
+++ b/crates/verifier/src/protocols/shift/monster.rs
@@ -3,7 +3,7 @@
 use std::iter;
 
 use binius_core::constraint_system::Operand;
-use binius_field::{BinaryField, FieldOps, util::powers};
+use binius_field::{BinaryField, FieldOps, WideMul, util::powers};
 use binius_math::{
 	inner_product::inner_product_scalars, multilinear::eq::eq_ind_partial_eval_scalars,
 };
@@ -131,18 +131,7 @@ where
 	}
 
 	let r_x_prime_tensor = eq_ind_partial_eval_scalars(r_x_prime);
-	let lambda_powers = powers(lambda).skip(1).take(arity).collect::<Vec<_>>();
-
-	// Fold the operand batching coefficients into the shared shift scalars, producing a table
-	// indexed by `(variant, amount, operand_id)` whose entry is
-	// `shift_scalars[variant * WORD_SIZE_BITS + amount] · λ^{operand_id + 1}` — the scalar that
-	// multiplies each shifted-value term.
-	let mut operand_shift_scalars = Vec::with_capacity(shift_scalars.len() * arity);
-	for shift_scalar in shift_scalars {
-		for lambda_power in &lambda_powers {
-			operand_shift_scalars.push(shift_scalar.clone() * lambda_power);
-		}
-	}
+	let operand_shift_scalars = operand_shift_scalar_table(shift_scalars, lambda, arity);
 
 	// Accumulate one contribution per constraint. Within a constraint, each shifted-value term over
 	// all operands is weighted by its operand shift scalar and the word-index tensor entry; the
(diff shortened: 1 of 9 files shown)


benji-potamus:
## Measurement / why rayon is deferred

**What was measured:** a direct timing harness (removed before commit) comparing `evaluate_matrices` (the existing scalar path that `call` uses) against `evaluate_matrices_wide` on the same synthetic input — `F = BinaryField128bGhash`, arity 3, 2¹⁷ constraints × 2 terms ≈ 786 K constraint-terms, over 2¹⁴ words, `-C target-cpu=native`.

| variant | time |
|---|---|
| scalar `evaluate_matrices` | ~4–6 ms |
| `evaluate_matrices_wide` (this PR, sequential) | ~4–5 ms |
| `evaluate_matrices_wide` **with a naive `par_iter().fold()`** | **~585 ms** |

So the `WideMul` path alone is ~neutral at this size (deferred-reduction saving ≈ the cost of the wider 256-bit accumulator). The interesting finding is the rayon result: a per-item `par_iter().zip().fold()` over the constraints is **~90× slower**, because the fold accumulator is a `[[<F as WideMul>::Output; 64]; 8]` (~16 KiB), and rayon oversplits into many tasks each zero-initializing and merging that 16 KiB — the merge/init cost swamps the tiny per-item work.

The **rayon follow-up** therefore needs coarse chunking (≈ a few chunks per thread via `par_chunks`, so only ~`num_threads` accumulators are created and merged), which is a distinct enough change — and perf claim — to review on its own. Happy to open that ticket/PR next if you'd like.

crates/verifier/src/protocols/shift/monster.rs
```diff
@@ -218,6 +218,78 @@ fn evaluate_matrices<F: BinaryField, E: FieldOps<Scalar = F> + From<F>>(
...
+					}
+				}
+				a
+			},
+		)
+}
+
+/// Maps a [`ShiftVariant`] to its index in `0..SHIFT_VARIANT_COUNT`.
+const fn shift_variant_id(shift_variant: ShiftVariant) -> usize {
+	match shift_variant {
```
jimpo: ShiftVariant is a repr(u8). So either `shift_variant as usize` or `shift_variant as u8 as usize` should work now.

benji-potamus:
Force-pushed the redo (`f062b860`, 2 commits on current main). Old commits/CI on this PR are superseded.

## Performance

Direct timing harness (removed before commit), `F = BinaryField128bGhash`, arity 3, 2¹⁸ constraints × 2 terms (~1.5 M term-muls), 2¹⁶ words, `-C target-cpu=native`, **4 cores**:

| variant | time |
|---|---|
| generic scalar (`evaluate_monster_multilinear_for_operation`) | ~46 ms |
| native wide + rayon (`..._native`) | ~45 ms |

So ~2–7% here. Two caveats worth being explicit about:
- **Memory-bound.** The hot access is `r_y_tensor[value_index]` — random reads into a ~1 MiB table — so 4 cores are bandwidth-limited and don't approach 4×. On a many-core verifier host it should scale further (up to memory bandwidth).
- **WideMul alone is ~neutral** at this size (the deferred-reduction saving ≈ the wider 256-bit accumulator cost); the parallelism is where the win comes from, which is why the rayon commit matters.

The key structural point from your steer held: with #1728's single-loop / single-accumulator shape, the rayon version is a clean `par_iter().map().sum::<Wide>()` — no oversplit pathology (the old `[[E; 64]; 8]`-per-task attempt was ~90× *slower*).

jimpo:
Complexity isn't worth it

jimpo (approved):
Changed my mind, it's a good idea.

PR #1730, opened 2026-07-07
BINIUS-233: reduce r_y tensor expansion work in MonsterEvalFn

benji-potamus:
Halves the `r_y` word-index tensor expansion in the verifier's `MonsterEvalFn`, and replaces the re-indexing map with a direct two-segment build.

Linear: https://linear.app/jimpo/issue/BINIUS-233

**Stacked on #1727** (base is that branch, not `main`) — it edits the `MonsterEvalFn::evaluate` helper introduced there. GitHub will retarget to `main` once #1727 merges.

### The problem

`MonsterEvalFn` receives `r_y` extended with the top `r_segment` coordinate (public vs hidden segment). The old code expanded the full extended point — `eq_ind_partial_eval_scalars(r_y_v)`, `2^r_y_len` entries — then re-indexed it into per-word weights, throwing away roughly half the work (the extension doubles the multiplications).

### Commit 1 — `scaled_eq_ind_partial_eval_scalars`

A scalar (`FieldOps`) analogue of the packed `scaled_eq_ind_partial_eval`, seeding the expansion with `scale` so it's folded in for free. `eq_ind_partial_eval_scalars` now delegates to it with `scale = 1` (mirroring how `eq_ind_partial_eval` calls the packed scaled version).

### Commit 2 — per-segment `r_y` tensor

Build each segment's portion directly from the shared `r_y` indicator instead of expanding the extended point:
- **hidden** = `scaled_eq_ind_partial_eval_scalars(r_y, r_segment)` — the high-half weight;
- **public** = `scaled_eq_ind_partial_eval_scalars(r_y[..log_public_words], (1 - r_segment) * eq_ind_zero(r_y[log_public_words..]))` — the public segment occupies a `log_public_words`-length prefix, scaled by the same `padded_public_eval` factor `check_eval` reconstructs with. `n_public_words` is a power of two, so this prefix indicator is exactly `n_public_words` entries.

Then the "garbage" re-indexing map becomes a `Vec::with_capacity(combined_len)` + two `extend_from_slice`s (public, then hidden).

## Correctness

New `scaled_eq_ind_partial_eval_scalars` unit test (equals unscaled × scale). The full prove→verify round-trip and all 8 `prove_verify` end-to-end tests (real public + hidden segments through `check_eval`) pass unchanged. Both commits independently green (build/clippy/doc/fmt).

## Performance

The `r_y` expansion drops from `2^r_y_len` to `2^(r_y_len-1) + 2^log_public_words` multiplications — ~2×. Micro-measured on the expansion alone at `r_y_len = 21`, `log_public = 10`, `F = BinaryField128bGhash`, `-C target-cpu=native`: **~27 ms → ~14 ms**. This is one component of `MonsterEvalFn` (which is dominated by the per-constraint loop), so the top-line verify impact is smaller, but the expansion itself is halved.

### DIFF
--- a/crates/verifier/src/protocols/shift/verify.rs
+++ b/crates/verifier/src/protocols/shift/verify.rs
@@ -15,7 +15,9 @@ use binius_ip::{
 use binius_math::{
 	BinarySubspace,
 	line::extrapolate_line,
-	multilinear::eq::{eq_ind_partial_eval_scalars, eq_ind_zero},
+	multilinear::eq::{
+		eq_ind_partial_eval_scalars, eq_ind_zero, eq_one_var, scaled_eq_ind_partial_eval_scalars,
+	},
 	univariate::{evaluate_univariate, lagrange_evals_scalars},
 };
 use getset::Getters;
@@ -291,7 +293,7 @@ where
 		let intmul_r_x_prime_len = intmul_data.r_x_prime.len();
 		let r_j_len = r_j.len();
 		let r_s_len = r_s.len();
-		let r_y_len = r_y.len() + 1;
+		let r_y_len = r_y.len();
 
 		let inputs: Vec<C::Elem> = iter::once(r_zhat_prime)
 			.chain(iter::once(bitand_lambda.clone()))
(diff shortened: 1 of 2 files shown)


crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -413,23 +415,33 @@ impl<F: BinaryField> MonsterEvalFn<'_, F> {
 		off += self.r_s_len;
 		let r_y_v = &vals[off..off + self.r_y_len];
```
jimpo, line 416: parse r_segment here. r_y_len should be 1 less.

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -413,23 +415,33 @@ impl<F: BinaryField> MonsterEvalFn<'_, F> {
...
-				} else {
-					witness_tensor[half + word_index - n_public_words].clone()
-				}
-			})
-			.collect::<Vec<_>>();
+		let log_public_words = layout.log_public_words();
+		let (r_y, r_segment) = r_y_v.split_at(self.r_y_len - 1);
+		let r_segment = r_segment[0].clone();
+
+		let public_scale = (E::one() - r_segment.clone()) * eq_ind_zero(&r_y[log_public_words..]);
```
jimpo: nit: use eq_one_var(r_segment.clone(), E::zero()) for the first part.

Also, drive-by: `eq_one_var` should have a conditional on `F::Scalar::CHARACTERISTIC == 2` and use the simplified expression if so.

jimpo (changes requested):

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done — `r_segment` is now parsed as its own trailing input element and `r_y_len` is just `r_y.len()` (the construction drops the `+ 1`). `r_y_v` is the word coordinates alone, so the `split_at` is gone.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Both done. Public scale now uses `eq_one_var(r_segment.clone(), E::zero())` for the `(1 - r_segment)` factor. And the drive-by: `eq_one_var` now branches on `F::Scalar::CHARACTERISTIC == 2` and returns `x + y + 1` in that case (a compile-time-constant condition, so only one arm is generated) — in a separate commit (`eq_one_var: use the simplified characteristic-2 expression`).

jimpo (approved):

Issue #1549, opened 2026-06-20
Private security contact for two frontend gate soundness findings?


kite-builds:
Hi maintainers — I've found two `CircuitBuilder` frontend gates whose **verifier-side** constraint systems accept witnesses that violate the gate's documented relation (a malicious prover can satisfy `verify_constraints` with a false statement). Each has a one-line fix and I have a runnable forge harness.

I don't see a `SECURITY.md`, a security email, or private vulnerability reporting enabled on this repo, so I'm **not** including any details here to avoid public disclosure. What's the best private channel to send the full write-up + forge? Happy to use email, a private advisory if you enable it, or anything you prefer.

— Kite (github.com/kite-builds)

jimpo:
Oops, I missed this. Thank you, you can post details here! This library is not used in production anywhere.

kite-builds:
Thanks @jimpo — posting the details here as requested. Reported purely as **library soundness** (no production/funds claim): two public `CircuitBuilder` gates emit constraint systems that `verify_constraints` accepts for witnesses that violate the gate's documented relation. Both are malicious-prover exploitable and both have a one-line fix. Verified against `binius64` @ `7c85579`.

## 1. `assert_non_zero` does not enforce `x ≠ 0`
`crates/frontend/src/compiler/gate/assert_non_zero.rs`. Docstring: "Enforces `x ≠ 0`." `constrain()` (L48–71) emits only the carry-defining AND `(x⊕cin) ∧ (all-1⊕cin) = cin⊕cout` (with `cin = cout<<1`); it never constrains `MSB(cout) = 1`. That MSB check lives only in `emit_assert_non_zero` (L105) — prover-side eval bytecode, **not** a verifier constraint.

**Forge:** `x = 0, cout = 0` ⇒ `0 ∧ all-1 = 0` ✓, so `verify_constraints` returns `Ok(())` for a claimed-nonzero `x` that is actually `0`. `test_assert_non_zero_fails_on_zero` (L170) only exercises the prover-side `populate_wire_witness` panic, so it does not catch this.

**Fix:** additionally force the MSB of `cout`, e.g. `cout ∧ msb_one = msb_one`.

## 2. `smul` low word is submask-constrained, not equality-constrained
[...]

PR #1731, opened 2026-07-07
Fix two frontend gate soundness bugs (assert_non_zero, smul)

benji-potamus:
Fixes a `CircuitBuilder` frontend gate whose **verifier-side** constraint system accepted witnesses that violate the gate's documented relation — a malicious prover could satisfy `verify_constraints` with a false statement. Reported by Kite ([github.com/kite-builds](https://github.com/kite-builds)) in #1549.

Linear: [BINIUS-115](https://linear.app/jimpo/issue/BINIUS-115)

> The `smul` low-word fix that was originally in this PR has been dropped — it's being handled in a separate PR.

### `assert_non_zero`: enforce `MSB(cout) = 1`
`constrain()` emitted only the carry-defining AND `(x ⊕ cin) ∧ (all-1 ⊕ cin) = cin ⊕ cout` and never constrained `MSB(cout) = 1`; that check lived only in the prover-side eval bytecode. Forge: `x = 0, cout = 0` satisfied the constraint, so `verify_constraints` accepted a false non-zero claim (the gate is not used by any in-tree gadget at this HEAD, so no shipped proof is affected).

The fix adds a second constraint forcing `MSB(cout) = 1`, expressed as an AND: `sar(cout, 63) ∧ all_one = all_one`. `sar(cout, 63)` sign-extends the MSB across all 64 bits, so it equals `all_one` iff `MSB(cout) = 1`. It's emitted as an AND (which defines no wire) rather than a linear constraint on purpose: a linear constraint with `dst` set to a constant gets inlined by the gate-fusion pass, which substitutes the constant back into Constraint 1 and silently drops the check. `ZERO` is no longer needed in `const_in` (the eval switches to the cin-less `emit_iadd_cout`), so the shape is now just `[ALL_ONE]`.

A regression test hand-crafts Kite's forge witness (`x = 0, cout = 0`) and asserts `verify_constraints` now returns `Err`; the pre-existing negative test only exercised the prover-side `populate_wire_witness` panic and never touched the verifier hole. Full `binius-frontend` lib suite + crate clippy are green.

### DIFF
--- a/crates/frontend/src/compiler/gate/assert_non_zero.rs
+++ b/crates/frontend/src/compiler/gate/assert_non_zero.rs
@@ -22,13 +22,14 @@
 //!
 //! # Constraints
 //!
-//! The gate generates 1 AND constraint:
-//! - `(x ⊕ cin) ∧ (all-1 ⊕ cin) = cin ⊕ cout`
+//! The gate generates 2 constraints:
+//! - AND: `(x ⊕ cin) ∧ (all-1 ⊕ cin) = cin ⊕ cout`
+//! - linear: `sar(cout, 63) ⊕ all-1 = 0` (forces `MSB(cout) = 1`, i.e. `x ≠ 0`)
 
 use binius_core::word::Word;
 
 use crate::compiler::{
-	constraint_builder::{ConstraintBuilder, sll, xor2},
+	constraint_builder::{ConstraintBuilder, sar, sll, xor2},
 	gate::opcode::OpcodeShape,
 	gate_graph::{Gate, GateData, GateParam, Wire},
 	pathspec::PathSpec,
@@ -36,7 +37,8 @@ use crate::compiler::{
 
 pub const fn shape() -> OpcodeShape {
 	OpcodeShape {
-		const_in: &[Word::ALL_ONE, Word::ZERO], // Need zero constant for cin
+		// ALL_ONE is the addend, ZERO is cin.
+		const_in: &[Word::ALL_ONE, Word::ZERO],
 		n_in: 1,
 		n_out: 0,
 		n_aux: 1,
@@ -52,7 +54,7 @@ pub fn constrain(_gate: Gate, data: &GateData, builder: &mut ConstraintBuilder) 
 		aux,
 		..
 	} = data.gate_param();
-	let [all_one, _zero] = constants else {
+	let [all_one, zero] = constants else {
 		unreachable!()
 	};
 	let [x] = inputs else { unreachable!() };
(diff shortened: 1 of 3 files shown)


crates/frontend/src/compiler/gate/assert_non_zero.rs
```diff
@@ -52,7 +54,7 @@ pub fn constrain(_gate: Gate, data: &GateData, builder: &mut ConstraintBuilder)
 		aux,
 		..
 	} = data.gate_param();
-	let [all_one, _zero] = constants else {
+	let [all_one, _zero, msb_one] = constants else {
```
jimpo: remove the unused _zero const from const_in.

crates/frontend/src/compiler/gate/assert_non_zero.rs
```diff
@@ -68,6 +70,10 @@ pub fn constrain(_gate: Gate, data: &GateData, builder: &mut ConstraintBuilder)
 		.b(xor2(*all_one, cin))
 		.c(xor2(cin, *cout))
 		.build();
+
+	// Constraint 2: Force MSB(cout) = 1 (i.e. x ≠ 0)
+	// cout ∧ msb_one = msb_one
+	builder.and().a(*cout).b(*msb_one).c(*msb_one).build();
```
jimpo: No, this should be a linear constraint on `sra(cout, 63) XOR all_one`. If that's 0, then it's not equal. Drop the `msb_one` const.

jimpo (changes requested):

crates/frontend/src/compiler/gate/assert_non_zero.rs
benji-potamus: Done. Constraint 2 is now the linear constraint `sar(cout, 63) ⊕ all_one = 0` (`.linear().rhs(xor2(sar(*cout, 63), *all_one)).dst(*zero)`), and `Word::MSB_ONE` is dropped. `sar(cout, 63)` sign-extends the MSB across all 64 bits, so it equals `all_one` iff `MSB(cout) = 1` (x ≠ 0), making the XOR zero; when `x = 0, cout = 0`, `sar = 0` and the XOR is `all_one ≠ 0`, so the forge is rejected. Forge test reconfirmed fail-before/pass-after.

crates/frontend/src/compiler/gate/assert_non_zero.rs
benji-potamus: `const_in` is back to `[Word::ALL_ONE, Word::ZERO]` — `ZERO` is used again as the `dst` of the new linear constraint (asserting the XOR expression is 0), so it's no longer unused, and `MSB_ONE` is gone. Net effect: the shape is unchanged from upstream and the fix is just the added linear constraint.

benji-potamus:
Superseded by #1736. This PR auto-closed when its branch was recreated while splitting the `smul` commit out; #1736 continues the `assert_non_zero` fix (now the AND form, `ZERO` dropped). The `smul` low-word fix is being handled separately.

PR #1732, opened 2026-07-07
BINIUS-119: batched shift prover reusing the single-instance verifier

tcoratger:
Part of [BINIUS-119](https://linear.app/jimpo/issue/BINIUS-119).
Prover-only split of #1722, following the review there: reuse the existing single-instance shift verifier and change only the prover.

Adds a batched shift-reduction prover for M4. The single-instance `verify` and `check_eval` are untouched.

### The problem

M4 proves $K = 2^k$ instances of one circuit at once, stacked instance-major.
The batched BitAnd/IntMul reductions leave operand claims whose evaluation point carries an extra instance part `r_kappa`.
The single-instance shift prover has no notion of that instance axis.

### The idea

The instance index enters an operand claim *only* through the committed witness bit — the shift structure is shared by every instance.
So the instance factor folds onto the witness, and every witness-derived buffer of the single-instance reduction becomes an `eq(r_kappa, .)`-weighted sum of the per-instance buffers:

```
g_batch    = sum_kappa eq(r_kappa, kappa) * g(instance_kappa)
fold_batch = sum_kappa eq(r_kappa, kappa) * fold(instance_kappa)
```

These combinations are exact: the `g` parts and the phase-2 folds are linear in the witness, and the shift kernels and monster are instance-independent.
So the batch prover is the single-instance reduction run on the instance-folded witness.

### The change

`prove_batch` reuses the single-instance building blocks verbatim, eps-combining their per-instance outputs:

- `build_g_parts` and `fold_words` per instance, combined with the `eq(r_kappa, .)` weights.
- `build_h_parts`, `build_monster_segments`, `run_phase_1_sumcheck`, `run_sumcheck` — used as-is (instance-independent).

The transcript is byte-identical in shape to a single-instance proof.
So the existing `verify` and `check_eval` accept it with no modification.
Their `witness_eval` output is the committed batch witness partially evaluated at `r_kappa` along the instance axis — exactly the review's framing.

### Why the verifier needs no change

`run_sumcheck` derives the committed-half evaluation by subtracting the public-segment contribution, and `check_eval` adds it back.
Prover and verifier use the same public words, so that term cancels and the verifier reconstructs the true folded trace.
The final identity `trace * monster == eval` then holds for the batch.

### Soundness

- The instance fold is an exact algebraic identity, not a probabilistic step.
- Both sumchecks and the closing monster identity are the single-instance reduction, run on the folded witness.
- The public reconstruction uses one shared public set (instance 0's). This binds correctly when the batch's public inputs agree, or when the full witness is committed (M4's current `ValueTable`), so the public is bound by the commitment and the cancellation is harmless. A hidden-only commitment with differing per-instance public would need the public folded across instances — a downstream commit/ring-switch point, not part of this reduction.

### Scope

- **new:** `crates/prover/src/protocols/shift/batch.rs` (`prove_batch`) and its `pub use`.
- **new:** a round-trip test in `crates/prover/tests/shift.rs`.
- **untouched:** the single-instance prover and the verifier (`verify`, `check_eval`).

### Verification

- `cargo check --workspace --all-targets`, `cargo clippy -p binius-prover --all-targets`, nightly `fmt` — clean.
- `cargo test -p binius-prover --test shift` — the new `test_batch_shift_prove_with_unmodified_verifier` (K=4 instances, distinct inputs, four shift variants at non-zero amounts) round-trips through the unmodified `verify` + `check_eval`, and the single-instance test still passes.

### DIFF
--- a/crates/prover/src/protocols/shift/batch.rs
+++ b/crates/prover/src/protocols/shift/batch.rs
@@ -1,0 +1,181 @@
+// Copyright 2025 Irreducible Inc.
+
+//! Batched shift reduction prover for the M4 data-parallel proof system.
+//!
+//! M4 proves `K = 2^k` instances of one circuit at once, stacked instance-major.
+//! The batched BitAnd/IntMul reductions leave operand claims at a local constraint point.
+//! Their evaluation point also carries an instance part `r_kappa`.
+//!
+//! The instance index enters an operand claim only through the committed witness bit.
+//! The shift structure is shared by every instance, so the instance factor folds onto the witness.
+//! Every witness-derived buffer of the single-instance reduction then becomes an eq-weighted sum:
+//!
+//! ```text
+//!   g_batch    = sum_kappa eq(r_kappa, kappa) * g(instance_kappa)
+//!   fold_batch = sum_kappa eq(r_kappa, kappa) * fold(instance_kappa)
+//! ```
+//!
+//! The g parts and the phase-2 folds are linear in the witness, so these combinations are exact.
+//! The shift kernels and the monster are instance-independent, so they are reused verbatim.
+//!
+//! So this is the single-instance reduction run on the instance-folded witness.
+//! Its transcript matches the single-instance prover, so `verify` and `check_eval` accept it as is.
+//! Their witness evaluation is the committed batch witness partially evaluated at `r_kappa`.
+
+use binius_core::word::Word;
+use binius_field::{BinaryField, PackedField};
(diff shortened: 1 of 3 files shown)


jimpo:
This is duplicating the build_g_parts call, which is the most expensive call.

https://github.com/binius-zk/binius64/pull/1735 implements the optimized algorithm that scales really well with the number of instances. As the number of instances grows in proportion to the size of the circuit, we will just need to pay for `fold_instances` essentially.

PR #1733, opened 2026-07-07
BINIUS-236: eliminate the per-layer witness copy in the fractional-addition prover

tcoratger:
Closes [BINIUS-236](https://linear.app/jimpo/issue/BINIUS-236).

Removes a full copy of the numerator+denominator layer that the fractional-addition prover made on every reduction layer. Proof output is bit-identical — no protocol change, no new soundness assumption.

### The problem

The fraction prover reduces a claim layer by layer, from the small summed layer back to the original witness.

At each layer it hands four multilinears to an MLE-check: the low and high halves of the numerator buffer and of the denominator buffer.

It built those four by **copying** — split each buffer by reference, then allocate four fresh owned buffers and memcpy the halves in.

That is one full copy of the num+den layer, allocated and freed, per layer per proof.

At lookup scale ($2^{20}$ to $2^{24}$ fractions) the widest layer is tens of megabytes, so the copy is real memory bandwidth and cache pressure.

### The idea

The MLE-check never needs to *own* those halves — it folds them in place.

A buffer split already exposes its two halves as borrowed mutable slices over the same backing store.

So hand the four borrowed halves straight to the prover instead of copying.

The sibling product-check prover already works this way; the fraction prover was the odd one out because its MLE-check constructor demanded four owned buffers.

### The change

```
layer_prover:
- split each buffer by ref, then allocate + memcpy four owned halves
+ split each owned buffer and hand the four borrowed halves over
```

### Soundness

- The reduction algorithm and all field arithmetic are unchanged.
- The prover folds the same values in the same order, so the proof output is bit-identical.
- The existing prove/verify round-trip tests (fraction, batched, logUp*) are the reference oracle and still pass.

### Scope

- **new:** a zero-copy adapter in `binius-math` presenting two split buffers as their four halves (first-low, first-high, second-low, second-high), plus a test pinning it to the borrow-only split.
- **changed:** the fraction MLE-check constructor now takes anything presenting four mutable slices; the per-layer step passes split halves directly.
- **untouched:** the other constructor call site (fresh padded buffers) — an owned array of buffers still presents as four slices.

### Verification

- `cargo check`/`clippy`/`test` on `binius-math` and `binius-ip-prover` (default and `rayon`) — clean, 0 clippy warnings.
- `binius-math` 33 + `binius-ip-prover` 58 tests pass, including all fraction/batched/logUp* prove-verify round-trips.
- nightly `fmt --check` — clean.
- Note: `cargo check --workspace --all-targets` is red on `main` for an unrelated reason (`FieldFn::call_native` feature gating in `binius-ip`); it reproduces on a clean checkout and is not touched here.

### Benchmark

End-to-end `prove` wall time, full reduction to a scalar, release build with rayon (median of repeated runs; speedup on min to filter noise):

| witness | before | after | speedup (min) |
|---|---|---|---|
| 2^18 fractions | 25.6 ms | 26.6 ms | ~0% (noise) |
| 2^20 fractions | 44.4 ms | 42.0 ms | ~9% |
| 2^22 fractions | 92.9 ms | 74 ms | ~23% |

The win scales with witness size: the copy cost is proportional to layer width, so it is lost in the arithmetic at small sizes and reaches ~20% at lookup scale. Measured with NEW running last (hottest) to rule out thermal-ordering bias.

### DIFF
--- a/crates/ip-prover/src/fracaddcheck.rs
+++ b/crates/ip-prover/src/fracaddcheck.rs
@@ -116,14 +116,11 @@ where
 		};
 
 		let (num, den) = layer;
-		let (num_0, num_1) = num.split_half_ref();
-		let (den_0, den_1) = den.split_half_ref();
-		let num_0 = FieldBuffer::new(num_0.log_len(), num_0.as_ref().into());
-		let num_1 = FieldBuffer::new(num_1.log_len(), num_1.as_ref().into());
-		let den_0 = FieldBuffer::new(den_0.log_len(), den_0.as_ref().into());
-		let den_1 = FieldBuffer::new(den_1.log_len(), den_1.as_ref().into());
+		// The MLE-check reduces four multilinears.
+		// These are the low and high halves of the numerator buffer and of the denominator buffer.
+		// Splitting the owned buffers hands those halves over by borrow, so the fold runs in place.
 		let prover = frac_add_mle::new(
-			[num_0, num_1, den_0, den_1],
+			[num.split_half(), den.split_half()],
 			num_claim.point.clone(),
 			[num_claim.eval, den_claim.eval],
 		);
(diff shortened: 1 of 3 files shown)


jimpo (approved):
Ahh, right, I forgot about AsSlicesMut! Nice!

PR #1734, opened 2026-07-07
BINIUS-234: replace the smul gate with a standard gadget

tcoratger:
Closes [BINIUS-234](https://linear.app/jimpo/issue/BINIUS-234).

Replaces the bespoke signed-multiply opcode with a gadget built from primitive gates. Pure refactor — the circuit produced is behaviorally identical.

### The problem

Signed 64-bit multiplication was its own opcode, and it carried a lot of weight for what it does:

- 7 aux wires and 1 scratch register in its shape.
- a hand-rolled modular-addition constraint helper (two `AND` constraints per add).
- a dedicated bytecode instruction (`0x31`) plus a matching interpreter path.
- a `Word::smul` witness primitive it leaned on.

All of that machinery existed only to express something the existing gates already can. A new gate is the most expensive way to add an operation: it touches the opcode enum, the shape table, the constrain dispatch, the bytecode emitter, and the interpreter — five places — for what is really just a fixed pattern of multiply, shift, and subtract.

### The idea

Signed multiplication is unsigned multiplication with a high-word correction (Hennessy & Patterson, App. J.2).

Read as unsigned bit patterns, the signed operands are $a - 2^{64} a_s$ and $b - 2^{64} b_s$, where $a_s$, $b_s$ are the sign bits. Expanding the product modulo $2^{128}$:

```
a_signed * b_signed = a * b
                    - 2^64 * (a_sign * b)
                    - 2^64 * (b_sign * a)
```

The $2^{128} a_s b_s$ term vanishes. So the low word is the unsigned low word untouched, and the high word just subtracts `b` when `a` is negative and `a` when `b` is negative.

That maps directly onto gates the frontend already has:

```
(hi_u, lo) = imul(a, b)              // unsigned 128-bit product
a_sign     = sar(a, 63)              // all-ones iff a < 0
b_sign     = sar(b, 63)              // all-ones iff b < 0
corr_a     = a_sign & b             // subtract b when a < 0
corr_b     = b_sign & a             // subtract a when b < 0
hi         = hi_u - corr_a - corr_b  // two isub, borrows discarded
```

### The change

```diff
- builder.smul(a, b)  // Opcode::Smul: 1 gate, 7 aux + 1 scratch, custom constrain + bytecode + interp
+ smul64(builder, a, b)  // imul + 2·sar + 2·band + 2·isub_bin_bout, all existing gates
```

`CircuitBuilder::smul` stays as the public entry point — it is reopened in the new `ops` module as a one-line wrapper over the gadget, so every existing caller is unchanged.

### Soundness

- The constraint set is the same shape as before: 1 `MUL` + 8 `AND` constraints, now emitted by stock gates instead of custom code.
- The two high-word subtractions discard their borrow-out, which is correct: bits past position 63 lie outside the 128-bit product.
- The full property-test suite (correctness vs native `i128`, commutativity, 0 / 1 / -1 identities, and the `i64::MIN` overflow boundary cases) moved with the gadget and still passes.

### Scope

- **new:** `crates/frontend/src/ops.rs` — the `smul64` gadget, the reopened `CircuitBuilder::smul`, and the test suite.
- **deleted:** `gate/smul.rs`, the `Smul` opcode + its shape/constrain/bytecode arms, `emit_smul` (bytecode builder), and `exec_smul` + the `0x31` dispatch (interpreter).
- **left untouched:** `Word::smul` in `binius-core` — it is now used only by its own tests, but it is a tested public primitive, so removing it is out of scope here.

### Verification

- `cargo check -p binius-frontend --all-targets`, `cargo clippy -p binius-frontend --all-targets`, nightly `cargo fmt --all` — clean.
- `cargo test -p binius-frontend` — 150 unit + 2 integration + doctests pass.
- Note: `cargo check --workspace --all-targets` currently fails on `main` due to a pre-existing feature-unification mismatch in `binius-core`'s `ValueVecLayout` (`n_hidden_words` / `combined_len`), unrelated to this change — verified by reproducing it on a clean checkout.


jimpo (approved):
Thank you! Much more maintainable.

PR #1735, opened 2026-07-07
BINIUS-119: batched M4 shift-reduction prover

jimpo:
## Summary

Starts the batched (data-parallel M4) shift-reduction prover in `binius-m4-prover`, building it up incrementally with a test at each step. The batched prover reduces the bitand/intmul evaluation claims for a whole batch of `2^k` instances of one circuit to a single multilinear claim on the witness, reusing the single-instance prover's own subroutines.

The work is organized as a sequence of small, reviewable commits:

- **Scaffold + CRC64 fixture**: a new `shift` module whose tests build a shift-heavy CRC-64/GO-ISO circuit over 4 private witness words and populate a `ValueTable` with `2^n` random instances.
- **`fold_instances`**: naively folds the committed witness of a batch table along the instance axis (`r_rho`), producing the oblong `(bit, word)` multilinear. Tested for commutativity of the fold vs. multilinear evaluation.
- **`prove` + `FoldedWord`**: the batched entry point, mirroring `binius_prover::protocols::shift::prove` but taking the witness folded over instances.
- **`build_g_parts`**: the batched analogue of the single-instance phase-1 g-parts builder — field multiplication by folded bits instead of masking — operating one `KeySegment` at a time. Validated by a phase-1 `g·h` identity test against the batched AND-check operand evals.
- **Public-word handling**: public words are constants shared across instances, so they are expanded (bit → `F::ONE`/`F::ZERO`) rather than folded over `r_rho`.
- **`prove` implementation**: phase 1 builds g-parts per segment and runs `run_phase_1_sumcheck`; phase 2 derives the hidden folded segment as a partial evaluation of `folded_witness` along the bit (`r_j`) axis and reuses `build_monster_segments` / `run_sumcheck` unchanged.

Also re-exports `Operation` from `binius_prover::protocols::shift` so the batched g-parts builder can select operator data by `key.operation`.

## Testing

`cargo test -p binius-m4-prover` — all pass, including:
- `fold_instances_commutes_with_evaluation`
- `phase_1_g_h_inner_product_matches_batched_evals`
- `prove_and_verify_round_trip` — runs the batched prover, then the **single-instance shift verifier** (`verify` + `check_eval` + `finalize`) on the same transcript, and asserts the reduced witness evaluation equals the direct evaluation of the instance-folded witness.

`cargo clippy` and `cargo fmt` are clean.

## Follow-ups

- **Optimize `fold_instances`**: the current implementation is intentionally naive (per-word, per-bit scan). It should be replaced with a method-of-four-Russians version like `fold_across_words`.
- **Use `PackedField` for the shift-reduction sumchecks**: `build_g_parts` and phase 1 currently run scalar (`F`). They should be generalized to packed `P` for SIMD, matching the single-instance prover.

### DIFF
--- a/crates/m4-prover/Cargo.toml
+++ b/crates/m4-prover/Cargo.toml
@@ -13,6 +13,7 @@ binius-field = { path = "../field" }
 binius-frontend = { path = "../frontend" }
 binius-hash = { path = "../hash" }
 binius-iop-prover = { path = "../iop-prover" }
+binius-ip = { path = "../ip" }
 binius-ip-prover = { path = "../ip-prover" }
 binius-m4-verifier = { path = "../m4-verifier" }
 binius-math = { path = "../math" }
@@ -24,7 +25,6 @@ binius-verifier = { path = "../verifier" }
 [dev-dependencies]
 assert_matches = { workspace = true }
 binius-iop = { path = "../iop" }
-binius-ip = { path = "../ip" }
 binius-math = { path = "../math", features = ["test-utils"] }
 proptest = { workspace = true }
 rand = { workspace = true }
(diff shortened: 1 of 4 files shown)


jimpo:
@tcoratger Please take a look

PR #1739, opened 2026-07-07
BINIUS-237: add ValueTable2, a wire-major M4 witness table

benji-potamus:
Implements [BINIUS-237](https://linear.app/jimpo/issue/BINIUS-237/valuetable2-for-m4).

`ValueTable2` is an alternative to `ValueTable` that stores the batch witness in the **transposed** order — each wire's values across all instances (wire-major), rather than each instance's words back to back (instance-major). It exists alongside `ValueTable` for now; once both are evaluated we can drop one. Transposing the value table this way is expected to pay off throughout the rest of the M4 protocol.

It is specialized for the M4 accelerator setting, per the issue's two requirements:
1. **No inout wires.** `populate` rejects a constraint system with `n_inout != 0`. Circuits keep their output wires alive with `force_commit` instead.
2. **Constants are not stored in the table.** They stay as a `Vec<Word>` on the `ConstraintSystem`; the batched interpreter reads them from that bank (broadcast into the working buffer) and only the hidden segment (witness + internal words) is stored.

Witness generation is done by a new `BatchInterpreter` — a structure-of-arrays counterpart to `Interpreter` that evaluates the circuit bytecode over every instance at once, treating each "register" as a whole row (a `StridedArray2DViewMut` over the value data, rows = value indices, columns = instances).

### Commits
1. **Fix gate fusion panic on force-committed non-linear outputs.** Gate fusion added *every* force-committed wire to the linear-commit set and then treated each as a linear definition, panicking (`no entry found for key`) when the pinned wire was an AND/MUL output. It now filters pinned wires to linear definitions only — non-linear outputs are already committed by their own constraints. This is a prerequisite: the M4 pattern (witness inputs, no inout, `force_commit`ed outputs) hit this panic on any circuit whose output is an AND. **Flagging for review** since it changes a foundational optimizer path; includes a focused regression test.
2. **Add the batched circuit interpreter.** `BatchInterpreter` + `EvalForm::evaluate_batched` + `Circuit::populate_wire_witness_batched`. Validated by a test asserting the batched result matches the single-instance interpreter for every instance across a spread of opcodes.
3. **Add `ValueTable2`.** The wire-major table + `Batch2WitnessFiller`, with tests verifying each reconstructed instance against `verify_constraints`, batch-vs-single-instance equivalence, failing-instance reporting, and inout rejection.

### Scope
This PR covers the transposed storage and witness generation only. A commitment/packing scheme for the transposed layout (the actual downstream protocol perf win) is deliberately left as follow-up.

### DIFF
--- a/crates/frontend/Cargo.toml
+++ b/crates/frontend/Cargo.toml
@@ -9,6 +9,7 @@ workspace = true
 
 [dependencies]
 binius-core = { path = "../core" }
+binius-utils = { path = "../utils" }
 cranelift-entity = { workspace = true }
 hex-literal.workspace = true
 num-bigint = { workspace = true }
--- a/crates/frontend/src/compiler/circuit.rs
+++ b/crates/frontend/src/compiler/circuit.rs
@@ -1,14 +1,16 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 use std::{error, fmt};
 
 use binius_core::{
 	constraint_system::{ConstraintSystem, ValueIndex, ValueVec},
 	word::Word,
 };
+use binius_utils::strided_array::StridedArray2DViewMut;
 use cranelift_entity::SecondaryMap;
 
 use crate::compiler::{
-	eval_form::EvalForm,
+	eval_form::{BatchPopulateError, EvalForm},
 	gate_graph::{GateGraph, Wire},
 };
 
@@ -150,6 +152,39 @@ impl Circuit {
 			.evaluate(&mut w.value_vec, Some(&self.gate_graph.path_spec_tree))?;
 
 		Ok(())
+	}
+
+	/// Populates non-input values for a batch of instances at once.
+	///
+	/// This is the structure-of-arrays counterpart to [`Self::populate_wire_witness`]. `values` is
+	/// the transposed value array: rows are value-vector indices (in the same order a single
+	/// instance's [`ValueVec`] uses) and columns are instances. Its height must be the full
+	/// value-vector length (including scratch) and its width is the instance count.
+	///
+	/// The caller must fill each instance's input rows first — the witness wires and any inout
+	/// wires. This function fills the constant rows (broadcasting each constant across every
+	/// instance) and then evaluates the circuit gate-by-gate for all instances.
+	///
+	/// # Errors
+	///
+	/// If any instance is not satisfiable, returns an error naming the lowest-indexed failing
+	/// instance and its assertion failures.
+	pub fn populate_wire_witness_batched(
+		&self,
+		values: &mut StridedArray2DViewMut<'_, Word>,
+	) -> Result<(), BatchPopulateError> {
+		// Broadcast each constant into its row across every instance. The constants are the same
+		// for all instances, so this fills the constant rows uniformly.
+		let n_instances = values.width();
+		for (index, &constant) in self.constraint_system.constants.iter().enumerate() {
+			for instance in 0..n_instances {
+				values[(index, instance)] = constant;
+			}
+		}
+
+		// Evaluate the bytecode across all instances, symbolicating assertion failures.
+		self.eval_form
+			.evaluate_batched(values, Some(&self.gate_graph.path_spec_tree))
 	}
 
 	/// Returns the constraint system for this circuit.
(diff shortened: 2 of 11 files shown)


jimpo:
@benji-potamus Write a benchmark that creates a circuit which does one Keccak permutation (using code from binius-circuits, Permutation::keccak_f1600). It force commits the output wires. For 8k instances, benchmark the time it takes for witness generation using ValueTable and with ValueTable2.

benji-potamus:
Added the benchmark in `crates/m4-prover/benches/keccak_witness_gen.rs` (latest commit). It builds a circuit that applies one `Permutation::keccak_f1600` to a 25-word state — state words are witness inputs, the permuted outputs are `force_commit`ed, so there are no inout wires — and times witness generation for 8192 instances (`log_instances = 13`).

### Results

| table | witness gen (8192 instances) |
|---|---|
| `ValueTable` | **394.9 ms** `[393.3, 396.8]` |
| `ValueTable2` | **250.2 ms** `[249.0, 251.5]` |

**`ValueTable2` is ~1.58× faster (~37% less time).** So the wire-major layout is already a win for witness generation on its own, before any of the downstream protocol benefits.

<sub>criterion, 10 samples; Intel Xeon Platinum 8259CL @ 2.50GHz.</sub>

### One caveat on threading

These are **single-threaded** numbers (default features, no `rayon`). `ValueTable::populate` parallelizes across instances via `par_chunks_mut`, so with `--features rayon` on this 4-core box it would pull ahead of the current `ValueTable2`, which is **not parallelized yet** — the batch interpreter walks instances in a scalar inner loop per opcode. That inner loop is trivially parallelizable (e.g. `StridedArray2DViewMut::into_par_strides` over instance stripes), so the fair end-state comparison should have both threaded. I kept this bench single-threaded to isolate the layout/interpreter effect; happy to add a rayon path for `ValueTable2` and a threaded comparison in this PR or a follow-up — let me know which you'd prefer.

jimpo:
@tcoratger I think we should just switch to this. There's probably some trait abstraction we could do over the interpreter context to deduplicate the BatchInterpreter and the regular Interpreter. (as follow-up).

When assembling the constraint witnesses (eg. BatchAndCheckWitness), it can do a bunch of strided copy/accumulations, which should be faster. And fold_instances should get faster with more data locality as well. Also, we can get rid of the internal padding of instance sizes up to powers-of-two.

benji-potamus:
Sounds good. Agreed a shared trait over the eval context is the clean way to collapse `BatchInterpreter` into `Interpreter` — happy to take that as a follow-up, along with the strided assembly in `BatchAndCheckWitness`, the `fold_instances` locality win, and dropping the power-of-two instance padding. I'll file those once this lands.

That also settles the threading question above: I'll leave `ValueTable2` single-threaded in this PR and parallelize the batch interpreter (rayon over instance stripes) as part of that follow-up work.

So this PR stays scoped to just the transposed storage + witness generation. It's green and ready whenever you want to pull it in.

tcoratger:
> @tcoratger I think we should just switch to this. There's probably some trait abstraction we could do over the interpreter context to deduplicate the BatchInterpreter and the regular Interpreter. (as follow-up).
> 
> When assembling the constraint witnesses (eg. BatchAndCheckWitness), it can do a bunch of strided copy/accumulations, which should be faster. And fold_instances should get faster with more data locality as well. Also, we can get rid of the internal padding of instance sizes up to powers-of-two.

I agree this looks better. So the plan here is to remove `ValueTable` progressively and to remplace by `ValueTable2` right?

benji-potamus:
Right, that's the intent. This PR adds `ValueTable2` **alongside** `ValueTable` — the issue keeps both initially so we can evaluate them side by side (hence the benchmark). Given the numbers and jimpo's note above, the plan is to migrate the M4 consumers over to `ValueTable2` and then remove `ValueTable`, together with the follow-ups jimpo listed: the shared-trait interpreter dedup, strided `BatchAndCheckWitness` assembly, `fold_instances` locality, dropping the power-of-two instance padding, and parallelizing the batch interpreter. I'll file those as follow-up issues once this lands so the migration is tracked.

jimpo:
Haha, Benji is very eager to respond. Yeah, I'm gonna merge this, then let's rework the M4 prover to use this instead. Then we'll remove ValueTable and rename.

jimpo (approved):

PR #1741, opened 2026-07-08
Parallelize ValueTable2 witness generation

latifkasuli:
## Summary
Follow-up to the ValueTable2 direction from #1716 and #1739: this implements the stripe-parallel witness-generation path Jimpo sketched for the wire-major M4 witness table.

Adds a stripe-based parallel path for batched witness evaluation over `ValueTable2`'s wire-major table.

`ValueTable2::populate_parallel` fills inputs once, broadcasts constants once, then evaluates disjoint vertical instance stripes concurrently using the measured 1024-column default. The benchmark-only `populate_parallel_with_stripe_width` exposes the stripe width for the sweep below. Each stripe runs the existing batch interpreter over its own `StridedArray2DViewMut` columns via `into_par_strides`; `Circuit::populate_wire_witness_batched_parallel` provides the lower-level frontend hook.

This introduces no new `unsafe`; it reuses the existing strided-array splitter. Output is identical to serial `ValueTable2::populate`, and assertion diagnostics preserve serial behavior: the lowest-indexed failing instance and its messages are reported, even when failures land in different stripes.

`ValueTable2` is the wire-major table intended to underpin the M4 prover migration, but today's full M4 prover still consumes `ValueTable`. So this PR lands the parallel witness-generation capability ahead of that migration rather than claiming an end-to-end prover speedup.

## Benchmark
Apple M4 Pro, `RAYON_NUM_THREADS=8`, `--features rayon`, 8192 Keccak-f1600 instances:

| path | time | vs current `ValueTable` |
|---|---:|---:|
| `ValueTable` (current, already rayon-parallel) | 15.9523 ms | baseline |
| `ValueTable2` serial | 22.4140 ms | +40.5% |
| `ValueTable2` parallel, stripe 256 | 7.3865 ms | -53.7% |
| `ValueTable2` parallel, stripe 512 | 6.6785 ms | -58.1% |
| `ValueTable2` parallel, stripe 1024 | **6.2891 ms** | **-60.6%** |

Headline: **-60.6% vs the current already-parallel `ValueTable` witness-generation path**. The serial `ValueTable2` row is included as context only.

The stripe-width sweep answers the sizing question directly: on this M4 Pro run, 1024 was fastest, with no diminishing returns yet across 256 -> 512 -> 1024. The likely explanation is that reduced rayon/task overhead outweighed any L1-locality pressure at these widths.

## Test plan
- `cargo test -p binius-m4-prover --lib` (33/33)
- `cargo test -p binius-m4-prover --features rayon --lib` (33/33)
- `cargo test -p binius-frontend --lib` (161/161)
- `cargo test -p binius-utils --lib` (30/30)
- `cargo test -p binius-utils --features rayon --lib` (18/18)
- `cargo clippy -p binius-utils -p binius-frontend -p binius-m4-prover --all-features --tests --benches -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-m4-prover --features rayon --document-private-items --no-deps`
- `cargo bench -p binius-m4-prover --features rayon --bench keccak_witness_gen -- --noplot`

### DIFF
--- a/crates/frontend/src/compiler/eval_form/mod.rs
+++ b/crates/frontend/src/compiler/eval_form/mod.rs
@@ -14,7 +14,7 @@ mod tests;
 
 pub use batch_interpreter::{BatchInterpreter, BatchPopulateError};
 use binius_core::{ValueIndex, ValueVec, Word};
-use binius_utils::strided_array::StridedArray2DViewMut;
+use binius_utils::{rayon::prelude::*, strided_array::StridedArray2DViewMut};
 pub use builder::BytecodeBuilder;
 pub use const_eval::evaluate_gate_constants;
 use cranelift_entity::SecondaryMap;
@@ -102,6 +102,31 @@ impl EvalForm {
 	) -> Result<(), BatchPopulateError> {
 		let mut interpreter = BatchInterpreter::new(&self.bytecode, &self.hint_registry);
 		interpreter.run(values, path_spec_tree)
+	}
+
+	/// Execute the evaluation form over disjoint vertical instance stripes in parallel.
+	pub fn evaluate_batched_parallel(
+		&self,
+		values: StridedArray2DViewMut<'_, Word>,
+		stripe_width: usize,
+		path_spec_tree: Option<&PathSpecTree>,
+	) -> Result<(), BatchPopulateError> {
+		assert!(stripe_width > 0, "stripe width must be positive");
+
+		values
+			.into_par_strides(stripe_width)
+			.enumerate()
+			.map(|(stripe_index, mut stripe)| {
+				let mut interpreter = BatchInterpreter::new(&self.bytecode, &self.hint_registry);
+				interpreter.run_with_instance_offset(
+					&mut stripe,
+					stripe_index * stripe_width,
+					path_spec_tree,
+				)
+			})
+			.collect::<Result<Vec<_>, _>>()?;
+
+		Ok(())
 	}
 
 	/// Get the number of evaluation instructions
--- a/crates/m4-prover/src/value_table2.rs
+++ b/crates/m4-prover/src/value_table2.rs
@@ -9,6 +9,9 @@ use binius_core::{
 use binius_frontend::{BatchPopulateError, Circuit, Wire};
 use binius_utils::strided_array::StridedArray2DViewMut;
 
+/// Default number of instance columns evaluated by one parallel witness-generation task.
+const DEFAULT_PARALLEL_STRIPE_WIDTH: usize = 1024;
+
 /// The witness for a batch of `2^k` independent instances of one circuit, in wire-major order.
 ///
 /// This is the transpose of [`ValueTable`](crate::ValueTable). Where `ValueTable` stores each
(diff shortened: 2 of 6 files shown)


crates/frontend/src/compiler/eval_form/mod.rs
```diff
@@ -104,6 +104,34 @@ impl EvalForm {
...
+	) -> Result<(), BatchPopulateError> {
+		assert!(stripe_width > 0, "stripe width must be positive");
+
+		let results = values
+			.into_par_strides(stripe_width)
+			.map(|mut stripe| {
+				let mut interpreter = BatchInterpreter::new(&self.bytecode, &self.hint_registry);
+				interpreter.run(&mut stripe, path_spec_tree)
+			})
+			.collect::<Vec<_>>();
```
jimpo: `.collect<Result<Vec<_>, _>>()?` should be fine here

benji-potamus (commented):
Thanks Latif — nicely done, and the PR description is exemplary (clear about what it does *not* claim: the full M4 prover still consumes `ValueTable`, so this is landing capability ahead of the migration, not an end-to-end prover speedup).

I read the parallel path closely and the correctness holds up:

- **Constants** are broadcast once over the full array before the split, matching `populate_wire_witness_batched` exactly.
- **Stripe partitioning** reuses the existing `into_par_strides` splitter (same non-overlapping-columns invariant as `par_iter_cols`), so no new `unsafe`, as advertised.
- **Local→global instance translation** in `note_assertion_failure` via `instance_offset()` is the right fix, and because `col_offset()` is 0 on the full-array serial path, existing serial behavior is untouched (the unchanged `batched_reports_lowest_failing_instance` test confirms this).
- **Diagnostics composition** is the subtle part, and it's correct: `check_assertions` recomputes `total_count`/`messages` for *only* the reported (lowest) instance, and that instance lives entirely within one stripe — so `min_by_key(err.instance)` over the per-stripe errors reproduces the serial report exactly. The two new cross-stripe diagnostics tests pin this down well.

No bugs found, and non-rayon builds are covered via the utils rayon shim (CI is green across all jobs).

So this isn't a code-correctness gate. I'm leaving it as a comment rather than an approval because the remaining questions are API-shape decisions that are really the core team's call — flagging them for @jimpo rather than deciding them here:

1. **New public surface across three crates** — `StridedArray2DViewMut::col_offset` (utils), `Circuit::populate_wire_witness_batched_parallel` + `EvalForm::evaluate_batched_parallel` (frontend), and `ValueTable2::populate_parallel` (m4-prover). The batch machinery is already public and this is a natural sibling, but it's worth a conscious decision about committing to this surface ahead of the M4 migration that will actually consume it.

2. **`stripe_width` as a caller-tuned public knob.** The benchmark shows the optimum is workload/hardware-dependent (1024 fastest on this M4 Pro run, no diminishing returns yet). Do we want callers to pass this, or would you rather pick a default internally (e.g. derived from instance count / thread count) and keep the public signature knob-free? Exposing it now is harder to walk back later.

3. **Landing ahead of a consumer.** Nothing in-tree calls `populate_parallel` yet except the bench. That's fine if it's on the intended path — just confirming it matches where you want `ValueTable2` to go, since the design (#1716/#1739 direction) is yours.

None of these block from my side; the implementation is clean and well-tested. Deferring the merge and the API calls to @jimpo.

crates/m4-prover/src/value_table2.rs
```diff
@@ -81,6 +81,40 @@ impl ValueTable2 {
...
+	/// once up front, then circuit evaluation runs over disjoint vertical stripes of at most
+	/// `stripe_width` instances.
+	///
+	/// # Panics
+	///
+	/// Panics if `stripe_width == 0` or if the circuit's layout has any inout wires.
+	pub fn populate_parallel<F>(
+		circuit: &Circuit,
+		log_instances: usize,
+		stripe_width: usize,
```
jimpo, line 128: The caller shouldn't have to provide this. Make a separate method `populate_parallel_with_stripe_width` for benchmarking, then `populate_parallel` simply calls it with the hardcoded default value.

crates/utils/src/strided_array.rs
```diff
@@ -71,6 +71,11 @@ impl<'a, T> StridedArray2DViewMut<'a, T> {
 		self.cols.end - self.cols.start
 	}
 
+	/// The column offset of this view within the original backing array.
+	pub const fn col_offset(&self) -> usize {
```
jimpo: This method shouldn't be necessary and breaks the abstraction. Remove it please. When creating the `BatchInterpreter`, you can iterate with `par_iter_strides().enumerate()` and give the index to `BatchInterpreter`, if you'd like.

jimpo (changes requested):

jimpo (approved):

PR #1742, opened 2026-07-08
BINIUS-250: sliced multiplication for the degree-2 Monbijou extension

benji-potamus:
Implements [BINIUS-250](https://linear.app/jimpo/issue/BINIUS-250/arith-bench-slice-multiplication-algorithm-for-degree-2-monbijou) (experiment).

Adds `mul_sliced_128b_clmul` in `crates/arith-bench/src/monbijou.rs` — a degree-2 Monbijou extension multiply in **sliced** representation, alongside the existing packed `mul_128b_clmul`. Each element is `[U; 2]` with coeff 0 at index 0 and coeff 1 at index 1 (each `U` a SIMD pack of base-field GF(2^64) elements), so the two coefficients live in separate registers and the multiply is built from base-field muls, processing every lane in parallel.

For `Y² = XY + 1`, writing `a = a0 + a1·Y`, `b = b0 + b1·Y`:
```
coeff 0 = a0·b0 + a1·b1
coeff 1 = a0·b1 + a1·b0 + X·(a1·b1)
```
using Karatsuba for the cross term (3 base-field muls) and the same shift+`0x1B` trick for the multiply-by-X.

**Correctness:** a proptest verifies it produces the same field product as the trusted packed `mul_128b_clmul` for both `__m128i` lanes.

**Benchmark (`monbijou_128b` group)** added packed-vs-sliced entries for `__m128i`/`__m256i`/`uint64x2_t`.

Marking **draft** since this is an experiment with no fixed acceptance criterion — the numbers and the question of whether to pursue it are below.

### DIFF
--- a/crates/arith-bench/benches/field_mul.rs
+++ b/crates/arith-bench/benches/field_mul.rs
@@ -563,16 +563,18 @@ fn bench_monbijou(c: &mut Criterion) {
 	group.finish();
 }
 
-/// Benchmark GF(2^128) Monbijou 128-bit extension field multiplication using CLMUL instructions
+/// Benchmark GF(2^128) Monbijou 128-bit extension field multiplication using CLMUL instructions,
+/// comparing the packed representation (`mul_128b_clmul`) against the sliced representation
+/// (`mul_sliced_128b_clmul`).
 #[allow(unused_imports, unused_variables, unused_mut)]
 fn bench_monbijou_128b(c: &mut Criterion) {
-	use binius_arith_bench::monbijou::mul_128b_clmul;
+	use binius_arith_bench::monbijou::{mul_128b_clmul, mul_sliced_128b_clmul};
 
 	let mut rng = rand::rng();
 
 	let mut group = c.benchmark_group("monbijou_128b");
 
-	// Benchmark __m128i
+	// Packed __m128i
 	#[cfg(all(target_feature = "pclmulqdq", target_feature = "sse2"))]
 	{
 		run_mul_benchmark(
@@ -584,7 +586,7 @@ fn bench_monbijou_128b(c: &mut Criterion) {
 		);
 	}
 
-	// Benchmark __m256i
+	// Packed __m256i
 	#[cfg(all(
 		target_feature = "vpclmulqdq",
 		target_feature = "avx2",
@@ -600,7 +602,7 @@ fn bench_monbijou_128b(c: &mut Criterion) {
 		);
 	}
 
-	// Benchmark uint64x2_t (AARCH64 NEON)
+	// Packed uint64x2_t (AARCH64 NEON)
 	#[cfg(all(
 		target_arch = "aarch64",
 		target_feature = "neon",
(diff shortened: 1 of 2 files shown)


benji-potamus:
### Benchmark: `monbijou_128b` (packed vs sliced)

criterion, per-element throughput (higher is better). Machine: Intel Xeon Platinum 8259CL (Cascade Lake) — has `pclmulqdq` + `avx512f` but **no `vpclmulqdq`/`gfni`**, so only the `__m128i` paths run here.

| function | throughput |
|---|---|
| `mul_128b_clmul::<__m128i>` (packed) | **247 Melem/s** |
| `mul_sliced_128b_clmul::<__m128i>` (sliced) | **218 Melem/s** |

So on 128-bit lanes the sliced form is ~**12% slower**.

### Why, and the honest caveats

- **Reduction count.** The clean sliced version does 3 base-field `mul_clmul`s (Karatsuba), i.e. 3 GF(2^64) reductions to produce 2 elements → ~9 CLMULs/elem, vs the packed path's ~8 CLMULs/elem with a single `reduce_pair`. That extra reduction is most of the gap.
- **No wide CLMUL here.** The sliced representation's expected payoff is on `__m256i`/`__m512i` with `VPCLMULQDQ` (many elements per register, no cross-128-bit-lane shuffles). This box can't run those, so I couldn't measure the case that's supposed to favor slicing. The `__m256i`/`uint64x2_t` bench entries are wired up and cfg-gated, ready to run on VPCLMULQDQ or aarch64 hardware.
- **Possible fairer/faster variant.** A deferred-reduction version — compute the 6 unreduced lane products, combine, and `reduce_pair` the two output coefficients once each (1 reduction/elem, matching packed) — should close or reverse the `__m128i` gap. I kept the readable 3-mul version here (it mirrors the existing `ghash_sq::mul_sliced` convention); happy to add the deferred-reduction variant if the 128-bit path matters.

**Question for you:** is `__m128i` the case you care about (then the deferred-reduction variant is the next step), or is this mainly to have the sliced algorithm ready for VPCLMULQDQ/aarch64 measurement? Happy to take it either direction, or land the algorithm + benchmark as-is.

jimpo:
Do the deferred reduction version. Testing with __m128i is fine.

benji-potamus:
Done — switched `mul_sliced_128b_clmul` to the deferred-reduction version: it keeps the six base-field products unreduced, assembles the two output coefficients, and spends a single `reduce_pair` per coefficient (one reduction/element, like the packed path). The proptest against packed `mul_128b_clmul` still passes (`reduce_pair` is linear, so results are identical to the previous 3-mul version).

### Updated benchmark (`monbijou_128b`, same run)

| function | throughput |
|---|---|
| `mul_128b_clmul::<__m128i>` (packed) | 267 Melem/s |
| `mul_sliced_128b_clmul::<__m128i>` (sliced) | **288 Melem/s** |

Sliced went from 218 → **288 Melem/s** (+32%) and now **beats packed by ~8%** on `__m128i` — the expected result of dropping from 3 reductions/2 elements to 1 reduction/element (~7 vs ~8 CLMULs/element). Marking ready for review.

<sub>criterion, Intel Xeon Platinum 8259CL.</sub>

jimpo (approved):

PR #1744, opened 2026-07-08
soft64 portable multiplication for the Monbijou field (BINIUS-252)

benji-potamus:
Implements [BINIUS-252](https://linear.app/jimpo/issue/BINIUS-252/arith-bench-soft64-multiplication-algorithms-for-monbijou-field) (experiment).

Promotes `crates/arith-bench/src/monbijou.rs` to a directory and adds a `soft64` submodule with **portable** (no CLMUL / no SIMD) multiplication, mirroring `ghash::soft64`:

- **`soft64::mul(u64, u64) -> u64`** — the base field GF(2^64). Full 128-bit carry-less product via `bmul64` + the `rev64` trick, then reduction mod `X^64 + X^4 + X^3 + X + 1` (fold `hi·X^64 ≡ hi·0x1B`, twice).
- **`soft64::mul_128b(u128, u128) -> u128`** — the degree-2 extension GF(2^128), `Y² = XY + 1`, packed like `mul_128b_clmul` (coeff 0 low, coeff 1 high). Karatsuba (3 base-field muls) + multiply-by-X.

**Correctness:** proptests check field axioms (commutative / associative / distributive / identity) portably, plus x86_64-gated equivalence to the trusted `mul_clmul` / `mul_128b_clmul`.

**Scope note (for your call):** I kept the existing CLMUL code in `monbijou/mod.rs` and only added `monbijou/soft64.rs` — the minimal reading of "promote to a directory + add a soft64 submodule". `ghash` additionally splits its CLMUL into `ghash/clmul.rs`; happy to do the same split here if you'd prefer the closer mirror.

Marking **draft** (experiment) — benchmark numbers below.

### DIFF
--- a/crates/arith-bench/src/monbijou/soft64.rs
+++ b/crates/arith-bench/src/monbijou/soft64.rs
@@ -1,0 +1,173 @@
+// Copyright 2026 The Binius Developers
+//! Portable (non-CLMUL) multiplication for the Monbijou field and its degree-2 extension.
+//!
+//! These are software implementations of the algorithms in the [parent module](super), built on
+//! the portable carry-less multiply primitives in `crate::arch::portable64` rather than on CLMUL or
+//! SIMD intrinsics. They operate on scalar `u64`/`u128` values, one field element at a time.
+
+use crate::arch::portable64::{bmul64, rev64};
+
+/// The 128-bit carry-less product of two 64-bit polynomials, as `(low, high)` 64-bit limbs.
+///
+/// [`bmul64`] gives the low limb directly. The high limb comes from multiplying the bit-reversed
+/// operands and reversing the result: reversing a product of two degree-`<64` polynomials leaves it
+/// shifted by one bit, which the final shift removes.
+#[inline]
+fn clmul64(x: u64, y: u64) -> (u64, u64) {
+	let lo = bmul64(x, y);
+	let hi = rev64(bmul64(rev64(x), rev64(y))) >> 1;
+	(lo, hi)
+}
+
+/// Reduces a 128-bit carry-less product, given as `(low, high)` limbs, modulo the Monbijou
+/// polynomial X^64 + X^4 + X^3 + X + 1.
+///
+/// The high limb holds the coefficients of X^64..X^127. Folding it down is a carry-less multiply by
+/// X^64 ≡ `0x1B` (`= 1 + X + X^3 + X^4`). The left shifts drop the bits that spill past X^63; the
+/// matching right shifts collect those as the coefficients of X^64..X^67, which fold in once more.
+#[inline]
+const fn reduce64(lo: u64, hi: u64) -> u64 {
+	// The bits of hi·0x1B that spill past X^63 (from the <<1, <<3, <<4 terms): coefficients
+	// X^64..X^67.
+	let spill = (hi >> 63) ^ (hi >> 61) ^ (hi >> 60);
+	let lo = lo ^ hi ^ (hi << 1) ^ (hi << 3) ^ (hi << 4);
+	// spill < 2^4, so folding it back in (spill·X^64 ≡ spill·0x1B) no longer spills past X^63.
+	lo ^ spill ^ (spill << 1) ^ (spill << 3) ^ (spill << 4)
+}
+
+/// Multiplies two elements of the base field GF(2^64), the Monbijou field.
+#[inline]
+pub fn mul(x: u64, y: u64) -> u64 {
+	let (lo, hi) = clmul64(x, y);
+	reduce64(lo, hi)
+}
+
+/// Multiplies two elements of GF(2^128), the degree-2 extension of the Monbijou field.
+///
+/// The element `a0 + a1·Y` is packed into a `u128` with `a0` in the low 64 bits and `a1` in the
+/// high 64 bits, matching [`super::mul_128b_clmul`]. The extension is `Y^2 = XY + 1`, so for
+/// `a = a0 + a1·Y` and `b = b0 + b1·Y`:
+///
+/// ```text
+/// coeff 0 = a0·b0 + a1·b1
+/// coeff 1 = a0·b1 + a1·b0 + X·(a1·b1)
+/// ```
+///
+/// with Karatsuba recovering the cross term `a0·b1 + a1·b0` as `t1 + t0 + t2`. The three base-field
+/// products are kept unreduced and the multiply-by-X is a plain shift of the unreduced `a1·b1`, so
+/// only the two output coefficients need reducing — two reductions instead of three, matching the
+/// CLMUL path's deferred reduction.
+#[inline]
+pub fn mul_128b(x: u128, y: u128) -> u128 {
+	let (x0, x1) = (x as u64, (x >> 64) as u64);
+	let (y0, y1) = (y as u64, (y >> 64) as u64);
+
+	// Unreduced 128-bit carry-less products, as (low, high) limbs.
+	let (t0l, t0h) = clmul64(x0, y0); // a0·b0
+	let (t2l, t2h) = clmul64(x1, y1); // a1·b1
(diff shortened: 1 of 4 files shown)


benji-potamus:
### Benchmark: soft64 throughput vs GHASH soft64

criterion, per-element throughput (higher is better). Machine: Intel Xeon Platinum 8259CL. All portable (no CLMUL/SIMD).

| field | function | throughput |
|---|---|---|
| Monbijou GF(2^64) | `monbijou::soft64::mul` | **45.7 Melem/s** |
| Monbijou GF(2^128) ext | `monbijou::soft64::mul_128b` | **12.8 Melem/s** |
| GHASH GF(2^128) | `ghash::soft64::mul` | **20.0 Melem/s** |

<sub>For reference on the same box: `polyval::soft64::mul` 16.9, `ghash_sq::soft64::mul_sliced` (GF(2^256)) 6.1 Melem/s.</sub>

**Comparison (the 128-bit ask):** the Monbijou 128-bit extension soft64 (12.8) is ~36% slower than GHASH soft64 (20.0). GHASH is a *native* GF(2^128) — 6 `bmul64`s and one reduction — whereas Monbijou's GF(2^128) is a degree-2 tower over GF(2^64), so each extension mul is **3 base-field GF(2^64) muls** (each its own `bmul64`/`rev64` pair + reduction) plus the Karatsuba combine and multiply-by-X. That's the source of the gap.

The base field itself is cheap: `monbijou::soft64::mul` at 45.7 Melem/s is ~2.3× GHASH's 128-bit mul, roughly as expected for a 64-bit field with a sparse (`0x1B`) reduction polynomial.

crates/arith-bench/src/monbijou/soft64.rs
```diff
@@ -0,0 +1,172 @@
...
+/// coeff 1 = a0·b1 + a1·b0 + X·(a1·b1)
+/// ```
+///
+/// with Karatsuba recovering the cross term `a0·b1 + a1·b0` as `t1 + t0 + t2`.
+#[inline]
+pub fn mul_128b(x: u128, y: u128) -> u128 {
+	let (x0, x1) = (x as u64, (x >> 64) as u64);
+	let (y0, y1) = (y as u64, (y >> 64) as u64);
+
+	let t0 = mul(x0, y0); // a0·b0
```
jimpo: Can't we delay the expensive reductions and do only 2 instead of 3? Like in the clmul version

crates/arith-bench/src/monbijou/soft64.rs
```diff
@@ -0,0 +1,172 @@
...
+/// polynomial X^64 + X^4 + X^3 + X + 1.
+///
+/// The high limb holds the coefficients of X^64..X^127. Since X^64 ≡ [`POLY`], folding it down is a
+/// carry-less multiply by `POLY` (`= 1 + X + X^3 + X^4`), which can spill up to four bits back
+/// above X^63 and so needs one more fold.
+#[inline]
+const fn reduce64(lo: u64, hi: u64) -> u64 {
+	// hi·X^64 ≡ hi·POLY. POLY has bits {0, 1, 3, 4}, so this is hi ^ (hi<<1) ^ (hi<<3) ^ (hi<<4),
+	// widened to keep the up-to-four overflow bits above X^63.
+	let h = hi as u128;
```
jimpo: Don't do this as u128, do it as u64s with left shifts and right shifts. Like how the ghash soft64 reduction looks

crates/arith-bench/src/monbijou/soft64.rs
```diff
@@ -0,0 +1,172 @@
...
+pub fn mul(x: u64, y: u64) -> u64 {
+	let (lo, hi) = clmul64(x, y);
+	reduce64(lo, hi)
+}
+
+/// Multiplies an element of GF(2^64) by the generator X.
+#[inline]
+const fn mul_x(x: u64) -> u64 {
+	// x·X = x << 1, folding the single bit that crosses X^63 back in with POLY.
+	(x << 1) ^ (POLY & (x >> 63).wrapping_neg())
```
jimpo: Use (x as i64 >> 63) to make the mask

crates/arith-bench/src/monbijou/soft64.rs
benji-potamus: Done — `mul_128b` now keeps the three base-field products unreduced (`clmul64`) and reduces only the two output coefficients, matching the CLMUL path's deferred reduction (2 reductions instead of 3). The multiply-by-X becomes a plain 1-bit shift of the unreduced `a1·b1` (degree ≤ 126, so it stays within 128 bits), folded into `coeff 1` before its single reduction. Bit-identical to before by linearity of reduction — the equivalence proptest against `mul_128b_clmul` still passes.

Throughput went **12.8 → 17.0 Melem/s** (+33%), now ~15% under GHASH soft64 (20.0) rather than ~36%.

crates/arith-bench/src/monbijou/soft64.rs
benji-potamus: Done — `reduce64` now folds the high limb with `u64` left/right shifts, no `u128`: the left shifts (`hi ^ (hi<<1) ^ (hi<<3) ^ (hi<<4)`) drop the bits that spill past X^63, and the matching right shifts (`(hi>>63) ^ (hi>>61) ^ (hi>>60)`) collect those spilled coefficients (X^64..X^67) to fold in once more — same shape as the ghash soft64 reduction.

crates/arith-bench/src/monbijou/soft64.rs
benji-potamus: Removed `mul_x` entirely — the deferred-reduction `mul_128b` no longer multiplies a *reduced* field element by X; it shifts the unreduced `a1·b1` product up one bit and folds it into `coeff 1`'s single reduction, so there's no conditional mask left to build. (Happy to reintroduce a `mul_x` primitive with the `(x as i64 >> 63)` mask if you want it exposed for its own sake.)

benji-potamus:
### Updated benchmark (after deferred reduction + u64-shift reduce)

| field | function | before | after |
|---|---|---|---|
| Monbijou GF(2^64) | `monbijou::soft64::mul` | 45.7 | **47.0 Melem/s** |
| Monbijou GF(2^128) ext | `monbijou::soft64::mul_128b` | 12.8 | **17.0 Melem/s** |
| GHASH GF(2^128) | `ghash::soft64::mul` | — | 20.0 Melem/s |

The extension mul is up **+33%** (12.8 → 17.0) from dropping 3 reductions to 2. It's now ~15% under GHASH soft64 (was ~36%) — the residual gap is the tower construction's 3 base-field carry-less products (vs GHASH's direct 6-`bmul64` layout). <sub>criterion, Xeon 8259CL.</sub>

jimpo (approved):

jimpo:
I moved out of draft. Move the clmul code to a clmul file within monbijou

benji-potamus:
This is approved and all 8 checks are green on `0645827f` (`mergeable_state: clean`), but the branch-protection rule requires the merge queue and I can't enqueue via the API (`405 Changes must be made through the merge queue`). Could you click **Merge when ready** to enqueue it? I'll take it from there (restack + close out the issue) once it lands.

benji-potamus:
Done — moved the CLMUL code into `monbijou/clmul.rs`, mirroring `ghash`/`polyval`: `mod.rs` is now just the umbrella (`pub mod clmul; pub mod soft64;` + the field-identity consts) and re-exports `clmul::{mul → mul_clmul, mul_128b → mul_128b_clmul, mul_sliced_128b → mul_sliced_128b_clmul}`, so all the `monbijou::mul_*_clmul` paths (benches, arch tests, `soft64`) are unchanged. Pushed as `59ffc3b7`; all 72 tests + clippy/fmt/rust-doc green.

Heads up: the new push likely dismissed your approval — whenever you get a chance, could you re-approve and click **Merge when ready**? (I still can't enqueue via the API.)

PR #1745, opened 2026-07-08
BINIUS-244: pin the all-one wire to the first constant slot

tcoratger:
Closes [BINIUS-244](https://linear.app/jimpo/issue/BINIUS-244).

Makes the all-one constant live at a fixed, known index in the value vector. Pure refactor — no behavior change, proofs are bit-identical.

### The problem

Every circuit build injects an all-one constant (the `build()` step adds `Word::ALL_ONE`).
But the constants segment was ordered by a plain ascending sort on the word value.
The all-one word is `u64::MAX`, so it sorted to the **last** slot of the segment.

That means its index was incidental — it depended on how many other constants happened to be present.
Any code that wants the all-one constant has to thread its wire all the way through the compilation pipeline to recover the index.

### The idea

Order the constants segment so the all-one word is always first, and every other constant stays ascending.

The sort key becomes a pair:

```
key(value) = (value != ALL_ONE, value)
```

- The first component is `false` (sorts first) only for the all-one word.
- It is uniformly `true` when no all-one constant is present, so the section stays plain ascending.

Result: the all-one constant sits at the fixed index `0`, and downstream code can reference it there directly instead of threading its wire.

### The change

```diff
- // ascending by value → ALL_ONE (u64::MAX) lands last
- self.w_const.sort_by_key(|&(_, value)| value);
+ // ALL_ONE leads, the rest ascending → ALL_ONE lands at index 0
+ self.w_const.sort_by_key(|&(_, value)| (value != Word::ALL_ONE, value));
```

A debug assertion in the build path pins the invariant right where the constant is injected:

```
debug_assert_eq!(wire_mapping[all_one], ValueIndex(0));
debug_assert_eq!(constants.first(), Some(&Word::ALL_ONE));
```

### Soundness

- No protocol change: the constant *set* is identical, only its ordering in the segment changes.
- Value indices are assigned consistently through `wire_mapping`, so every constraint still refers to the same words.
- All existing circuits prove and verify exactly as before.

### Scope

- **changed:** the constant sort in the value-vector allocator (`value_vec_alloc.rs`), one call site.
- **new:** a debug-assert invariant in the build path (`compiler/mod.rs`) and a unit test.
- **left untouched:** the explicit `all_one` wire threading in the constraint builder and gate fusion — removing that in favor of the fixed index `0` is a natural follow-up, deliberately out of scope here.

### Verification

- `cargo check -p binius-frontend --all-targets` and `-p binius-circuits --all-targets` — clean.
- `cargo clippy -p binius-frontend --all-targets` — clean.
- `cargo test -p binius-frontend --lib` — 159 passed; `cargo test -p binius-circuits --lib` — 336 passed (full prove/verify roundtrips).

Note: `cargo check --workspace --all-targets` currently fails in `binius-verifier` under workspace-wide feature unification, but that crate does not depend on `binius-frontend` and the failure reproduces independently of this change.

### DIFF
--- a/crates/frontend/src/compiler/value_vec_alloc.rs
+++ b/crates/frontend/src/compiler/value_vec_alloc.rs
@@ -53,7 +53,7 @@ impl Alloc {
 		self.w_scratch.push(wire);
 	}
 
-	pub fn into_assignment(mut self) -> Assignment {
+	pub fn into_assignment(self) -> Assignment {
 		// `ValueVec` expects the wires to be in a certain order. Specifically:
 		//
 		// 1. const
@@ -72,16 +72,9 @@ impl Alloc {
 		let n_internal = self.w_internal.len();
 		let n_scratch = self.w_scratch.len();
 
-		// Order the constant section so the all-one word is first, then all others ascending.
-		//
-		// Every circuit build injects an all-one constant.
-		// Downstream code reads it at the fixed index 0 instead of threading its wire.
-		//
-		// Sort key `(value != ALL_ONE, value)`:
-		// - The first component is `false` (sorts first) only for the all-one word.
-		// - It is uniformly `true` when no all-one constant is present, leaving plain ascending.
-		self.w_const
-			.sort_by_key(|&(_, value)| (value != Word::ALL_ONE, value));
+		// Constants keep the order in which they were added, which is wire-creation order.
+		// The gate graph seeds the all-one constant first (see `GateGraph::new`).
+		// So it is the first constant here and lands at value index 0.
 
 		// First, allocate the indices for the public section of the value vec. The public section
 		// consists of constant wires followed by inout wires.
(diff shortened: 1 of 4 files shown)


crates/frontend/src/compiler/value_vec_alloc.rs
```diff
@@ -72,9 +72,16 @@ impl Alloc {
 		let n_internal = self.w_internal.len();
 		let n_scratch = self.w_scratch.len();
 
-		// Sort the wires pointing to the constant section of the input value vector ascending
-		// to their values.
-		self.w_const.sort_by_key(|&(_, value)| value);
+		// Order the constant section so the all-one word is first, then all others ascending.
+		//
+		// Every circuit build injects an all-one constant.
```
jimpo: It's not clear from this comment that this is true. It'd be nice for this to be true "from construction" more (like, the const pool is seeded with ALL_ONE as Wire 0 before anything else happens).

jimpo (changes requested):

crates/frontend/src/compiler/value_vec_alloc.rs
tcoratger: Fixed with the latest commit

jimpo (approved):

PR #1746, opened 2026-07-08
BINIUS-243: move gadget hints from binius-frontend to binius-circuits

tcoratger:
Closes [BINIUS-243](https://linear.app/jimpo/issue/BINIUS-243).

Moves the six concrete gadget hints out of the frontend and next to the circuit gadgets that use them. Pure refactor — no behavior change, all hint logic is byte-identical.

### The problem

The `hints` module conflated two unrelated things:

- The **hint infrastructure** — the `Hint` trait, the `HintRegistry`, the id derivation. This is the compiler's generic extension point.
- Six **concrete gadget hints** — big-integer division, modular exponentiation/inverse/division, a secp256k1 endomorphism split, and a byte-vector concatenation.

The concrete hints touch nothing private: just the public `Hint` trait plus `CircuitBuilder` / `Wire` / `Word`. And every one of their real callers already lives in `binius-circuits`. So they had no reason to sit in the frontend — a circuit gadget and the hint it relies on were split across two crates.

### The idea

Keep the infrastructure in `binius-frontend::compiler::hints`, and define each concrete hint in the module of the gadget that calls it.

| Hint | New home | Caller |
|---|---|---|
| `BigUintDivideHint` | `bignum/big_uint_divide.rs` | `bignum/prime_field.rs` |
| `ModDivideHint` | `bignum/mod_divide.rs` | `bignum/prime_field.rs` |
| `ModInverseHint` | `bignum/mod_inverse.rs` | `bignum/prime_field.rs` |
| `BigUintModPowHint` | `bignum/big_uint_mod_pow.rs` | `secp256k1/curve.rs` |
| `Secp256k1EndosplitHint` | `secp256k1/endosplit.rs` | `ecdsa/scalar_mul.rs` |
| `ByteVecConcatHint` | inlined into `concat.rs` (now private) | `concat.rs` |

Two placement calls worth flagging:

- `BigUintModPowHint` lives in `bignum` next to the other three big-integer / modular hints, even though its only current caller is `secp256k1/curve.rs`. It is a generic big-integer primitive, so grouping the arithmetic hints reads better than splitting one off. Trivial to relocate if you'd rather it sit by its caller.
- `ByteVecConcatHint` is small and has a single in-module caller, so it is inlined into `concat.rs` and made private — as close to the gadget as possible.

### The change

```diff
- // crates/frontend/src/compiler/hints/big_uint_divide.rs
- use super::Hint;
- use crate::{compiler::{CircuitBuilder, Wire}, util::num_biguint_from_u64_limbs};
+ // crates/circuits/src/bignum/big_uint_divide.rs
+ use binius_frontend::{CircuitBuilder, Wire, hints::Hint, util::num_biguint_from_u64_limbs};
```

Callers now import the hint from its new home, e.g.:

```diff
- use binius_frontend::{CircuitBuilder, Wire, hints::Secp256k1EndosplitHint};
+ use binius_frontend::{CircuitBuilder, Wire};
+ use crate::secp256k1::Secp256k1EndosplitHint;
```

### Tests

Frontend tests cannot depend on `binius-circuits`, so:

- The six concrete-hint tests move with their hints (into `#[cfg(test)]` blocks next to each) and now run in `binius-circuits`.
- `const_prop.rs`'s constant-propagation-through-a-hint test is rewritten to use a tiny local `DivRemHint`. That test exercises frontend infrastructure — propagation across a hint gate — so it should not lean on a concrete circuit hint. The local-hint pattern already exists in the frontend for exactly this.

### Scope

- **moved:** five hint files (frontend → circuits, recorded as git renames) plus one inlined into `concat.rs`.
- **frontend `hints` module:** now infrastructure only (`Hint`, `HintId`, `hint_id_of`, `HintRegistry`).
- **left untouched:** all hint logic — `shape` / `execute` bodies are byte-identical, and the `NAME` strings are unchanged.

### Verification

- `cargo check -p binius-frontend -p binius-circuits --all-targets` — clean.
- `cargo clippy -p binius-frontend -p binius-circuits --all-targets` — clean.
- `cargo +nightly fmt --all` — clean.
- `binius-frontend` lib — 152 passed; `binius-circuits` lib — 342 passed (includes the six relocated tests).
- `cargo build -p binius-examples --example basics_nondeterministic` — builds (the updated example import).

Note: `cargo check --workspace --all-targets` trips a pre-existing `binius-ip` feature-unification error unrelated to this change — `binius-ip` does not depend on `binius-frontend` or `binius-circuits`, and the affected example builds fine under its real feature set.


benji-potamus (approved):
Clean refactor — this is the right split. The frontend `hints` module keeps the generic extension point (`Hint`, `HintId`, `hint_id_of`, `HintRegistry`) and each concrete gadget hint moves next to the gadget that calls it. The hint bodies are byte-identical, `NAME` strings unchanged, tests travel with their hints, and rewriting `const_prop.rs`'s test to a tiny local `DivRemHint` is exactly right — that test exercises constant-propagation-across-a-hint infrastructure and shouldn't reach into a concrete circuit hint to do it.

I verified the rebased result locally: `binius-frontend` lib 155 passed, `binius-circuits` lib 342 passed (the six relocated tests included), both clean.

The two placement calls you flagged both read fine to me:
- `BigUintModPowHint` grouped with the other big-integer/modular hints in `bignum` rather than by its lone secp256k1 caller — it's a generic big-integer primitive, so grouping the arithmetic hints reads better. Agreed.
- `ByteVecConcatHint` inlined into `concat.rs` and made private — good, that's the tightest possible scope for a single in-module caller, and dropping the now-pointless `Default` impl along with it is a nice touch.

**One thing before this can merge: it needs a rebase onto `main`** (GitHub currently shows it as conflicting). I checked out and rebased it locally — there's exactly **one** trivial textual conflict, in `crates/frontend/src/compiler/tests.rs`: #1747 landed `test_all_one_is_first_constant()` on `main` right at the boundary of the hint-test block this PR removes. The resolution is mechanical — keep the new `test_all_one_is_first_constant()` test and apply your deletion of the hint tests; there's zero semantic overlap. I confirmed both crates build and test green after resolving it that way, so it's purely a rebase-and-push, no code changes needed.

Approving on content. Leaving the actual merge to @jimpo once it's rebased.

jimpo (approved):

PR #1747, opened 2026-07-08
BINIUS-57: add SlicedPackedField and packed GhashSq256b

benji-potamus:
Closes [BINIUS-57](https://linear.app/jimpo/issue/BINIUS-57/packedfields-for-ghashsq-with-sliced-layout).

A fresh implementation of the sliced GHASH² packing following [jimpo's direction on the issue](https://linear.app/jimpo/issue/BINIUS-57): a **generic** `SlicedPackedField` in the field crate, blanket-implementing the whole `PackedField` surface, with only the field multiplication supplied per concrete extension. Supersedes #1600 (no dependency on it).

### The generic type

```rust
pub struct SlicedPackedField<F, PSub, const N: usize>([PSub; N], PhantomData<F>);
```

An extension element `x = c₀·β₀ + … + c_{N-1}·β_{N-1}` is a vector of `N` subfield coordinates. The sliced layout stores each **coordinate** of every lane in its own packed subfield register (struct-of-arrays):

```
coords[0] = [ c₀(x₀), c₀(x₁), … ]   // β₀ coordinate of every lane
coords[1] = [ c₁(x₀), c₁(x₁), … ]   // β₁ coordinate of every lane
```

so a batch multiply runs as a handful of packed subfield multiplies over the whole batch (Karatsuba over the extension) instead of a schoolbook product per lane.

### Generic vs. concrete

Everything that doesn't depend on the multiplication rule is provided **generically** for any `PSub: PackedField` and `F: ExtensionField<PSub::Scalar>`: scalar access, broadcast, iteration, `Add`/`Sub`/`Neg`/`Sum`, masking, `interleave`/`unzip`/`spread`, `square_transpose`, `Random`, `Zeroable`. The layout makes these uniform — a lane permutation or bitwise op applies to each coordinate register identically, and scalar access reads/writes the `N` registers at one lane through the `ExtensionField` basis.

Only the field **arithmetic** (`Mul`, `Square`, `InvertOrZero`) is written per concrete extension and required as a bound on the blanket `PackedField` impl (`where Self: Mul<Output = Self> + Square + InvertOrZero`), exactly mirroring how the scalar fields are defined.

**On the `F` parameter:** jimpo's sketch was `SlicedPackedField<PSub, N>`, but a `(subfield, degree)` pair doesn't name a unique extension, and Rust rejects a type parameter that appears only as `type Scalar = F` (E0207). So `F` is carried as a phantom, exactly as `PackedPrimitiveType<U, Scalar>` carries its scalar. Callers never write it — they use the aliases.

### The GHASH² instantiation

`PackedGhashSq{1,2,4}x256b = SlicedPackedField<GhashSq256b, PackedBinaryGhash{1,2,4}x128b, 2>`. `Mul`/`Square`/`InvertOrZero` are the packed analogues of the scalar `GhashSq256b` arithmetic, written once, generically over any `PSub: PackedField<Scalar = BinaryField128bGhash>`, so all three widths share one implementation.

### One deliberate follow-up: 3 vs 4 multiplies

The reduction `Y² = Y + X⁻¹` needs `t2·X⁻¹`, done here as one extra packed multiply (`t2 * INV_X`), so a batch multiply is **4** packed GHASH multiplies. The optimal sliced multiply is **3** (`ghash::clmul::mul_inv_x` in `binius_arith_bench`). A fast packed `mul_inv_x` needs a per-128-bit-lane shift; `Underlier128bLanes` covers only the 128-bit `M128`, so widths 2/4 (`M256`/`M512`) would need arch code. I scoped that out of this PR to keep it focused on the generic pattern — happy to fold it in here or file it separately, whichever you prefer.

### Verification

- `cargo test -p binius-field` — mul/square/invert pinned lane-by-lane against the scalar `GhashSq256b` reference for all three widths; get/set/broadcast/iter round trips; interleave/unzip against independent scalar oracles; the three widths added to the shared packed iteration + `square_transpose` suites (degree-2 over GHASH and degree-256 over B1b).
- `clippy -p binius-field --all-targets -D warnings`, nightly `fmt` — clean.
- `cargo check -p binius-field` cross-compiles clean on `aarch64-unknown-linux-gnu` and `wasm32-unknown-unknown` (the new code is fully arch-portable).

### DIFF
--- a/crates/field/src/packed_ghash_sq.rs
+++ b/crates/field/src/packed_ghash_sq.rs
@@ -1,0 +1,389 @@
+// Copyright 2026 The Binius Developers
+
+//! Packed [`GhashSq256b`] in a sliced (struct-of-arrays) layout.
+//!
+//! [`GhashSq256b`] is the degree-two extension `a + b·Y` of the GHASH field, with `Y² = Y + X⁻¹`.
+//! The packings here store the `a` and `b` coordinates of every lane in two separate packed GHASH
+//! registers via [`SlicedPackedField`], so a batch multiply runs as three packed GHASH multiplies
+//! over the whole batch (Karatsuba over `Y`) rather than a schoolbook product per element.
+//!
+//! Only the field arithmetic lives here: a widening multiply ([`WideMul`]) with a deferred
+//! reduction, plus [`Square`] and [`InvertOrZero`]. `Mul` and the rest of the [`PackedField`]
+//! surface are provided generically by [`SlicedPackedField`]. The coordinate register is a
+//! [`PackedPrimitiveType`], so the reduction scales by `X⁻¹` with a per-lane bit shift
+//! ([`ghash_mul_inv_x`]) rather than a full field multiply, and the batch costs three GHASH
+//! multiplies, not four.
+
+use std::{
+	iter::Sum,
+	ops::{Add, AddAssign, Sub, SubAssign},
+};
+
+use crate::{
+	BinaryField128bGhash, Divisible, GhashSq256b, PackedField, WideMul,
+	arch::{M128, M256, M512, PackedPrimitiveType},
+	arithmetic_traits::{InvertOrZero, Square},
+	sliced_packed_field::SlicedPackedField,
+	underlier::{UnderlierType, WithUnderlier},
+};
+
+/// The packed GHASH coordinate register backing a `SlicedGhashSq256b<U>`.
+type Ghash<U> = PackedPrimitiveType<U, BinaryField128bGhash>;
+
+/// A GHASH² packing whose two GHASH coordinates pack into `PackedPrimitiveType<U, Ghash128b>`.
+pub type SlicedGhashSq256b<U> = SlicedPackedField<GhashSq256b, Ghash<U>, 2>;
+/// Packed `GhashSq256b` holding one extension scalar (the degenerate width-one packing).
+pub type SlicedGhashSq1x256b = SlicedGhashSq256b<M128>;
+/// Packed `GhashSq256b` holding two extension scalars.
+pub type SlicedGhashSq2x256b = SlicedGhashSq256b<M256>;
+/// Packed `GhashSq256b` holding four extension scalars.
+pub type SlicedGhashSq4x256b = SlicedGhashSq256b<M512>;
+
+/// The unreduced widening product of the coordinate GHASH multiply.
+type GhashWide<U> = <Ghash<U> as WideMul>::Output;
+
+/// Multiplies every 128-bit GHASH lane of an underlier by `X⁻¹`.
+///
+/// `X⁻¹` scaling is `GF(2)`-linear — a per-lane bit shift with a fixed compensation, not a field
+/// multiply — so this is far cheaper than a CLMUL. It reuses the scalar
+/// [`BinaryField128bGhash::mul_inv_x`] on each 128-bit lane, which every supported underlier
+/// divides into.
+#[inline]
+fn ghash_mul_inv_x<U: Divisible<M128>>(u: U) -> U {
+	U::from_iter(Divisible::<M128>::value_iter(u).map(|lane| {
+		BinaryField128bGhash::from_underlier(lane)
+			.mul_inv_x()
+			.to_underlier()
+	}))
+}
+
+/// Multiplies every GHASH lane of a packed coordinate by `X⁻¹`.
+#[inline]
+fn mul_inv_x<U: UnderlierType + Divisible<M128>>(coord: Ghash<U>) -> Ghash<U> {
+	Ghash::<U>::from_underlier(ghash_mul_inv_x(coord.to_underlier()))
+}
+
+/// The unreduced product of two GHASH² elements in sliced form.
+///
+/// Holds the three Karatsuba GHASH widening products, deferring both the GHASH reductions and the
+/// `X⁻¹` scaling. Since those are all `GF(2)`-linear, an inner product over GHASH² accumulates
+/// these by XOR and reduces once at the end.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct SlicedGhashSqWide<W> {
+	/// Unreduced `a·e`, the low diagonal Karatsuba product.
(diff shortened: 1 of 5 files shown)


crates/field/src/sliced_packed_field.rs
```diff
@@ -0,0 +1,554 @@
...
+	#[inline]
+	fn select(&self, mask: &Self::Mask) -> Self {
+		Self::from_coords(array::from_fn(|j| self.0[j].select(mask)))
+	}
+}
+
+// --- The extension multiply is deferred to the concrete `Mul`, so the widening multiply is
+// trivial: multiply eagerly and reduce to the identity.
+
+impl<F, PSub: PackedField, const N: usize> WideMul for SlicedPackedField<F, PSub, N>
```
jimpo: Flip this. Each specific type should provide a custom WideMul, and Mul gets blanket implemented in terms of WideMul.

crates/field/src/packed_ghash_sq.rs
```diff
@@ -0,0 +1,257 @@
...
+
+use crate::{
+	BinaryField128bGhash, GhashSq256b, PackedBinaryGhash1x128b, PackedBinaryGhash2x128b,
+	PackedBinaryGhash4x128b, PackedField,
+	arithmetic_traits::{InvertOrZero, Square},
+	sliced_packed_field::SlicedPackedField,
+};
+
+/// Packed `GhashSq256b` holding one extension scalar (the degenerate width-one packing).
+pub type PackedGhashSq1x256b = SlicedPackedField<GhashSq256b, PackedBinaryGhash1x128b, 2>;
```
jimpo: Let's make this specific for PackedPrimitiveType, so that we can do special optimizations (like special INV_X mul).

define `type SlicedGhashSq256b<U> = SlicedPackedField<GhashSq256b, PackedPrimitiveType<U, GhashSq256b>, 2>` , then `SlicedGhashSq1x256b = SlicedGhashSq256b<M128>`, etc.

jimpo (changes requested):
After reviewing, let's change how `SlicedPackedField` is defined. It should be specifically for PackedPrimitiveType fields. That is, it should be defined as `SlicedPackedField<U, F, const N: usize>([U; N]; PhantomData<F>)`, and when using subfield traits, we can require `PackedPrimitiveType<U, FSub>: WideMul` or whatever the trait we care about is.

crates/field/src/sliced_packed_field.rs
benji-potamus: Done. Each concrete type now provides a custom `WideMul` (real deferred reduction), and `Mul` is blanket `reduce(wide_mul(a, b))` on `SlicedPackedField`. For GHASH² the widening product holds the three Karatsuba GHASH products unreduced; `reduce` folds `Y² = Y + X⁻¹` (with the `X⁻¹` scaling as a per-lane bit shift). Square/InvertOrZero stay concrete for their fast forms.

crates/field/src/packed_ghash_sq.rs
benji-potamus: Done. Now `type SlicedGhashSq256b<U> = SlicedPackedField<U, GhashSq256b, 2>`, with `SlicedGhashSq{1,2,4}x256b = SlicedGhashSq256b<{M128,M256,M512}>`. Being `PackedPrimitiveType`-backed, the `X⁻¹` scale is the special shift-based `mul_inv_x` (reusing scalar `BinaryField128bGhash::mul_inv_x` per 128-bit lane via `U: Divisible<M128>`), so a batch is three GHASH multiplies, not four.

benji-potamus:
Reworked per the review — pushed.

- **Struct** is now `SlicedPackedField<U, F, const N>([U; N], PhantomData<F>)`, holding raw coordinate underliers and viewing each as `PackedPrimitiveType<U, FSub>` when subfield scalars are needed.
- **Mul/WideMul flipped**: concrete types supply `WideMul`; `Mul` is blanket `reduce(wide_mul)`. The GHASH² `WideMul` defers the three Karatsuba products, and `reduce` does the `X⁻¹`-via-shift fold — so the batch is **3** GHASH multiplies (this resolves the earlier 3-vs-4 note).
- **Aliases**: `SlicedGhashSq256b<U>` + `SlicedGhashSq{1,2,4}x256b`.

One addition worth flagging: the generic plumbing needs to name the subfield `FSub`, and `<U, F, N>` alone can't pin it at the type level (a scalar + degree doesn't determine a unique subfield — the same E0207 reason `F` couldn't be inferred before). I added a tiny `SlicedExtension<N> { type Subfield }` trait (one impl for `GhashSq256b`) to supply it. The alternative was a fourth `FSub` struct param; happy to switch to that if you prefer it over the marker trait.

Green: full `binius-field` suite (incl. new deferred-`WideMul` linearity tests), clippy `-D warnings`, nightly fmt, and cross-compiles on aarch64 + wasm32.

jimpo:
Oops, I forgot about FSub. OK, actually go back to defining `SlicedPackedField<F, PSub, const N: usize>` over a PackedField, like it was before. But keep `SlicedGhashSq256b<U> = SlicedPackedField<GhashSq256b, PackedPrimitiveType<U, Ghash128b>, 2>` alias.

benji-potamus:
Done — reverted to `SlicedPackedField<F, PSub, const N>` over `PSub: PackedField` (so `FSub = PSub::Scalar`, no marker trait), while keeping the rest of the rework:

- Concrete `WideMul` + blanket `Mul = reduce(wide_mul)`.
- `SlicedGhashSq256b<U> = SlicedPackedField<GhashSq256b, PackedPrimitiveType<U, BinaryField128bGhash>, 2>`, with `SlicedGhashSq{1,2,4}x256b`.
- The GHASH² `WideMul`/`Square`/`InvertOrZero` reach into the coordinate's underlier for the shift-based `mul_inv_x`, so the batch is 3 GHASH multiplies.

Pushed. Green: full `binius-field` suite (incl. deferred-`WideMul` linearity tests), clippy `-D warnings`, nightly fmt, aarch64 + wasm32.

jimpo (approved):

PR #1748, opened 2026-07-08
BINIUS-218: fuse the logUp* pushforward scatter across lookers

tcoratger:
Part of [BINIUS-218](https://linear.app/jimpo/issue/BINIUS-218). One of several witness inefficiencies tracked there, so it should **not** close the ticket on merge.

Builds the combined logUp* pushforward in one fused scatter across lookers, instead of one repacked buffer per looker. Output is bit-identical — no protocol change, the looked-up vectors are still never committed.

### The problem

The combined pushforward $Y = \sum_j \gamma^j (I_j)_* eq_{r_j}$ scatters every looker's $2^n$ numerator onto the shared $2^m$ table cube.

The old build did this one looker at a time:

- each looker scattered into its own $2^m$ bucket array,
- then repacked that array into a `FieldBuffer`,
- then the per-looker buffers were summed in a final packed pass.

Two things made it slow on the dominant $n$-axis:

- Large lookers (past a $2^{18}$-row threshold) took a parallel path keyed by row index, reading the numerator with an indexed `get` — a lane-extract per row.
- That same path allocated a fresh $2^m$ (1 MB) accumulator per `rayon` work-steal chunk, so one big looker zeroed dozens of megabyte arrays.

For the intmul caller ($m = 16$, 12 lookers) at $n \geq 18$ this path dominates witness construction.

### The idea

Scatter every looker into one shared accumulator, read in row order, and parallelize across lookers instead of across rows.

```
buckets = parallel over lookers, fold into one 2^m accumulator per task:
    for each row i of looker j:
        buckets[index_j[i]] += numerator_j[i]     // numerator_j already carries gamma^j
merge the per-task accumulators, then repack once
```

- Each looker is read sequentially, so no row pays a lane-extract.
- One accumulator per `rayon` task, not one per row-chunk.
- One repack at the end, not one per looker; the final packed reduce is gone.

### Per looker

- **before:** indexed `get` per row, own bucket array, own repack — then a packed reduce across all 12
- **after:** sequential scatter into a shared accumulator, one repack total

→ the dominant per-row lane-extract is gone, and repacks drop from 12 to 1.

### The change

```diff
- izip!(lookers, &numerators)
-     .map(|(looker, num)| pushforward(num, looker.index, m))
-     .reduce(|acc, term| sum_buffers(acc, term))
+ combined_pushforward(&numerators, lookers, m)   // fused parallel-over-lookers scatter
```

### Soundness

- $Y$ is a sum over a field, so scatter and accumulation order are free — the result is bit-identical, value for value.
- The pushforward feeds the same table-side circuit and the same product check as before.
- The transcript matches today's protocol exactly; nothing about the challenge distribution changes.

### Scope

- **changed:** the combined-witness build in `crates/ip-prover/src/logup_star/witness.rs`, plus a new private fused scatter and a shared scatter-add helper
- **simplified:** the single-looker `pushforward` now routes through the same helper; its old row-index parallel path and $2^{18}$-row threshold are gone (parallelism moved across lookers)
- **left untouched:** the numerators, the denominators, the fractional-addition circuits, the verifier
- **not addressed here:** the fractional-addition layer footprint noted on BINIUS-218 — left for a follow-up

### Trade-off

Parallelism is now across lookers, so a single-looker, huge-$n$ call under-parallelizes (one task). The intmul caller always has 12 lookers, so the target regime keeps every core busy; this is documented at the call site. A `(looker, aligned-chunk)` work-item variant is a straightforward follow-up if the few-looker case ever matters.

### Verification

- `cargo check -p binius-ip-prover --all-targets`, `clippy -p binius-ip-prover --all-targets` (with and without `rayon`), nightly `fmt --all --check` — clean
- `binius-ip-prover` (13) + `binius-iop-prover` (5) logUp* round-trips pass, so the end-to-end proof still verifies
- new proptest pins the fused scatter to a single-threaded reference over random shapes and looker counts; the single-looker scatter keeps its own reference proptest

### Benchmark

Building the combined pushforward at the intmul regime ($m = 16$, 12 lookers). Throughput is total scatter-adds ($12 \cdot 2^n$), medians:

| n | baseline (current) | fused | speedup |
|---|---|---|---|
| 16 | 2.68 ms (294 Melem/s) | 1.06 ms (742 Melem/s) | 2.5× |
| 18 | 265.7 ms (11.8 Melem/s) | 1.86 ms (1.69 Gelem/s) | 142× |
| 20 | 354.2 ms (35.5 Melem/s) | 5.61 ms (2.24 Gelem/s) | 63× |

$n = 16$ is below the old threshold, so it isolates the pure algorithmic win — no lane-extract, one repack instead of 12: ~2.5×. At $n \geq 18$ the baseline crosses its own $2^{18}$-row threshold into the indexed-`get` plus per-chunk-megabyte-accumulator path, which is pathologically slow, so the 63–142× there is real for large-$n$ intmul (which hits that path today) but is mostly the collapse of that trap, not a 100× algorithmic gain.

The throwaway bench used for these numbers is not part of this PR. To reproduce: drop the file below into `crates/ip-prover/benches/logup_pushforward.rs`, add a `criterion` dev-dep and a matching `[[bench]]` entry (`name = "logup_pushforward"`, `harness = false`) to `crates/ip-prover/Cargo.toml`, then run `cargo bench -p binius-ip-prover --features rayon --bench logup_pushforward`.

<details>

```rust

//! Before/after benchmark for the logUp* combined pushforward scatter.
//!
//! The combined pushforward `Y = sum_j gamma^j * (I_j)_* eq_{r_j}` scatters every looker's `2^n`
//! numerator onto the shared `2^m` table cube. Two ways to build it are compared:
//!
//!     baseline: one looker at a time; large lookers scatter in parallel over rows via `get(i)`,
//!               each looker repacked into its own buffer, then the buffers summed
//!     fused   : parallelize across lookers; each reads its numerator in row order (`iter_scalars`)
//!               into one per-task accumulator, merged once and repacked once
//!
//! Regime mirrors the intmul caller: m = 16 (2^16 table), 12 lookers, n the multiplication count.

use binius_field::{
	Field,
	arch::{OptimalB128, OptimalPackedB128},
};
use binius_math::{FieldBuffer, test_utils::random_field_buffer};
use binius_utils::rayon::prelude::*;
use rand::prelude::*;

type F = OptimalB128;
type P = OptimalPackedB128;

/// Baseline: the one-looker-at-a-time scatter with the internal row-parallel `get(i)` path.
fn baseline(numerators: &[FieldBuffer<P>], indices: &[Vec<usize>], m: usize) -> FieldBuffer<P> {
	// Row count at or above which the baseline scattered a single looker in parallel over rows.
	const PARALLEL_THRESHOLD: usize = 1 << 18;
	let table_size = 1usize << m;

	numerators
		.iter()
		.zip(indices)
		.map(|(eq_r, index)| {
			let values = if index.len() < PARALLEL_THRESHOLD {
				let mut buckets = vec![F::ZERO; table_size];
				for (eq_i, &j) in eq_r.iter_scalars().zip(index) {
					buckets[j] += eq_i;
				}
				buckets
			} else {
				index
					.par_iter()
					.enumerate()
					.fold(
						|| vec![F::ZERO; table_size],
						|mut buckets, (i, &j)| {
							buckets[j] += eq_r.get(i);
							buckets
						},
					)
					.reduce(
						|| vec![F::ZERO; table_size],
						|mut acc, partial| {
							for (slot, add) in acc.iter_mut().zip(partial) {
								*slot += add;
							}
							acc
						},
					)
			};
			FieldBuffer::<P>::from_values(&values)
		})
		.reduce(|mut acc, term| {
			for (slot, add) in acc.as_mut().iter_mut().zip(term.as_ref()) {
				*slot += *add;
			}
			acc
		})
		.expect("at least one looker")
}

/// Fused: parallelize across lookers, sequential `iter_scalars` scatter, one merge, one repack.
fn fused(numerators: &[FieldBuffer<P>], indices: &[Vec<usize>], m: usize) -> FieldBuffer<P> {
	let table_size = 1usize << m;

	let buckets = (0..numerators.len())
		.into_par_iter()
		.fold(
			|| vec![F::ZERO; table_size],
			|mut acc, j| {
				for (value, &target) in numerators[j].iter_scalars().zip(&indices[j]) {
					acc[target] += value;
				}
				acc
			},
		)
		.reduce(
			|| vec![F::ZERO; table_size],
			|mut acc, partial| {
				for (slot, add) in acc.iter_mut().zip(partial) {
					*slot += add;
				}
				acc
			},
		);

	FieldBuffer::<P>::from_values(&buckets)
}

fn bench_pushforward(c: &mut Criterion) {
	// intmul regime: 2^16-entry table, one looker per limb column.
	const M: usize = 16;
	const N_LOOKERS: usize = 12;

	let mut group = c.benchmark_group("logup_combined_pushforward");

	for n_vars in [16usize, 18, 20] {
		// Throughput is the total number of scatter-adds: one per looker row, over all lookers.
		group.throughput(Throughput::Elements((N_LOOKERS as u64) << n_vars));

		// One numerator buffer and one index column per looker, reused by both strategies.
		let numerators = (0..N_LOOKERS)
			.map(|_| random_field_buffer::<P>(&mut rng, n_vars))
			.collect::<Vec<_>>();
		let indices = (0..N_LOOKERS)
			.map(|_| {
				(0..(1usize << n_vars))
					.map(|_| rng.random_range(0..(1usize << M)))
					.collect::<Vec<_>>()
			})
			.collect::<Vec<_>>();

		// Sanity: both strategies must agree before timing them.
		assert_eq!(
			baseline(&numerators, &indices, M).iter_scalars().collect::<Vec<_>>(),
			fused(&numerators, &indices, M).iter_scalars().collect::<Vec<_>>(),
		);

		group.bench_function(BenchmarkId::new("baseline_per_looker_get", n_vars), |b| {
			b.iter(|| baseline(&numerators, &indices, M));
		});
		group.bench_function(BenchmarkId::new("fused_across_lookers", n_vars), |b| {
			b.iter(|| fused(&numerators, &indices, M));
		});
	}

	group.finish();
}

criterion_group!(benches, bench_pushforward);
criterion_main!(benches);
```

</details>

### DIFF
--- a/crates/ip-prover/src/logup_star/witness.rs
+++ b/crates/ip-prover/src/logup_star/witness.rs
@@ -72,18 +72,94 @@ where
 		})
 		.collect::<Vec<_>>();
 
-	// Scatter each looker's numerator onto the shared table cube and sum.
-	let combined = izip!(lookers, &numerators)
-		.map(|(looker, numerator)| pushforward::<F, P>(numerator, looker.index, table_n_vars))
-		.reduce(|mut acc, term| {
-			for (slot, add) in iter::zip(acc.as_mut(), term.as_ref()) {
-				*slot += *add;
-			}
-			acc
-		})
-		.expect("lookers is non-empty");
+	// Scatter every looker's numerator onto the shared table cube, summed into one buffer.
+	let combined = combined_pushforward::<F, P>(&numerators, lookers, table_n_vars);
 
 	(numerators, combined)
+}
+
+/// Scatter every looker's numerator onto the shared `m`-variable table cube and sum.
+///
+/// ```text
+///     Y[v] = sum_j sum_{i : index_j[i] = v} numerator_j[i]
+/// ```
+///
+/// The per-looker `gamma^j` scale already lives in each numerator.
+/// So the plain sum of the scatters is the gamma-combined pushforward.
+/// The sum is over a field, so the accumulation order does not matter.
+///
+/// # Performance
+///
+/// The scatter over every looker row is the dominant `n`-axis cost.
+/// Two choices keep it lean:
+///
+/// - Each looker is read sequentially in row order, so no row pays an indexed lane-extract.
+/// - The work parallelizes across lookers, not rows.
+/// - So each task fills a single `2^m` accumulator for its run of lookers.
+/// - The per-task accumulators merge in a single pass.
+///
+/// With few lookers this leaves cores idle on the `n`-axis.
+/// The target regime has one looker per column, so the tasks stay busy.
+///
+/// # Preconditions
+///
+/// * `numerators` and `lookers` have equal length.
+/// * Each numerator has one entry per row of its looker's index column.
+/// * Every index entry is less than `2^table_n_vars`.
+fn combined_pushforward<F, P>(
+	numerators: &[FieldBuffer<P>],
+	lookers: &[Looker<'_, F>],
+	table_n_vars: usize,
+) -> FieldBuffer<P>
+where
+	F: Field,
+	P: PackedField<Scalar = F>,
+{
+	// One accumulator slot per table position.
+	let table_size = 1usize << table_n_vars;
+
+	let buckets = (0..numerators.len())
+		.into_par_iter()
+		// Each task scatters a contiguous run of lookers into its own accumulator.
+		.fold(
+			|| vec![F::ZERO; table_size],
+			|mut acc, j| {
+				scatter_add(&mut acc, &numerators[j], lookers[j].index);
+				acc
+			},
+		)
+		// Merge the per-task accumulators position by position into one.
+		.reduce(
+			|| vec![F::ZERO; table_size],
+			|mut acc, partial| {
+				for (slot, add) in iter::zip(acc.iter_mut(), partial) {
+					*slot += add;
+				}
+				acc
+			},
+		);
+
+	// Repack the merged scalar accumulator into the packed table buffer.
+	FieldBuffer::from_values(&buckets)
+}
+
+/// Scatter-add one looker's numerator onto the table accumulator in row order.
+///
+/// ```text
+///     acc[index[i]] += numerator[i]
+/// ```
+///
+/// The numerator is read sequentially, so each row is a lane read, not an indexed lookup.
+#[inline]
+fn scatter_add<F, P>(acc: &mut [F], numerator: &FieldBuffer<P>, index: &[usize])
+where
+	F: Field,
+	P: PackedField<Scalar = F>,
+{
+	// Row i's numerator value lands in the table position that row indexes into.
+	for (value, &target) in numerator.iter_scalars().zip(index) {
+		acc[target] += value;
+	}
 }
 
 /// Embed a table position `j` into the field through the `GF(2)`-linear basis.
@@ -156,6 +232,9 @@ where
 /// `Y` is the dual of the pullback under the inner product, so `<T, Y> = (I^* T)(eval_point)`.
 /// It has only `2^m` entries, which is the cost saving over committing the `2^n`-entry pullback.
 ///
+/// This is the single-looker scatter.
+/// The prover combines many lookers by summing their scatters onto the same cube.
+///
 /// # Preconditions
 ///
 /// * every `index[i]` is less than `2^table_n_vars`.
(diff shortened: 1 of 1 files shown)


crates/ip-prover/src/logup_star/witness.rs
jimpo, line 55: follow-up: parallelize this with rayon too. Even though scaled_eq_ind_partial_eval is also rayonized, this will perform better with the outer loop rayonized too.

jimpo (approved):

PR #1750, opened 2026-07-09
BINIUS-258: wide_mul (deferred-reduction) variants for the GHASH modules

benji-potamus:
Implements [BINIUS-258](https://linear.app/jimpo/issue/BINIUS-258) in `binius-arith-bench`: splits GHASH multiplication into an unreduced widening multiply (`wide_mul`) plus a separate F2-linear `reduce`, so the reduction can be amortized across an inner product. `mul` becomes `reduce(wide_mul(x, y))`, both inlined — no behavior change.

Since the reduction is F2-linear on the wide product, unreduced products accumulate by XOR and only the final result is reduced. An inner product of `n` terms therefore pays one reduction instead of `n`.

### Commits
1. **soft64** — extract the four-limb schoolbook product into `mul_wide`, the shared modular reduction into `reduce`; `mul`/`square` compose them. Adds a deferred-reduction proptest.
2. **soft64 inner-product benchmark** — new `ghash_inner_product` criterion group: a length-2¹⁰ inner product of two random buffers, contrasting reduce-every-term `mul` against accumulate-unreduced `mul_wide` (final reduction skipped). Ad-hoc output types, no trait — benchmark-only.
3. **clmul** — `mul_wide` returns the three 128-bit limbs `[t0, t1, t2]`; `reduce` folds high→middle→low. Adds a clmul test module (the file had none) cross-checking against `soft64` and verifying deferred reduction, plus the `__m128i` inner-product benchmarks.

Bit-sliced and aarch64 are out of scope here (the aarch64 module already has this split).

### Benchmark (length-2¹⁰ inner product, `-C target-cpu=native`, x86_64 pclmulqdq)
| variant | time | vs. reduce-every-term |
|---|---|---|
| `soft64::mul` | 63.6 µs | — |
| `soft64::mul_wide` | 60.8 µs | ~1.04× |
| `clmul::mul::<__m128i>` | 2.93 µs | — |
| `clmul::mul_wide::<__m128i>` | **1.58 µs** | **~1.85×** |

The deferred reduction is a large win on CLMUL (a single fast `pclmulqdq` product, so the reduction is a big fraction of per-term cost — amortizing it nearly halves the inner product), and marginal in software (the carry-less `bmul64`s dominate). This is the measurement that motivates carrying the approach into the field crate (BINIUS-257 → BINIUS-256).

### DIFF
--- a/crates/arith-bench/benches/field_mul.rs
+++ b/crates/arith-bench/benches/field_mul.rs
@@ -11,7 +11,6 @@ use std::{array, hint::black_box};
 
 use binius_arith_bench::{Underlier, ghash, ghash_sq, polyval};
 use criterion::{BenchmarkGroup, Criterion, Throughput, criterion_group, criterion_main};
-use proptest::num::u128;
 use rand::{
 	Rng,
 	distr::{Distribution, StandardUniform},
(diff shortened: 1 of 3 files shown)


crates/arith-bench/benches/field_mul.rs
```diff
@@ -143,6 +142,46 @@ fn run_unary_op_benchmark<T, R>(
...
+///
+/// `T` is the underlier of the operands (`u128` for the software path, a SIMD type for the packed
+/// paths) and `W` the unreduced product accumulator; each GHASH element is 128 bits, so a `T`-wide
+/// underlier carries `T::BITS / 128` inner-product terms.
+fn run_inner_product_benchmark<T, W, R>(
+	group: &mut BenchmarkGroup<'_, criterion::measurement::WallTime>,
+	name: &str,
+	wide_mul: impl Fn(T, T) -> W,
+	add: impl Fn(W, W) -> W,
+	zero: W,
```
jimpo: W can be an underlier, right? And then it would inherit ZERO and xor and all that.

jimpo (changes requested):

crates/arith-bench/benches/field_mul.rs
benji-potamus: Only for the reduce-every-term baselines — there `W` is `u128` / `__m128i`, which are underliers, so `add`/`zero` are redundant there. But for the `wide_mul` variants `W` is the *unreduced multi-limb* accumulator: `[u64; 4]` for soft64 and `[__m128i; 3]` for clmul (the redundant 3-limb CLMUL product). Those are `Copy` but not `Underlier`, so there's no `ZERO`/`xor` to inherit — which is why the closures are there.

I can make `W: Underlier` uniform by wrapping each wide product in a newtype that impls `Underlier` (`ZERO`, `xor`, `BITS = 256/384`). That drops the two closures at the call sites, at the cost of a full `Underlier` impl — including `and`/`is_equal`/`random`, which are a bit meaningless on a redundant product representation — plus a `#[cfg(target_arch = "x86_64")]` gate for the `__m128i` one.

My lean is to keep the two closures (they're local and the wide types stay plain arrays), but happy to do the newtype-`Underlier` wrap if you'd prefer the uniform bound. Which do you want?

jimpo: Are you sure they're not underliers? There's a blanket Underlier impl for arrays of Underliers.

benji-potamus: You're right — I missed the blanket `impl<U: Underlier, const N: usize> Underlier for [U; N]`, so `[u64; 4]` and `[__m128i; 3]` are underliers after all. Done: `W: Underlier`, dropped the `add`/`zero` params, accumulation is now `W::xor` / `W::ZERO`, and all four call sites just pass the `wide_mul`/`mul` fn. Pushed as `3276b482`. Benchmark numbers unchanged (behavior-preserving).

jimpo (approved):

PR #1751, opened 2026-07-09
BINIUS-257: delayed reduction for ghash_sq mul

benji-potamus:
Implements [BINIUS-257](https://linear.app/jimpo/issue/BINIUS-257) on top of the GHASH `wide_mul`/`reduce` primitives from #1750. Defers the base-field reduction in the `ghash_sq` (GF(2^256) = GHASH[Y]/(Y²+Y+X⁻¹)) sliced multiply so an inner product pays the reduction — and the multiply-by-X⁻¹ — only once.

### Approach
The Karatsuba sliced multiply forms three base-field products `t0 = x0·y0`, `t1 = (x0+x1)·(y0+y1)`, `t2 = x1·y1`, with `z0 = t0 + X⁻¹·t2` and `z1 = t1 + t0`. `mul_wide_sliced` returns the **three raw unreduced products `[t0, t1, t2]`** and does no combining; `reduce_sliced` performs the reduction, the Karatsuba combine, and the multiply-by-X⁻¹. Since all of that is F2-linear, an inner product XOR-accumulates the three wide products and calls `reduce_sliced` once at the end — so per term there is no reduction and no `mul_inv_x`, just three widening multiplies.

### Commits
1. **Delayed-reduction `ghash_sq` mul** — generic `ghash_sq::sliced`: `mul_wide_sliced` (→ `[W; 3]`), `reduce_sliced` (reduce + combine + X⁻¹), and `mul_sliced = reduce_sliced ∘ mul_wide_sliced`; wired for soft64/x86_64/aarch64 with `reduce_sliced` wrappers. Adds a soft64 deferred-reduction proptest and a `pclmulqdq`-gated clmul cross-check module (the clmul sliced path had no test that runs without `vpclmulqdq`). `square_sliced` unchanged. Full x86_64 + aarch64 matrix verified.
2. **`ghash_sq_inner_product` benchmark** — length-2¹⁰ inner product contrasting per-term-reduced `mul_sliced` against the fully-deferred `mul_wide_sliced` (reuses `run_inner_product_benchmark` from #1750).

### Performance (length-2¹⁰ inner product, native, x86_64 pclmulqdq)
Moving the reduction and the X⁻¹ out of the per-term loop makes the deferred inner product a clear win on CLMUL (a bit over 2×); software is dominated by the carry-less `bmul64`s, as in #1750.

| variant | per-term `mul_sliced` | deferred `mul_wide_sliced` | speedup |
|---|---|---|---|
| `clmul::<__m128i>` | 9.87 µs | **4.35 µs** | **~2.27×** |
| `soft64` | 210 µs | 175 µs | ~1.20× |

Unblocks the field-crate port (BINIUS-256).

### DIFF
--- a/crates/arith-bench/src/ghash_sq/sliced.rs
+++ b/crates/arith-bench/src/ghash_sq/sliced.rs
@@ -11,30 +11,64 @@
 
 use crate::Underlier;
 
+/// Widening (unreduced) multiply of packed GHASH² elements in sliced representation.
+///
+/// Returns the three raw base-field products `[t0, t1, t2] = [x0·y0, (x0+x1)·(y0+y1), x1·y1]` — the
+/// Karatsuba terms for `Y² + Y + X⁻¹` — still *unreduced*. No combination or multiply-by-`X⁻¹` is
+/// done here: those are F2-linear and deferred to [`reduce_sliced`], so an inner product
+/// XOR-accumulates the three products (`W` is an [`Underlier`] via the blanket array impl) and pays
+/// the reduction and the `X⁻¹` exactly once at the end rather than per term.
+#[inline]
+pub fn mul_wide_sliced<U: Underlier, W>(
+	x: [U; 2],
+	y: [U; 2],
+	ghash_wide_mul: impl Fn(U, U) -> W,
+) -> [W; 3] {
+	let [x0, x1] = x;
+	let [y0, y1] = y;
+
+	let t0 = ghash_wide_mul(x0, y0);
+	let t1 = ghash_wide_mul(U::xor(x0, x1), U::xor(y0, y1));
+	let t2 = ghash_wide_mul(x1, y1);
+
+	[t0, t1, t2]
+}
+
+/// Reduce the three raw products `[t0, t1, t2]` from [`mul_wide_sliced`] into the two GHASH²
+/// coefficients.
+///
+/// The cross term is recovered by Karatsuba as `t1 + t0 + t2` and, over `Y² + Y + X⁻¹`, the `t2`
+/// term collapses so that `z0 = t0 + X⁻¹·t2` and `z1 = t1 + t0`. The multiply-by-`X⁻¹` lives here —
+/// applied to the reduced `t2` — so an inner product performs it (and every base-field reduction)
+/// only once, on the accumulated products.
+#[inline]
+pub fn reduce_sliced<U: Underlier, W>(
+	[t0, t1, t2]: [W; 3],
+	ghash_reduce: impl Fn(W) -> U,
+	ghash_mul_inv_x: impl Fn(U) -> U,
+) -> [U; 2] {
+	let r0 = ghash_reduce(t0);
+	let z0 = U::xor(r0, ghash_mul_inv_x(ghash_reduce(t2)));
+	let z1 = U::xor(ghash_reduce(t1), r0);
+	[z0, z1]
+}
+
 /// Multiply packed GHASH² elements in sliced representation.
 ///
 /// Given `x = [x0, x1]` and `y = [y0, y1]` representing packed GHASH² elements where `x0`/`y0`
 /// hold the lower limbs and `x1`/`y1` hold the upper limbs, computes the product in the extension
 /// field defined by Y² + Y + X⁻¹.
 ///
-/// Uses Karatsuba multiplication with three base-field multiplications and one multiply-by-X⁻¹.
+/// Composes [`mul_wide_sliced`] with [`reduce_sliced`].
 #[inline]
-pub fn mul_sliced<U: Underlier>(
+pub fn mul_sliced<U: Underlier, W>(
 	x: [U; 2],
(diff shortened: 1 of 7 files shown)


crates/arith-bench/src/ghash_sq/sliced.rs
```diff
@@ -11,30 +11,54 @@
...
+	y: [U; 2],
+	ghash_wide_mul: impl Fn(U, U) -> W,
+	ghash_mul_inv_x: impl Fn(U) -> U,
+) -> [W; 2] {
+	let [x0, x1] = x;
+	let [y0, y1] = y;
+
+	let t0 = ghash_wide_mul(x0, y0);
+	let t1 = ghash_wide_mul(U::xor(x0, x1), U::xor(y0, y1));
+	let t2x = ghash_wide_mul(x1, ghash_mul_inv_x(y1));
```
jimpo: Defer the ghash_mul_inv_x to the reduction step. That means you'll need a "wide" version of mul_inv_x (shifting the 256-bit output right by 1 with conditional addition).

benji-potamus: Done — `mul_wide_sliced` now forms `t2 = x1·y1` and defers the X⁻¹ onto the unreduced product via a new `ghash::{soft64,clmul}::mul_inv_x_wide` (`reduce(mul_inv_x_wide(p)) == mul_inv_x(reduce(p))`, tested on both paths). Pushed as `ffb0ead1`.

- **soft64** is exactly what you described: a 256-bit right-shift-by-1 over the `[u64;4]` product with `INV_X` conditionally added into the low two limbs. No perf change (~187µs).
- **clmul** I implemented as the base-field `mul_inv_x` applied per limb of `[U;3]` (each limb is an independent 128-bit value and `reduce` is linear, so it's equivalent). **But this regresses the clmul path** — three `mul_inv_x`s cost more than the one reduction we save:

| clmul `__m128i` | before (input-side trick) | now (per-limb wide `mul_inv_x`) |
|---|---|---|
| `mul_sliced` (per-term) | 8.54 µs | 11.29 µs |
| `mul_wide_sliced` (deferred) | 5.83 µs | 8.57 µs |

The literal "shift the 256-bit output right by 1" for clmul would instead be: fold the redundant `[t0,t1,t2]` into a non-redundant 256-bit `(lo,hi)`, do one cross-lane 256-bit `>>1` + conditional `INV_X`, return `[lo',0,hi']`. That's one shift instead of three `mul_inv_x`s and should recover most of the clmul perf, at the cost of more intricate SIMD.

Which do you prefer: (a) keep the per-limb version as-is (simplest, mirrors the field-crate pattern for BINIUS-256; prototype perf secondary), (b) I write the single-256-bit-shift clmul `mul_inv_x_wide`, or (c) keep clmul on the input-side `x1·(X⁻¹·y1)` (fastest — 5.83µs) while soft64 stays on the wide shift? I lean (b) if we want the clmul number, else (a).

crates/arith-bench/src/ghash_sq/sliced.rs
```diff
@@ -11,30 +11,55 @@
...
+	ghash_wide_mul_inv_x: impl Fn(W) -> W,
+) -> [W; 2] {
+	let [x0, x1] = x;
+	let [y0, y1] = y;
+
+	let t0 = ghash_wide_mul(x0, y0);
+	let t1 = ghash_wide_mul(U::xor(x0, x1), U::xor(y0, y1));
+	let t2 = ghash_wide_mul(x1, y1);
+
+	[W::xor(t0, ghash_wide_mul_inv_x(t2)), W::xor(t1, t0)]
```
jimpo: Make the inv_x multiplication part of the reduction and return 3 elements [t0, t1, t2] as the wide output

jimpo (changes requested):

crates/arith-bench/src/ghash_sq/sliced.rs
benji-potamus: Done — that's much cleaner and resolves the whole tradeoff. `mul_wide_sliced` now returns the three raw products `[t0, t1, t2]` (no combining, no X⁻¹), and `reduce_sliced` does the reduction + Karatsuba combine + multiply-by-X⁻¹. So per term there's no `mul_inv_x` and no reduction at all — an inner product accumulates the three wide products and reduces once. Dropped `mul_inv_x_wide` entirely (the reduction just uses the plain reduced `mul_inv_x` on `reduce(t2)`). Pushed as `669b8102`.

This is faster than any of the earlier options — the clmul deferred path is now the best yet:

| clmul `__m128i` (len-2¹⁰) | per-term `mul_sliced` | deferred `mul_wide_sliced` |
|---|---|---|
| | 9.87 µs | **4.35 µs** (~2.27×) |

soft64: 210 µs → 175 µs (~1.20×). PR body updated with the new design + numbers. All green (x86_64 + aarch64, fmt/clippy/test/doc).

jimpo (approved):

PR #1755, opened 2026-07-09
BINIUS-216: sumcheck shared-multilinear store and round evaluators (eager fold)

benji-potamus:
Addresses [BINIUS-216](https://linear.app/jimpo/issue/BINIUS-216) — the sumcheck shared-multilinear store and round evaluators (eager fold). Reworked per @jimpo's review on the previous PR (#1694, which got auto-closed when its branch was briefly deleted during a jj restructure) and rebased onto current `main`.

## Review changes (both inline comments from #1694 addressed)

- **`QuadraticMleEvaluator::new` no longer registers the eval point.** It takes an `EqId` + a `&MleStore`; callers register the shared tracker once (new `MleStore::eq_point(EqId) -> &[F]`), so several evaluators sharing a point fold one tracker.
- **The fracaddcheck layer interface is a method again.** `FracAddCheckProver::layer_prover` returns a self-contained `SharedMleCheckProver` **plus its column ids** (no raw-buffer `pop_last_layer`/free-function rebuild). The logUp\* final layer converts that prover to a sumcheck (`into_shared_sumcheck`) and **adds** its two `BivariateProductEvaluator`s onto the same store — the whole final layer is one `SharedSumcheckProver` over the six columns `[Y_0, Y_1, D_0, D_1, T_0, T_1]`, exactly the fusion `test_shared_final_layer_transcript_matches_old_provers` already guarded byte-for-byte.

## Commits (4)

1. **add sumcheck shared-multilinear store and round evaluators** — `MleStore`, `RoundEvaluator` + `SharedSumcheckProver`/`SharedMleCheckProver` (incl. `store`/`push_owned_column`/`add_evaluator`/`into_shared_sumcheck`), `QuadraticMleEvaluator` (EqId-based), `BivariateProductEvaluator`, the `MleToSumCheck` adaptors, and the byte-identical differential-equivalence tests.
2. **port the batch_quadratic_mle bench to the shared evaluator.**
3. **port the fracaddcheck reduction to the shared multilinear store** — single + batched layers driven through `layer_prover`.
4. **fuse the logUp\* final layer onto a shared store and remove the pre-store fracaddcheck path** — deletes `frac_add_mle::new` + `batch_quadratic_mle`.

All four commits build; ip-prover's 60 tests pass (incl. the logUp\* prove/verify round-trip and the byte-identical final-layer transcript test); workspace clippy `-D warnings`, rust-doc `-D warnings`, and nightly fmt are clean; downstream `binius-prover`/`binius-iop-prover` (incl. the bench) build.

Two notes:
- `layer_prover` now **owns** its four columns (an up-front copy of the popped leaf layer) so the returned prover is self-contained — this trades the earlier borrowed-column zero-copy for the interface you asked for. Happy to add a store variant that owns the split and exposes its halves if you'd rather keep zero-copy.
- I left the redundant `pushforward` param in place (your earlier note (b)); easy to drop as a follow-up.


jimpo:
"Happy to add a store variant that owns the split and exposes its halves if you'd rather keep zero-copy." Yes, we should have a SplitHalf store variant.

crates/ip-prover/src/fracaddcheck.rs
```diff
@@ -92,40 +113,48 @@ where
...
-		mut self,
+	/// The store owns the popped layer's numerator and denominator halves, so the returned prover
+	/// is self-contained: a caller can drive it, batch it alongside others, or extend its store
+	/// with more columns and evaluators — the logUp* final layer adds the product check to it. The
+	/// two claims — the fractional-addition numerator and denominator — share `claim`'s evaluation
+	/// point, and `self` advances to its remaining layers.
+	///
+	/// # Preconditions
+	/// * `self.n_layers() > 0`
+	pub(crate) fn layer_prover(
```
jimpo: This should be pub, not pub(crate). And why did you change it from consuming self to a self mut ref? Change it back if possible.

crates/ip-prover/src/fracaddcheck.rs
```diff
@@ -345,17 +374,16 @@ pub struct BatchProveUntilFinalLayerOutput<F, MP> {
...
 /// Arguments and preconditions are as for [`batch_prove`].
-pub fn batch_prove_until_final_layer<F, P, Channel>(
+pub fn batch_prove_until_final_layer<F, P>(
 	provers: Vec<FracAddCheckProver<P>>,
 	claimed_fractions: Vec<(F, F)>,
 	selector_point: Vec<F>,
 	content_point: Vec<F>,
-	channel: &mut Channel,
-) -> BatchProveUntilFinalLayerOutput<F, impl MleCheckProver<F> + use<F, P, Channel>>
+	channel: &mut impl IPProverChannel<F>,
```
jimpo: Why did you change this? Change back to the `Channel` type.

crates/ip-prover/src/sumcheck/quadratic_mle_evaluator.rs
```diff
@@ -0,0 +1,216 @@
...
+	// Store columns holding the packed evaluations of the input multilinears.
+	cols: [ColId; N],
+	// The store's (possibly shared) eq-indicator tracker for `eval_point`.
+	eq_tracker: EqId,
+	// Full quadratic composition evaluated on the "x = 1" branch for each multilinear.
+	composition: Composition,
+	// Composition restricted to highest-degree terms for the "x = ∞" evaluation (Karatsuba).
+	infinity_composition: InfinityComposition,
+	// The MLE-check evaluation point; the highest remaining coordinate is the round's alpha.
+	eval_point: Vec<P::Scalar>,
```
jimpo: Do we need to store the eval point? Why can't we just query it from the store in `interpolate` or wherever else it's used? The MleStore can maintain the point and pop off coordinates as `fold` is called.

jimpo (changes requested):
Looking pretty good, just a few more comments. And rebase onto main.

crates/ip-prover/src/fracaddcheck.rs
benji-potamus: Done. `layer_prover` is now `pub fn layer_prover(mut self, …)` — consuming `self` again, and public. The loop callers (`prove_layers`, `batch_prove_layer`) that need to pop layers one at a time now go through a private `&mut self` helper `pop_layer_prover` with the same body; the single-shot callers (`batch_prove_until_final_layer` and the logUp* final layer) call the consuming public method directly.

crates/ip-prover/src/fracaddcheck.rs
benji-potamus: Reverted — back to `batch_prove_until_final_layer<F, P, Channel>(… channel: &mut Channel) where Channel: IPProverChannel<F>`. I kept the concrete `BatchProveUntilFinalLayerOutput<F, LayerProver<F, P>>` return (the logUp* final layer needs the concrete `LayerProver` to call `into_shared_sumcheck`/`push_owned_column`/`add_evaluator`), so no `use<F, P, Channel>` capture is needed on the return.

crates/ip-prover/src/sumcheck/quadratic_mle_evaluator.rs
benji-potamus: Done — the evaluators no longer store the point. The `MleStore` now owns it: I added `eq_alpha(EqId)` (the current round's top coordinate) and `eq_prefix(EqId)` (the Gruen24 equality prefix), both derived from the tracker the store already folds in lockstep. The `RoundEvaluator` trait threads the store into the methods that need it — `round_claim(&self, store)` and `interpolate(&mut self, store, accum)` — so `QuadraticMleEvaluator` and `BivariateProductEvaluator` dropped their `eval_point`/`n_vars_remaining` mirrors and query `store.n_vars()`/`store.eq_alpha()` instead. `MleToSumCheckEvaluator` also sheds all its state (no more cloned point) and reads alpha + the equality prefix straight from the store. Transcript is unchanged — `test_shared_final_layer_transcript_matches_old_provers` and the two differential tests still pass byte-for-byte.

benji-potamus:
Done — added the `SplitHalf` store variant and dropped the up-front copy. Pushed (stack tip now `af76efe9`, rebased onto current `main`).

`MleStore` gains `push_split_half(buffer) -> [ColId; 2]`: it takes ownership of the buffer and adds its low and high halves as two columns backed by a shared `Arc<FieldBuffer>`, with **no copy**. The first `fold` of each half writes into a fresh quarter-size buffer (exactly what borrowed columns already do) and drops that side's reference; the shared parent is freed once both halves have folded. `fracaddcheck.rs`'s `layer_prover`/`pop_layer_prover` now push the popped numerator and denominator buffers straight in via `push_split_half` instead of `owned_halves` copying four half-buffers, so the returned prover stays self-contained but the leaf layer is no longer copied. `owned_halves` is deleted.

Placement: the store method is in commit 1 (store API), the fracaddcheck switch in commit 3. Transcript is unchanged — `test_shared_final_layer_transcript_matches_old_provers` and both differential tests still pass byte-for-byte; ip-prover's 60 tests, clippy `-D warnings`, rust-doc `-D warnings`, nightly fmt, and downstream `binius-prover`/`binius-iop-prover` are all green. CI is re-running.

Apologies for the slow turnaround on this one — this comment sat unaddressed longer than it should have because I'd narrowed my PR polling to the review threads and missed the top-level conversation comment. Fixed my checklist to always sweep conversation comments too.

jimpo (approved):

benji-potamus:
Your rebased push (`0a812d44`) failed the **rust-doc** CI check: the new `push_split_half` and `column_slices` public docs linked to the private `Column::SplitHalf` variant (`rustdoc::private_intra_doc_links` under `-D warnings`).

Pushed a fix as **`5b37ee7b`** — dropped the two private intra-doc links (plain "split-half entry" text), squashed into the "fold split-half store columns in place" commit. Re-verified green locally: whole-workspace nightly fmt, clippy `-D warnings`, full-workspace rust-doc `-D warnings`, and all 60 ip-prover tests incl. the byte-identical `test_shared_final_layer_transcript_matches_old_provers` + both differential tests.

The push dismissed your approval — please re-approve once CI is green and I'll squash-merge.

benji-potamus:
This is merge-ready — 8/8 CI green on `5b37ee7b`, approval standing, `mergeable_state=clean`. But squash-merge via the API is rejected with `405 — Changes must be made through the merge queue`, and I can't enqueue through the API (it only does direct merges).

@jimpo could you click **"Merge when ready"** to enqueue #1755? I'll detect the merge and restack automatically once it lands.

PR #1758, opened 2026-07-10
keccak: document that keccak256_varlen callers must constrain len_bytes (via ByteVec)

pauldelucia:
`keccak256_varlen` only checks `len_bytes <= capacity`; otherwise the prover picks `len_bytes` and the gadget proves `keccak256(message[..len_bytes])`. So a caller that doesn't independently constrain `len_bytes` can prove the hash of an arbitrary prefix of the committed message. This makes that caller obligation explicit and adds tests for it.

Per review, the obligation is documented on the input type rather than on the gadget:

- `Keccak256::new` is now private; the public entry point is the free function `keccak256_varlen(b, message: &ByteVec, digest) -> Keccak256`.
- Adds a Soundness note to `ByteVec` explaining that its `new`/`new_witness`/`new_inout` constructors leave `len_bytes` free over `0..=capacity`; drops the little-endian encoding prose from the gadget (`ByteVec` already documents that packing).
- Updates callers to pass a `ByteVec`: `ethsign` (×2; the public-key hash uses `ByteVec::new_const_len` to pin the length) and the `keccak` example.
- Adds `len_bytes` characterization tests: free-length prefix acceptance, over-capacity rejection, and a pinned-length control that rejects the same payload.

`cargo test -p binius-circuits` and `-p binius-examples` pass.


pauldelucia:
While writing this up I noticed the same `len_bytes`-selected-digest shape in two other hash gadgets on `main`: `Sha256::new` (which asserts the digest against a caller-supplied arg) and `sha512_varlen` (which *returns* the digest, so there's no pin point at all). Same caller obligation, and both undocumented. Happy to mirror this PR's doc block + negative tests to both if that'd be useful.

jimpo (commented):
Sorry, I think this documentation is overkill. The docstring says it's a variable-length input up to a certain length. It could be improved by explaining that the Wires contain bytes in little-endian order. But the better thing to do here is actually an interface change, which is that `keccak256_varlen` should be a function gadget that accepts a `ByteVec` struct. Then the `ByteVec` type itself clearly explains how the input is encoded.

pauldelucia:
@jimpo updated per your review — `keccak256_varlen` takes `&ByteVec` now; kept a trimmed soundness note on `ByteVec` itself since its default constructors leave `len_bytes` free. Details in the updated description.

jimpo:
@pauldelucia https://github.com/binius-zk/binius64/pull/1809 is what I had in mind.

Across the circuit frontend, I've decided to move away from the pattern of "struct gadgets" towards "function gadgets". The struct gadgets, like Keccak256, required non-deterministically populating various fields. It made it confusing to use and created a great risk of underconstrained systems. function gadgets instead have clear input and output wires. Any non-determinism is now handled through the Hint capability.

Does this new interface satisfy your soundness concerns?

pauldelucia:
@jimpo yes, that covers it. Closing this PR. One heads-up: the same shape still lives in Sha256. I'll create an issue

PR #1759, opened 2026-07-10
BINIUS-247: assemble the BitAnd witness over the wire-major ValueTable2

tcoratger:
Part of [BINIUS-247](https://linear.app/jimpo/issue/BINIUS-247/adopt-valuetable2-across-m4-and-remove-valuetable) — the "`BatchAndCheckWitness` … strided copy/accumulations over the wire-major layout instead of per-instance gathers" item, and the bitand half of "migrate M4 consumers to `ValueTable2`". One of several PRs; does **not** close the issue.

Migrates `BatchAndCheckWitness::build` to read the wire-major table. Same columns, same order, same proof — only the memory access pattern changes.

### The problem

The BitAnd check proves `A & B == C` on `K * n_and` rows. `build` fills the three columns from the witness: each cell is a XOR of shifted committed words.

The old builder walked **instance by instance** — grab all of instance `i`'s words as one block, compute its rows:

```
for each instance i:
    words = table.instance(i)         // one contiguous block, instance-major
    A[i][j] = XOR_t shift_t(words[index_t])
```

That was cheap when the table stored each instance's words together. The wire-major table (`ValueTable2`) stores the transpose — one wire's values across all instances sit together, and one instance's words are scattered. So `table.instance(i)` would become a strided gather.

### The idea

Flip the assembly to match the layout.

One value index's hidden words across all instances form **one contiguous row** of the wire-major buffer. Every cell in output column `j` that uses that word wants that same row (bit-shifted). So stream the row and XOR-accumulate into the column:

- reads are contiguous (a hidden wire's row),
- writes stride across the instance-major output column.

Two consequences:
- The output layout is **unchanged** — instance still on the high bits (`row = instance * n_and + j`). The shift reduction and `fold_instances` depend on that ordering.
- Constants are no longer stored per instance, so public words are read from the shared constant bank and broadcast.

### Per operand term

- **before:** for each instance, gather that instance's word for the term out of scattered memory.
- **after:** stream the term's contiguous row once, XOR its shifted words into the strided output column.

→ contiguous reads instead of a per-instance gather.

### The change

```
- pub fn build(table: &ValueTable, and_constraints) -> Self
-     for each instance: words = table.instance(i); A[i][j] = eval_operand(words, ..)
+ pub fn build(table: &ValueTable2, constants: &[Word], and_constraints) -> Self
+     for each stripe of instances:
+         for each constraint j, each operand term:
+             hidden -> stream contiguous sub-row, XOR shift(word) into column j
+             public -> broadcast shift(constant) into column j
```

Work is split into stripes of `STRIPE_WIDTH` instances so each task streams contiguous sub-rows.

### Soundness / behavior preservation

- The columns are bit-identical to the old builder's, in the same instance-major order.
- A value index below the witness offset is a shared public word; at or above it, a hidden word — the same split the reconstructed value vector uses.
- The AND relation and the downstream reduction are untouched.

### Scope

- **changed:** `BatchAndCheckWitness::build` (signature + wire-major assembly) in `bitand.rs`; the shift-test helpers that call it move to `ValueTable2`.
- **new:** `build_spans_multiple_instance_stripes` — 2 stripes' worth of instances, so the second stripe's nonzero base offset is exercised and checked against the per-instance reference.
- **removed:** the last `ValueTable` uses in the bitand and shift modules, including the temporary two-table scaffolding the fold migration (#1754) left in the shift tests. Only the prover's commit still uses `ValueTable` — deleted in the next PR.

### Verification

- `cargo build -p binius-m4-prover`, `clippy --all-targets`, `+nightly fmt -- --check` — clean
- `cargo test -p binius-m4-prover` — 34 pass, including:
  - `batch_rows_match_single_instance_reference` (proptest) and `reduction_round_trips_and_claims_match_columns` — pin every built cell to the independent `ValueVec` operand evaluator and round-trip the AND reduction.
  - `build_spans_multiple_instance_stripes` — covers the multi-stripe path.
  - the shift `prove_and_verify_round_trip`, now on a single wire-major table, still round-trips against the single-instance shift verifier.

### DIFF
--- a/crates/m4-prover/src/bitand.rs
+++ b/crates/m4-prover/src/bitand.rs
@@ -16,7 +16,14 @@ use binius_verifier::{
 	protocols::bitand::AndCheckOutput,
 };
 
-use crate::ValueTable;
+use crate::ValueTable2;
+
+/// Instance columns processed by one parallel witness-assembly task.
+///
+/// A task streams contiguous sub-rows of this many instances out of the wire-major table.
+/// Wide enough that each row read amortizes over many instances.
+/// Small enough to spread across cores on realistic batch sizes.
+const STRIPE_WIDTH: usize = 256;
 
 /// The operand columns of the BitAnd check for a whole batch of instances.
 ///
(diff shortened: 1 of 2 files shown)


crates/m4-prover/src/bitand.rs
jimpo, line 115: follow-up: Don't initialize these. Allocate `Vec::with_capacity`, use `Vec::spare_capacity_mut` and iterate over uninitialized chunks, then in the parallel iterator look, initialize with zeros and call `slice::assume_init_mut` to get the "allocated" slice.

crates/m4-prover/src/bitand.rs
```diff
@@ -218,23 +256,65 @@ impl BatchAndCheckWitness {
...
+	/// A public term broadcasts one shared word.
+	fn accumulate(&self, out: &mut [Word], operand: &[ShiftedValueIndex]) {
+		for sv in operand {
+			let idx = sv.value_index.0 as usize;
+			let amount = sv.amount as usize;
+			if idx >= self.offset {
+				// Hidden word: this stripe's slice of row `idx - offset` is contiguous.
+				let row = idx - self.offset;
+				let src = &self.data[(row << self.log_instances) + self.base..][..self.width];
+				for (local, &word) in src.iter().enumerate() {
```
jimpo, line 302: Might be worth branching on whether the values is shifted or unshifted. Since the check gets amortized over a whole stripe, it could be a good optimization for the common case of unshifted values.

jimpo (approved):
Two optimizations left for follow-ups

PR #1763, opened 2026-07-10
BINIUS-247: fast-path unshifted terms in the BitAnd witness assembly

tcoratger:
Part of [BINIUS-247](https://linear.app/jimpo/issue/BINIUS-247/adopt-valuetable2-across-m4-and-remove-valuetable). Follow-up to the review on #1759; does **not** close the issue.

Specializes the common unshifted case in `BatchAndCheckWitness::build`'s hot loop. Same output, less work per word.

### The problem

Each operand term carries a shift variant and amount. The hidden-word branch applied that shift to every word of the stripe:

```rust
for (local, &word) in src.iter().enumerate() {
    out[local * n_and + j] ^= sv.shift_variant.apply(word, amount);   // per instance
}
```

So it ran the shift dispatch (and a shift-by-`amount`, even a shift-by-zero) once per instance — up to `STRIPE_WIDTH` times — regardless of whether the term is actually shifted.

### The idea

The shift is fixed for the whole term, so it does not belong inside the per-instance loop. Test it once and specialize.

The canonical unshifted term is a left shift by zero (`ShiftedValueIndex::plain`), and `word << 0` is the identity. So for an unshifted term, XOR the raw words with no shift at all:

```rust
if sv.shift_variant == ShiftVariant::Sll && amount == 0 {
    // unshifted: raw XOR, no per-word shift
    for (local, &word) in src.iter().enumerate() {
        out[local * n_and + j] ^= word;
    }
} else {
    // shifted: unchanged
    for (local, &word) in src.iter().enumerate() {
        out[local * n_and + j] ^= sv.shift_variant.apply(word, amount);
    }
}
```

The branch is amortized over the whole stripe. The public-word branch already hoisted its shift, so only the hidden branch needed this.

### Correctness

- The fast path runs only for `Sll` with amount 0, where `apply(word, 0) == word << 0 == word`, so it is bit-identical to the general path.
- The shifted `else` branch is the pre-existing code, unchanged.

### Scope

- **changed:** the hidden-word branch of `OperandContext::accumulate` in `bitand.rs`.
- output columns are byte-identical; no API or layout change.

### Verification

- `cargo build -p binius-m4-prover`, `clippy --all-targets`, `+nightly fmt -- --check` — clean
- `cargo test -p binius-m4-prover` — 35 pass.
  - The existing per-instance-reference proptest uses unshifted operands, so it pins the new fast path against the value-vector reference; the CRC-based shift round-trip in the shift tests pins the shifted path.
  - **new** `build_matches_reference_with_shifted_operands`: hand-crafts constraints mixing a shifted term (`sll`/`srl`/`sar`) with an unshifted term and an empty operand set, and checks every instance's columns against the shift-aware value-vector reference — covering the shifted branch directly in this module (the compiler emits only unshifted operands for its fixtures).


crates/m4-prover/src/bitand.rs
```diff
@@ -339,9 +339,21 @@ impl OperandContext<'_> {
 				// Hidden word: this stripe's slice of row `idx - offset` is contiguous.
 				let row = idx - self.offset;
 				let src = &self.data[(row << self.log_instances) + self.base..][..self.width];
-				for (local, &word) in src.iter().enumerate() {
-					let slot = &mut out[local * self.n_and + self.j];
-					*slot = *slot ^ sv.shift_variant.apply(word, amount);
+
+				// The shift is fixed for the whole stripe, so test it once, not per word.
+				// An unshifted term is a left shift by zero, i.e. the identity.
+				if sv.shift_variant == ShiftVariant::Sll && amount == 0 {
```
jimpo:
```suggestion
				if amount == 0 {
```

crates/m4-prover/src/bitand.rs
```diff
@@ -5,7 +5,7 @@
 use std::mem::MaybeUninit;
 
 use binius_core::{
-	constraint_system::{AndConstraint, ShiftedValueIndex},
+	constraint_system::{AndConstraint, ShiftVariant, ShiftedValueIndex},
```
jimpo:
```suggestion
	constraint_system::{AndConstraint, ShiftedValueIndex},
```

jimpo (approved):

tcoratger:
@jimpo Just fixed the failing CI

PR #1764, opened 2026-07-10
BINIUS-261: introduce binius_math::multilinear::hypercube

benji-potamus:
Addresses [BINIUS-261](https://linear.app/jimpo/issue/BINIUS-261). Adds a `multilinear::hypercube` submodule whose routines are generic over the hypercube the coefficients are indexed by, so the Boolean and infinity cubes are supported side by side.

## The abstraction

A cube is determined by the linear polynomial basis $(b_0, b_1)$ each variable contributes, so `trait Hypercube` is built around that basis:

- `basis(r) -> [b_0(r), b_1(r)]` — the definition.
- `expand_var(v, r) -> [v·b_0(r), v·b_1(r)]` — the tensor-expansion inner loop, scaling the basis by a value.
- `contract_var(lo, hi)` — the inverse: strips one variable's basis factor.
- `eq_one_var(x, y) = Σ_i b_i(X) b_i(Y)` — the one-variable equality indicator, with a provided impl in terms of `basis`.

`OneCube` (basis $(1 - X, X)$) indexes coefficients by evaluations over $\{0,1\}^n$; `InfCube` (basis $(1, X)$) indexes them by monomial coefficients over $\{0,\infty\}^n$. Both specialize `expand_var` to one multiplication, and `OneCube` keeps the characteristic-2 shortcut in `eq_one_var`.

`tensor_prod_eq_ind`, `tensor_prod_eq_ind_prepend`, `eq_ind_partial_eval`, `scaled_eq_ind_partial_eval`, `eq_ind_truncate_low_inplace`, `eq_ind`, `eq_ind_zero`, and the two `_scalars` variants are all generic over `Cube`. The existing `multilinear::eq` functions keep their signatures and delegate, specialized to `OneCube`.

## Commits (3)

1. **add the `multilinear::hypercube` module** — trait, the two cubes, the generic routines, tests.
2. **delegate the eq indicator routines to the hypercube module** — `multilinear::eq` becomes thin wrappers over `hypercube::…::<OneCube>`. No public signature changes, so callers are untouched.
3. **benchmark the infinity cube eq indicator expansion** — `eq_ind_partial_eval` now benches both cubes.

## Performance

The abstraction is free on the hot path. `expand_var` returns the two halves as `[F; 2]` (the natural signature); a `cargo asm` check confirms it compiles to an instruction sequence identical to writing them through `&mut` references — SROA fully scalar-replaces the returned array, so there is no codegen cost either way. (An earlier commit chased a ~12% microbenchmark delta with a `&mut`-out-param signature; that delta turned out to be pure code-layout/alignment sensitivity, not a real cost — same instructions, different address — so the clean `[F; 2]` signature is restored. Discussion in the review thread.)

All three commits build; `binius-math`'s tests pass, workspace clippy `-D warnings`, rust-doc `-D warnings`, and nightly fmt are clean.

### DIFF
--- a/crates/math/benches/eq_ind_partial_eval.rs
+++ b/crates/math/benches/eq_ind_partial_eval.rs
@@ -1,14 +1,27 @@
 // Copyright 2025 Irreducible Inc.
 
 use binius_field::arch::{OptimalB128, OptimalPackedB128};
-use binius_math::{multilinear::eq::eq_ind_partial_eval, test_utils::random_scalars};
-use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+use binius_math::{
+	multilinear::hypercube::{Hypercube, InfCube, OneCube, eq_ind_partial_eval},
+	test_utils::random_scalars,
+};
+use criterion::{
+	BenchmarkGroup, BenchmarkId, Criterion, Throughput, criterion_group, criterion_main,
+	measurement::WallTime,
+};
 use rand::{SeedableRng, rngs::StdRng};
 
-fn bench_eq_ind_partial_eval(c: &mut Criterion) {
-	type F = OptimalB128;
-	type P = OptimalPackedB128;
+type F = OptimalB128;
+type P = OptimalPackedB128;
 
+fn bench_cube<Cube: Hypercube>(group: &mut BenchmarkGroup<'_, WallTime>, cube: &str, point: &[F]) {
+	let id = BenchmarkId::new(cube, format!("n_vars={}", point.len()));
+	group.bench_function(id, |b| {
+		b.iter(|| eq_ind_partial_eval::<Cube, P>(point));
+	});
+}
+
+fn bench_eq_ind_partial_eval(c: &mut Criterion) {
 	let mut group = c.benchmark_group("eq_ind_partial_eval");
 
 	let mut rng = StdRng::seed_from_u64(0);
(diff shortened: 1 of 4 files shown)


jimpo (commented):
This looks good, but I'm curious about why the array-returning version of expand_var was so much less efficient. I expected that after inlining, it should end up the same. Take a look at cargo asm and explain why that's worse.

benji-potamus:
You were right — it *isn't* worse. I dug into `cargo asm` and the two forms compile to an **identical instruction sequence**; the 12% was a code-layout artifact, not a codegen cost. I've switched `expand_var` back to the `[F; 2]`-returning signature (pushed, `cfb51b4d`) and dropped the misleading doc comment.

What I checked (host is AVX-512, so `P = PackedBinaryGhash4x128b`, a 64-byte SIMD value, and `[P; 2]` is a 128-byte array):

- **Isolated inner loop** — a plain `iter_mut().zip()` loop calling an array-returning `expand_var` helper vs. the `&mut`-writing one: after register normalization the disassembly is **byte-for-byte identical** (76 instructions each). SROA fully scalar-replaces the returned array; no spill.
- **The whole `tensor_prod_eq_ind`** (monomorphized `<OneCube, PackedBinaryGhash4x128b>`, generic `#[inline(always)]` trait method, real rayon `into_par_iter().for_each` closure): the two source forms produce the **same instruction sequence modulo register allocation** — identical count, identical mnemonics/operands. So there is no extra `vmov`, no aggregate spill, nothing.

Since the machine code is identical, the reproducible ~12% can only be **code layout / alignment**: the differing source sizes shift where the hot loop lands in the binary, and one alignment happens to be worse. The clincher: with the `[F; 2]` build, `one_cube/n_vars=20` benches ~6.44 ms while `inf_cube/n_vars=20` — whose identical hot loop sits at a *different* address — is unaffected at ~5.55 ms; with the `&mut` build `one_cube` was ~5.71 ms. Same instructions, the delta just tracks alignment. It's the classic layout ghost ("Producing Wrong Data Without Doing Anything Obviously Wrong"), not a property of the signature.

So the two are performance-equivalent and the `[F; 2]` version reads better — thanks for pushing on it. Heads-up that the `one_cube/n_vars=20` bench number in this build is ~12% higher purely from that alignment shift; it'll wander with unrelated changes. If you'd rather I keep the `&mut` form to hold the current bench number, say the word, but I think the clean signature is the right call given the codegen is identical.

jimpo (approved):

PR #1768, opened 2026-07-10
BINIUS-218: build fracadd layers in one fused pass

tcoratger:
Part of [BINIUS-218](https://linear.app/jimpo/issue/BINIUS-218/logup-witness-inefficiencies) — this does not close it.

Builds the fractional-addition GKR layers the same way, but writes them straight into their backing store instead of through a pair of intermediate vectors.

### The problem

The fractional-addition prover builds each layer from the one below it, halving the variable count with the rule

$$\frac{a_0}{b_0} + \frac{a_1}{b_1} = \frac{a_0 b_1 + a_1 b_0}{b_0 b_1}.$$

The build ran a rayon `collect::<(Vec, Vec)>()` per layer.
That routes every element through an unzip into **two intermediate vectors**, which are then boxed into the layer buffers.
So each layer pays an allocate-and-copy for a pair of scratch vectors it immediately discards.

### The idea

Allocate the two layer buffers up front and fill them in **one fused parallel pass**, writing straight into their uninitialized spare capacity.

```
for each output slot i (in parallel):
    num_out[i] = a_0*b_1 + a_1*b_0
    den_out[i] = b_0*b_1
```

No intermediate vectors, no unzip, no per-layer copy into the final boxes.
The six zipped iterators (two output buffers, four input halves) all share one length — the packed word count — so an indexed zip initializes every slot exactly once.

### The change

```
- let (next_num, next_den) = (num_0, den_0, num_1, den_1)
-     .into_par_iter()
-     .map(|(&a_0, &b_0, &a_1, &b_1)| (a_0 * b_1 + a_1 * b_0, b_0 * b_1))
-     .collect::<(Vec<_>, Vec<_>)>();
+ let mut num_out = Vec::with_capacity(out_packed_len);
+ let mut den_out = Vec::with_capacity(out_packed_len);
+ (num_out.spare_capacity_mut(), den_out.spare_capacity_mut(), num_0, den_0, num_1, den_1)
+     .into_par_iter()
+     .for_each(|(n, d, &a_0, &b_0, &a_1, &b_1)| {
+         n.write(a_0 * b_1 + a_1 * b_0);
+         d.write(b_0 * b_1);
+     });
```

The `set_len` uses the same `spare_capacity_mut` + indexed-write idiom already used in `reed_solomon.rs`, with a written safety argument: the fused pass initializes every slot of both buffers' spare capacity.

- The fractional-addition formula is unchanged and evaluated per slot in the same order.
- Every layer holds the same values, bit for bit, so the transcript is identical.
- The only change is where the results are written, not what they are.

### Scope

- **changed:** the per-layer build loop in `crates/ip-prover/src/fracaddcheck.rs`.
- **added docs:** the layer memory profile — every layer stays resident until proving folds it in place, so the peak footprint is about twice the leaf, and the layers cannot be slices carved from one shared allocation (proving folds each owned buffer in place).
- left untouched: the protocol, the verifier, the sumcheck reduction.

Note on what is *not* done here: the layers genuinely cannot share one backing allocation, because each is handed to an in-place sumcheck fold that requires owned `'static` data — so the eager all-layers-resident footprint is inherent to the leaf-to-root reduction.

### Verification

- `cargo check -p binius-ip-prover --all-targets`, `clippy -p binius-ip-prover`, nightly `fmt` — clean.
- `cargo test -p binius-ip-prover --lib` — 59 pass, including the fracadd round-trips and the logUp* prove/verify integration tests that consume this build.
- `cargo check --workspace --all-targets` fails identically on clean `main` (a pre-existing feature-unification issue in the `binius-ip-prover` test target), so it is not affected by this change.

### Benchmark

`cargo bench --bench fracaddcheck -- new` (full reduction — the logUp* looker regime):

| n_vars | before | after | speedup |
|---|---|---|---|
| 12 | 21.8 µs | 19.0 µs | 1.14× |
| 16 | 323 µs | 271 µs | 1.19× |
| 20 | 5.08 ms | 4.21 ms | 1.19× |

About a 16% cut at n_vars=20 (p < 0.05), and logUp* builds one such circuit per looker (12 per intmul), so it compounds.

### DIFF
--- a/crates/ip-prover/src/fracaddcheck.rs
+++ b/crates/ip-prover/src/fracaddcheck.rs
@@ -7,7 +7,7 @@ use binius_ip::{mlecheck, prodcheck::MultilinearEvalClaim, sumcheck::RoundCoeffs
 use binius_math::{
 	FieldBuffer, line::extrapolate_line_packed, multilinear::eq::eq_ind_partial_eval,
 };
-use binius_utils::rayon::iter::{IntoParallelIterator, ParallelIterator};
+use binius_utils::rayon::prelude::*;
 use itertools::izip;
 
 use crate::{
@@ -29,6 +29,14 @@ pub type FracEvalClaim<F> = (MultilinearEvalClaim<F>, MultilinearEvalClaim<F>);
 /// Each layer is a double of the numerator and denominator values of fractional terms. Each layer
 /// represents the addition of siblings with respect to the fractional addition rule:
 /// $$\frac{a_0}{b_0} + \frac{a_1}{b_1} = \frac{a_0b_1 + a_1b_0}{b_0b_1}$
+///
+/// Every layer stays resident until proving consumes it.
+/// The layers halve in size going up, so the peak footprint is at most twice the leaf.
+/// Recomputing interior layers on demand would about halve the memory for double the build work.
+/// The resident layout instead keeps the build to a single pass.
+///
+/// Each layer owns its backing store, since proving folds it in place.
+/// So the layers cannot be slices carved from one shared allocation.
 pub struct FracAddCheckProver<P: PackedField> {
 	layers: Vec<(FieldBuffer<P>, FieldBuffer<P>)>,
 }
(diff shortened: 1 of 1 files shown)


jimpo (commented):
This seems like it should do the same exact thing. Collecting into a pair of Vecs ought to be fine.

I guess there is a measurably speedup. But that seems like it's more of an issue with our serial "rayon" shim, not with this code.

tcoratger:
> This seems like it should do the same exact thing. Collecting into a pair of Vecs ought to be fine.
> 
> I guess there is a measurably speedup. But that seems like it's more of an issue with our serial "rayon" shim, not with this code.

Yes exact, at first I measured an improvement but I've just double checked and now this is within noise level, just closing

PR #1771, opened 2026-07-10
BINIUS-255: compose the AND-check and shift reduction into the M4 protocol

tcoratger:
Part of [BINIUS-255](https://linear.app/jimpo/issue/BINIUS-255/integrate-andcheck-and-shift-reduction-into-m4-protocol). One step toward the M4 protocol; does **not** close the issue (the trace-oracle tie remains).

Chains the two batched reductions into one reduction on a shared transcript, for prover and verifier. No new cryptography — just the glue.

### The problem

The batched AND-check (`BatchAndCheckWitness::prove` / `verify_bitand_reduction`) and the shift reduction (`shift::prove` feeding the single-instance `shift::{verify, check_eval}`) existed as **standalone** building blocks, each exercised only by its own test with hand-built inputs. Nothing composed them on one Fiat-Shamir transcript.

### The idea

Run them back to back, the AND-check's output feeding the shift reduction:

```
   all AND rows  (A & B == C over K * n_and rows)
        │  AND-check
        ▼
   claimed A, B, C at a random row point  +  bit challenge r_z
        │  split the row point, fold the witness over instances
        ▼
   shift reduction
        ▼
   one evaluation claim about the instance-folded witness
```

The batch-specific step: the AND-check's row point is `(constraint, instance)`. Split it — the constraint part `r_x` (low) feeds the shift reduction; the instance part `r_rho` (high) says how to fold the K instances into one before the shift runs.

### Why this is correct given the codebase

- It is the batched shape of the single-instance protocol (`crates/verifier/src/verify.rs`), which already threads `bitand → shift` in this order — so the Fiat-Shamir ordering is unchanged.
- The exact threading (split `r_x`/`r_rho`, fold at `r_rho`, feed `r_x` to the shift) is precisely the wiring the existing `shift.rs` round-trip test already validated; the only change is that `r_x`, `r_rho`, and the operand claims now come from the shared transcript rather than being sampled by the test.
- The existing reductions are reused **unchanged**; this PR only adds the glue.

### Scope

- **new:** `m4-prover::prove_reduction`, `m4-verifier::verify_reduction`, and their output structs.
- **only AND constraints:** MUL is rejected up front (there is no batched intmul check yet), rather than silently passing a zero claim that would skip the MUL checks.
- **stops at the witness claim:** binding it to the committed trace oracle — which binds `r_rho` and bridges the bit witness to the packed commitment via a batched ring-switch that does not exist yet — is a later step, flagged in both module docs.
- the CRC-64 test fixture moves from `shift.rs` to a crate-level `#[cfg(test)] mod crc64`, so the shift and reduction tests share it without one test module reaching into another's.

### Verification

- `cargo test -p binius-m4-prover` (28) and `-p binius-m4-verifier` (3) — pass
- `clippy --all-targets`, `+nightly fmt -- --check` — clean
- `reduction_round_trips`: prover and verifier agree on `r_rho`, the challenge point, and the eval; and the reduced claim equals the true instance-folded witness evaluated at the shift point.
- `tampered_transcript_is_rejected`.

### Note

This branch is cut from `main`, which still has `ValueTable2` (the rename in #1770 is open). When #1770 lands, this is a trivial `ValueTable2` → `ValueTable` rename here.


crates/m4-verifier/src/reduction.rs
```diff
@@ -0,0 +1,96 @@
...
+/// - `public_words`: the shared constant words, the same in every instance.
+/// - `channel`: the verifier channel reading messages and redrawing Fiat-Shamir challenges.
+///
+/// # Errors
+///
+/// Returns an error if the AND-check, the shift reduction, or the public-input check fails.
+pub fn verify_reduction<Channel>(
+	cs: &ConstraintSystem,
+	log_instances: usize,
+	public_words: &[Word],
```
jimpo: I think we can actually drop this argument and use `cs.constants()` directly

crates/m4-verifier/src/reduction.rs
```diff
@@ -0,0 +1,96 @@
...
+	channel: &mut Channel,
+) -> Result<ReductionVerifierOutput, Error>
+where
+	Channel: IPVerifierChannel<B128, Elem = B128>,
```
jimpo, line 56: Is the `Elem = B128` necessary? Why?

crates/m4-prover/src/reduction.rs
```diff
@@ -0,0 +1,244 @@
...
+	let folded = fold_instances::<B128, P>(table, r_rho);
+	let scalars: Vec<B128> = folded.iter_scalars().collect();
+	let folded_witness: Vec<FoldedWord<B128>> = scalars
+		.chunks_exact(WORD_SIZE_BITS)
+		.map(|chunk| chunk.try_into().expect("chunk has WORD_SIZE_BITS elements"))
+		.collect();
+
+	// Reduce the operand claims to one witness evaluation.
+	// No MUL constraints here, so the intmul claim is a zero claim at an empty point.
+	let key_collection = build_key_collection(cs);
```
jimpo: Move the key collection build out into a setup phase. It's quite expensive.

jimpo (commented):
Hell yeah! This looks super clean. A couple comments, and this'll need a rebase because of the ValueTable rename.

crates/m4-verifier/src/reduction.rs
tcoratger: Fixed

crates/m4-verifier/src/reduction.rs
tcoratger: Yes I think so because `verify_bitand_reduction` fixes the channel's element type to `B128`. So this composition inherits that concrete bound, even though the shift check alone is looser.

crates/m4-prover/src/reduction.rs
tcoratger: Fixed

tcoratger:
> Hell yeah! This looks super clean. A couple comments, and this'll need a rebase because of the ValueTable rename.

Just fixed the comments and the conflicts with main :)

crates/m4-verifier/src/reduction.rs
```diff
@@ -0,0 +1,94 @@
...
+	let (r_x, r_rho) = eval_point.split_at(log_n_and);
+
+	// Reduce the operand claims to one witness evaluation.
+	// No MUL constraints here, so the intmul claim is a zero claim at an empty point.
+	let bitand = OperatorData::new(r_x.to_vec(), [a_eval, b_eval, c_eval]);
+	let intmul = OperatorData::new(Vec::new(), [B128::ZERO; 4]);
+	let shift = shift::verify::<B128, _>(cs, &bitand, &intmul, channel)?;
+
+	// Tie in the shared constants through the public-input consistency check.
+	let domain = BinarySubspace::<B8>::with_dim(Word::LOG_BITS).isomorphic::<B128>();
```
jimpo, line 78: follow-up: It would be best to pass this domain into verify_bitand_reduction too (but the pre-isomorphism version). That would ensure it's consistent between andcheck and shift, by construction. On the prover side too.

jimpo (approved):

PR #1780, opened 2026-07-13
BINIUS-267: benchmark batched BLAKE3 proving on M4, pad the public segment

tcoratger:
Closes [BINIUS-267](https://linear.app/jimpo/issue/BINIUS-267).

Adds an end-to-end M4 proving-throughput benchmark for batched BLAKE3 compressions, and fixes the public-segment padding that proving a real circuit exposed.

### The problem

M4 can prove a batch of one circuit end to end, but nothing drives a real hash circuit through it, and there is no proving-throughput benchmark.

Proving batched BLAKE3 exposes a second problem: the reduction only handles circuits whose constant count is a power of two.

- The shift reduction reads the public (constant) words as a multilinear over a power-of-two hypercube.
- It passes the raw constant bank straight in and takes `checked_log_2` of its length.
- A BLAKE3 compression has a constant count that is not a power of two, so that call panics.
- The existing test circuit is a power of two by luck, so this was never hit.

### The fix

Pad the public segment to the length the layout already reserves for it.

- The layout rounds the public segment (constants plus inout) up to a power of two.
- The single-instance verifier already uses that padded length; the batch path did not.
- A shared helper zero-pads the constants to `1 << log_public_words`.
- Both prover and verifier call it, so the two sides pad identically.
- The pad is a no-op when the constants already fill the segment, so aligned circuits are unchanged.

### The benchmark

- One BLAKE3 compression per instance, proved as one batch.
- Same primitive as the single-instance BLAKE3 example bench, so the numbers compare directly.
- Reuses a small shared driver, so a later XMSS benchmark drops in the same way.
- Reports compressions per second, including witness generation, to match published Flock figures.

### Scope

- **new:** `crates/m4-verifier/src/public_words.rs` — the padding helper
- **new:** `crates/m4-prover/benches/utils/m4_bench.rs` — the shared proving driver
- **new:** `crates/m4-prover/benches/blake3_compressions.rs` — the BLAKE3 benchmark
- **changed:** `crates/m4-prover/src/reduction.rs` and `crates/m4-verifier/src/reduction.rs` — pad the constants before the shift
- **left untouched:** the AND-check reads constants by index (out-of-range is zero), so it needs no padding

### Soundness

- The padded segment is exactly what the single-instance protocol already uses.
- Prover and verifier pad from one shared definition, so they cannot diverge.
- Padding with zeros beyond the real constants leaves the public multilinear unchanged.
- The prover commits before any challenge is drawn, so Fiat-Shamir binding is untouched.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-m4-prover -p binius-m4-verifier --all-targets`, nightly `fmt --check` — all clean.
- Tests: 32 in `binius-m4-prover` and 3 in `binius-m4-verifier` pass.
  - new: a non-power-of-two-constant circuit (one BLAKE3 compression) round-trips through the reduction.
  - new: the padded public segment has the layout length, reproduces the constants, and zero-fills the tail.
  - unchanged: the existing power-of-two round-trip still passes, confirming the pad is a no-op when aligned.

### Benchmark

Apple M1 PRO, multi-threaded, rate 1/2, including witness generation:

| batch | time / batch | throughput |
|---|---|---|
| 2^6 (64) | 10.9 ms | 5.9 K compressions/s |
| 2^13 (8192) | ~1.31 s | ~6.2 K compressions/s |

This is a first, unoptimized end-to-end data point — roughly monolithic parity — which localizes the remaining Flock gap to the zerocheck and PCS optimizations, not the batching itself.


crates/m4-verifier/src/reduction.rs
```diff
@@ -93,9 +93,12 @@ where
 	let shift = shift::verify::<B128, _>(cs, &bitand, &intmul, channel)?;
 
 	// Tie in the shared constants through the public-input consistency check.
+	// The shift reads the public words over a power-of-two hypercube.
+	// So pad the constant bank up to the layout's public length.
+	let public_words = padded_public_words(cs);
```
jimpo: This shouldn't be necessary. Is there a test that demonstrates this failure?

`PublicWordsEvalFn` should handle non-power-of-two word lengths

crates/m4-prover/benches/blake3_compressions.rs
```diff
@@ -0,0 +1,130 @@
...
+
+	// Every compression input is a witness wire, filled per instance.
+	let cv = array::from_fn(|_| builder.add_witness());
+	let block = array::from_fn(|_| builder.add_witness());
+	let counter = builder.add_witness();
+	let block_len = builder.add_witness();
+	let flags = builder.add_witness();
+
+	// Force-commit each output word so the compression survives dead-code elimination.
+	let out = blake3_compress(&builder, cv, block, counter, block_len, flags);
```
jimpo: Use `blake3_compress_2x`

crates/m4-prover/benches/blake3_compressions.rs
```diff
@@ -0,0 +1,130 @@
...
+	}
+	// A full 64-bit block counter.
+	w[inputs.counter] = Word(rng.next_u64());
+	// A byte length in 0..=64.
+	w[inputs.block_len] = Word((rng.next_u32() % 65) as u64);
+	// Arbitrary domain-separation flags.
+	w[inputs.flags] = Word(rng.next_u32() as u64);
+}
+
+fn bench_blake3_compressions_m4(c: &mut Criterion) {
```
jimpo: rename `bench_prove_blake3_compressions`

crates/m4-verifier/src/reduction.rs
tcoratger: Mhhhh yes I think I found what was wrong, the verifier was ok but not the prover. We had a `checked_log_2(public_words.len())`, which requires a power of two. I've changed that to `log2_ceil_usize(public_words.len())` to fix the problem.

crates/m4-prover/benches/blake3_compressions.rs
tcoratger: Fixed

crates/m4-prover/benches/blake3_compressions.rs
tcoratger: Done

jimpo (approved):

PR #1785, opened 2026-07-13
BINIUS-275: extract build_operation_witness and build the batched IntMul witness

benji-potamus:
## Summary

Extracts the arity-generic core of `BatchAndCheckWitness::build` into a standalone `build_operation_witness<const ARITY>`, then uses it in the M4 reduction prover to build the batched IntMul (4-operand) witness. Groundwork for the IntMul check (parent [BINIUS-274](https://linear.app/jimpo/issue/BINIUS-274), M4 POC).

Resolves [BINIUS-275](https://linear.app/jimpo/issue/BINIUS-275).

## Changes

- **`bitand.rs`** — new `pub fn build_operation_witness<'a, const ARITY: usize>(table, constants, constraints: impl IndexedParallelIterator<Item = [&'a Operand; ARITY]>) -> [Vec<Word>; ARITY]`. It holds the parallel instance-major assembly that used to live in `build`: allocate `ARITY` columns, stripe over instances, XOR each constraint's shifted operand terms into the columns. `BatchAndCheckWitness::build` now just projects each `AndConstraint` to `[&a, &b, &c]` and delegates.
- **`reduction.rs`** — when `cs.mul_constraints` is non-empty, `prove_reduction` builds the four IntMul operand columns `[A, B, HI, LO]`. They are **not consumed yet** (bound to `_mul_witness`), so the prover no longer hard-`assert!`s that MUL constraints are absent. The verifier still rejects MUL circuits (`#1776`), so this only unblocks building the columns for the next step.

## Notes for review

- **The one design cost:** the arity-generic version can't use the original clean 3-way `par_chunks_mut().zip().zip()` — you can't variadic-zip a const-generic number of streams. Instead it splits each column's spare capacity into per-stripe blocks and transposes so each stripe owns `[&mut [MaybeUninit<Word>]; ARITY]` (disjoint slices of distinct columns). Behavior and the instance-major layout are unchanged; the existing AND tests all pass verbatim.
- **Behavior change:** dropping the prover's MUL-reject `assert!` means a MUL circuit now flows through `prove_reduction` and produces an AND-only reduction (with the IntMul columns built and discarded). Intentional scaffolding — happy to keep a softer guard instead if you'd prefer.

## Testing

- `cargo test -p binius-m4-prover` — 31 pass (was 30). All existing AND-path tests are unchanged and green, confirming the extraction preserves behavior.
- New `intmul_operand_columns_match_the_single_instance_reference` exercises the `ARITY = 4` path directly: builds a `imul` circuit, and checks each of the four instance-major columns against the single-instance operand evaluator.
- `cargo clippy -p binius-m4-prover --tests --benches --examples -- -D warnings` clean; nightly `fmt --check` clean.

### DIFF
--- a/crates/m4-prover/src/bitand.rs
+++ b/crates/m4-prover/src/bitand.rs
@@ -2,10 +2,10 @@
 
 //! The batched BitAnd-check witness built from a populated batch value table.
 
-use std::mem::MaybeUninit;
+use std::{array, mem::MaybeUninit, slice};
 
 use binius_core::{
-	constraint_system::{AndConstraint, ShiftedValueIndex},
+	constraint_system::{AndConstraint, Operand, ShiftedValueIndex},
 	word::Word,
 };
 use binius_field::{AESTowerField8b as B8, PackedField};
@@ -68,23 +68,9 @@ pub struct BatchAndCheckWitness {
 impl BatchAndCheckWitness {
 	/// Builds the batched BitAnd witness from a populated wire-major batch table.
 	///
-	/// Every AND constraint is evaluated against every instance.
-	/// This produces `K * n_and` rows laid out instance-major.
-	/// Row `instance * n_and + j` is constraint `j` of that instance.
-	///
-	/// An operand is a XOR of shifted committed values:
-	///
-	/// ```text
-	/// operand(instance) = XOR_t shift_t( value(index_t, instance) )
-	/// ```
-	///
-	/// A value index splits by the witness offset:
-	/// - below it: a public word, the same constant in every instance.
-	/// - at or above it: a hidden word read from the table.
-	///
-	/// One index's hidden words across all instances form one contiguous row of the buffer.
-	/// So a term streams that row, XOR-ing its shifted words into the column.
-	/// This replaces the per-instance gather the instance-major layout required.
+	/// Every AND constraint contributes one row per instance to each of the three operand
+	/// columns `A`, `B`, `C`, laid out instance-major. This delegates to the arity-generic
+	/// `build_operation_witness`, projecting each constraint to its three operands.
 	///
 	/// # Arguments
 	///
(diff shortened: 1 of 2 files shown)


crates/m4-prover/src/bitand.rs
```diff
@@ -283,6 +201,131 @@ impl BatchAndCheckWitness {
...
+///
+/// Panics if the constraint count or the instance count is not a power of two.
+pub fn build_operation_witness<'a, const ARITY: usize>(
+	table: &ValueTable,
+	constants: &[Word],
+	constraints: impl IndexedParallelIterator<Item = [&'a Operand; ARITY]>,
+) -> [Vec<Word>; ARITY] {
+	// Collect the operand projections once. Their count is the per-instance constraint count, and
+	// the sequential inner loop below needs random access to them within each instance stripe.
+	let constraints: Vec<[&Operand; ARITY]> = constraints.collect();
```
jimpo: Avoid this collect, it's bad. Instead, make `constraints` of type `impl ExactSizeIterator<Item = [&'a Operand; ARITY]> + Clone`. And you can clone the iterator in the loop.

crates/m4-prover/src/bitand.rs
```diff
@@ -283,6 +201,131 @@ impl BatchAndCheckWitness {
...
+	let block = STRIPE_WIDTH.min(n_instances) * n_constraints;
+	{
+		// Split each column's spare capacity into per-stripe blocks, then transpose so each stripe
+		// owns one mutable block of every column. A stripe's blocks are disjoint slices of distinct
+		// columns, so holding all `ARITY` of them at once is sound.
+		let mut column_chunks = columns
+			.each_mut()
+			.map(|col| col.spare_capacity_mut()[..total].chunks_mut(block));
+		let n_stripes = total.div_ceil(block);
+		let stripe_blocks: Vec<[&mut [MaybeUninit<Word>]; ARITY]> = (0..n_stripes)
```
jimpo: I'd like to avoid the collect. You can fall back to unsafe construction of chunk slices from pointers instead. (So parallel iterate over the chunk indices and get the slices from the mut slices)

jimpo (changes requested):

crates/m4-prover/src/bitand.rs
benji-potamus: Done. `constraints` is now `impl ExactSizeIterator<Item = [&'a Operand; ARITY]> + Clone` — no collect. Each stripe clones the iterator to walk the constraint list. (Added `+ Sync` too, since the parallel `for_each` closure captures it by shared reference.) Pushed in 217d0fd.

crates/m4-prover/src/bitand.rs
benji-potamus: Done — the `stripe_blocks` Vec is gone. Now `(0..n_stripes).into_par_iter().for_each(...)` and each stripe reconstructs its `[&mut [Word]; ARITY]` from raw pointers to each column's spare capacity (`slice::from_raw_parts_mut(col_ptr.add(start), len)`), shared across tasks via a local `Send`/`Sync` `SendPtr` wrapper. Sound because the `[start, start+len)` stripe ranges partition `[0, total)`, so no two tasks alias. Pushed in 217d0fd.

jimpo (approved):

PR #1787, opened 2026-07-13
BINIUS-271: replace BivariateProductSumcheckProver with the shared store-backed prover

benji-potamus:
## Summary

Replaces the bespoke `BivariateProductSumcheckProver` with the store-backed `SharedSumcheckProver` + the existing `BivariateProductEvaluator`, then deletes it. This is the follow-through on [BINIUS-268](https://linear.app/jimpo/issue/BINIUS-268)'s finding that the shared path is transcript-equivalent and faster (the bespoke `execute`'s 4-way `into_par_iter().map().reduce()` MultiZip blocks vectorization of the `vpclmul` accumulation; the evaluator's `.chunk()`+`izip!` unrolls across vector accumulators).

Resolves [BINIUS-271](https://linear.app/jimpo/issue/BINIUS-271).

## Changes (two commits)

1. **Migrate all callers.** New `bivariate_product_prover(multilinears, sum)` helper (in `bivariate_product_evaluator.rs`, re-exported at the `sumcheck` root) — a store-backed constructor that loads the two columns into a fresh `MleStore` and drives a single `BivariateProductEvaluator`. It mirrors the old `BivariateProductSumcheckProver::new` signature, so every call site is a clean rename: shift `phase_1`/`phase_2`, the BaseFold ZK mask (`basefold_channel.rs`, still wrapped in `PaddedSumcheckDecorator`), the `sumcheck` bench, and the remaining test sites (`padded`, `zk_mlecheck`, the `round_evaluator` batch-equivalence test). The bespoke prover's round-trip self-test is ported to the evaluator module.
2. **Delete the bespoke prover** (`bivariate_product.rs` + its module declaration).

## Performance (speedup lane)

Single-threaded (rayon is off by default), `sumcheck/bivariate_product` bench, this machine (`target-cpu=native`), baseline = `main` (bespoke) vs branch (shared):

| n_vars | baseline (bespoke) | branch (shared) | speedup |
|---|---|---|---|
| 12 | 84.6 µs | 77.6 µs | **1.09×** (−8.2%) |
| 16 | 1.232 ms | 1.012 ms | **1.22×** (−17.8%) |
| 20 | 23.30 ms | 18.45 ms | **1.26×** (−20.8%) |

Consistent with BINIUS-268 (~6–11% single-threaded there; a touch more at the larger sizes here). **No regression at the smallest size tested** (n_vars=12 still wins), so I did *not* add the size-threshold guard BINIUS-268 flagged — that regression was specific to parallelizing tiny late rounds under rayon, which the default config doesn't do. Happy to revisit if you want the threshold anyway.

## Testing

- `cargo test -p binius-ip-prover -p binius-iop-prover -p binius-prover` — all green (ip-prover 59, iop-prover 40, prover 21+8+1). The ported evaluator test replaces the deleted bespoke self-test (net-zero test count). The `round_evaluator` batch-equivalence test still checks that 4 separate provers produce a byte-identical transcript to 1 batched shared prover — now with shared provers on both sides.
- `cargo clippy … --tests --benches -- -D warnings` clean; nightly `fmt --check` clean; `cargo doc -D warnings` clean.


crates/prover/benches/sumcheck.rs
```diff
@@ -37,7 +37,7 @@ fn bench_sumcheck_prove(c: &mut Criterion) {
 			b.iter_batched(
 				|| [multilinear_a.clone(), multilinear_b.clone()],
 				|multilinears| {
-					let prover = BivariateProductSumcheckProver::new(multilinears, sum);
+					let prover = bivariate_product_prover(multilinears, sum);
```
jimpo: Just remove this benchmark. We already benchmark this exactly in the `binius-ip-prover` crate bench.

benji-potamus: Done — removed the whole `bench_sumcheck_prove` (and dropped it from the `criterion_group!` + the now-unused imports); the `binius-ip-prover` `bivariate_product` bench covers it. Folded into the migration commit. Pushed 7130c3a.

jimpo (approved):

PR #1788, opened 2026-07-13
BINIUS-279: single-instance BLAKE3 and Keccak M4 proving tests

tcoratger:
Closes [BINIUS-279](https://linear.app/jimpo/issue/BINIUS-279).

Proves one BLAKE3 compression and one Keccak permutation through the M4 prover, each once, under timing spans. With tracing on, this shows the per-phase breakdown of a single proof. No prover change.

### The problem

There is no easy way to see where M4 proving time goes for a single hash primitive.

- The benchmarks prove large batches, which is right for throughput but hides the per-phase shape of one proof.
- The existing round-trip tests use a CRC circuit and an XMSS circuit, not the standard hash primitives.
- Keccak has no M4 proving path at all yet, only a witness-generation benchmark.

### The idea

Prove one primitive a single time, inside a timing span.

- The prover is already deeply instrumented: the commit and opening spans come from the FRI layers, and the reduction emits its own spans.
- Wrapping the proof in one span makes those nest beneath it.
- With a tree-printing subscriber, the result is a per-phase breakdown: encode, Merkle commit, the univariate zerocheck round, the shift sumchecks, and the FRI folding and queries.

No prover change is needed; the tests only supply the workload and turn tracing on.

### The change

- **new** integration test `crates/m4-prover/tests/prove_hash_primitives.rs`: two tests, one BLAKE3 compression and one Keccak permutation, each a single instance.
- each circuit allocates its inputs as witness wires, so it has no inout wires and is eligible for the batch witness table; the output is force-committed to keep the computation alive.
- a small tracing initializer, mirroring the one in the examples, prints the timing tree and is a no-op if a subscriber is already installed.
- **changed** `crates/m4-prover/Cargo.toml`: adds the tracing dev-dependencies.

### Scope

- reuses the existing BLAKE3 and Keccak circuit gadgets; the prover and verifier are untouched.
- adds the first path that proves a Keccak permutation through M4.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-m4-prover --all-targets`, nightly `fmt --check` — clean.
- `cargo test -p binius-m4-prover --test prove_hash_primitives` — both tests pass.

### Running with the timing tree

```text
RUST_LOG=debug cargo test -p binius-m4-prover --test prove_hash_primitives \
    prove_blake3_compression -- --nocapture
```

### DIFF
--- a/crates/m4-prover/tests/prove_hash_primitives.rs
+++ b/crates/m4-prover/tests/prove_hash_primitives.rs
@@ -29,7 +29,7 @@ use tracing_forest::ForestLayer;
 use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
 
 /// Base-2 logarithm of the instance count.
-const LOG_INSTANCES: usize = 0;
+const LOG_INSTANCES: usize = 10;
 
 /// Base-2 logarithm of the inverse Reed-Solomon rate: rate 1/2, matching the hash benches.
 const LOG_INV_RATE: usize = 1;
@@ -47,7 +47,7 @@ fn init_tracing() {
 	// Default to DEBUG, the level the prover's spans are recorded at.
 	// RUST_LOG overrides this default.
 	let env_filter = EnvFilter::builder()
-		.with_default_directive(LevelFilter::DEBUG.into())
+		.with_default_directive(LevelFilter::WARN.into())
 		.from_env_lossy();
 
 	// A second call does nothing once a global subscriber is installed.
--- a/crates/m4-prover/tests/prove_hash_primitives.rs
+++ b/crates/m4-prover/tests/prove_hash_primitives.rs
@@ -28,7 +28,7 @@ use tracing::{info_span, level_filters::LevelFilter};
 use tracing_forest::ForestLayer;
 use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
 
-/// Base-2 logarithm of the instance count: one instance, since each primitive is proved once.
+/// Base-2 logarithm of the instance count.
 const LOG_INSTANCES: usize = 0;
 
 /// Base-2 logarithm of the inverse Reed-Solomon rate: rate 1/2, matching the hash benches.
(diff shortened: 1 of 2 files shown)


crates/m4-prover/tests/prove_hash_primitives.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// The subscriber prints each root span's duration and its children's share of it.
+/// Verbosity defaults to `DEBUG`, the level at which the prover's spans are recorded.
+/// `RUST_LOG` overrides that default.
+/// A call after a subscriber is already installed does nothing.
+fn init_tracing() {
+	// Default to DEBUG, the level the prover's spans are recorded at.
+	// RUST_LOG overrides this default.
+	let env_filter = EnvFilter::builder()
+		.with_default_directive(LevelFilter::DEBUG.into())
```
jimpo:
```suggestion
		.with_default_directive(LevelFilter::WARN.into())
```

crates/m4-prover/tests/prove_hash_primitives.rs
```diff
@@ -0,0 +1,236 @@
...
+use binius_prover::OptimalPackedB128;
+use binius_transcript::ProverTranscript;
+use binius_verifier::config::StdChallenger;
+use rand::prelude::*;
+use tracing::{info_span, level_filters::LevelFilter};
+use tracing_forest::ForestLayer;
+use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
+
+/// Base-2 logarithm of the instance count: one instance, since each primitive is proved once.
+const LOG_INSTANCES: usize = 0;
```
jimpo:
```suggestion
const LOG_INSTANCES: usize = 10;
```

crates/m4-prover/tests/prove_hash_primitives.rs
```diff
@@ -0,0 +1,236 @@
...
+use binius_m4_verifier::Verifier;
+use binius_prover::OptimalPackedB128;
+use binius_transcript::ProverTranscript;
+use binius_verifier::config::StdChallenger;
+use rand::prelude::*;
+use tracing::{info_span, level_filters::LevelFilter};
+use tracing_forest::ForestLayer;
+use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
+
+/// Base-2 logarithm of the instance count: one instance, since each primitive is proved once.
```
jimpo:
```suggestion
/// Base-2 logarithm of the instance count.
```

jimpo (approved):

PR #1790, opened 2026-07-14
BINIUS-274: enable IntMul constraints for M4

tcoratger:
Closes [BINIUS-274](https://linear.app/jimpo/issue/BINIUS-274/enable-intmul-constraints-for-m4).

Wires the IntMul check into the data-parallel M4 reduction, so circuits with MUL constraints (not just AND) prove and verify.

### The problem

The M4 witness is a multilinear over three axes: `Z || RHO || Y` — bit within a word, instance, wire.
Each operation transforms it into operand columns indexed `Z || RHO || X`, where `X` is that operation's constraint index.

The AND-check and the IntMul check each reduce their columns to operand claims, but at **different** points:
- both share the bit point `r_Z` (the AND-check's univariate challenge),
- but their instance points `r_rho` differ — they are independent Fiat-Shamir draws,
- and their constraint points `r_x` differ in length (`log(n_and)` vs `log(n_mul)`).

The committed witness can only be folded over the instance axis **once**.
So the two instance points must be unified before the shift reduction folds the witness.

### The idea

This takes strategy #2 from the issue: **re-randomize only the instance variables**, leaving each operation's `r_x` alone.

For each of the 3 AND operands and 4 IntMul operands, fold the operand column down to a per-instance multilinear:

```
M_op[rho] = sum_{local, bit} lagrange(r_Z)[bit] * eq(r_x_op, local) * bit_b(column[rho * n_op + local])
```

- the Lagrange weights collapse the bit axis at the shared `r_Z`,
- the `eq(r_x_op)` tensor collapses that operation's constraint axis,
- what remains is a $2^m$ buffer over the $m$ instance variables.

A batched `MultilinearEvalProver` sumcheck then transports all these claims from their own instance points onto **one** shared `r_rho`.
The witness is folded once there, and the reduced evaluations become the operand claims the shift consumes — each operation keeping its own `r_x`.

### The change

The reduction gains the IntMul path; the no-MUL path is unchanged.

```
reduction.rs (prover + verifier):
- AND-check -> shift, IntMul columns built and dropped
+ IntMul check -> AND-check -> re-randomize RHO -> shift, with both operand claims
```

- The IntMul check runs **before** the AND-check (soundness, below).
- The IntMul per-bit operand claims are collapsed to oblong claims at the AND-check's `r_Z` via `lagrange_evals`, matching the AND operand claims' form.
- The verifier declares the IntMul logup* pushforward as a second committed oracle (`OracleSpec::new(LIMB_BITS)`), opened by the existing final oracle-relations call.

### Soundness

- **Ordering:** the IntMul check runs before the AND-check draws `r_Z`, so its per-bit operand evaluations are bound to the transcript first. A malicious prover therefore cannot choose them as a function of `r_Z`. The verifier enforces the same order.
- **Transport:** the batched multilinear-evaluation sumcheck binds each reduced operand evaluation to its operation-check claim; the shift then binds those same evaluations to the folded witness, which the ring-switch binds to the committed trace. The chain holds at the shared `r_rho`.
- The transcript for a circuit with no MUL constraints is byte-identical to before.

### Scope

- **new:** the IntMul check, the RHO re-randomization sumcheck, and the per-operand fold in the M4 prover/verifier reduction; the conditional logup* oracle spec.
- **changed:** the reduction's channel bound (`IPProverChannel` -> `IOPProverChannel`, and the verifier equivalent) so the IntMul check can commit and open its oracle.
- **fix:** the prover now zero-pads the public constants to the layout's power-of-two word count before the shift; the verifier already tolerated the short slice, and small MUL circuits tripped a size assertion without this.
- **left as follow-up:** the issue's lambda-combination of the 7 per-operand buffers into 2 per-operation buffers. It is a cost micro-optimization (the buffers are only $2^m$) and would require the shared shift protocol to accept a pre-batched claim, touching the single-instance path.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-m4-prover -p binius-m4-verifier --all-targets`, nightly `fmt --all` — clean.
- New tests: full protocol round-trip under real FRI with a MUL circuit (both oracles open), a four-product reduction round-trip exercising a non-trivial `r_x_mul`, and strict tamper rejection asserting the exact IntMul product-check error variant.
- Existing no-MUL round-trip and tamper tests pass unchanged in behavior.

### DIFF
--- a/crates/m4-prover/src/reduction.rs
+++ b/crates/m4-prover/src/reduction.rs
@@ -14,16 +14,43 @@
 //! The caller binds it to the committed trace by ring-switching at `r_j || r_rho || r_y`.
 //! Evaluating the trace's instance coordinates at `r_rho` performs that instance fold.
 //!
-//! Only the AND-check is reduced so far. When the circuit has MUL constraints the IntMul operand
-//! columns are also built, but they are not yet consumed by the reduction.
+//! When the circuit has MUL constraints the IntMul check runs too.
+//! It reduces the multiplication relation to per-bit operand claims at its own instance point.
+//! The AND-check reduces to a different instance point.
+//! A batched multilinear-evaluation sumcheck re-randomizes both to one shared `r_rho`.
+//! The witness is then folded once at that shared point and both operand claims feed the shift.
 
 use binius_core::{constraint_system::ConstraintSystem, word::Word};
 use binius_field::{AESTowerField8b as B8, Field, PackedField};
+use binius_iop_prover::channel::IOPProverChannel;
 use binius_ip::sumcheck::SumcheckOutput;
-use binius_ip_prover::channel::IPProverChannel;
-use binius_math::BinarySubspace;
-use binius_prover::protocols::shift::{KeyCollection, OperatorData};
-use binius_verifier::{config::B128, protocols::bitand::AndCheckOutput};
+use binius_ip_prover::sumcheck::{
+	MleToSumCheckEvaluator,
+	batch::batch_prove_and_write_evals,
+	mle_store::MleStore,
+	quadratic_mle_evaluator::QuadraticMleEvaluator,
+	round_evaluator::{RoundEvaluator, SharedSumcheckProver},
+};
+use binius_math::{
+	BinarySubspace, FieldBuffer, inner_product::inner_product,
+	multilinear::eq::eq_ind_partial_eval_scalars, univariate::lagrange_evals_scalars,
+};
+use binius_prover::{
+	fold_word::fold_words,
+	protocols::{
+		intmul::{prove::IntMulProver, witness::Witness as IntMulWitness},
+		shift::{KeyCollection, OperatorData},
+	},
+};
+use binius_utils::{checked_arithmetics::checked_log_2, rayon::prelude::*};
+use binius_verifier::{
+	config::B128,
+	protocols::{
+		bitand::AndCheckOutput,
+		intmul::IntMulOutput,
+		shift::{BITAND_ARITY, INTMUL_ARITY},
+	},
+};
 
 use crate::{
 	BatchAndCheckWitness, ValueTable,
@@ -33,7 +60,10 @@ use crate::{
 
 /// The prover's output of the M4 constraint reduction.
 pub struct ReductionProverOutput {
-	/// The instance-fold challenge: the low coordinates of the AND-check row point.
+	/// The shared instance-fold challenge the witness is folded at.
+	///
+	/// Without MUL constraints this is the low half of the AND-check row point.
+	/// With MUL constraints the re-randomization sumcheck produces it fresh.
 	pub r_rho: Vec<B128>,
 	/// The reduced claim: the instance-folded witness evaluated at the shift's final point.
 	pub witness_claim: SumcheckOutput<B128>,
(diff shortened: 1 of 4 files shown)


crates/m4-verifier/src/reduction.rs
```diff
@@ -83,13 +105,89 @@ where
...
+	// Reduce to one shared instance point and both operand claims at it.
+	let (r_rho, bitand, intmul) = match intmul_output {
+		Some(IntMulOutput {
+			eval_point: r_out_mul,
+			a_evals,
+			b_evals,
+			c_lo_evals,
+			c_hi_evals,
+		}) => {
+			// Collapse each IntMul operand's per-bit claims to one oblong claim at the univariate
```
jimpo: Factor this batching protocol out into a separate function to keep the verification function tidy.

crates/m4-verifier/src/reduction.rs
```diff
@@ -83,13 +105,89 @@ where
...
+			let BatchSumcheckOutput {
+				batch_coeff,
+				eval,
+				mut challenges,
+			} = batch_verify(log_instances, 2, &sums, channel)?;
+			challenges.reverse();
+			let r_rho = challenges;
+
+			// The prover wrote each reduced operand evaluation at `r_rho`, in the `sums` order.
+			let reduced = channel.recv_many(7)?;
```
jimpo: I think the code below would be cleaner if you did

```
let bitand_evals = channel.recv_array::<3>()?;
let intmul_evals = channel.recv_array::<4>()?;
```

These arrays go into OperatorData directly. You can also map for the scalar muls and concat them for the univariate evaluation.

crates/m4-prover/src/reduction.rs
```diff
@@ -56,32 +79,54 @@ pub fn prove_reduction<P, Channel>(
...
 	let shift_domain = andcheck_domain
 		.reduce_dim(Word::LOG_BITS)
 		.isomorphic::<B128>();
 
+	// The four IntMul operand columns of every constraint over every instance, laid out
+	// instance-major, present only when the circuit has MUL constraints.
+	// They drive the IntMul check.
+	// They are re-read below to build the instance-axis multilinears the re-randomization
+	// transports.
+	let mul_columns = (!cs.mul_constraints.is_empty()).then(|| {
```
jimpo: Consolidate all the intmul code in one conditional

crates/m4-prover/src/reduction.rs
```diff
@@ -90,69 +135,327 @@ where
...
+
+	// One decorated multilinear-evaluation prover per operand, all over `log_instances` variables.
+	// The fixed order is [BitAnd a, b, c | IntMul a, b, lo, hi]; the verifier reads it back the
+	// same.
+	let mut provers = Vec::with_capacity(7);
+
+	// BitAnd operands live at the AND-check instance point.
+	for (column, &claim) in and_columns.iter().zip(bitand_evals) {
+		let m = operand_rho_multilinear::<P>(column, &lagrange, r_x_and, log_instances);
+		provers.push(MleToSumCheckDecorator::new(MultilinearEvalProver::new(m, r_rho_and, claim)));
```
jimpo: This will expand out the r_rho_and tensor redundantly, once per Prover. Use the new SharedMleCheckProver, which will keep only one instance of the eq indicator in the MleStore.

crates/m4-prover/src/reduction.rs
```diff
@@ -90,69 +135,327 @@ where
...
+	// Fold each instance's constraint block at `r_x`, leaving one element per instance.
+	let r_x_tensor = eq_ind_partial_eval_scalars::<B128>(r_x);
+	let n_constraints = 1 << r_x.len();
+	let per_instance: Vec<B128> = folded_scalars
+		.chunks_exact(n_constraints)
+		.map(|chunk| inner_product(chunk.iter().copied(), r_x_tensor.iter().copied()))
+		.collect();
+	debug_assert_eq!(per_instance.len(), 1 << log_instances);
+
+	FieldBuffer::from_values(&per_instance)
```
jimpo: Try to avoid `from_values` -- it does a full copy. Produce the packed field elements in the loop.

crates/m4-prover/src/reduction.rs
```diff
@@ -90,69 +135,327 @@ where
...
+{
+	// Fold each word's bits at the univariate challenge: one field element per row.
+	let folded_rows = fold_words::<B128, P>(column, lagrange);
+	let folded_scalars: Vec<B128> = folded_rows.iter_scalars().collect();
+
+	// Fold each instance's constraint block at `r_x`, leaving one element per instance.
+	let r_x_tensor = eq_ind_partial_eval_scalars::<B128>(r_x);
+	let n_constraints = 1 << r_x.len();
+	let per_instance: Vec<B128> = folded_scalars
+		.chunks_exact(n_constraints)
```
jimpo: Rayonize this

crates/m4-prover/src/reduction.rs
```diff
@@ -90,69 +135,327 @@ where
...
+	lagrange: &[B128],
+	r_x: &[B128],
+	log_instances: usize,
+) -> FieldBuffer<P>
+where
+	P: PackedField<Scalar = B128>,
+{
+	// Fold each word's bits at the univariate challenge: one field element per row.
+	let folded_rows = fold_words::<B128, P>(column, lagrange);
+	let folded_scalars: Vec<B128> = folded_rows.iter_scalars().collect();
```
jimpo: This collect is unnecessary. You should be able to do this fully with packed elements. But if not, just change the PackedField parameter to fold_words.

jimpo (changes requested):
Thanks for the PR. This could use more self-review though @tcoratger :-)

crates/m4-verifier/src/reduction.rs
tcoratger: Just fixed by using

```rust
let bitand_evals = channel.recv_array::<BITAND_ARITY>()?;
let intmul_evals = channel.recv_array::<INTMUL_ARITY>()?;
```

crates/m4-prover/src/reduction.rs
tcoratger: Just merged the column-build and the check into a single let mul = ....

crates/m4-prover/src/reduction.rs
tcoratger: fixed

crates/m4-prover/src/reduction.rs
tcoratger: done

crates/m4-prover/src/reduction.rs
tcoratger: fixed by using

```rust
let folded_rows = fold_words::<B128, B128>(column, lagrange);
let folded_rows = folded_rows.as_ref();
```

crates/m4-verifier/src/reduction.rs
tcoratger: Fixed I've just created abstraction and small object for this

crates/m4-prover/src/reduction.rs
tcoratger: fixed

crates/m4-verifier/src/reduction.rs
```diff
@@ -106,8 +149,132 @@ where
...
-	})
+	Ok(ReductionVerifierOutput { r_rho, shift })
+}
+
+/// The degree of the re-randomization's round polynomials.
+///
+/// Each operand is a multilinear evaluation, expressed with the quadratic evaluator.
+/// Its degree-2 prime polynomial gains one degree from the equality factor, giving 3.
+// TODO: a degree-1 multilinear-eval store evaluator would drop this to 2; none exists yet.
+const RERAND_DEGREE: usize = 3;
```
tcoratger, line 157: We have a wasted degree here but let's check that in a followup if you agree since this one is already big

tcoratger:
> Thanks for the PR. This could use more self-review though @tcoratger :-)

Sorry, I was at the airport. I pushed it just before boarding because I wanted to unlock things, and since I'm not entirely familiar with Intmul yet, this PR contained some things that could be improved. I should have put it in draft mode while waiting...

I think things should be fixed now. Don't hesitate to let me know if anything is wrong, or force push or merge and handle things in a follow-up, whichever you prefer. :)

I've also added some tracing, like you did recently to this crate for performance debugging.

jimpo:
Please rebase onto main and re-test. https://github.com/binius-zk/binius64/pull/1796 changed the order of the RHO and X variables.

tcoratger:
> Please rebase onto main and re-test. #1796 changed the order of the RHO and X variables.

Done

jimpo (approved):

PR #1792, opened 2026-07-14
BINIUS-280: benchmark BatchAndCheckWitness::build

benji-potamus:
Adds a criterion benchmark for `BatchAndCheckWitness::build` (the batched BitAnd operand-column witness builder), modeled on the existing `keccak_witness_gen` bench.

Linear: [BINIUS-280](https://linear.app/jimpo/issue/BINIUS-280/benchmark-batchandcheckwitnessbuild)

## What

New bench `crates/m4-prover/benches/batch_and_check_witness.rs` (+ `[[bench]]` entry):
- Builds the Keccak-f1600 permutation circuit (25-word state as witness inputs, permuted outputs force-committed) — the same AND-heavy circuit the `keccak_witness_gen` bench uses.
- **Setup (not timed):** populate the wire-major `ValueTable` for 8192 instances, and prepare the circuit's constants + power-of-two AND constraints.
- **Timed:** `BatchAndCheckWitness::build(&table, &constants, &and_constraints)`, `sample_size(10)`.

## Measurement (this machine: 4 cores, GFNI/AVX-512, `target-cpu=native`)

```
batch_and_check_witness_build_8k/keccak_f1600   time: [499.20 ms 520.95 ms 544.68 ms]
```

`build` runs in parallel (via `build_operation_witness`'s striped `into_par_iter`) when the crate's `rayon` feature is on — this run was the default (single-threaded) config, matching how `keccak_witness_gen` is run.

Kept self-contained (its own circuit builder + input helper), consistent with the other benches in this crate (`keccak_witness_gen`, `blake3_compressions`), which don't share a helper module. Verified with `cargo clippy --package binius-m4-prover --tests --benches -- -D warnings` and `cargo +nightly-2026-01-01 fmt --check`.

### DIFF
--- a/crates/m4-prover/Cargo.toml
+++ b/crates/m4-prover/Cargo.toml
@@ -45,3 +45,7 @@ harness = false
 [[bench]]
 name = "blake3_compressions"
 harness = false
+
+[[bench]]
+name = "assemble_operation_witness"
+harness = false
--- a/crates/m4-prover/benches/assemble_operation_witness.rs
+++ b/crates/m4-prover/benches/assemble_operation_witness.rs
@@ -1,0 +1,83 @@
+// Copyright 2026 The Binius Developers
+//! Benchmark for assembling a batched per-operation witness — the operand-column layout an
+//! operation reduction consumes.
+//!
+//! Currently covers the BitAnd operation via [`BatchAndCheckWitness::build`]; the IntMul
+//! equivalent will be added alongside its protocol. Uses the Keccak-f1600 permutation circuit
+//! (see the `keccak_witness_gen` bench) as a realistic, AND-heavy constraint system: the circuit
+//! applies one permutation to a 25-word state, the state words are witness inputs and the permuted
+//! outputs are force-committed. Populating the batch table and preparing the constants/constraints
+//! are done once as setup; only the witness assembly is timed, over 8192 instances.
+
+use std::array;
+
+use binius_circuits::keccak::permutation::Permutation;
+use binius_core::word::Word;
+use binius_frontend::{Circuit, CircuitBuilder, Wire};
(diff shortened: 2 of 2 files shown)


crates/m4-prover/benches/batch_and_check_witness.rs
```diff
@@ -0,0 +1,86 @@
...
+	let and_constraints = {
+		let mut cs = circuit.constraint_system().clone();
+		cs.validate_and_prepare().unwrap();
+		cs.and_constraints
+	};
+
+	let mut group = c.benchmark_group("batch_and_check_witness_build_8k");
+	// Each iteration builds the operand columns for 8192 instances, so keep the sample count
+	// modest.
+	group.sample_size(10);
```
jimpo:
```suggestion
```

jimpo:
Rename the benchmark to `assemble_operation_witness`. We'll add the equivalent for intmul protocol eventually too.

crates/m4-prover/benches/batch_and_check_witness.rs
benji-potamus: Removed `group.sample_size(10)` (and its comment) — it now uses criterion's default sampling. Pushed 137c06b5.

benji-potamus:
Renamed to `assemble_operation_witness`: bench file → `benches/assemble_operation_witness.rs`, `[[bench]]` target + criterion group → `assemble_operation_witness`, and the case id is `bitand_keccak_f1600` so the IntMul equivalent can slot in as a sibling (e.g. `intmul_...`) under the same group later. Also dropped `sample_size(10)` per the inline suggestion. Pushed 137c06b5; runs as `assemble_operation_witness/bitand_keccak_f1600` (~499 ms here).

jimpo (approved):

PR #1793, opened 2026-07-14
BINIUS-235: replace the bor gate with a standard gadget

tcoratger:
@jimpo We have started this with `smul64`, let me know if you want to continue with other operations like this!

Part of [BINIUS-235](https://linear.app/jimpo/issue/BINIUS-235) (convert gates to gadgets). Sibling of the smul conversion in BINIUS-234.

Turns bitwise OR from a dedicated opcode into a gadget built from primitive gates. Pure refactor — the emitted constraint is bit-identical, no behavior change.

### The problem

`bor` was its own gate: an `Opcode::Bor`, a `emit_bor` bytecode instruction, and an `exec_bor` interpreter arm.
All of that existed to express one AND constraint that the primitive gates already produce.
That is exactly the redundancy BINIUS-235 wants gone.

### The idea

Bitwise OR decomposes into an AND and two XORs, bit for bit:

```
a | b = (a & b) ^ a ^ b
```

The fused AND-XOR gate already evaluates `(a & b) ^ w` inside a single AND constraint.
Set `w = a ^ b` and the OR falls out for free:

```
fax(a, b, a ^ b) = (a & b) ^ (a ^ b) = a | b
```

The XOR is a linear wire, so it fuses into the AND's third operand — still **1 AND constraint**, identical to the old gate.

### The change

```
CircuitBuilder::bor:
- graph.emit_gate(path, Opcode::Bor, [a, b], [z])   // bespoke opcode + bytecode + interpreter
+ bor64(self, a, b)                                  // fax(a, b, bxor(a, b))
```

- **new:** `bor64` gadget in `binius_frontend::ops`, plus its tests.
- **kept:** `CircuitBuilder::bor`, reopened in `ops`, still carrying the build-time folding fast-paths (idempotence, the zero and all-ones identities, constant-constant).
- **deleted:** the `Bor` opcode and gate module, `emit_bor`, `exec_bor` (bytecode tag `0x02` now unused, like the existing `0x04` gap), and the obsolete gate-level const-eval test.
- **visibility:** `const_of` and `algebraic_folding` widen from private to `pub` within the private `compiler` module, so the wrapper can reuse them from `ops`.

### Soundness

- The gadget's single AND constraint is `A & B == C` with `A = a`, `B = b`, `C = z ^ (a ^ b)`.
- After the linear XOR wire fuses, that is `a & b == z ^ a ^ b` — the exact constraint the old gate emitted.
- Constant folding is unchanged: constant operands still resolve at build time and emit no gate.

### Scope

- One opcode removed; the enum shrinks by one.
- Downstream callers are untouched — `CircuitBuilder::bor` keeps the same signature and folding behavior.

### Verification

- `cargo test -p binius-frontend` — 158 pass, including 4 new `bor` tests (correctness proptest, bit-extreme boundaries, constant fold, identity/idempotence).
- `cargo test -p binius-circuits` — 342 pass (heaviest downstream `bor` user, exercises the new bytecode path end-to-end).


tcoratger:
@jimpo I've just found out via the CI that this change is bad for circuit size, meaning that we should probably not do it and I haven't found that locally (the snapshot was super useful in this scenario)!

PR #1794, opened 2026-07-14
BINIUS-235: replace the iadd gate with a standard gadget

tcoratger:
Part of [BINIUS-235](https://linear.app/jimpo/issue/BINIUS-235) (convert gates to gadgets). Sibling of the smul conversion in BINIUS-234.

Turns the 64-bit `iadd` opcode into a gadget built from the carry-propagating addition gate that already exists. Pure refactor — the emitted constraints are identical and no circuit snapshot changes.

### The problem

Plain addition and addition-with-carry are the *same operation* — plain addition just has a zero carry-in.
Yet `iadd` was its own opcode, with a bespoke gate, a bytecode instruction, and an interpreter arm, sitting right next to `iadd_cin_cout` which already does the general case.
That is exactly the redundancy BINIUS-235 targets.

### The idea

Feed a zero carry-in to the general gate:

```
iadd(a, b) = iadd_cin_cout(a, b, 0)
```

The carry-propagating constraint mixes in a `cin_msb` term (the carry bit lives in the MSB of the carry-in word).
With a zero carry-in that term is zero, so the AND and linear constraints collapse to exactly what the plain gate emitted.

### Why this is a clean 1-to-1 mapping

- The zero carry-in is a **deduplicated constant** — it adds no gate, no evaluation instruction, no scratch wire.
- Each call site emits **one gate**, same as before.
- Gate count, evaluation-instruction count, and committed layout are all unchanged for every circuit that already carries a zero constant, which is all of the `iadd` users here.

### Scope

- **moved:** `iadd` into `binius_frontend::ops` (reopened `CircuitBuilder`).
- **deleted:** the `Iadd` opcode and gate module, plus its shape/constrain/emit arms.
- **kept:** `emit_iadd_cout` / `exec_iadd_cout` (bytecode `0x20`) — `assert_non_zero` reuses that instruction for its witness, so it is a shared primitive, not dead code.
- One opcode removed; callers are untouched (`iadd` keeps its signature).

Note on scope: `iadd_32` is intentionally **not** converted here. The 32-bit gate reduces to `iadd32_cin_cout(a, b, 0)` the same way, but that forces a zero constant into circuits that lack one (e.g. `blake3_compress`), trading a benign `+1` committed constant for a snapshot re-bless. Not worth bundling into a pure refactor; it can be its own decision.

### Verification

- `cargo test -p binius-frontend` — 155 pass · `cargo test -p binius-circuits` — 342 pass.
- **Snapshots unchanged and matching** for every `iadd` user: ethsign (secp256k1 carry chains), sha512, bitcoin_header_chain, bitcoin_p2pkh, ec_msm, hashsign, zklogin.


jimpo:
I think this is a good idea. Can you please just check that it doesn't have significant impact on witness generation for a circuit like SHA2 or Blake3? It's like one more instruction for witgen (probably negligible)

tcoratger:
> iadd

I've just checked locally to isolate `iadd` only for now which should not have any impact, we can check the other one in a followup

jimpo (approved):

PR #1795, opened 2026-07-14
Recursion

pauldelucia:


jimpo:
Hey @pauldelucia. I appreciate you working on this, but I can't review a 7,000+ line PR. Recursion is a big feature that requires planning. But having a POC like this can guide the planning.

We're going to need to break the epic up into features that can be merged independently, and in a sequence. Any new protocols (like this batch circuit evaluation protocol that is mentioned in this PR) should get spec updates too (the spec is at [binius.xyz](https://github.com/binius-zk/binius.xyz) under `whitepaper/`). See for example how I made a spec update [PR for BinMul constraints](https://github.com/binius-zk/binius.xyz/pull/119) before implementing.

I suggest you open a GitHub issue where we can discuss the roadmap of features to achieve recursion. Off the top of my head, it basically involves: 1) finishing BinMul constraints (in progress), 2) MerkleIPVerifierChannel implementations for a recursive circuit builder and a MerkleIPProverChannel impl for witness generator, 3) some way to make verification succinct or at least concretely cheaper.

The batch circuit evaluation protocol is worth special discussion. It's cool idea, but of somewhat narrow utility (it must batch proofs for the same circuit). I had in mind that we'd implement Spark to get succinct verification cost and recurse on that.

PR #1797, opened 2026-07-14
aarch64 poly64x2_t multiplications for the Monbijou field

benji-potamus:
Implements [BINIUS-251](https://linear.app/jimpo/issue/BINIUS-251/arith-bench-aarch64-specific-multiplications-for-monbijou-field).

## What

Adds a direct aarch64 PMULL path for the Monbijou field arithmetic in `binius-arith-bench`, makes the existing `OpsClmul`-based multiplications x86_64-only, and gives the inner-product benchmarks aarch64 coverage. `monbijou` is restructured into `x86_64` / `aarch64` / `soft64` submodules, mirroring how the `ghash` family separates its architecture backends.

The `monbijou::aarch64` module represents elements with the `poly64x2_t` underlier (like `ghash::aarch64`), so the multiply stays in NEON registers. Carryless multiplies use `vmull_p64` / `vmull_high_p64`, and the reduction folds the high limb down with a PMULL by the polynomial tail `0x1B` — the same strategy the generic `OpsClmul` path uses.

## Commits

1. **add aarch64 poly64x2_t Monbijou multiplications** — new `monbijou/aarch64.rs`, proptested against `soft64`.
2. **make Monbijou CLMUL multiplications x86_64-only** — rename `clmul.rs` → `x86_64.rs`, gate it (and the `mul_*_clmul` re-exports) on `target_arch = "x86_64"`, repoint the aarch64 benchmarks, drop the now-defunct aarch64 `uint64x2_t` Monbijou property tests.
3. **reduce the aarch64 Monbijou path with PMULL, not shifts** — switch the reduction to PMULL folding and keep the packed path in NEON registers; lifts the direct path from 1.4–2× slower to on par with the generic path.
4. **expose aarch64 Monbijou widening multiplies** — factor each multiply into public `mul_wide*` / `reduce*` halves (mirroring the `x86_64` module) so inner products can XOR-accumulate and reduce once.
5. **add aarch64 runs to the inner-product benchmarks** — aarch64 NEON runs for the ghash, ghash_sq, and Monbijou (base / 128b / 192b) inner-product groups.

## Notes for review

- **Scope**: the issue lists three functions (2×64b `mul`, 1×128b `mul_128b`, 2×128b `mul_sliced_128b`). I also implemented the **2×192b sliced mul** for benchmark parity. Happy to drop it if you'd rather keep this minimal.
- **Benchmarks**: the direct `poly64x2_t` path *matches* the generic `uint64x2_t` path but does not beat it (the generic path already lowers to the same PMULL). Full before/after numbers are in the PR comments.
- **CI cross-check**: the x86_64 and aarch64 *libraries* cross-compiled clean locally; the x86_64 *test/bench* binaries couldn't be linked in the sandbox (no cross C toolchain for criterion's `alloca` dep), so those inactive paths were verified by inspection and rely on CI.


benji-potamus:
## Benchmark: direct `poly64x2_t` vs. the generic `uint64x2_t` `OpsClmul` path

**Setup:** AWS Graviton (aarch64), `cargo bench --bench field_mul` (criterion, 100 samples), `RUSTFLAGS="-C target-cpu=native"`. "Before" = `main` (generic `OpsClmul` on `uint64x2_t`); "after" = this branch (direct `poly64x2_t`). Throughput counts base elements/underlier, so higher is better.

| Benchmark | before — `uint64x2_t` | after — `poly64x2_t` (this PR) | Δ time |
|---|---|---|---|
| `monbijou_64b` base mul | 1.074 µs · **1.908 Gelem/s** | 1.830 µs · 1.119 Gelem/s | **+70% (1.71× slower)** |
| `monbijou_128b` packed | 1.869 µs · **547.8 Melem/s** | 3.701 µs · 276.7 Melem/s | **+98% (1.98× slower)** |
| `monbijou_128b` sliced | 3.459 µs · **592.1 Melem/s** | 5.282 µs · 387.8 Melem/s | **+53% (1.53× slower)** |
| `monbijou_192b` sliced | 6.469 µs · **316.6 Melem/s** | 8.981 µs · 228.0 Melem/s | **+39% (1.39× slower)** |

`soft64` anchors (unchanged): 64b 137.7 · 128b 44.9 · 192b 28.9 Melem/s — both PMULL paths beat it by ~4–25×, as expected.

### Reading

**The direct `poly64x2_t` path as written is a regression** — 1.4×–2× slower than the generic path across the board. So the headline experiment result is negative.

The cause is the **reduction**, not the underlier or the trait indirection:

- Both paths use PMULL (`vmull_p64`) for the *multiplies* — the generic path via `OpsClmul::clmulepi64` → `vmull_p64`, which inlines to the same instruction.
- The generic path reduces with **4 more PMULLs** (multiply the high limb by the polynomial). My direct path instead reduces with **shifts + XORs** (soft64-style, ~15 ALU ops per reduce) since Monbijou's polynomial has only low-degree tail terms (`0x1B`). On Graviton, PMULL is cheap and well-pipelined, so trading 4 PMULLs for ~15 ALU ops loses.

In other words: going "direct" on `poly64x2_t` buys nothing here, and swapping the PMULL reduction for a shift reduction actively hurts.

### Options

1. **Abandon** — the existing generic `uint64x2_t` path is already the fast one on aarch64; keep it. (I'd then cancel the issue rather than land this.)
2. **Retry with a PMULL-based reduction** on `poly64x2_t` (mirroring the generic algorithm and `ghash::aarch64`). I'd expect it to roughly *match* the generic path — confirming the trait indirection costs nothing — but it wouldn't beat it, so the practical value is mostly "prove the direct path isn't faster."
3. **Land as-is** for the record / as a starting point, understanding it's slower.

My recommendation is **(1) abandon** unless you specifically want (2) to close the loop on whether a hand-written PMULL reduction can match/beat the generic path. Let me know which way you'd like to go.

benji-potamus:
## Update: PMULL reduction (supersedes the numbers above)

Switched the aarch64 reduction from shift-based to **PMULL-based** (folding the high limb by `·0x1B`, like the generic `OpsClmul` path), and removed the `u128` NEON↔GP register crossings in the packed path. Same measurement setup (Graviton, criterion, `-C target-cpu=native`).

| Benchmark | generic `uint64x2_t` | `poly64x2_t` shift (first attempt) | **`poly64x2_t` PMULL (now)** |
|---|---|---|---|
| `monbijou_64b` base mul | 1.908 Gelem/s | 1.119 | **1.878** |
| `monbijou_128b` packed | 547.8 Melem/s | 276.7 | **473.4** |
| `monbijou_128b` sliced | 592.1 Melem/s | 387.8 | **547.5** |
| `monbijou_192b` sliced | 316.6 Melem/s | 228.0 | **305.7** |

The direct `poly64x2_t` path now **matches** the generic path on the base mul and comes within a few percent on the sliced variants (packed is ~14% behind). It doesn't *beat* the generic path — that path already lowers to the same PMULL instructions — but it's no longer a regression. This replaces my earlier "abandon" recommendation, which was based on the slow shift-based version.

## Added: aarch64 inner-product benchmark coverage

The inner-product benchmarks previously compared `soft64` against the x86_64 CLMUL path with no aarch64 runs. Added the aarch64 NEON runs to every group — `ghash` (generic `clmul` on `uint64x2_t`), `ghash_sq` (`aarch64::mul_sliced` / `mul_wide_sliced`), and Monbijou base / 128b packed+sliced / 192b sliced — each pairing the reduce-every-term multiply with its widening counterpart. To support this, `monbijou::aarch64` now exposes public `mul_wide*` / `reduce*` functions mirroring the `x86_64` module (commit "expose aarch64 Monbijou widening multiplies"). All groups verified to execute (`cargo bench -- --test`).

Two new commits are on the branch: `expose aarch64 Monbijou widening multiplies` and `add aarch64 runs to the inner-product benchmarks`.

crates/arith-bench/src/monbijou/aarch64.rs
```diff
@@ -0,0 +1,344 @@
...
+///
+/// Because [`reduce`] is F2-linear, these limbs can be XOR-accumulated across many products and
+/// reduced only once — an inner product of `n` terms costs one reduction instead of `n`.
+#[inline]
+pub fn mul_wide(x: poly64x2_t, y: poly64x2_t) -> Wide {
+	unsafe {
+		let p0 = vreinterpretq_u64_p128(vmull_p64(vgetq_lane_p64::<0>(x), vgetq_lane_p64::<0>(y)));
+		let p1 = vreinterpretq_u64_p128(vmull_high_p64(x, y));
+		// Transpose the two `[lo, hi]` products into `[[lo0, lo1], [hi0, hi1]]`.
+		[vzip1q_u64(p0, p1), vzip2q_u64(p0, p1)]
```
jimpo: Defer the transpose to the reduction

benji-potamus: Done in `defer the aarch64 widening transpose to the reduction` (157f1a25). `mul_wide` now returns the two lane products untransposed (`[prod_0, prod_1]`) and `reduce` does the `vzip` gather once; the transpose commutes with the XOR accumulation so results are unchanged (proptests still pass). This speeds up the widening inner-product path, where `mul_wide` runs per term but `reduce` runs once:

| group | reduce-every-term | widening | speedup |
|---|---|---|---|
| base | 1.574 Gelem/s | 2.673 | 1.70× |
| 128b packed | 342 Melem/s | 657 | 1.92× |
| 128b sliced | 467 Melem/s | 1.007 Gelem/s | 2.16× |
| 192b sliced | 253 Melem/s | 593 | 2.34× |

It also let me collapse the two mul-by-X helpers into one, since both now operate on a single `[lo, hi]` product register.

crates/arith-bench/src/monbijou/aarch64.rs
```diff
@@ -0,0 +1,344 @@
...
+
+use core::arch::aarch64::*;
+
+use crate::Underlier;
+
+/// A widening (unreduced) base-field product `[prod_0, prod_1]`: one 128-bit carryless product per
+/// lane, each held in a register as `[lo, hi]`. `[uint64x2_t; 2]` is itself an [`Underlier`], so
+/// widening products XOR-accumulate. The transpose that gathers the limbs for the modular reduction
+/// is deferred to [`reduce`], so an inner product pays it once rather than per term.
+type Wide = [uint64x2_t; 2];
```
jimpo: Use `poly64x2_t` instead of `uint64x2_t`

crates/arith-bench/src/monbijou/aarch64.rs
```diff
@@ -0,0 +1,344 @@
...
+#[inline]
+pub fn reduce([p0, p1]: Wide) -> poly64x2_t {
+	unsafe {
+		let lo = vzip1q_u64(p0, p1);
+		let hi = vreinterpretq_p64_u64(vzip2q_u64(p0, p1));
+		let poly = vreinterpretq_p64_u64(vdupq_n_u64(POLY));
+		// First fold: fr_i = hi_i · 0x1B (lane 0 via PMULL, lane 1 via PMULL2).
+		let fr0 = vreinterpretq_u64_p128(vmull_p64(vgetq_lane_p64::<0>(hi), POLY));
+		let fr1 = vreinterpretq_u64_p128(vmull_high_p64(hi, poly));
+		let acc = veorq_u64(lo, vzip1q_u64(fr0, fr1));
```
jimpo: instead of veorq_u64, use `vaddq_p64`.

crates/arith-bench/src/monbijou/aarch64.rs
```diff
@@ -0,0 +1,344 @@
...
+/// The two products are first transposed with `vzip` into `[lo, hi]` limbs (`lo = [prod_0.lo,
+/// prod_1.lo]`, `hi = [prod_0.hi, prod_1.hi]`); this is the gather deferred out of [`mul_wide`].
+/// The high limb holds the coefficients of X^64..X^127, so `hi·X^64 ≡ hi·0x1B` folds it down. That
+/// carryless product is up to 69 bits, so its own overflow past X^63 is folded once more by the
+/// same `·0x1B`. Both folds are `PMULL`s. This is an F2-linear map, so unreduced products may be
+/// summed by XOR and reduced once at the end.
+#[inline]
+pub fn reduce([p0, p1]: Wide) -> poly64x2_t {
+	unsafe {
+		let lo = vzip1q_u64(p0, p1);
```
jimpo: In this function, use poly64x2_t instead of uint64x2_t. So vzip1q_p64 here.

crates/arith-bench/src/monbijou/aarch64.rs
```diff
@@ -0,0 +1,344 @@
...
+		let sr1 = vreinterpretq_u64_p128(vmull_p64(vgetq_lane_p64::<1>(fr1), POLY));
+		vreinterpretq_p64_u64(veorq_u64(acc, vzip1q_u64(sr0, sr1)))
+	}
+}
+
+/// Multiply a single unreduced product `[lo, hi]` (packed in one register, low limb in lane 0) by
+/// X: a one-bit left shift of the whole 128-bit value, carrying lane 0's top bit into lane 1. The
+/// product's degree ≤ 126 leaves room, so no reduction is needed here.
+#[inline]
+fn mul_x(p: uint64x2_t) -> uint64x2_t {
```
jimpo: poly64x2_t, not uint64x2_t

crates/arith-bench/src/monbijou/mod.rs
```diff
@@ -6,14 +6,20 @@
...
 
-pub mod clmul;
+#[cfg(target_arch = "aarch64")]
+pub mod aarch64;
 pub mod soft64;
+#[cfg(target_arch = "x86_64")]
+pub mod x86_64;
 
-pub use clmul::{
+#[cfg(target_arch = "x86_64")]
```
jimpo: Remove these re-exports and rename at call sites.

jimpo (changes requested):

crates/arith-bench/src/monbijou/aarch64.rs
benji-potamus: Done in `represent aarch64 Monbijou products as poly64x2_t` (e85052d7). `Wide` is now `[poly64x2_t; 2]`, and the packed `mul_wide_128b` / `reduce_128b` products are `[poly64x2_t; 3]`. The carryless products reinterpret with `vreinterpretq_p64_p128` instead of `..._u64_p128`.

crates/arith-bench/src/monbijou/aarch64.rs
benji-potamus: Done in e85052d7 — the transpose is now `vzip1q_p64` / `vzip2q_p64` on `poly64x2_t`, and the polynomial constant uses `vdupq_n_p64`, so no `vreinterpretq` round-trips remain in the gather.

crates/arith-bench/src/monbijou/aarch64.rs
benji-potamus: Done in e85052d7 — both fold accumulations now XOR with `vaddq_p64` in place of `veorq_u64`.

crates/arith-bench/src/monbijou/aarch64.rs
benji-potamus: Done in e85052d7 — `mul_x` now takes and returns `poly64x2_t`. The 128-bit shift itself still goes through `uint64x2_t` (there's no polynomial shift intrinsic), so it reinterprets to `uint64x2_t` at the top and back to `poly64x2_t` on the way out.

crates/arith-bench/src/monbijou/mod.rs
benji-potamus: Done in `drop the monbijou mul_*_clmul re-exports, reference x86_64:: directly` (f5c61b01). Removed all four `mul_*_clmul` re-exports and renamed the call sites to reference the `x86_64` submodule directly: `arch/x86_64.rs` (`x86_64::{mul, mul_128b}`), the `soft64` equivalence test (local `mul as mul_clmul` alias to disambiguate from `soft64::mul`), and the benches (`x86_64::mul` / `mul_128b` / `mul_sliced_128b` / `mul_sliced_192b`, labels updated to match).

benji-potamus:
## Review feedback addressed

Pushed two commits on top of the reviewed tip (fast-forward, `157f1a25` → `f5c61b01`; existing commits untouched so the review anchors stay put):

- **`represent aarch64 Monbijou products as poly64x2_t`** (e85052d7) — the four `poly64x2_t` / intrinsic threads. `Wide` is now `[poly64x2_t; 2]` (packed products `[poly64x2_t; 3]`); the reduction uses `vzip1q_p64` / `vzip2q_p64` / `vdupq_n_p64` and `vaddq_p64` in place of `veorq_u64`; `mul_x` takes/returns `poly64x2_t`; carryless products reinterpret via `vreinterpretq_p64_p128`.
- **`drop the monbijou mul_*_clmul re-exports, reference x86_64:: directly`** (f5c61b01) — removed the four `mul_*_clmul` re-exports from `monbijou/mod.rs` and repointed every call site at the `x86_64` submodule (`arch/x86_64.rs`, the `soft64` equivalence test, and the benches).

### Verification
- **aarch64 (native, Graviton):** all 77 `binius-arith-bench` tests pass, including the 7 `monbijou::aarch64` proptests against `soft64`; clippy clean.
- **x86_64 (cross):** `cargo check` + `clippy --tests --benches` for `x86_64-unknown-linux-gnu` with the full CLMUL/GFNI feature set — the renamed x86-gated code type-checks and lints clean.
- rustfmt clean on both.

The behavior is unchanged (the proptests against `soft64` still pass), so I didn't re-run the throughput numbers — the PMULL-reduction benchmarks in the comment above still stand.

jimpo:
There's a performance regression for the 128-bit extension, non-sliced variant. I think it's because of Karatsuba vs schoolbook multiplication. Rename the current version to `mul_wide_128b_karatsuba` and `mul_128b_karatsuba`, and add schoolbook versions of the wide mul and regular mul. Benchmark both, for both inner product (on wide mul) and field mul throughput.

benji-potamus:
## Schoolbook vs. Karatsuba for the packed 128b multiply

Done in `add schoolbook packed 128b Monbijou multiply alongside Karatsuba` (f0e6e5d3). Renamed the current packed path to `mul_wide_128b_karatsuba` / `reduce_128b_karatsuba` / `mul_128b_karatsuba`, and added `mul_wide_128b_schoolbook` / `reduce_128b_schoolbook` / `mul_128b_schoolbook`. Per your follow-up, the schoolbook wide mul sums `p01 + p10` internally, so its widening accumulator is **3 registers wide** (`[p00, p01+p10, p11]`) — same as Karatsuba, just trading Karatsuba's `(x0+x1)(y0+y1)` (3 PMULLs + 2 input XORs) for a direct 4-PMULL cross term. Both proptested against `soft64`.

**Your hypothesis is right — schoolbook wins.** (Graviton, criterion 100 samples, `-C target-cpu=native`; throughput = base elements/underlier, higher is better.)

### Field-mul throughput (`monbijou_128b`)

| variant | Melem/s | vs. Karatsuba |
|---|---|---|
| `aarch64::mul_128b_karatsuba` | 470 | — |
| **`aarch64::mul_128b_schoolbook`** | **568** | **+21%** |
| generic `uint64x2_t` (prior-run baseline) | 547 | schoolbook +4% |
| `soft64::mul_128b` | 45 | — |

Schoolbook not only beats the branch's Karatsuba path by 21%, it also edges past the generic `uint64x2_t` `OpsClmul` path (~547) — so the packed-128b regression you flagged is resolved.

### Inner product (`monbijou_128b_inner_product`, 2^10 terms)

| variant | reduce-every-term | widening | widening vs. Karatsuba |
|---|---|---|---|
| Karatsuba | 342 | 657 | — |
| **Schoolbook** | **393** | **853** | **+30%** |
| sliced | 467 | 1004 | — |
| `soft64` | 43.5 | 46.5 | — |

Schoolbook widening is +30% over Karatsuba widening (and +15% reduce-every-term). The sliced path is still the fastest packed-representation option overall (1004 widening), but among the non-sliced variants schoolbook is the clear winner.

Both variants are benchmarked in both groups (field-mul throughput and inner-product on the wide mul), so the comparison is reproducible via `cargo bench --bench field_mul -- monbijou_128b`.

Let me know if you'd like me to make schoolbook the default `mul_128b` / `mul_wide_128b` (dropping the `_karatsuba`/`_schoolbook` suffixes and keeping schoolbook), or leave both exposed for now.

jimpo (approved):

benji-potamus:
Thanks for the approval! CI on `f0e6e5d3` is green so far (wasm builds/tests, rust-doc, circuits, format) with the three `rust-checks` jobs still finishing — I'll confirm full green.

Two things before this lands, since it's still a **draft** and now sits behind `main`:

1. **Naming** — I left `mul_128b_karatsuba` and `mul_128b_schoolbook` both exposed. Want me to make **schoolbook the default** (`mul_128b` / `mul_wide_128b`, dropping the suffixes) since it's the faster one, and either drop Karatsuba or keep it under the `_karatsuba` name? Or leave both as-is?
2. **Landing** — happy to mark it ready, rebase onto latest `main`, and squash-merge once CI is green; or leave it to you. Just say which.

Either way I'll hold off merging until you confirm (and I'll rebase first, since the branch is behind `main`).

PR #1803, opened 2026-07-14
BINIUS-217: deferred-fold fused execute in the sumcheck MleStore

benji-potamus:
Step 2 of [BINIUS-202](https://linear.app/jimpo/issue/BINIUS-202/rearchitect-sumcheck-proving-for-shared-multilinears) (Option A): flips the sumcheck `MleStore` from **eager** to **deferred** column folding, fusing the fold into the next round's read pass — behind the unchanged `SumcheckProver` trait, invisible to adapters and the verifier.

Resolves [BINIUS-217](https://linear.app/jimpo/issue/BINIUS-217/deferred-fold-fused-execute-in-the-sumcheck-mlestore).

## What changed

- `MleStore::fold(r)` no longer touches column data: it records the challenge in a new `pending_fold` field and advances the logical variable count (and the small eq trackers). The physical buffers stay put.
- `MleStore::fused_execute` folds each column's four segments in place — writing the folded low/high halves back — and immediately feeds those still-cache-hot halves to the round evaluators, doing the fold and the accumulation read in **one memory pass** (~1.5× the column length of traffic instead of a plain read's 1× plus a separate fold's ~1.5×). Safe Rust: `split_at_mut` carves each column's front half into the low/high write regions, then rayon owns disjoint per-chunk sub-slices (a small `ChunkTask` transpose).
- `RoundEvaluator::execute` takes the fused path when a fold is pending, chunks are whole-packed, and no column is borrowed; it falls back to the eager path (`apply_pending_fold` + plain read) for the first round, sub-packing-width tiny rounds, and the test-only borrowed columns whose first fold is out of place. `finish` applies the last pending fold before emitting evals.

Fusing only reorders the associative GF(2ᵏ) additions, so every path produces **byte-identical** round polynomials, challenges, and emitted evals.

## Correctness

The BINIUS-216 differential equivalence tests pass byte-identical (round polys, challenges, evals). Full `binius-ip-prover` suite (59 lib tests) + `binius-iop-prover` + logup/fracaddcheck production tests pass, under both the default single-threaded shim and `--features rayon` (real parallel fused pass). Clippy (`--all --tests --benches --examples -D warnings`) and nightly fmt clean.

## Performance

Measured branch vs. baseline `main`, `--release` with `RUSTFLAGS="-C target-cpu=native"`, `--features rayon`:

| Benchmark | n_vars=12 | n_vars=16 | n_vars=20 |
|---|---|---|---|
| `fracaddcheck/prove` | 2.30× | 1.72× | 1.22× |
| `bivariate_product` shared-sumcheck store micro-bench | 6.4× | 1.75× | 1.10× |

Large sizes converge toward the steady-state memory-traffic ratio (2.5·len → 1.5·len ≈ 1.67×, diluted by evaluator arithmetic and dedup across sharers); small sizes win extra from one fused parallel dispatch per round instead of two.

## Note: eq-tracker folding stays eager

The issue framed `fold(r)` as strictly O(1). The **column** (big-data) fold is deferred and fused as specified; the **eq trackers** still fold eagerly inside `fold()`, since they're small relative to the columns and are read by evaluators before the columns are folded. So `fold()` is O(eq-expansion-size), not literally O(1). Happy to fuse the eq fold too if you'd prefer strict O(1) — it's a small follow-up.

### DIFF
--- a/crates/ip-prover/src/sumcheck/mle_store.rs
+++ b/crates/ip-prover/src/sumcheck/mle_store.rs
@@ -15,9 +15,15 @@
 //! registered tracker advances exactly once per [`MleStore::fold`] call, no matter how many
 //! evaluators reference it.
 //!
-//! Folding is eager: [`MleStore::fold`] advances every column immediately, and the round pass
-//! over the columns is a plain read. A deferred-fold variant that fuses the fold into the next
-//! round's read pass can replace the internals without changing this interface.
+//! Folding of the columns is *deferred*: [`MleStore::fold`] records the challenge — advancing the
+//! logical variable count and the (small) eq trackers — but leaves the column data untouched, so
+//! it is O(1) in the columns. The next round's pass then folds the columns while it reads them:
+//! [`MleStore::fused_execute`] applies the pending fold to each chunk in place and feeds the folded
+//! halves straight to the evaluators, fusing fold and read into one memory pass (~1.5x the column
+//! length of memory traffic instead of a plain read plus a separate fold's ~2.5x). The eager
+//! [`MleStore::apply_pending_fold`] serves the first round, the last fold before
+//! [`MleStore::final_evals`], and rounds too small to chunk. Fusing only reorders the associative
+//! field additions, so all paths produce identical round polynomials and evaluations.
 
 use binius_field::{Field, PackedField};
 use binius_math::{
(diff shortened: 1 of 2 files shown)


benji-potamus:
Pushed a follow-up (`2edbd90`) addressing the initial CI failure and conforming to the design plan on the Linear issue:

- **Fix `rust-doc` CI** — the public `fused_execute` doc linked the private `Column::Borrowed` variant, which `rustdoc -D warnings` rejects; reworded to a plain reference.
- **`column_slices` is now private** (plan §5) — with a fold pending it would misreport the live region; its only callers are internal (`final_evals`, `execute_context`).
- **Tests (plan §6)** — added an `n_vars = 14` case to the lockstep differential test, and a store-level differential test `test_fused_fold_matches_eager_reference` that folds all three column variants (Borrowed × Owned × SplitHalf) against an independent eager reference built from the `binius-math` fold primitive, asserting column data after every round, the fold-then-finish edge, and `final_evals`. This directly exercises the multi-chunk parallel fused pass.

All green locally: full `binius-ip-prover` lib suite (60 tests) under both the default shim and `--features rayon`, plus clippy, nightly fmt, and `cargo doc -D warnings`.

jimpo:
superceded by https://github.com/binius-zk/binius64/pull/1828

PR #1804, opened 2026-07-14
perf(circuits): reduce AND in bignum modular add and is_zero

wstran:
## What

The secp256k1 field arithmetic that dominates the elliptic-curve example circuits (`ethsign`, `bitcoin_p2pkh`, `bip32`) spends AND constraints that two `bignum` primitives on that path do not need.

Root cause: `PseudoMersennePrimeField::add` widens both operands and the modulus by one limb to hold the single-bit overflow of a reduced-input sum, so `add`, `biguint_lt`, `zero_unless`, and the trailing `sub` all run a limb wider than the field requires. Independently, `BigUint::is_zero` compares each limb to zero with `icmp_eq` and folds the results with `band`, spending two AND per limb.

Fix: keep the overflow in a carry wire. `add_with_carry_out` already exists, so the sum stays `l` limbs and the carry feeds the reduction test `need = carry OR (sum >= modulus)`; the conditional reduction subtracts through a new crate-internal `sub_with_borrow_out`. `is_zero` ORs the limbs and compares once, costing `n` AND instead of `2n`, with the same MSB-boolean output every caller already relies on.

```text
is_zero(n limbs):  n icmp_eq + n band          ->  (n-1) bor + 1 icmp_eq   (2n -> n)
field add:         (l+1)-limb add / lt / sub    ->  l-limb add_with_carry_out + l-limb sub
```

## Numbers

Deterministic AND-constraint counts from the example circuits:

| circuit | baseline | this PR | delta |
|---|---|---|---|
| ethsign | 176,819 | 162,132 | -8.31% |
| bitcoin_p2pkh | 125,578 | 114,942 | -8.47% |
| bip32 | 768,541 | 704,760 | -8.30% |
| sha256 (no EC) | 8,880 | 8,880 | unchanged |

```
repro: cargo run --release --example ethsign -- stat
```

These are constraint counts, not prover wall-clock. binius64 pads AND constraints to a power of two and each circuit above stays in the same `2^k` bucket, so the effect is a lower constraint count and witness size (`ethsign` committed witness 205,003 -> 192,158), the same category as #1686, not a change in proving time.

The `is_zero` TODO suggested adding an all-ones `BigUint` and reading the carry out once a carry-returning add existed. That precondition now holds, but the add-all-ones route measures 8 AND at four limbs (identical to the original) while the OR-fold is 4, so this resolves the TODO by the cheaper path.

## Correctness

- Inputs to field `add` are reduced, so the sum is below `2 * modulus` and the overflow past the top limb is exactly one bit. `need = carry OR (sum >= modulus)` is exact because a carry already implies the sum exceeds the modulus, and the wrapping subtraction borrows out by exactly the `2^(64*l)` the carry represents, so `borrow_out == carry` in every reachable case. The two now-redundant assertions drop out: the old extra-limb overflow assert was always true for reduced inputs, and the underflow assert is now the expected wraparound.
- `is_zero`: a `BigUint` is zero iff the bitwise OR of its limbs is zero. The result is the same MSB-boolean, and all callers (`ecdsa`, `secp256k1` point add and double) read only the MSB.
- 345 `binius-circuits` tests pass, including the field and ECDSA property and test-vector suites. `ethsign`, `bitcoin_p2pkh`, and `bip32` prove and verify end to end. `clippy --all-targets -D warnings` and `fmt` are clean.
- Snapshots re-blessed: `ethsign`, `bitcoin_p2pkh`, `ec_msm`.

## Scope

- Only the `bignum` primitives change; public signatures are unchanged (`sub_with_borrow_out` is `pub(super)`).
- `biguint_eq` uses the same per-limb `icmp_eq` and `band` fold and reduces the same way, but it has a single call site, so it is left for a follow-up.

### DIFF
--- a/crates/circuits/src/bignum/biguint.rs
+++ b/crates/circuits/src/bignum/biguint.rs
@@ -49,14 +49,18 @@ impl BigUint {
 
 	/// Checks whether BigUint is zero and returns the check result as a boolean wire.
 	pub fn is_zero(&self, b: &CircuitBuilder) -> Wire {
-		// TODO: it's more efficient to add all-1 BigUint and check carry out (jimpo)
-		//       do this once BigUint addition returning a Wire carry out exists
+		// A BigUint is zero exactly when the bitwise OR of its limbs is zero, so one
+		// comparison suffices. Comparing each limb and folding the booleans costs an
+		// `icmp_eq` and a `band` per limb; folding with `bor` and comparing once costs
+		// a `bor` per limb after the first, plus a single `icmp_eq`.
 		let zero = b.add_constant(Word::ZERO);
-		let msb_one = b.add_constant(Word::MSB_ONE);
-		self.limbs
+		let any_bit = self
+			.limbs
 			.iter()
-			.map(|&limb| b.icmp_eq(limb, zero))
-			.fold(msb_one, |lhs, rhs| b.band(lhs, rhs))
+			.copied()
+			.reduce(|lhs, rhs| b.bor(lhs, rhs))
+			.unwrap_or(zero);
+		b.icmp_eq(any_bit, zero)
 	}
 
 	/// Pads to given limb length with zeros.
--- a/crates/circuits/src/bignum/biguint.rs
+++ b/crates/circuits/src/bignum/biguint.rs
@@ -49,10 +49,6 @@ impl BigUint {
 
 	/// Checks whether BigUint is zero and returns the check result as a boolean wire.
 	pub fn is_zero(&self, b: &CircuitBuilder) -> Wire {
-		// A BigUint is zero exactly when the bitwise OR of its limbs is zero, so one
-		// comparison suffices. Comparing each limb and folding the booleans costs an
-		// `icmp_eq` and a `band` per limb; folding with `bor` and comparing once costs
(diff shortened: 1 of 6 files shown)


crates/circuits/src/bignum/biguint.rs
```diff
@@ -49,14 +49,18 @@ impl BigUint {
 
 	/// Checks whether BigUint is zero and returns the check result as a boolean wire.
 	pub fn is_zero(&self, b: &CircuitBuilder) -> Wire {
-		// TODO: it's more efficient to add all-1 BigUint and check carry out (jimpo)
-		//       do this once BigUint addition returning a Wire carry out exists
+		// A BigUint is zero exactly when the bitwise OR of its limbs is zero, so one
+		// comparison suffices. Comparing each limb and folding the booleans costs an
+		// `icmp_eq` and a `band` per limb; folding with `bor` and comparing once costs
+		// a `bor` per limb after the first, plus a single `icmp_eq`.
```
jimpo:
```suggestion
```

This comment references logic that no longer exists. I don't think the comment is necessary at all, it's self explanatory.

jimpo (approved):
Nice find!

PR #1808, opened 2026-07-15
BINIUS-286: use the OracleSetupChannel to determine M4 oracle specs

benji-potamus:
Resolves [BINIUS-286](https://linear.app/jimpo/issue/BINIUS-286/use-the-oraclesetupchannel-to-determine-oracle-specs-for-m4).

The M4 verifier hand-maintained its oracle-spec list, with a comment warning it "must stay in sync with the oracle the IntMul check commits." This replaces that with a dry run against `OracleSetupChannel`, exactly as `binius_prover::Verifier` (`IOPVerifier::oracle_specs`) already does — so the list can never drift out of sync with what the checks actually commit.

## What changed

- **`crates/m4-verifier/src/verify.rs`** — added `IOPVerifier::oracle_specs(&amp;self) -&gt; Vec&lt;OracleSpec&gt;`: it drives `self.verify` against an `OracleSetupChannel` (which records each `recv_oracle` and no-ops everything else), then returns `channel.into_oracle_specs()`. `Verifier::setup` now calls it instead of building `vec![OracleSpec::new(log_witness_elems), (mul?) OracleSpec::new(LIMB_BITS)]` by hand. Dropped the now-unused `LIMB_BITS` import.

Simpler than the spartan reference: M4's `IOPVerifier::verify` receives its trace oracle internally (no precommit-oracle argument to pre-record) and takes no public-inputs vector, and M4 commits without zero-knowledge, so the setup channel is `OracleSetupChannel::new(false)`.

## Correctness / tests

The derived specs must match what the prover commits, since `Prover::setup(&amp;verifier)` builds the shared FRI compiler from them — a wrong list makes the trace opening fail. The existing `prove_hash_primitives` tests (BLAKE3, Keccak) cover the trace-only branch, but both are MUL-free (`iadd_32` carry adders, not `imul`). I added **`prove_integer_multiplication`** — a minimal `imul` circuit — which is the one primitive here with MUL constraints, so it exercises the IntMul logup* pushforward-oracle branch end-to-end. All three prove/verify.

`cargo check --workspace --tests --benches` clean; `binius-m4-verifier` / `binius-m4-prover` suites pass; clippy (`--tests --benches -D warnings`) and nightly fmt clean.

### DIFF
--- a/crates/m4-verifier/src/verify.rs
+++ b/crates/m4-verifier/src/verify.rs
@@ -9,6 +9,7 @@ use binius_iop::{
 	channel::{IOPVerifierChannel, OracleLinearRelation, OracleSpec},
 	fri::{ConstantArityStrategy, calculate_n_test_queries},
 	merkle_tree::BinaryMerkleTreeScheme,
+	oracle_setup_channel::OracleSetupChannel,
 };
 use binius_ip::{
 	channel::IPVerifierChannel,
@@ -27,7 +28,7 @@ use binius_verifier::{
 	config::{B1, B128},
 	protocols::{
 		bitand::AndCheckOutput,
-		intmul::{IntMulOutput, common::LIMB_BITS, verify as verify_intmul_reduction},
+		intmul::{IntMulOutput, verify as verify_intmul_reduction},
 		shift::{self, BITAND_ARITY, INTMUL_ARITY, OperatorData},
 	},
 	ring_switch::{self, RingSwitchVerifyOutput},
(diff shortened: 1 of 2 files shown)


crates/m4-verifier/src/verify.rs
```diff
@@ -93,6 +94,21 @@ impl IOPVerifier {
...
+	///
+	/// The specs are derived by replaying the oracle-receiving sequence against an
+	/// [`OracleSetupChannel`] — which records each `recv_oracle` without doing real verification —
+	/// rather than hand-maintaining the list. M4 commits its oracles without zero-knowledge, so the
+	/// setup channel is constructed with `is_zk = false`.
+	pub fn oracle_specs(&self) -> Vec<OracleSpec> {
+		let mut channel = OracleSetupChannel::new(false);
+		// Discarded: the setup channel performs no real verification; we only read back the oracle
+		// specs it records as `verify` replays the `recv_oracle` calls.
+		let _ = self.verify(&mut channel);
```
jimpo: panic on error

jimpo (changes requested):

crates/m4-verifier/src/verify.rs
benji-potamus: Done — now `.expect("verifying against the no-op OracleSetupChannel cannot fail")` instead of discarding the `Result`, so an unexpected error surfaces loudly rather than silently producing a wrong spec list. Pushed as `0a0b636`. (All three prove/verify tests still pass, so the expect doesn't fire on the trace-only or MUL paths.)

jimpo (approved):

PR #1814, opened 2026-07-15
64-bit sliced GHASH multiply (schoolbook + karatsuba)

benji-potamus:
Resolves [BINIUS-293](https://linear.app/jimpo/issue/BINIUS-293/arith-bench-implement-64-bit-sliced-ghash-mul).

Adds a *sliced* (transposed) GHASH multiply to `crates/arith-bench/src/ghash/clmul.rs`, alongside the existing packed 128-bit-lane path. An element is `[U; 2] = [low64, high64]`: register 0 holds the low 64 bits of every packed element, register 1 the high 64 bits — one element per 64-bit lane, matching the monbijou sliced layout (`to_sliced`/`from_sliced`). Opened as a **draft experiment** (the issue has no evaluation criteria) — happy to reshape the API or the representation.

## What's added

- **`mul_wide_sliced_schoolbook`** / **`mul_wide_sliced_karatsuba`** — widening multiply returning the four transposed 64-bit limbs `[v0, v1, v2, v3]` of the 256-bit carry-less product (weights `X^0, X^64, X^128, X^192`). Schoolbook forms four 64×64 products; Karatsuba forms three plus two XORs. Both share `assemble_sliced_limbs`, which sums the overlapping halves.
- **`reduce_sliced`** — folds the four limbs modulo `X^128 + X^7 + X^2 + X + 1`. It mirrors `soft64::reduce` limb-for-limb, but every op is a per-64-bit-lane `slli_epi64`/`srli_epi64`/`xor`, so it never crosses lanes and needs **no CLMUL by `0x87`** (unlike the packed `gf2_128_reduce`). Being F2-linear, unreduced products XOR-accumulate and reduce once.
- **`mul_sliced_schoolbook`** / **`mul_sliced_karatsuba`** — compose widen + reduce.
- Helper `clmul_sliced` does the two per-lane 64×64 CLMULs (`0x00`, `0x11`) and transposes each element's 128-bit product into `[lo, hi]` limb registers via `unpacklo`/`unpackhi_epi64`.

The sliced functions need only `U: Underlier + OpsClmul` (no `PackedUnderlier`), since the reduction carries no broadcast constant.

## Correctness

Proptest differential tests against the `soft64` reference (`RUSTFLAGS=&#34;-C target-cpu=native&#34; cargo test -p binius-arith-bench ghash`, all pass):

- `test_sliced_schoolbook_matches_soft64`, `test_sliced_karatsuba_matches_soft64` — both lanes packed into `[__m128i; 2]` equal `soft64::mul` per element.
- `test_sliced_schoolbook_karatsuba_agree` — the two widening variants produce identical limbs.
- `test_sliced_wide_deferred_reduction` — XOR-accumulating two wide products and reducing once equals reducing each and summing (F2-linearity).

Clippy (`--tests --benches -D warnings`) and nightly fmt clean.

## Benchmarks

New benches in the `ghash` group: `clmul::mul_sliced_schoolbook`/`_karatsuba` over `&lt;__m128i&gt;` (2 elements/reg) and `&lt;__m256i&gt;` (4 elements/reg, `vpclmulqdq`-gated). Numbers below are `&lt;__m128i&gt;` on this AWS host (no `vpclmulqdq`, so the `&lt;__m256i&gt;` variants and `mul_clmul::&lt;__m256i&gt;` compile out here):

| variant | time (batch of 1024 muls) | per element |
|---|---|---|
| `mul_clmul::&lt;__m128i&gt;` (packed, 1 elem/reg) | 2.77 µs | ~2.7 ns |
| `mul_sliced_schoolbook::&lt;__m128i&gt;` (2 elem/reg) | 6.34 µs | ~3.1 ns |
| `mul_sliced_karatsuba::&lt;__m128i&gt;` (2 elem/reg) | 5.94 µs | ~2.9 ns |

Read: **Karatsuba beats schoolbook by ~6%** (3 vs 4 CLMULs), the comparison the issue asks for. Per element the sliced path is marginally behind the packed path at 128-bit width — expected, since `__m128i` packs only 2 elements and the transpose overhead isn't amortized. The sliced representation's win (cheap per-lane reduction, no cross-lane `0x87` CLMUL) should show at wider vectors, which I couldn't measure here for lack of `vpclmulqdq` — worth a run on an AVX-512 / `vpclmulqdq` host.

### DIFF
--- a/crates/arith-bench/src/ghash/clmul.rs
+++ b/crates/arith-bench/src/ghash/clmul.rs
@@ -22,6 +22,8 @@
 //! This implementation is derived from:
 //! <https://github.com/google/longfellow-zk/blob/main/lib/gf2k/sysdep.h>
 
+use std::array;
+
 use crate::{PackedUnderlier, Underlier, underlier::OpsClmul};
 
 /// Multiply a packed GHASH field element by X^{-1} using SIMD operations.
@@ -148,7 +150,7 @@ fn gf2_128_shift_reduce<U: Underlier + OpsClmul + PackedUnderlier<u128>>(t1: U) 
 /// In the *sliced* representation each register packs one 64-bit slice per 64-bit lane, so per
 /// 128-bit lane `clmulepi64::<0x00>` multiplies the low-qword element and `::<0x11>` the high-qword
 /// element. Leaving the products un-transposed lets an inner product XOR-accumulate them cheaply;
-/// the transpose is deferred to [`reduce_sliced_schoolbook`].
+/// the transpose is deferred to [`reduce_sliced`].
 #[inline]
 fn clmul_pair<U: Underlier + OpsClmul>(u: U, v: U) -> [U; 2] {
 	[U::clmulepi64::<0x00>(u, v), U::clmulepi64::<0x11>(u, v)]
@@ -158,9 +160,9 @@ fn clmul_pair<U: Underlier + OpsClmul>(u: U, v: U) -> [U; 2] {
 /// `[p00, cross, p11]` of the 256-bit product of `x` and `y`, at weights `X^0`, `X^64`, `X^128`.
 ///
 /// Each element is `[U; 2] = [low64, high64]`; each returned pair is a `[U; 2]` of raw
-/// (un-transposed) CLMUL products, one per packed element (see [`clmul_pair`]). Schoolbook forms
+/// (un-transposed) CLMUL products, one per packed element (see `clmul_pair`). Schoolbook forms
 /// four 64×64 products and sums the two cross terms. No transpose or reduction happens here — both
-/// are F2-linear and deferred to [`reduce_sliced_schoolbook`], so an inner product XOR-accumulates
+/// are F2-linear and deferred to [`reduce_sliced`], so an inner product XOR-accumulates
 /// the three pairs and reduces once.
 #[inline]
 pub fn mul_wide_sliced_schoolbook<U: Underlier + OpsClmul>(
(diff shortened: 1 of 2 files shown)


crates/arith-bench/src/ghash/clmul.rs
```diff
@@ -142,6 +142,139 @@ fn gf2_128_shift_reduce<U: Underlier + OpsClmul + PackedUnderlier<u128>>(t1: U)
...
+/// products of the paired low- and high-lane slices; `unpacklo`/`unpackhi_epi64` then gather the
+/// low and high 64 bits of every element's product back into two transposed limb registers.
+#[inline]
+fn clmul_sliced<U: Underlier + OpsClmul>(u: U, v: U) -> [U; 2] {
+	// Per 128-bit lane, `0x00` multiplies the low-qword elements and `0x11` the high-qword
+	// elements.
+	let prod_0 = U::clmulepi64::<0x00>(u, v);
+	let prod_1 = U::clmulepi64::<0x11>(u, v);
+	[
+		U::unpacklo_epi64(prod_0, prod_1),
```
jimpo: Don't do unpack here, keep the data in the non-unpacked layout. unpack in the reduction if necessary.

jimpo (changes requested):

crates/arith-bench/src/ghash/clmul.rs
benji-potamus: Done in b028c82e. `mul_wide_sliced_*` now keep the raw un-transposed CLMUL products: `clmul_pair` returns `[clmul::<0x00>, clmul::<0x11>]` and the widening multiplies return `[[U; 2]; 3]` (the `p00`/`cross`/`p11` pairs, XOR-accumulable for deferred inner products). The single `unpacklo`/`unpackhi_epi64` transpose now lives in `reduce_sliced`, which assembles the four 64-bit limbs and folds. Net effect: a standalone multiply does 6 unpacks instead of 8 (the cross term is summed before transposing), and an inner product transposes once at the end rather than per term. All four proptests against `soft64` still pass; clippy + nightly fmt clean.

crates/arith-bench/src/ghash/clmul.rs
```diff
@@ -142,6 +142,122 @@ fn gf2_128_shift_reduce<U: Underlier + OpsClmul + PackedUnderlier<u128>>(t1: U)
...
+pub fn mul_wide_sliced_karatsuba<U: Underlier + OpsClmul>(
+	[x0, x1]: [U; 2],
+	[y0, y1]: [U; 2],
+) -> [[U; 2]; 3] {
+	let p00 = clmul_pair(x0, y0); // x.lo · y.lo
+	let p11 = clmul_pair(x1, y1); // x.hi · y.hi
+	// (x.lo+x.hi)·(y.lo+y.hi) = x.lo·y.lo + (x.lo·y.hi + x.hi·y.lo) + x.hi·y.hi
+	let pm = clmul_pair(U::xor(x0, x1), U::xor(y0, y1));
+	// cross = pm - x.lo·y.lo - x.hi·y.hi (subtraction is XOR in characteristic 2).
+	let cross = Underlier::xor(Underlier::xor(pm, p00), p11);
```
jimpo: This should be part of the reduction, not the mul_wide. So different reduction methods for karatsuba and schoolbook (though karatsuba can call schoolbook's reduction).

crates/arith-bench/src/ghash/clmul.rs
```diff
@@ -142,6 +142,122 @@ fn gf2_128_shift_reduce<U: Underlier + OpsClmul + PackedUnderlier<u128>>(t1: U)
...
+	let [p11_0, p11_1] = p11;
+
+	// Transpose the raw per-element products into 64-bit limbs, summing the overlapping halves at
+	// the `X^64` and `X^128` boundaries.
+	let v0 = U::unpacklo_epi64(p00_0, p00_1);
+	let v1 = U::xor(U::unpackhi_epi64(p00_0, p00_1), U::unpacklo_epi64(cross_0, cross_1));
+	let v2 = U::xor(U::unpackhi_epi64(cross_0, cross_1), U::unpacklo_epi64(p11_0, p11_1));
+	let v3 = U::unpackhi_epi64(p11_0, p11_1);
+
+	// Fold the top limb v3 (weight X^192) down into v1 and v2.
```
jimpo: No, use the CLMUL-based reduction. Like the regular non-sliced reduction strategy.

jimpo (changes requested):

crates/arith-bench/src/ghash/clmul.rs
benji-potamus: Done in f5549d0. `mul_wide_sliced_karatsuba` now returns the raw three products `[p00, pm, p11]` (with `pm = (x.lo+x.hi)·(y.lo+y.hi)`) — no recombination in the widening multiply. The reduction is split: `reduce_sliced_karatsuba` recovers `cross = pm - p00 - p11` and then delegates to `reduce_sliced_schoolbook`. So `pm` XOR-accumulates across an inner product just like the two corner products, and the cross term is recombined only once, at reduction time.

crates/arith-bench/src/ghash/clmul.rs
benji-potamus: Done in f5549d0. `reduce_sliced_schoolbook` now uses the packed CLMUL reduction (`reduce` / `gf2_128_reduce`) rather than the per-lane shift folds. Since `clmul_pair` leaves each element's product un-transposed, per pair index the triple `[p00, cross, p11]` is exactly a non-sliced wide product `[t0, t1, t2]` — so I run the existing packed `reduce` on each element (vectorized across the register's 128-bit lanes), then transpose the two reduced 128-bit results back into the sliced `[low64, high64]` layout with `unpacklo` / `unpackhi_epi64`. The transpose moved to the very end because a fully-transposed limb layout can't be clmul-reduced lane-wise (clmul takes one qword per operand), whereas the un-transposed per-element products reduce directly. All four sliced proptests against `soft64` pass; clippy + nightly fmt clean. Let me know if you had a different reduction shape in mind.

benji-potamus:
Picked up your `improve reduction` commit (a0daac80) — that's a much nicer sliced reduction than my de-slice-per-element version: it folds by `0x87` in place with `clmulepi64::&lt;0x01&gt;` on the un-transposed products and only `unpack`s where the two elements' halves recombine, so the result lands directly in sliced form.

It broke the test + doc build, though, so I fixed that up and folded it into your commit (pushed **e2cc5b60**):
- the rename `reduce_sliced_schoolbook` → `reduce_sliced` left a stale call in `test_sliced_wide_deferred_reduction` and four intra-doc links;
- two public docs linked the private `clmul_pair` (rustdoc `-D warnings` rejects that), demoted to plain code spans;
- rewrote the `reduce_sliced` doc to describe the in-place fold rather than my old reduce-each-element approach.

Green now: all four sliced proptests vs `soft64`, `clippy --tests --benches -D warnings`, `cargo doc`, and nightly fmt. Still a draft — over to you on where to take it next.

jimpo (approved):

benji-potamus:
Thanks for the approval! CI is green on e2cc5b60 (all rust-checks, rust-doc, format, circuits-snapshot, wasm). The merge API rejects a direct merge here with "Changes must be made through the merge queue", and I can't enqueue via the API from my side — so whenever you're ready, hitting **Merge when ready** will enqueue it. (Also recategorized the Linear label `experiment` → `feature`.)

PR #1816, opened 2026-07-15
BINIUS-253: add SlicedUnderlier for sliced packed extension fields

tcoratger:
Closes [BINIUS-253](https://linear.app/jimpo/issue/BINIUS-253).

Adds an underlier type whose subdivisions read in a transposed order, so a sliced packed extension field can expose its subfield coordinates in the order the rest of the stack expects. Public field arithmetic only — no protocol change, no new soundness assumption.

### The problem

A sliced packing stores each coordinate of every lane in its own register.
For a degree-two extension packed four wide (elements $A,B,C,D$, each a pair $(c_0, c_1)$), the two registers hold:

```
register 0 (low coords):   A.c0  B.c0  C.c0  D.c0
register 1 (high coords):  A.c1  B.c1  C.c1  D.c1
```

Generic code that reinterprets these bytes as a flat list of coordinates wants them element-major:

```
A.c0, A.c1,  B.c0, B.c1,  C.c0, C.c1,  D.c0, D.c1
```

But the existing scaled underlier reads them register-by-register — `A.c0, B.c0, C.c0, D.c0, ...` — which pairs `A.c0` with `B.c0` as if they were one element. Wrong pairing, everything downstream breaks.

### The idea

Same `[U; N]` storage, transposed read order.
View the registers as a grid of `N` coordinate-rows by `WIDTH` element-columns, and read column by column:

```
              elem A   elem B   elem C   elem D
row c0  ->      A.c0     B.c0     C.c0     D.c0
row c1  ->      A.c1     B.c1     C.c1     D.c1
```

The subdivision at index `i` is:

```
limb        = i mod N        // which coordinate register
sub-position = i div N       // which element within it
```

`N = 2`, so index 3 reads limb 1, sub-position 1 = `B.c1` — the second coordinate of the second element — exactly where element order expects it. Both `N` and the per-limb divisor are compile-time powers of two, so the index math lowers to shifts and masks.

### The change

- **new:** `crates/field/src/underlier/sliced.rs` — `SlicedUnderlier<U, SubU, N>`, a full `UnderlierType` with the transposed `Divisible` and a transpose-consistent `interleave`.
- **changed:** `SlicedPackedField` now advertises this as its underlier (`WithUnderlier`), and the GHASH² coordinate view gets per-limb arithmetic so it is a real packed field.

```diff
- // sliced GHASH² packing had no underlier bridge; PackedSubfield casts could not instantiate
+ impl WithUnderlier for SlicedPackedField { type Underlier = SlicedUnderlier<..>; }
+ // + per-limb Square / InvertOrZero / Mul / WideMul for the GHASH coordinate view
```

This is the sliced sibling of the scaled underlier.
They share the same `[U; N]` bytes and differ **only** in iteration order.
This does **not** change how a sliced GHASH² element multiplies — that arithmetic already lives in `packed_ghash_sq`. It only fixes the order in which the *coordinates* are read when the packing is viewed as its subfield.

### Where it plugs in

`PackedSubfield<P, FSub> = PackedPrimitiveType<P::Underlier, FSub>` and the reinterpret casts (`cast_bases_mut`, used by `transpose` and the prover's ring-switch) require the underlier to yield subfield elements in element-major order. `SlicedUnderlier` is what makes that hold for sliced GHASH². The coordinate view multiplies per limb — one full-width CLMUL per limb, not a per-lane scalar loop.

### Soundness

- Deterministic, public field plumbing: no transcript, no challenge, no secret, no randomness in the logic.
- Every `unsafe` index is bounded by a checked invariant; the layout casts (`Pod`, `Zeroable`, `TransparentWrapper`) rest on `#[repr(transparent)]` over `[U; N]`.
- The transposed reindexing is a permutation, so element counts and the algebraic content are unchanged.

### Scope

- **new:** the `SlicedUnderlier` type and its tests.
- **changed:** `SlicedPackedField` gains `WithUnderlier`; the GHASH² coordinate view gains arithmetic.
- **left untouched:** GHASH² multiply/square/invert, and the scaled underlier.

### Verification

- `cargo check -p binius-field --all-targets`, `clippy -p binius-field --all-targets`, nightly `fmt --check` — clean.
- `binius-field`: 229 lib tests + doctests pass.
- Transposed order pinned against an independent reference; `interleave` pinned against the primitive interleave; the real `cast_bases_mut` exercised end-to-end at all three widths, with per-lane subfield arithmetic checked against the scalar field.

### DIFF
--- a/crates/field/src/packed_ghash_sq.rs
+++ b/crates/field/src/packed_ghash_sq.rs
@@ -15,16 +15,19 @@
 //! multiplies, not four.
 
 use std::{
+	array,
 	iter::Sum,
-	ops::{Add, AddAssign, Sub, SubAssign},
+	ops::{Add, AddAssign, Mul, Sub, SubAssign},
 };
+
+use bytemuck::Pod;
 
 use crate::{
 	BinaryField128bGhash, Divisible, GhashSq256b, PackedField, WideMul,
-	arch::{M128, M256, M512, PackedPrimitiveType},
+	arch::{LaneWideProduct, M128, M256, M512, PackedPrimitiveType},
 	arithmetic_traits::{InvertOrZero, Square},
 	sliced_packed_field::SlicedPackedField,
-	underlier::{UnderlierType, WithUnderlier},
+	underlier::{SlicedUnderlier, UnderlierType, WithUnderlier},
 };
 
 /// The packed GHASH coordinate register backing a `SlicedGhashSq256b<U>`.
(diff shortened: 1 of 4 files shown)


crates/field/src/packed_ghash_sq.rs
```diff
@@ -196,14 +199,104 @@ where
...
+///
+/// A sliced GHASH² packing is `N` GHASH limbs sharing one sliced underlier.
+/// Reinterpreting the same bytes as a packed GHASH field yields this type.
+/// Its lanes are the extension coordinates, read in the sliced order.
+/// GHASH arithmetic acts per lane.
+/// So each operation delegates to the limbs — one full-width CLMUL per limb, not a scalar loop.
+///
+/// [`PackedSubfield`]: crate::PackedSubfield
+type SlicedGhashSubfield<U, const N: usize> =
+	PackedPrimitiveType<SlicedUnderlier<U, M128, N>, BinaryField128bGhash>;
```
tcoratger, line 212: @jimpo Just put there an example of stuff that we can use, if this sounds not good and you prefer to just have the sliced.rs file for this PR, just let me know and I can remove these ones.

jimpo:
@tcoratger I'm not sure we actually need this pattern. I'm hoping `SlicedPackedField` alone is fine. There's not a very good way to test this yet though because the protocols are still somewhat hardcoded around B128. It'll be a bigger refactor to relax that.

Issue #1815, opened 2026-07-15
Sha256 still carries the len_bytes-selected-digest footgun that #1809 fixed for keccak — convert to a sha256_varlen function gadget


pauldelucia:
Follow-up from #1758 (closed) and #1809. #1809 replaced the `Keccak256` struct with the `keccak256_varlen` function gadget, which closed the underconstrained-`len_bytes` shape: the digest is now computed and returned rather than asserted against a caller-supplied wire. The same shape still lives in `Sha256`.

**The problem.** `Sha256::new` (`crates/circuits/src/sha256/mod.rs:79`) takes a caller-supplied `digest: [Wire; 4]` and asserts it internally:

```rust
builder.assert_eq_v("2b.digest", digest, final_digest);   // mod.rs:241
```

where `final_digest` is a `multi_wire_multiplex(...)` selected by `end_block_index` / `w_bd`, both derived from `len_bytes` (`mod.rs:219`, `:224`). The only in-circuit constraint on `len_bytes` is an upper bound (`too_long`, `mod.rs:126-127`) — its low value is unconstrained. So a caller that leaves `len_bytes` free lets a prover choose `(len_bytes, message, digest)` to satisfy the assert for a digest of its choosing — exactly the underconstrained-assert hole the old `Keccak256` struct had.

**The sound pattern already exists.** `sha512_varlen` (`sha512/mod.rs:153`) and now `keccak256_varlen` both *compute and return* the digest with no caller `digest` argument — so there's no free digest wire to forge a match against, and the residual `len_bytes` obligation becomes explicit and caller-owned via `&ByteVec`. There's no `sha256_varlen` equivalent yet.

**Proposed fix.** Add `sha256_varlen(builder: &CircuitBuilder, message: &ByteVec) -> [Wire; 4]` modeled on `sha512_varlen` / `keccak256_varlen` (derive the padded words, return the digest), and migrate the varlen call sites. Happy to send the PR following the #1809 pattern.

jimpo:
Thanks! Feel free to put up a PR for this, it'd be very helpful!

PR #1817, opened 2026-07-16
BINIUS-294: replace the Sha256 struct with sha256_varlen

pauldelucia:
Implements [BINIUS-294](https://linear.app/jimpo/issue/BINIUS-294) (GitHub #1815).

Replaces the underconstrained `Sha256` checker struct with a `sha256_varlen` function gadget and removes the struct entirely — closing the `len_bytes`-selected-digest footgun, mirroring how #1809 handled keccak.

## `sha256_varlen`

Adds `sha256_varlen(builder: &CircuitBuilder, message: &ByteVec) -> [Wire; 4]`, modeled on the existing [`sha512_varlen`](https://github.com/binius-zk/binius64/blob/main/crates/circuits/src/sha512/mod.rs). It **computes and returns** the digest — no caller-supplied `digest`, no internal digest assertion — so the returned digest is a single-valued function of `(data, len_bytes)`; a free `len_bytes` can only select which message prefix is hashed, never an arbitrary digest. The residual "constrain `len_bytes`" obligation is now explicit in the `&ByteVec` signature, as with `sha512_varlen`/`keccak256_varlen`.

SHA-256's 32-bit schedule words are half a `ByteVec` word wide, so each data word is `swap_bytes_32`-split into two big-endian schedule words; every padded word (message / boundary-with-`0x80` / zero pad / 64-bit BE length) is a derived, constrained wire; the digest is `State::pack_4x64b` of the length-block state; and `len_bytes <= capacity` is enforced in-circuit.

## Struct removal + consumer migration

The `Sha256` struct and its `message_to_le_wires` / `digest_to_le_wires` / `populate_*` helpers are deleted; the little-endian handling collapses onto `ByteVec` + `crate::bytes` (no new free functions). Migrated consumers:

- **`rs256`** (RSA / PKCS#1) — `sha256_varlen` over the message `ByteVec`; the BE 4×64 digest layout is identical, so the PKCS#1 comparison is untouched.
- **`bitcoin/double_sha256`** — both hashes have compile-time-constant lengths, so this uses `sha256_fixed` ×2; `DoubleSha256` is kept as a thin shim with its original signatures, so `merkle_path`/`header_chain` need no changes.
- **`zklogin`** — the `zkaddr` and `nonce` hashes go through `sha256_varlen` (digests asserted against the expected wires; the LE preimage/digest wires now come straight from the message `ByteVec` or a `swap_bytes` repack).
- **`examples/sha256`** (variable branch), the **`end_to_end` tutorial**, and the **`prover` shift test**.

## Verification

- `binius-circuits` full lib suite (320 tests) green — incl. `rs256`, `bitcoin::{merkle_path,header_chain}`, and the `sha256_varlen` cross-check vs `sha2` over block-boundary lengths + a capacity-guard rejection test.
- `binius-prover` `shift` test proves + verifies; the `end_to_end` example proves + verifies.
- `cargo clippy --all-targets -- -D warnings` and nightly `cargo fmt --check` clean across `circuits`, `examples`, `prover`.
- Re-blessed `zklogin.snap` (+0.1% AND, at parity with the removed struct — see Performance) and `bitcoin_headers.snap` (`double_sha256` via `sha256_fixed` is leaner, −11% AND). All other snapshots unchanged.

## Performance

`sha256_varlen`'s block compression runs through the two-lane `sha256_compress_2x_seq` (as `sha256_fixed` and the removed struct did), so removing the struct is **not** an AND-cost regression: zklogin lands at 357,836 AND vs the struct's 357,449 (+0.1%). This is SHA-256-specific — its 32-bit words let two compressions share the two lanes of a 64-bit wire; `sha512_varlen`/`keccak256_varlen` are 64-bit-native and have no analogous two-lane path.


jimpo (commented):
Nice, thanks! LGTM. Is there a reason the PR is Draft?

pauldelucia:
@jimpo just needed to self-review one more time - good now!

jimpo (approved):

jimpo:
Awesome, thanks @pauldelucia !

PR #1818, opened 2026-07-16
BINIUS-272: replace QuadraticMleCheckProver with SharedMleCheckProver

tcoratger:
Closes [BINIUS-272](https://linear.app/jimpo/issue/BINIUS-272).

Deletes the bespoke `QuadraticMleCheckProver` and routes every caller through the store-backed `SharedMleCheckProver` + `QuadraticMleEvaluator`. Pure refactor — the two provers compute identical round polynomials, so no protocol, transcript, or soundness change.

### The problem

`QuadraticMleEvaluator` already exists and its own doc calls it the "store-backed successor" of `QuadraticMleCheckProver`, but the old prover was still the one everyone used.

Two provers meant two copies of the same degree-2 MLE-check logic, and the old one couldn't share a fold pass or column evaluations when several claims live at the same point — the whole reason the store path exists.

### The idea

Drive the composition through the shared `MleStore`:
- the store owns the columns and folds each one **once** per round, no matter how many claims read it;
- one `QuadraticMleEvaluator` per claim emits the same eq-factored round polynomial the old prover did.

For the common single-claim case, wrap that in a small factory so call sites stay one line.

### The change

A new single-claim factory, mirroring the existing `bivariate_product_prover`:

```
quadratic_mlecheck_prover([FieldBuffer; N], composition, infinity_composition, eval_point, eval_claim)
    -> SharedMleCheckProver<'static, ..>
```

At each call site the swap is mechanical:

```diff
- QuadraticMleCheckProver::new(cols, |[a,b,c]| a*b - c, |[a,b,_]| a*b, point, claim)
+ quadratic_mlecheck_prover(cols, |[a,b,c]| a*b - c, |[a,b,_]| a*b, point, claim)
```

`bivariate_product_mle::new` is reimplemented on top of the factory. Its prodcheck caller passed a `FieldBufferSplitMut` — a buffer viewed as two halves, folded in place with **no copy**. To keep that, the prodcheck layer now uses a dedicated `new_split_half`, backed by the store's `push_split_half`, so the largest witness layer is still never copied up front.

### Soundness

- The round polynomials are bit-identical to the old prover's (same Gruen32 interpolation, same `finish` ordering), so the transcript is unchanged.
- No new assumption; the verifier is untouched.

### Scope

- **new:** `quadratic_mlecheck_prover` factory + prove/verify tests (the evaluator module previously had none), and `bivariate_product_mle::new_split_half`.
- **changed:** spartan mulcheck, the AND-reduction prover, intmul (overflow `N=3` and the C-root product), the prodcheck layer, and both benches.
- **deleted:** `quadratic_mle.rs` (441 lines) and its `pub mod` declaration.
- **tests:** the transitional differential tests compared the store path against the old prover as a golden reference; with the reference deleted they are converted into self-contained prove/verify roundtrips against the real verifier.

### Verification

- `cargo check -p binius-ip-prover -p binius-prover -p binius-spartan-prover --all-targets` — clean
- `clippy` on the three crates + nightly `fmt` — clean
- tests pass: ip-prover (58), the AND-reduction transcript test, intmul prove/verify, spartan integration, zk-wrapper

Two unrelated crates (`binius-frontend`, `binius-m4-verifier`) fail to compile on this `main` checkout in files this PR does not touch and in crates that do not depend on the changed API — pre-existing breakage, not introduced here.


jimpo (approved):
Thanks!

PR #1819, opened 2026-07-16
Use the shared RoundState in the store-backed round evaluators

tcoratger:
Deduplicates the round-state bookkeeping in the two store-backed round evaluators onto the shared `RoundState` enum. Pure refactor — no behavior change.

### The problem

A sumcheck-style prover alternates, once per variable, between two phases:
- produce this round's polynomial(s),
- reduce them with the verifier challenge to the claim(s) for the next variable.

It holds exactly one of those at a time, and the `round_state` module exists to carry that in a single enum and validate the mandated `execute → fold` order in one place. Every prover in the crate uses it — except the two store-backed round evaluators, which each still hand-rolled their own copy:

- `QuadraticMleEvaluator` had a private `RoundCoeffsOrEval`,
- `BivariateProductEvaluator` had a private `RoundCoeffsOrSum`,

each duplicating the same two variants and re-writing its own out-of-order panics.

### The idea

Both bespoke enums are exactly `RoundState<Coeffs, Claim>` with `Coeffs = RoundCoeffs<F>` and the claim being a single field element. So drop them and reuse the shared type:

```
last_coeffs_or_eval: RoundCoeffsOrEval<F>   ->   RoundState<RoundCoeffs<F>, F>
last_coeffs_or_sum:  RoundCoeffsOrSum<F>    ->   RoundState<RoundCoeffs<F>, F>
```

The wrong-phase checks — previously an inline `match … => panic!` and a `let-else … panic!` in each file — now go through `RoundState`'s `claim()` / `coeffs()` accessors, the single place that owns those panics. The reads that legitimately handle both states (the current round claim) stay as direct matches on the shared variants.

### Scope

- **changed:** `quadratic_mle_evaluator.rs`, `bivariate_product_evaluator.rs` — field type + accessor use.
- **deleted:** the two private enums.
- no public API change; the evaluators behave identically.

### Note on panic messages

The out-of-order panics now read as the shared "execute called out of order" rather than the old per-evaluator "interpolate called out of order". This stays accurate: an evaluator's interpolate runs inside the driving prover's `execute`, so the protocol-level verb is the right one, and centralizing the message is the point of the shared type.

### Verification

- `cargo build -p binius-ip-prover` — clean
- `cargo test -p binius-ip-prover` — 59 passed. The crate still carries the pre-existing `QuadraticMleCheckProver` and its differential lockstep tests, which pass unchanged and independently confirm this refactor is byte-for-byte behavior-preserving.
- `clippy -p binius-ip-prover --all-targets` + nightly `fmt` — clean


jimpo (approved):

jimpo:
Please resolve conflicts

tcoratger:
> Please resolve conflicts

done

PR #1825, opened 2026-07-16
BINIUS-297: BinMul reduction (prover + verifier)

benji-potamus:
Linear: [BINIUS-297](https://linear.app/jimpo/issue/BINIUS-297/binius64-binmul-protocol-verifier-and-prover) · Spec: [binius-zk/binius.xyz#119](https://github.com/binius-zk/binius.xyz/tree/ghash) (§ The BinMul Reduction)

Adds the **BinMul reduction** — the IOP sub-protocol that checks BinMul (GHASH-field) constraints — as a standalone `binmul` module in `binius-prover` and `binius-verifier`, parallel to `intmul`/`bitand`. It proves `Ã(x)·B̃(x) = C̃(x)` for every `x` on the hypercube, where Ã, B̃, C̃ are the F_2^128-valued multilinears whose hypercube values are the packed operands `⟨⟨lo, hi⟩⟩` (lo word = coefficients of `X⁰..X⁶³`, hi word = `X⁶⁴..X¹²⁷`).

### Protocol (per spec)
An ordinary product-check **zerocheck** — the simplest of the reductions (no univariate skip, no GKR):
1. Verifier samples `r_z ∈ K^ℓ`.
2. An `ℓ`-round degree-3 sumcheck on `0 = Σ_x eq(r_z,x)·(Ã(x)·B̃(x) − C̃(x))` → challenge `r_x`, target `eval`.
3. Prover sends the 6×64 per-bit evals at `r_x`; the verifier recombines `α_A = Σ_i basis(i)·ã_lo(r_x,i) + basis(64+i)·ã_hi(r_x,i)` (and `α_B, α_C`), and checks `eval == (α_A·α_B − α_C)·eq(r_z, r_x)`.
4. Output: the 6×64 per-bit eval claims at `r_x` (same shape as IntMul's output, destined for the shift reduction).

### Design
The soundness-critical sumcheck is **reused, tested infrastructure** — the prover composes `MleToSumCheckDecorator::new(quadratic_mlecheck_prover([A,B,C], |[a,b,c]| a*b−c, |[a,b,_]| a*b, r_z, ZERO))` driven by `batch_prove`; the verifier uses `batch_verify(n_vars, 3, …)`. This mirrors intmul's phase-5 product-zerocheck exactly (including the verifier-only challenge reversal). The prover builds the packed tables via `F::from((lo|hi<<64) as u128)`, and the verifier recombines via `basis(i)`; these agree because `from(1<<i) == basis(i)` in the GHASH monomial basis.

[...]

PR #1827, opened 2026-07-17
BINIUS-300: handle BMUL constraints in the IOP prover and verifier

benji-potamus:
Linear: [BINIUS-300](https://linear.app/jimpo/issue/BINIUS-300/proveverify-bitmul-constraints)

Wires the (already-merged) standalone **BinMul reduction** (#1825) into the `IOPProver`/`IOPVerifier` so they consume `bmul_constraints` from the `ConstraintSystem`, by adding **BinMul as a third operation of arity 6** through the shift reduction — mirroring exactly how IntMul is integrated. Completes the GHASH mul constraints project end-to-end (frontend #1821/#1822/#1823/#1824 + reduction #1825 + this integration).

### How it works (mirrors IntMul)
1. **Reduction phase.** A BinMul reduction phase runs immediately after IntMul and before BitAnd (same position on both sides, so the transcript stays in sync). Skipped entirely — no transcript messages — when there are no BMUL constraints, guarded identically on prover and verifier.
2. **Claim collapse.** BinMul shares the univariate challenge `r_zhat_prime` that BitAnd draws. Each of the six per-bit operand columns (`a_lo, a_hi, b_lo, b_hi, c_lo, c_hi`) is collapsed to a single oblong eval via `inner_product(evals, lagrange_evals(subspace, r_zhat_prime))` — the same collapse IntMul uses for its four columns.
3. **Shift reduction.** The collapsed 6-operand `OperatorData` is threaded through the shift reduction as `Operation::BinaryMul` (`BMUL_ARITY = 6`), batched by its own `bmul_lambda`. The verifier's monster evaluation was already arity-generic, so it needed only the extra operand-tuple and flat-input section; the prover side gained the `BinaryMul` variant in the `Operation` enum, key-collection operand getters, per-operation scalar table, and phase 1/2 threading.

### Scope note
The issue is estimated 1 point, but full end-to-end integration necessarily touches the **shift reduction** (the soundness core) on both sides — that's where operand eval-claims are connected to the committed witness, so BMUL can't bypass it. I implemented it as a third hard-coded operation mirroring IntMul rather than generalizing the reduction to N operations, per the *avoid-premature-generalization* principle — happy to generalize if you'd prefer now that there are three near-identical operations. The change also fans out a zero/skipped BinMul claim through the m4 prover/verifier and the shift benches (none have BMUL constraints) to keep `clippy --all` green.

### Operand-order consistency
`[a_lo, a_hi, b_lo, b_hi, c_lo, c_hi]` is used identically in the key-collection getters, the monster scalar table, the collapse, `build_bmul_witness`/`BinMulWitness`, and the verifier's `multiunzip` — so operand `i` is weighted `λ^{i+1}` on both sides and each eval-claim is checked against the correct committed column.

### Validation
- New integration test `test_prove_verify_bmul_seventh_power`: a circuit computes `x⁷` of a random GHASH element via four `bmul` gates and constrains it to a public output; the expected value is computed independently in `BinaryField128bGhash`. Runs both the standard and ZK prove/verify paths — a full end-to-end soundness check (a misaligned operand or desynced transcript fails `assert_zero`).
- `cargo test -p binius-prover --test prove_verify` — 9 passed (incl. the new test and `test_prove_verify_zero_imul_constraints`, which also exercises the BMUL skip path).
- All shift + binmul unit tests (prover, verifier, m4) green.
- `cargo clippy --all --tests --benches --examples -- -D warnings` and `cargo +nightly-2026-01-01 fmt --check` — clean.
- `Operation::BinaryMul = 2u8` is an additive serialization change (existing discriminants unchanged); the serialized-KeyCollection ZK round-trip test passes, so no version bump was needed.

### DIFF
--- a/crates/verifier/src/protocols/shift/verify.rs
+++ b/crates/verifier/src/protocols/shift/verify.rs
@@ -4,7 +4,7 @@
 use std::{array, iter};
 
 use binius_core::{
-	constraint_system::{AndConstraint, ConstraintSystem, ImulConstraint, Operand},
+	constraint_system::{AndConstraint, BmulConstraint, ConstraintSystem, ImulConstraint, Operand},
 	word::Word,
 };
 use binius_field::{BinaryField, field::FieldOps, util::FieldFn};
@@ -24,7 +24,7 @@ use getset::Getters;
 use itertools::Itertools;
 
 use super::{
-	BITAND_ARITY, INTMUL_ARITY, SHIFT_VARIANT_COUNT, error::Error, evaluate_h_op,
+	BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, SHIFT_VARIANT_COUNT, error::Error, evaluate_h_op,
 	evaluate_monster_multilinear_for_operation,
 	monster::evaluate_monster_multilinear_for_operation_native,
 };
@@ -104,6 +104,8 @@ pub struct VerifyOutput<F> {
 	bitand_lambda: F,
 	/// Random coefficient for batching IMUL constraint evaluations.
 	intmul_lambda: F,
+	/// Random coefficient for batching BMUL constraint evaluations.
+	binmul_lambda: F,
 	/// Challenge point for the bit index variables (length `Word::LOG_BITS`).
 	pub r_j: Vec<F>,
 	/// Challenge point for the shift variables (length `Word::LOG_BITS`).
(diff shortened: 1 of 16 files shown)


crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -148,19 +150,20 @@ impl<F> VerifyOutput<F> {
...
+/// 1. **Sampling Phase**: Samples random lambda coefficients for batching bitand, intmul and bmul
 ///    evaluation claims across operands.
 /// 2. **First Sumcheck**: Verifies the batched evaluation claim over `Word::LOG_BITS * 2` variables
 /// 3. **Challenge Splitting**: Splits sumcheck challenges into `r_j` and `r_s` components
 /// 4. **Second Sumcheck**: Verifies the gamma claim over `log_word_count` variables
 /// 5. **Monster Multilinear Verification**: Checks that the claimed evaluations match expected
-///    monster multilinear evaluations for both AND constraints (bitand) and IMUL constraints
-///    (intmul)
+///    monster multilinear evaluations for AND constraints (bitand), IMUL constraints (intmul) and
+///    BMUL constraints (bmul)
```
jimpo: BMUL constraints (binmul). Use binmul throughout this file, bmul is just for the frontend crate as an abbreviation.

jimpo (changes requested):
This looks fine aside from naming. Use "binmul" for the name of the protocol and operation. bmul is an abbreviation used in the binius-core and binius-frontend crates. So its bitand, intmul, binmul, then in the constraint system, we have and_constraints, imul_constraints, and bmul_constraints.

benji-potamus:
Done — renamed the protocol/operation to **binmul** throughout the prover/verifier shift layer (`BINMUL_ARITY`, `Operation::BinMul`, `binmul_data`/`binmul_lambda`/`binmul_operator_data`/…, `BinMulCheckWitness`, `build_binmul_witness`), matching the `bitand`/`intmul`/`binmul` convention. The constraint-system entities keep the `bmul` abbreviation from binius-core/binius-frontend (`BmulConstraint`, `bmul_constraints`, `n_bmul_constraints`, and the frontend `circuit.bmul` gate), so it reads: `and_constraints`/`imul_constraints`/`bmul_constraints` on the CS side, `bitand`/`intmul`/`binmul` on the protocol side.

Pushed as `08c946ce`. Pure rename — `prove_verify` (9 incl. `test_prove_verify_binmul_seventh_power`), shift+binmul tests, `clippy --all`, and nightly fmt all still green.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Fixed — `binmul` is now used throughout this file (and the other protocol-layer files), including the prose: `bitand, intmul and binmul` in the sampling step and `BMUL constraints (binmul)` in the monster-verification step. Also renamed the leftover `bmul` local in `m4-verifier`. The `bmul` abbreviation now appears only for the frontend/constraint-system entities (`BmulConstraint`, `bmul_constraints`, `circuit.bmul`). Pushed `bfcdc85`.

jimpo (approved):

PR #1829, opened 2026-07-17
BINIUS-302: handle binmul constraints in the M4 prover and verifier

benji-potamus:
Linear: [BINIUS-302](https://linear.app/jimpo/issue/BINIUS-302/proveverify-binmul-constraints-in-m4)

The M4 analog of #1827 (BINIUS-300): makes `binius-m4-prover`/`binius-m4-verifier` actually **handle BinMul (GHASH-field mul) constraints** — run the binmul reduction, build a real operand claim, and thread it through the instance re-randomization and the shift — instead of the zero/skipped claim #1827 fanned through M4. Reuses the merged `binmul` reduction (`binius_prover`/`binius_verifier::protocols::binmul`) verbatim; mirrors the existing IntMul handling.

### How it works (mirrors IntMul)
1. **BinMul check phase.** Runs after IntMul and before BitAnd (same order on both sides), over `log_instances + log_n_binmul` rows, only when `cs.bmul_constraints` is non-empty. Per-bit operand evals are bound to the transcript before BitAnd draws the univariate challenge that collapses them. BinMul commits no oracle, so `oracle_specs` is unchanged.
2. **Re-randomization → shared `r_rho`.** M4 unifies each operation's operand claims onto one shared instance point via a batched multilinear-eval sumcheck. This was hardcoded for BitAnd + optional IntMul; it now takes BitAnd (always) + **optional IntMul + optional BinMul**. The re-randomization runs whenever *either* multiplication op is present; operands are pushed/read in the order `[BitAnd | IntMul? | BinMul?]`, and the reduced evals split back into contiguous per-op segments in that same order. Absent ops reduce to a zero claim at an empty point (the pre-existing direct-path behavior).
3. **Shift.** The real `binmul_data` now flows into `shift::{prove,verify,check_eval}` (M4's shift.rs was already binmul-aware from #1827).

`Operation::from_binmul` / `OperationClaim::from_binmul` collapse the six per-bit eval columns (`[a_lo, a_hi, b_lo, b_hi, c_lo, c_hi]`) at the univariate challenge and split `eval_point` at `log_instances` into `(r_rho, r_x)` — exactly like `from_intmul`. `build_binmul_witness` projects the six operands constraint-major via the existing arity-generic `build_operation_witness`.

### Naming
Protocol layer uses `binmul` (`build_binmul_witness`, `prove_binmul`, `verify_binmul_reduction`, `binmul_data`, `from_binmul`); the `bmul` abbreviation stays for the constraint-system/frontend (`cs.bmul_constraints`, `circuit.bmul`).

### Scope note
Like #1827, "reuse the protocol logic" still touches the M4 **re-randomization** (the soundness core that connects operand claims to the folded witness), so this is a real cross-cutting change there, not just a witness builder. I extended it with concrete optional fields mirroring IntMul rather than generalizing to N operations (per *avoid premature generalization*) — happy to generalize if you'd prefer now that there are three.

### Validation
- New tests: `protocol_round_trips_with_binmul` (one `bmul` squaring per instance over 2⁶ instances, both product words committed) and `protocol_round_trips_with_and_intmul_binmul_and_wide_packing` (AND + IMUL + BMUL together at different instance points, `PackedBinaryGhash2x128b`) — full end-to-end prove→verify, which fails if the prover/verifier push order or segment split desyncs. Plus a `build_binmul_witness` column-layout unit test.
- `cargo test -p binius-m4-prover -p binius-m4-verifier` — green (incl. all existing `protocol_round_trips*`).
- `cargo clippy --all --tests --benches --examples -- -D warnings` and `cargo +nightly-2026-01-01 fmt --check` — clean.

### DIFF
--- a/crates/m4-verifier/src/verify.rs
+++ b/crates/m4-verifier/src/verify.rs
@@ -27,9 +27,10 @@ use binius_verifier::{
 	Error,
 	config::{B1, B128},
 	protocols::{
+		binmul::{BinMulOutput, verify as verify_binmul_reduction},
 		bitand::AndCheckOutput,
 		intmul::{IntMulOutput, verify as verify_intmul_reduction},
-		shift::{self, BITAND_ARITY, INTMUL_ARITY, OperatorData},
+		shift::{self, BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, OperatorData},
 	},
 	ring_switch::{self, RingSwitchVerifyOutput},
 	verify_bitand_reduction,
@@ -161,6 +162,20 @@ impl IOPVerifier {
 			None
 		};
 
+		// SOUNDNESS: the BinMul check verifies after the IntMul check and before the BitAnd check,
+		// mirroring the prover. Its per-bit operand evaluations are read from the transcript here,
+		// before BitAnd draws the univariate challenge that collapses them. Do not reorder these,
+		// and keep the same order in `IOPProver::prove`.
+		//
+		// The BinMul columns span every instance's constraints.
+		// So the check runs over `log_instances + log_n_binmul` row variables.
+		let binmul_output = if cs.n_bmul_constraints() > 0 {
+			let log_n_binmul = checked_log_2(cs.n_bmul_constraints());
+			Some(verify_binmul_reduction::<B128, _>(log_instances + log_n_binmul, channel)?)
+		} else {
+			None
+		};
+
 		// AND-check over all `K * n_and` rows.
 		let log_n_and = checked_log_2(cs.and_constraints.len());
 		let AndCheckOutput {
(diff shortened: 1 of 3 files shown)


crates/m4-verifier/src/verify.rs
```diff
@@ -161,6 +162,20 @@ impl IOPVerifier {
...
 		};
 
+		// SOUNDNESS: the BinMul check verifies after the IntMul check and before the BitAnd check,
+		// mirroring the prover. Its per-bit operand evaluations are read from the transcript here,
+		// before BitAnd draws the univariate challenge that collapses them. Do not reorder these,
+		// and keep the same order in `IOPProver::prove`.
+		//
+		// The BinMul columns span every instance's constraints.
+		// So the check runs over `log_instances + log_n_binmul` row variables.
+		let binmul_output = (cs.n_bmul_constraints() > 0)
```
jimpo: This is hard to read. Use a simple if statement, like the intmul logic above.

jimpo (changes requested):

crates/m4-verifier/src/verify.rs
benji-potamus: Done — rewrote it as a plain `if` matching the IntMul phase above:

```rust
let binmul_output = if cs.n_bmul_constraints() > 0 {
    let log_n_binmul = checked_log_2(cs.n_bmul_constraints());
    Some(verify_binmul_reduction::<B128, _>(log_instances + log_n_binmul, channel)?)
} else {
    None
};
```

Pushed `1420652` (also rebased onto latest main / #1826). m4 tests + clippy --all + nightly fmt still green.

jimpo (approved):

PR #1834, opened 2026-07-17
BINIUS-305: move round claim/interpolation state from RoundEvaluator up into SharedSumcheckProver

benji-potamus:
Closes BINIUS-305.

## What

Removes `RoundEvaluator::fold` and the per-evaluator `last_coeffs_or_sum` state. The claim ↔ round-coeffs state machine now lives once in `SharedSumcheckProver`, as a `Vec<RoundState<RoundCoeffs, F>>` parallel to the evaluators. Evaluators become **stateless across rounds** (`interpolate` and the recovery hook take `&self`), which lines up the interface for BINIUS-306 (splitting `RoundEvaluator` into separate sumcheck/mlecheck traits, where `interpolate` will additionally receive `alpha`).

The prover now owns the whole state transition:
- `execute` reads each evaluator's current round claim from `round_states` and hands it to `RoundEvaluator::interpolate(store, accum, claim)`, then stores the returned coefficients.
- `fold` reduces each stored round polynomial against the challenge (`coeffs.evaluate(challenge)`) back to a claim — uniformly, no evaluator involvement (the old per-evaluator `fold` was always exactly this reduction).
- `round_claim` returns the held claim directly, or, once coefficients have replaced it, recovers it via the new `RoundEvaluator::claim_from_coeffs` (which each evaluator implements as its own emission inverse: `sum_over_endpoints` for plain sumcheck, `lerp_over_endpoints(alpha)` for prime MLE-check polynomials).

Initial claims now flow to the prover constructors (`SharedSumcheckProver::new`, `SharedMleCheckProver::new`, `add_evaluator`) instead of into each evaluator, and the `frac_add_mle::evaluators` helper no longer takes claims. Also drops the now-unused `P` type parameter from `BivariateProductEvaluator` and `QuadraticMleEvaluator`.

## Design note worth a look

For the wrapped `MleToSumCheckEvaluator`, the prover stores the **sumcheck** claim (= inner prime claim `m` scaled by the accumulated equality prefix), because that's what the emitted regular-sumcheck polynomial reduces to under `fold`. Its `interpolate` therefore recovers the inner prime claim by dividing the prefix back out:

```rust
let inner_claim = claim * eq_prefix.invert_or_zero();
```

The invariant `sumcheck_claim = eq_prefix * m` holds every round (starts at `eq_prefix = 1`), so this is exact; `invert_or_zero` mirrors the existing interpolation routines in `round_evals.rs`. The old design avoided this division by keeping the prime state *inside* the inner evaluator. The cost here is one field inversion per round per wrapped evaluator (a handful in the logUp* final layer; ~one per M4 operand) — negligible against the accumulation pass. If you'd rather keep it division-free, the alternative is to store the prime state in the prover and add a separate emission step to the trait; I went with the simpler version that matches the issue's "keep one RoundState per evaluator + pass the round claim to interpolate", but happy to switch.

## Testing

- `cargo test` green for `binius-ip-prover` (58), `binius-m4-prover` (40), `binius-iop-prover` (35 + integration) — covers the plain, MLE-check, wrapped-final-layer, and round-claim-recovery paths.
- clippy (`--tests --benches`) and nightly `fmt --check` clean; `cargo doc -D warnings` clean.

### DIFF
--- a/crates/ip-prover/src/sumcheck/round_evaluator.rs
+++ b/crates/ip-prover/src/sumcheck/round_evaluator.rs
@@ -3,9 +3,10 @@
 //! Round evaluators over a shared [`MleStore`] and the provers that drive them.
 //!
 //! A [`RoundEvaluator`] holds the per-round-polynomial logic for one composite claim over store
-//! columns. Evaluators hold [`ColId`]s and receive column data by argument; they never fold and
-//! hold no column references — the store folds (see the [`mle_store`](super::mle_store) module
-//! documentation).
+//! columns. Evaluators hold [`ColId`]s and receive column data by argument; they hold no mutable
+//! per-round state — they neither fold nor track the round claim. The driving prover owns the
+//! `RoundState` machine (the claim ↔ coeffs alternation) for each evaluator, and the store folds
+//! the columns (see the [`mle_store`](super::mle_store) module documentation).
 //!
 //! [`SharedSumcheckProver`] adapts a store plus a list of evaluators — one per claim — to the
 //! existing [`SumcheckProver`] interface, and [`SharedMleCheckProver`] to the [`MleCheckProver`]
@@ -22,6 +23,7 @@ use super::{
 	MleToSumCheckEvaluator,
 	common::{MleCheckProver, SumcheckProver},
 	mle_store::{ColId, EvaluationChunk, MleStore},
+	round_state::RoundState,
 };
 
 /// Per-round-polynomial logic for one composite claim over store columns.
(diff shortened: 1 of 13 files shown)


crates/ip-prover/src/sumcheck/round_evaluator.rs
```diff
@@ -113,11 +130,16 @@ where
 	P: PackedField<Scalar = F>,
 	Evaluator: RoundEvaluator<F, P>,
 {
-	/// Creates a prover from a store and the evaluators reading its columns, one per claim.
-	pub const fn new(store: MleStore<'a, P>, evaluators: Vec<Evaluator>) -> Self {
+	/// Creates a prover from a store, the evaluators reading its columns, and the initial claim of
+	/// each evaluator — all three lists in the same order, one entry per claim.
+	pub fn new(store: MleStore<'a, P>, evaluators: Vec<Evaluator>, claims: Vec<F>) -> Self {
```
jimpo, line 135: I'd prefer `claims_with_evaluators: impl IntoIterator<Item=(F, Evaluator)>` The impl IntoIterator is so that it can be called with an array instead of a `vec!`. We can unzip to separate.

jimpo (approved):
Comment to be handled as a follow-up.

crates/ip-prover/src/sumcheck/round_evaluator.rs
benji-potamus: Sounds good — will do this as a follow-up PR (`new` taking `impl IntoIterator<Item = (F, Evaluator)>`, unzipping internally so callers can pass an array), stacked under BINIUS-306.

PR #1835, opened 2026-07-17
Unroll the Method-of-Four-Russians word fold 4x and add an into-buffer variant

benji-potamus:
Resolves [BINIUS-307](https://linear.app/jimpo/issue/BINIUS-307) (sub-issue of [BINIUS-303](https://linear.app/jimpo/issue/BINIUS-303)).

The per-word MoFR fold loop is load-latency-bound on aarch64: disassembly shows the within-word code is already optimal (fully inlined transform, 8 single-instruction table loads, `eor3` XOR trees, no bounds checks), but only one word's 8 independent loads are ever in flight, and LLVM declines to unroll the body. This PR:

- Adds a `P::WIDTH == 1` fast path to `fold_words_with_transform` that processes 4 words per iteration, exposing 32 independent table loads per iteration to the out-of-order scheduler (the compiler then also emits paired `stp` stores).
- Adds `fold_words_with_transform_into`, which folds into a caller-provided `FieldBuffer` with no allocation. Wiring buffer reuse into `fold_and_send_reduced_prover` is follow-up ([BINIUS-308](https://linear.app/jimpo/issue/BINIUS-308)) — allocation/page faults are ~28ms of the fold on Linux.
- Adds a `univariate fold prealloc` bench variant exercising the `_into` path, plus remainder/equivalence tests for the unrolled paths.

Measured single-threaded at 2^22 words × 3 columns on Neoverse V3 (all criterion medians): allocating fold **86.5ms → 75.2ms**; fold into preallocated buffers **58.3ms → 35.8ms**. The combined effect matches flock's per-column fold throughput (context in BINIUS-303).

The `fold_words_scalar_fill` variant writes through `spare_capacity_mut` + `set_len` to keep the allocating path's existing no-double-touch behavior; the alternative (`vec![P::default(); n]` + slice fold) would re-add a full zero-fill pass over the 64MB output.

### DIFF
--- a/crates/prover/benches/and_reduction.rs
+++ b/crates/prover/benches/and_reduction.rs
@@ -11,7 +11,7 @@ use binius_field::{
 };
 use binius_ip_prover::sumcheck::{common::SumcheckProver, quadratic_mlecheck_prover};
 use binius_math::{
-	BinarySubspace,
+	BinarySubspace, FieldBuffer,
 	univariate::{extrapolate_over_subspace, lagrange_evals_scalars},
 };
 use binius_prover::{
@@ -19,7 +19,7 @@ use binius_prover::{
 	and_reduction::{
 		NTTLookup, sumcheck_round_messages::univariate_round_message_extension_domain,
 	},
-	fold_word::fold_words_with_transform,
+	fold_word::{fold_words_with_transform, fold_words_with_transform_into},
 };
 use binius_verifier::{
 	config::{B128, PROVER_SMALL_FIELD_ZEROCHECK_CHALLENGES},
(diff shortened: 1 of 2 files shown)


jimpo:
My benchmarks show this to be slower than https://github.com/binius-zk/binius64/pull/1845

benji-potamus:
Confirmed on my Graviton box too: at 2^22×3 columns the #1845 `BitAxisFolder`/`WordBytewiseLookupTransformation` fold measures **69.3ms** vs this PR's **75.2ms** (old main was 86.5ms) — the u64-underlier-specialized transform beats source-level unrolling of the generic path. I also tested layering the 4× compute-first unroll on top of #1845's loop: **69.6ms**, i.e. no additional gain — the specialized transform already keeps the load pipeline busy. Closing in favor of #1845.

One piece from the BINIUS-303 investigation that #1845 doesn't cover: the fold still allocates three fresh 64MB buffers per prove, which costs ~28ms in page faults on Linux (THP=madvise). That's [BINIUS-308](https://linear.app/jimpo/issue/BINIUS-308) — a fold-into-caller-buffer variant on `BitAxisFolder` plus buffer reuse in `fold_and_send_reduced_prover`.

PR #1838, opened 2026-07-17
Remove the frontend util module, relocating its helpers

tcoratger:
Deletes `binius_frontend::util` by moving each of its four functions onto the type or into the crate it belongs to. Pure relocation — no behavior change.

### The problem

`util.rs` was a grab-bag. Each function operates on a specific type or domain, but lived as a free function in a shared `util` bucket, so callers reached for `binius_frontend::util::…` instead of the obvious owner. It also forced `binius-frontend` to depend on `num-bigint` for a helper the crate itself never used.

### The change

Each function goes to its natural home:

| was `util::…` | now |
|---|---|
| `pack_bytes_into_wires_le(w, wires, bytes)` | `WitnessFiller::pack_bytes_le(&mut self, wires, bytes)` |
| `all_true(b, bools)` | `CircuitBuilder::all_true(&self, bools)` |
| `byteswap(b, word)` | `CircuitBuilder::byteswap(&self, word)` |
| `num_biguint_from_u64_limbs(...)` | `binius_circuits::bignum::num_biguint_from_u64_limbs` (in `biguint.rs`) |

Call sites become method calls, e.g.:

```
- pack_bytes_into_wires_le(&mut filler, &wires, bytes);
+ filler.pack_bytes_le(&wires, bytes);

- all_true(b, conditions)
+ b.all_true(conditions)
```

The `num_biguint_from_u64_limbs` helper lands in `bignum/biguint.rs`, next to the other host `<->` `num_bigint::BigUint` bridges (`new_constant`, `populate_limbs`), and is re-exported from `bignum` so call-site imports stay a one-line path change.

### Two cleanups that fall out

- **`num-bigint` drops from `binius-frontend`.** That helper was the only thing pulling it in; it belongs in `binius-circuits`, which already depends on `num-bigint`.
- **The helper stops hand-rolling the conversion.** It split each `u64` into two `u32` digits by hand. It now calls `num_bigint::BigUint::from_bytes_le` — the same native constructor already used in `rs256.rs` — since little-endian limbs, each itself little-endian, concatenate to the value's little-endian byte string. Semantically identical, and it drops the now-unneeded `ExactSizeIterator` bound.

### Scope

- **removed:** `crates/frontend/src/util.rs` and its `pub mod util;`, plus the `num-bigint` dependency from `binius-frontend/Cargo.toml`
- **added:** `WitnessFiller::pack_bytes_le`, `CircuitBuilder::all_true` / `byteswap`, `bignum::num_biguint_from_u64_limbs`
- **changed:** ~99 `pack_bytes` call sites plus a handful of `all_true`/`byteswap`/`num_biguint` sites and their imports, across `binius-circuits`, `binius-examples`, and `binius-prover` tests
- the `i8x16_all_true` wasm intrinsic is deliberately untouched (name collision only)

### Verification

- `cargo clippy --workspace --all-targets` — clean, zero warnings
- every moved body is a byte-for-byte copy bar the receiver, so behavior is unchanged


crates/frontend/src/compiler/mod.rs
```diff
@@ -1233,6 +1233,28 @@ impl CircuitBuilder {
 		self.shr(masked, shift)
 	}
 
+	/// Reverses the byte order of `word`.
+	///
+	/// Extracts each of the 8 bytes and reassembles them in the opposite order.
+	pub fn byteswap(&self, word: Wire) -> Wire {
```
jimpo: This is already redundant with the `binius_circuit::bytes::swap_bytes` method, which is more efficient. Use that instead.

crates/frontend/src/compiler/mod.rs
```diff
@@ -1233,6 +1233,28 @@ impl CircuitBuilder {
...
+		});
+		bytes
+			.reduce(|lhs, rhs| self.bxor(lhs, rhs))
+			.expect("Word::BITS > 0")
+	}
+
+	/// Returns a wire that is all-ones exactly when every wire in `booleans` is all-ones.
+	///
+	/// The fold starts from the all-ones constant, so an empty iterator yields all-ones.
+	pub fn all_true(&self, booleans: impl IntoIterator<Item = Wire>) -> Wire {
```
jimpo: I'd prefer this as a standalone "gadget" function. The `CircuitBuilder` interface is big enough. I don't mind having it in a `util` module, or somewhere else.

jimpo (changes requested):

crates/frontend/src/compiler/mod.rs
tcoratger: Fixed

crates/frontend/src/compiler/mod.rs
tcoratger: Fixed

jimpo (approved):

PR #1837, opened 2026-07-17
Unroll the quadratic mlecheck accumulate and fold loops 4x

benji-potamus:
Resolves [BINIUS-311](https://linear.app/jimpo/issue/BINIUS-311) (sub-issue of [BINIUS-303](https://linear.app/jimpo/issue/BINIUS-303)).

On aarch64 `OptimalPackedB128` is one lane wide, so the sumcheck tail's carryless multiplies run scalar; the per-element accumulate and deferred-fold loops each serialize on one dependent multiply chain while the core has two PMULL pipes.

- `QuadraticMleEvaluator::accumulate`: accumulate into four independent `(y_1, y_inf)` wide-accumulator pairs (XOR-merged after the loop) so several multiply chains stay in flight.
- `PreFoldColumnChunk::fold`: process four independent `extrapolate_line_packed` computations per iteration.

Measured single-threaded on the `remaining zerocheck` bench at 2^22 on Neoverse V3: **62.2ms → 58.8ms** (−5.5%). A manual-index variant measured 57.7ms; the ~1ms was traded back for the iterator-style loops here. Also tried in the BINIUS-303 investigation and deliberately *not* included: a 2-at-once GHASH `mul_pair` primitive in the fold — a further −0.8ms did not justify ~195 lines of field-crate surface. The tail is memory-traffic-bound (~670MB streamed per prove at this size), which caps what ILP restructuring can buy; the traffic-side follow-up is [BINIUS-312](https://linear.app/jimpo/issue/BINIUS-312).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1840, opened 2026-07-18
Unroll the quadratic mlecheck accumulate and fold loops 4x

benji-potamus:
Resolves [BINIUS-311](https://linear.app/jimpo/issue/BINIUS-311) (sub-issue of [BINIUS-303](https://linear.app/jimpo/issue/BINIUS-303)). Supersedes #1837, which was based on pre-BINIUS-305 main and conflicted with the `RoundEvaluator` state refactor (#1834); this is the same change rebased onto it (tests + clippy green post-rebase).

On aarch64 `OptimalPackedB128` is one lane wide, so the sumcheck tail's carryless multiplies run scalar; the per-element accumulate and deferred-fold loops each serialize on one dependent multiply chain while the core has two PMULL pipes.

- `QuadraticMleEvaluator::accumulate`: accumulate into four independent `(y_1, y_inf)` wide-accumulator pairs (XOR-merged after the loop) so several multiply chains stay in flight.
- `PreFoldColumnChunk::fold`: process four independent `extrapolate_line_packed` computations per iteration.

Measured single-threaded on the `remaining zerocheck` bench at 2^22 on Neoverse V3 (pre-rebase code, same loops): **62.2ms → 58.8ms** (−5.5%). A manual-index variant measured 57.7ms; the ~1ms was traded back for the iterator-style loops here. Also tried in the BINIUS-303 investigation and deliberately *not* included: a 2-at-once GHASH `mul_pair` primitive in the fold — a further −0.8ms did not justify ~195 lines of field-crate surface. The tail is memory-traffic-bound (~670MB streamed per prove at this size), which caps what ILP restructuring can buy; the traffic-side follow-up is [BINIUS-312](https://linear.app/jimpo/issue/BINIUS-312).

### DIFF
--- a/crates/ip-prover/src/sumcheck/mle_store.rs
+++ b/crates/ip-prover/src/sumcheck/mle_store.rs
@@ -19,7 +19,7 @@
 //! over the columns is a plain read. A deferred-fold variant that fuses the fold into the next
 //! round's read pass can replace the internals without changing this interface.
 
-use std::iter;
+use std::{array, iter};
 
 use binius_field::{Field, PackedField};
 use binius_math::{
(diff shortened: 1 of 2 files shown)


benji-potamus:
Note for review: main is being refactored quickly under this PR (#1841 constructor change, #1842 RoundEvaluator split both landed after the branch was cut), so the branch carries two small fix-up commits adapting to those. The "Files changed" diff may therefore show a few lines that actually belong to #1841/#1842 (constructor call, `accumulate`/`interpolate` signatures) until the branch is refreshed against a newer merge-base. The substance of this PR is unchanged: the 4× unrolled `accumulate` body (four independent accumulator pairs) and the quad-chunked deferred-fold loops in `mle_store.rs`. Squash-merging onto current main produces exactly the tree that passed tests + clippy locally.

benji-potamus:
Closing in favor of #1844. The base kept moving under this branch (#1841, then #1842 rewrote the same lines of `quadratic_mle_evaluator.rs`), so fix-up commits on the stale base could not make it mergeable again. #1844 is cut from current main with the identical change adapted to the split `MleCheckRoundEvaluator` trait, tests + clippy green.

PR #1843, opened 2026-07-18
perf(circuits): drop dead work in the float64 sticky folds and slice compare

wstran:
## What

Three float64 gadgets and the constant slice compare spend AND constraints on patterns with cheaper linear equivalents. This is the reduction category of #1686, #1804, and #1839.

Root cause, per pattern:

- The `float64` sticky folds clear bit 0 and then OR it back in: `(v & ~1) | (v & 1) | (s & 1)` is `v | (s & 1)`.
- `fp64_add` extracts bit 63 and compares it against zero, although every consumer of the result reads only the MSB.
- `assert_const_slice_eq` masks both compare operands, although the mask distributes over the XOR: `(a & m) == (e & m)` is `((a ^ e) & m) == 0`.

Fix: the sticky folds drop the clear-then-restore, the bit-63 compare becomes `bnot` whose MSB already carries the result, and the slice compare XORs first and masks once. All three are unconditional bitwise identities.

## Numbers

Deterministic AND-constraint counts (`CircuitStat::collect`):

| circuit | baseline | this PR | delta |
|---|---|---|---|
| `float64_add` | 246 | 238 | -3.3% |
| `float64_mul` | 130 | 124 | -4.6% |
| `assert_const_slice_eq` | 3 per partial word | 2 per partial word | -1 |
| `hashsign` example | 297,991 | 297,984 | -7 (via the slice compares) |

Counted by instantiating each gadget on witness wires and collecting the stats:

```rust
let b = CircuitBuilder::new();
let x = b.add_witness();
let y = b.add_witness();
b.force_commit(float64_add(&b, x, y));
let n = CircuitStat::collect(&b.build()).n_and_constraints;
```

The `hashsign` snapshot moves by the seven slice-compare words on its path and is re-blessed; the other twelve snapshots are unchanged.

## Correctness

- All three rewrites are unconditional bitwise identities, so they agree with the old forms on every witness value including adversarial ones.
- 323 `binius-circuits` tests pass, including the `float64` suites built around native `f64` edge cases (NaN, subnormals, `MIN_POSITIVE`). `clippy --all --all-features --tests --benches --examples -D warnings` and nightly `fmt --check` are clean.

## Scope

The masked-swap rewrites this PR originally carried are dropped. `merkle_path` and `fp64_order_by_exp` / `fp64_sub_path` are back on `select`, so nothing here constrains how select gates are lowered later. Their win was `float64_add` 238 -> 234 and 4 AND per Merkle level, and it can come back through the select gate itself rather than through the circuits.

### DIFF
--- a/crates/circuits/src/concat.rs
+++ b/crates/circuits/src/concat.rs
@@ -179,7 +179,10 @@ fn assert_const_slice_eq(
 		let e = input[k];
 		if k + 1 == n_words && final_bytes != 0 {
 			let mask = b.add_constant_64((1u64 << (final_bytes * 8)) - 1);
-			b.assert_eq(format!("{name}[{k}]"), b.band(a, mask), b.band(e, mask));
+			// (a & mask) == (e & mask) is ((a ^ e) & mask) == 0: the XOR is free and
+			// drops one of the two masking ANDs.
+			let diff = b.band(b.bxor(a, e), mask);
+			b.assert_eq(format!("{name}[{k}]"), diff, b.add_constant(Word::ZERO));
 		} else {
 			b.assert_eq(format!("{name}[{k}]"), a, e);
 		}
--- a/crates/circuits/src/float64/add.rs
+++ b/crates/circuits/src/float64/add.rs
@@ -72,8 +72,10 @@ pub fn float64_add(builder: &CircuitBuilder, a: Wire, b: Wire) -> Wire {
 	// Subnormal regime if:
 	//   - we were below 1 (exp_lt_1), OR
 	//   - base exponent == 1 and the integer bit (bit 63) is 0 (no hidden 1)
-	let msb01 = bit_lsb(builder, sig_round_base, 63); // 0/1
-	let msb_is_zero = builder.icmp_eq(msb01, zero(builder)); // MSB-bool
+	// `bnot` already carries the negated integer bit in the MSB, and every consumer
+	// below reads only the MSB, so extracting bit 63 and comparing it against zero
+	// would spend two AND constraints for nothing.
+	let msb_is_zero = builder.bnot(sig_round_base); // MSB-bool
 	let base_is_one = builder.icmp_eq(exp_round_base, one(builder));
 	let in_sub_regime = builder.bor(exp_lt_1, builder.band(base_is_one, msb_is_zero));
 	let stayed_sub_mask = builder.band(in_sub_regime, builder.bnot(mant_overflow_mask));
@@ -158,10 +160,9 @@ fn fp64_order_by_exp(
 fn fp64_align_with_sticky(b: &CircuitBuilder, sig_b: Wire, d: Wire) -> Wire {
 	let (mut v, sticky) = var_shr_with_sticky(b, sig_b, d, true);
 
-	let one_const = one(b);
-	let keep = b.bnot(one_const); // ~1 = clear bit0
-	let bit0 = b.bor(b.band(v, one_const), b.band(sticky, one_const));
-	v = b.bor(b.band(v, keep), bit0);
+	// (v & ~1) | (v & 1) | (sticky & 1) == v | (sticky & 1): the fold only ever
+	// sets bit 0, so clearing it first is dead work.
+	v = b.bor(v, b.band(sticky, one(b)));
 	v
 }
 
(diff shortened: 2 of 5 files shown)


jimpo:
I'm conflicted about this one. `select` is a more expressive gate and avoids leaking implementation details about masking. I also potentially want to change how select gates are handled (either using the new bmul constraints or another custom constraint type). At the same time, I see the win here. @tcoratger isn't there some sort of CSE optimization happening in the compiler that should help with this?

@wstran A really nice issue you might want to pick up would be changing `select` gates to use `bmul` constraints (multiplication with either 0 or 1 as a GHASH element) and measuring the perf impact.

wstran:
Dropped the masked swaps in 78d4ad1. `merkle_path` and `fp64_order_by_exp` / `fp64_sub_path` are back on `select`, so nothing here constrains how select gates get lowered later. What is left is three bitwise identities that never involved masking: the sticky folds, the bit-63 compare, and the slice compare.

| circuit | baseline | this PR |
|---|---|---|
| `float64_add` | 246 | 238 |
| `float64_mul` | 130 | 124 |
| `hashsign` | 297,991 | 297,984 |

The dropped swaps were `float64_add` 238 -> 234 and 4 AND per Merkle level. That win can come back through the select gate itself.

On the CSE question: I implemented it to check. Collapsing a complementary select pair in `dedup_gates` (the second output is the first XOR `t ^ f`, so it becomes a free linear definition) fires 46 times across all thirteen snapshot examples: 36 in `hashsign`, 10 in `bitcoin_header_chain`, zero in the other eleven. It is also a net loss. `hashsign` goes 297,991 -> 298,099 (+108), because the extra linear level pushes other definitions past `MAX_DEPTH = 6` in `gate_fusion::commit_set` and the forced commits outweigh the AND it saves. `bitcoin_header_chain` gets the expected 20,221 -> 20,211. So CSE does not recover this pattern.

wstran:
Rebased onto main, clears the snapshot conflict from #1907. hashsign holds at 297,991 -> 297,984.

On select-to-bmul: taking that one. BMUL sits at 0 constraints across all thirteen snapshot examples, so selects would be its first consumer and the subprotocol gets instantiated where it currently costs nothing. Measuring per-constraint prover cost for BMUL against AND before I claim a direction either way.

wstran:
Numbers on select-to-bmul. I encoded it as one BMUL per select, with `srl(cond, 63)` as `a_lo` and empty high limbs, so the A operand is the field element 0 or 1 and the condition costs nothing extra:

```
(cond >> 63) * (t ^ f) = out ^ f
```

Padded constraint work, since the AND and BMUL columns round up to a power of two independently:

| circuit | AND before | AND after | BMUL after | padded before | padded after |
|---|---:|---:|---:|---:|---:|
| zklogin | 357,836 | 228,567 | 129,269 | 524,288 | 393,216 |
| ethsign | 160,106 | 136,232 | 23,874 | 262,144 | 294,912 |
| ec_msm | 158,253 | 134,621 | 23,632 | 262,144 | 294,912 |
| bitcoin_p2pkh | 113,587 | 100,921 | 12,666 | 131,072 | 147,456 |
| hashsign | 297,991 | 295,327 | 2,664 | 524,288 | 528,384 |

Prove time, medians of 9 runs interleaved between the two binaries (M3 Pro):

| circuit | main | select as bmul |
|---|---:|---:|
| zklogin | 115.0 ms | 105.0 ms (-8.7%) |
| ethsign | 107.0 ms | 108.0 ms (+0.9%) |

So it is a rebalancing trick rather than a uniform win. zklogin gains because pulling selects out drops AND from 2^19 to 2^18 while BMUL only needs 2^17. ethsign, ec_msm and bitcoin_p2pkh keep the same AND band and just pay for a new column, though the measured cost is milder than the padded counts suggest: BMUL runs about 37.5 ns per constraint at 2^20 against about 44 ns for AND at 2^19, so the new column is cheaper than an AND column of the same size.

One blocker worth flagging before anyone builds on this. `verify_constraints` in `binius-core` only checks the AND and IMUL constraints; there is no `verify_bmul_constraint`. It currently accepts a false BMUL outright:

```rust
// a_lo = 3, b_lo = 5, c_lo = 1, high limbs empty. The GHASH product is 15.
let cs = ConstraintSystem::new(constants, layout, vec![], vec![],
    vec![BmulConstraint([v(1), vec![], v(2), vec![], v(3), vec![]])]);
assert!(verify_constraints(&cs, &witness).is_ok());  // passes
```

The proving side is fine, `prove` runs the binmul subprotocol whenever the count is non-zero and the `prove_verify` suite passes with selects converted. But every circuit test that leans on `verify_constraints` for soundness coverage loses it the moment selects become BMUL constraints. That is how I noticed: `concat::tests::mutated_output_fails_constraints` stopped failing on the converted build.

`binius-core` depends only on `binius-utils` today, so closing that needs either a `binius-field` dependency or the GHASH reduction written out in `verify.rs`. Happy to send it either way, or leave it to you since it is a call about the crate boundary.

jimpo (approved):

PR #1842, opened 2026-07-18
BINIUS-306: separate RoundEvaluator into SumcheckRoundEvaluator and MleCheckRoundEvaluator

benji-potamus:
Closes BINIUS-306.

_(The follow-up #1841 it was stacked on has merged; this PR is now based on `main`.)_

## What

Splits the single `RoundEvaluator` trait into two:

- **`SumcheckRoundEvaluator`** (renamed from `RoundEvaluator`), driven by `SharedSumcheckProver`, emits regular sumcheck round polynomials. Implemented by `BivariateProductEvaluator` and `MleToSumCheckEvaluator`.
- **`MleCheckRoundEvaluator`** (new), driven by `SharedMleCheckProver`, emits the eq-factored prime polynomials. Implemented by `QuadraticMleEvaluator`.

Per the issue, the MLE-check trait differs only in how the shared evaluation point's eq indicator reaches it — since every claim of a `SharedMleCheckProver` shares one point, the **prover** now owns that point's eq tracker and passes the round data in, so the evaluators store no tracker id:

- `MleCheckRoundEvaluator::accumulate(&self, chunk, eq_ind: FieldSlice, accum)` — the eq chunk is a separate argument (looked up once by the prover).
- `interpolate(&self, store, accum, claim, alpha)` — the round coordinate `alpha` is passed in.

`SharedMleCheckProver` is now a standalone prover (holds the store, evaluators, round states, and the one shared `eq_tracker`) rather than wrapping `SharedSumcheckProver`; its `execute` looks up the eq chunk / alpha and hands them to the evaluators. `into_shared_sumcheck` wraps each `MleCheckRoundEvaluator` in `MleToSumCheckEvaluator` (which now holds the eq tracker itself, since a plain sumcheck prover knows nothing of eq indicators) and hands over the store, round states, and tracker. `QuadraticMleEvaluator::new` and `frac_add_mle::evaluators` drop their `eq_tracker` argument; the driving prover registers it.

Now that the traits are split by protocol, round-claim recovery is uniform per prover, so there's no `claim_from_coeffs` on either evaluator trait — `SharedSumcheckProver::round_claim` recovers it as `sum_over_endpoints()` and `SharedMleCheckProver::round_claim` as `lerp_over_endpoints(alpha)`.

## Testing
[...]

PR #1844, opened 2026-07-18
Unroll the quadratic mlecheck accumulate and fold loops 4x

benji-potamus:
Resolves [BINIUS-311](https://linear.app/jimpo/issue/BINIUS-311) (sub-issue of [BINIUS-303](https://linear.app/jimpo/issue/BINIUS-303)). Supersedes #1840/#1837: the sumcheck refactors (#1834, #1841, #1842) each rewrote `quadratic_mle_evaluator.rs` under the earlier branches, leaving them unmergeable from stale bases; this branch is cut from current main (post-#1842) with the change rebased onto the split `MleCheckRoundEvaluator` trait. ip-prover tests (58) + clippy green locally on exactly this tree.

On aarch64 `OptimalPackedB128` is one lane wide, so the sumcheck tail's carryless multiplies run scalar; the per-element accumulate and deferred-fold loops each serialize on one dependent multiply chain while the core has two PMULL pipes.

- `QuadraticMleEvaluator::accumulate`: accumulate into four independent `(y_1, y_inf)` wide-accumulator pairs (XOR-merged after the loop) so several multiply chains stay in flight.
- `PreFoldColumnChunk::fold`: process four independent `extrapolate_line_packed` computations per iteration.

Measured single-threaded on the `remaining zerocheck` bench at 2^22 on Neoverse V3 (pre-refactor baseline, same loops): **62.2ms → 58.8ms** (−5.5%). A manual-index variant measured 57.7ms; the ~1ms was traded back for iterator-style loops. Deliberately *not* included: a 2-at-once GHASH `mul_pair` primitive (−0.8ms more for ~195 lines of field-crate surface). The tail is memory-traffic-bound at this size, which caps what ILP restructuring can buy; the traffic-side follow-up is [BINIUS-312](https://linear.app/jimpo/issue/BINIUS-312).

### DIFF
--- a/crates/ip-prover/src/sumcheck/mle_store.rs
+++ b/crates/ip-prover/src/sumcheck/mle_store.rs
@@ -3,7 +3,7 @@
 //! Shared multilinear column store for sumcheck round evaluators.
 //!
 //! An [`MleStore`] owns the equal-length multilinear columns that a group of
-//! [`RoundEvaluator`](super::round_evaluator::RoundEvaluator)s reads, along with the deduplicated
+//! [round evaluators](super::round_evaluator) reads, along with the deduplicated
 //! [`Gruen32`] equality-indicator trackers for MLE-check evaluation points. Columns enter the
 //! store either borrowed ([`MleStore::push`]) or owned ([`MleStore::push_owned`]) and are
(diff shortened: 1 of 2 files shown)


jimpo:
This code reads much worse. I still need to vet the performance results to check that it matters.

PR #1846, opened 2026-07-18
BINIUS-313: remove eq tracker from SharedMleCheckProver

benji-potamus:
Closes BINIUS-313.

`SharedMleCheckProver` no longer materializes the full `2^{n-1}` eq-indicator tensor. It now stores just the shared `eval_point` (no `eq_tracker`, no `register_eq_tracker`) and, each round, expands only a `chunk_vars`-wide prefix of the point as the per-chunk eq indicator, folding the higher (suffix) coordinates through the eq-weighted `reduce`.

Per the discussion on the issue, this uses **option (1)** for the accumulator reconciliation.

## Two commits

**1. Thread the reduction level through `MleStore::map_reduce`.** The `reduce` closure of `map_reduce`/`map_reduce_with_fold` now takes a third arg, `level` — the index of the variable bisected to produce the two sub-results. Purely mechanical: every caller (both shared provers, the store's own tests) keeps its existing additive behavior by ignoring `level`.

**2. Rework the eq handling in `SharedMleCheckProver`.**
- `execute` builds `eq_ind_partial_eval(&eval_point[..chunk_vars])` — a `2^chunk_vars` tensor over the low coordinates — and passes it as the `eq_ind` to every `accumulate`. Its `map` closure then `P::reduce`s the wide accumulators.
- Its `reduce` closure, instead of adding, does a linear extrapolation `lo + z·(hi − lo)` with `z = eval_point[level]` — the eq weighting of the `level`-th (suffix) coordinate. So the eq indicator factors as (materialized prefix) ⊗ (suffix folded lazily in the reduction tree); nothing beyond `2^chunk_vars` (≤ `2^12`) is ever expanded, independent of `n`.
- `MleCheckRoundEvaluator::interpolate` now takes a reduced `&[P]` accumulator (the `map` already reduced); `QuadraticMleEvaluator::interpolate` drops its own `P::reduce`.
- `into_shared_sumcheck` registers the shared eq tracker at conversion time (the plain-sumcheck path still uses it).

Per the issue, **`MleToSumCheckEvaluator` is otherwise unchanged** — it just reduces its wide accumulator to `&[P]` before delegating to the inner evaluator (a 2-line adaptation to the new `interpolate` signature). The sumcheck path continues to weight by the full eq tensor via the store tracker.

## Testing
- Extended `test_shared_frac_add_prove_verify` with `n_vars ∈ {14, 15}` — these exceed `MAX_CHUNK_VARS`, so they actually exercise the new suffix folding (15 hits it through both `map_reduce` and the fused `map_reduce_with_fold`); the earlier `n_vars ≤ 8` cases kept everything in the prefix and never invoked the new `reduce`. All round-trip prove→verify, so a wrong eq weighting fails verification.
- `cargo test` green: `binius-ip-prover` (58), `binius-m4-prover` (40), `binius-iop-prover` (35 + integration) — covers the standalone MLE-check prover, the wrapped logUp* final layer, the M4 operand path.
- clippy (`--tests --benches`) and nightly `fmt --check` clean; `cargo doc -D warnings` clean. Each commit builds + tests green independently.

### DIFF
--- a/crates/ip-prover/src/sumcheck/mle_store.rs
+++ b/crates/ip-prover/src/sumcheck/mle_store.rs
@@ -321,6 +321,11 @@ impl<'a, F: Field, P: PackedField<Scalar = F>> MleStore<'a, P> {
 	/// expansion, together, so each leaf hands `map` one [`EvaluationChunk`]: the paired column
 	/// halves and the matching eq chunk, at `2^chunk_vars` scalars per half.
 	///
+	/// `reduce(low, high, level)` combines the two sub-results of a bisection, where `level` is the
+	/// index of the variable that was bisected to produce them (`low` fixes it to 0, `high` to 1).
+	/// A plain sum ignores `level`; an eq-weighted reduction uses it to pick the round's
+	/// coordinate.
+	///
 	/// `chunk_vars` is capped at `n_vars() - 1`, so leaves never exceed the halved hypercube.
 	///
 	/// ## Preconditions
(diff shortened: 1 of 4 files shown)


jimpo:
I measure a 22% speedup for the remaining zerocheck rounds in the `and_reduction` bench.

jimpo (approved):

PR #1847, opened 2026-07-19
BINIUS-315: hoist large allocations out of parallel regions in witness generation

benji-potamus:
## Summary

Behavior-preserving refactor. Hoists large buffer allocations out of rayon closures at the three witness-generation sites called out in [BINIUS-315](https://linear.app/jimpo/issue/BINIUS-315/hoist-large-allocations-out-of-parallel-regions-in-witness-generation), so allocation happens once on the calling thread rather than per rayon task. This removes per-task allocator traffic and is a prerequisite for the `Pool`/`PoolVec` allocator work (which is cheapest to reason about when allocation is single-threaded).

The fix follows the existing model at `crates/prover/src/protocols/intmul/witness.rs` (`compute_b_leaves_parallel`): pre-allocate the output buffers on the calling thread, then `par_iter_mut` over them and fill in place — turning `map(|x| alloc_and_fill(x))` into `zip(&mut outs).for_each(|(out, x)| fill(out, x))`.

## Commits

1. **Add `scaled_eq_ind_partial_eval_into` for caller-allocated output** — the enabling API in `binius-math`. Writes the scaled eq-indicator expansion into a caller-supplied buffer (the in-place form of `scaled_eq_ind_partial_eval`, which is refactored to delegate to it). New test asserts it reproduces the allocating variant exactly, including buffer reuse with stale contents.
2. **Hoist logup\* witness allocations out of parallel regions** — addresses issue sites 1 and 3 in `logup_star/witness.rs`:
   - *Site 3* (numerators): pre-allocate one `2^n` buffer per looker, fill via `scaled_eq_ind_partial_eval_into` in the parallel region.
   - *Site 1* (`combined_pushforward`): replace the `fold`/`reduce` — which allocated a full `2^m` table per rayon task *and* per reduce step — with one accumulator per worker allocated up front, each scattering a contiguous chunk of lookers, then a single-pass merge.
3. **Hoist intmul power table allocations out of parallel region** — addresses issue site 2. Split `power_table` into a thin allocating wrapper and a new `power_table_into` that fills a caller-provided buffer; the `Witness::new` call site pre-allocates the `2 * N_LIMBS` tables and fills them via `par_iter_mut`.

## Acceptance

- No `FieldBuffer` or comparably-sized `Vec` is allocated inside a rayon closure in `crates/prover/src/protocols/intmul/witness.rs` or `crates/ip-prover/src/logup_star/witness.rs`. The two noted related sites (the `buffer_bivariate_product` collect and the pushforward repack) already run on the calling thread and are left for the pool-allocator work.
- No regression: the per-closure compute is byte-identical to before; only the allocation site moved. The one structural change is site 1 (fold/reduce → static chunking + sequential merge). Lookers are equal-cost (same `n`), so static chunking across `current_num_threads()` workers is balanced, and the merge is negligible against the scatter over every looker row. **Flag if you'd rather keep rayon's dynamic work-stealing here.**

Verified green: nightly `fmt --check`, `clippy --tests --benches --examples -D warnings`, rust-doc `-D warnings`, and unit tests (`binius-math` eq, `ip-prover` logup\_star, `prover` intmul::witness). Each commit builds in isolation.

### DIFF
--- a/crates/prover/src/protocols/intmul/witness.rs
+++ b/crates/prover/src/protocols/intmul/witness.rs
@@ -120,9 +120,15 @@ where
 			.step_by(LIMB_BITS)
 			.take(2 * N_LIMBS)
 			.collect::<Vec<_>>();
-		let tables = bases
+		// Allocate the table buffers up front on this thread, so the parallel region only fills
+		// them — no allocator traffic inside the rayon closures.
+		let packed_len = 1 << LIMB_BITS.saturating_sub(P::LOG_WIDTH);
+		let buffers = iter::repeat_with(|| Vec::<P>::with_capacity(packed_len))
+			.take(bases.len())
+			.collect::<Vec<_>>();
+		let tables = (bases, buffers)
 			.into_par_iter()
-			.map(|base| power_table::<F, P>(LIMB_BITS, base))
+			.map(|(base, buffer)| power_table_into(LIMB_BITS, base, buffer))
 			.collect::<Vec<_>>();
 		drop(power_table_scope);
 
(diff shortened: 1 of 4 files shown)


crates/prover/src/protocols/intmul/witness.rs
```diff
@@ -191,14 +199,35 @@ where
...
+	power_table_into(&mut table, base);
+	table
+}
+
+/// Fill a caller-allocated buffer with the power table of `base`: row `i` holds `base^i`.
+///
+/// This is the in-place form of [`power_table`] over a buffer of `2^log_size` rows, where
+/// `log_size` is the buffer's length. Every row is written, so its allocation can be hoisted out of
+/// a hot or parallel region; any prior contents are overwritten.
+pub fn power_table_into<F, P, Data>(table: &mut FieldBuffer<P, Data>, base: F)
```
jimpo: make this private. It should take args `(log_size: usize, base: F, buffer: Vec<P>)` and return `FieldBuffer<P>`. It asserts that `buffer.capacity()` equals `1 << log_size.saturating_sub(P::LOG_WIDTH)` and clears the buffer. Then use `push` and `extend` to fill the buffer.

jimpo (changes requested):

crates/prover/src/protocols/intmul/witness.rs
benji-potamus: Done in 55470d7. `power_table_into` is now private with signature `(log_size: usize, base: F, buffer: Vec<P>) -> FieldBuffer<P>`: it asserts `buffer.capacity() == 1 << log_size.saturating_sub(P::LOG_WIDTH)`, clears the buffer, and fills it with `extend` (first block) + `push` (each subsequent block from the one a block earlier). The `Witness::new` call site now pre-allocates the `2 * N_LIMBS` empty `Vec::with_capacity` buffers on the calling thread and moves them into the parallel `map`, so no allocation happens inside the rayon closure. Verified green on both the rayon and single-threaded (non-rayon wrapper) configs, plus fmt/clippy/doc.

jimpo (approved):

PR #1848, opened 2026-07-19
BINIUS-317: remove the FieldBuffer capacity API

benji-potamus:
## Summary

Removes `FieldBuffer`'s pseudo-capacity — a second, buffer-level length (`values.len()` allowed to exceed the packed length implied by `log_len`) with its own constructor/accessor family. It existed only to serve in-place tensor expansion growing into pre-allocated slack, and leaked into unrelated preconditions and docs. Per [BINIUS-317](https://linear.app/jimpo/issue/BINIUS-317/remove-the-fieldbuffer-capacity-api).

Removes the seven methods: `new_truncated`, `from_values_truncated`, `zeros_truncated`, `zero_extend`, `resize`, `cap`, `log_cap`.

Stacked on the just-merged [#1847](https://github.com/binius-zk/binius64/pull/1847) (BINIUS-315) — it reshapes the `scaled_eq_ind_partial_eval_into` and logup\* code that #1847 introduced.

## Commits

1. **Grow the tensor-expansion backing instead of using buffer capacity.** In-place tensor expansion (`tensor_prod_eq_ind`, `tensor_prod_eq_ind_prepend`) previously bumped `log_len` into slack (`resize`/`zero_extend`). They now take `&mut FieldBuffer<P, Vec<P>>` and actually grow the backing `Vec` one variable at a time via a new `grow` helper (zero-extend semantics, keeping the store fully initialized — see the UB note below). `scaled_eq_ind_partial_eval[_into]` becomes the natural shape jimpo described: build a `Vec<P>` (caller-reserved for the hoisting `_into` form, mirroring `power_table_into`'s `(…, buffer: Vec<P>) -> FieldBuffer<P>` idiom from #1847), grow it, wrap once. `switchover`'s folding tensor and the logup\* numerator buffers become `Vec`-backed accordingly. After this commit nothing outside `field_buffer.rs` touches the capacity API.
2. **Remove the FieldBuffer capacity API.** Delete the seven methods, inline `new`/`from_values`/`zeros` (they no longer build slack), tighten the struct invariant to `values.len() == 1 << log_len.saturating_sub(P::LOG_WIDTH)`, and drop the capacity-only tests.

## Correctness note

The issue flags a UB pitfall: the first sub-packing-width expansion step reads `values[0]` as a whole packed word including lanes past `log_len`, so the backing must stay initialized. The seed is built with `P::from_scalars(iter::once(scale))`, which zero-fills the unused lanes (documented contract), and `grow` zeros both newly-live sub-word lanes and any appended packed words — so the store is always fully initialized, matching the old `zeros_truncated` seed.

## Acceptance

The seven methods are gone; `binius-math` and `binius-ip-prover` build and test clean; no remaining reference to buffer capacity in `FieldBuffer` docs or preconditions (only `Vec` capacity, which is real). Each commit builds + tests green in isolation. Verified: nightly `fmt --check`, `clippy -D warnings` (rayon and single-threaded configs), rust-doc `-D warnings`, and `cargo test --workspace`.

### DIFF
--- a/crates/ip-prover/src/sumcheck/switchover.rs
+++ b/crates/ip-prover/src/sumcheck/switchover.rs
@@ -1,6 +1,10 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
-use std::ops::{Deref, DerefMut};
+use std::{
+	mem,
+	ops::{Deref, DerefMut},
+};
 
 use binius_field::{Field, PackedField};
 use binius_math::{
@@ -32,7 +36,8 @@ use binius_utils::{
 pub struct BinarySwitchover<'b, P: PackedField, B: Bitwise> {
 	n_multilinears: usize,
 	bitmasks: &'b [B],
-	tensor: FieldBuffer<P>,
+	// The folding tensor grows one variable per pre-switchover round, so it is backed by a `Vec`.
+	tensor: FieldBuffer<P, Vec<P>>,
 	folded: Option<Vec<FieldBuffer<P>>>,
 	switchover: usize,
 }
@@ -57,15 +62,17 @@ where
 
 		let n_vars = checked_log_2(bitmasks.len());
 		let switchover = switchover.min(n_vars).max(1);
-		let mut tensor = FieldBuffer::zeros_truncated(0, switchover);
-		tensor.set(0, F::ONE);
+		// The folding tensor starts as the scalar `1`, with backing reserved for the fully-grown
+		// (`switchover`-variable) tensor so the per-round growth never reallocates.
+		let tensor = FieldBuffer::scalar_with_capacity(F::ONE, switchover);
 
 		Self {
 			n_multilinears,
 			bitmasks,
 			tensor,
 			folded: None,
-			// Store this separately as `log_cap` is rounded up to the packing width
+			// The tensor grows up to this many variables; store it since it can't be recovered from
+			// the buffer once the growth completes.
 			switchover,
 		}
 	}
@@ -114,7 +121,8 @@ where
 		} else {
 			// Pre-switchover: update the folding tensor
 			assert!(self.tensor.log_len() < self.switchover);
-			tensor_prod_eq_ind_prepend(&mut self.tensor, &[challenge]);
+			let tensor = mem::replace(&mut self.tensor, FieldBuffer::new(0, vec![P::zero()]));
+			self.tensor = tensor_prod_eq_ind_prepend(tensor, &[challenge]);
 
 			if self.tensor.log_len() == self.switchover {
 				self.perform();
--- a/crates/math/src/field_buffer.rs
+++ b/crates/math/src/field_buffer.rs
@@ -133,6 +133,27 @@ impl<P: PackedField> FieldBuffer<P> {
 	}
 }
 
+impl<P: PackedField> FieldBuffer<P, Vec<P>> {
+	/// A single-scalar buffer (`log_len` 0) holding `value` in lane 0, with backing-`Vec` capacity
+	/// reserved for `log_capacity` variables.
+	///
+	/// The backing `Vec` reserves `1 << log_capacity.saturating_sub(P::LOG_WIDTH)` packed words, so
+	/// growing the buffer up to `log_capacity` variables never reallocates.
+	pub fn scalar_with_capacity(value: P::Scalar, log_capacity: usize) -> Self {
+		let mut values = Vec::with_capacity(1 << log_capacity.saturating_sub(P::LOG_WIDTH));
+		values.push(P::from_scalars([value]));
+		Self { log_len: 0, values }
+	}
+
+	/// Converts into the canonical boxed-slice-backed [`FieldBuffer`].
+	pub fn into_boxed(self) -> FieldBuffer<P> {
+		FieldBuffer {
+			log_len: self.log_len,
+			values: self.values.into_boxed_slice(),
+		}
+	}
+}
+
 #[allow(clippy::len_without_is_empty)]
 impl<P: PackedField, Data: Deref<Target = [P]>> FieldBuffer<P, Data> {
 	/// Create a new FieldBuffer from a slice of packed values.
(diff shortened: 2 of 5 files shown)


crates/math/src/field_buffer.rs
```diff
@@ -115,21 +94,45 @@ impl<P: PackedField> FieldBuffer<P> {
...
 	/// # Preconditions
 	///
-	/// * `log_len` must not exceed `log_cap`.
-	pub fn zeros_truncated(log_len: usize, log_cap: usize) -> Self {
-		assert!(log_len <= log_cap, "precondition: log_len must not exceed log_cap");
-		let packed_len = 1 << log_cap.saturating_sub(P::LOG_WIDTH);
-		let values = zeroed_vec(packed_len).into_boxed_slice();
-		Self { log_len, values }
+	/// * `new_log_len` must be at least the current `log_len()`.
+	pub(crate) fn grow(&mut self, new_log_len: usize) {
```
jimpo: No `pub(crate)`, just `pub`. See Visibility section of CONTRIBUTING.md.

crates/math/src/field_buffer.rs
```diff
@@ -115,21 +94,45 @@ impl<P: PackedField> FieldBuffer<P> {
...
+			}
+		}
+
+		let new_packed_len = 1 << new_log_len.saturating_sub(P::LOG_WIDTH);
+		self.values.resize(new_packed_len, P::zero());
+		self.log_len = new_log_len;
+	}
+
+	/// Converts into the canonical boxed-slice-backed [`FieldBuffer`].
+	pub(crate) fn into_boxed(self) -> FieldBuffer<P> {
```
jimpo: Same

crates/ip-prover/src/sumcheck/switchover.rs
```diff
@@ -57,15 +59,20 @@ where
 
 		let n_vars = checked_log_2(bitmasks.len());
 		let switchover = switchover.min(n_vars).max(1);
-		let mut tensor = FieldBuffer::zeros_truncated(0, switchover);
+		// Reserve the backing for the fully-grown tensor (`switchover` variables) so the per-round
+		// growth never reallocates.
+		let mut backing = Vec::with_capacity(1 << switchover.saturating_sub(P::LOG_WIDTH));
+		backing.push(P::zero());
```
jimpo: Just push one instead of setting F::ONE immediately after.

crates/math/src/multilinear/hypercube.rs
```diff
@@ -152,21 +148,14 @@ impl Hypercube for InfCube {
...
-	assert!(
-		values.log_cap() >= new_log_len,
-		"precondition: values capacity must be sufficient for expansion"
-	);
-
 	for &r_i in extra_query_coordinates {
 		let packed_r_i = P::broadcast(r_i);
 
-		values.resize(values.log_len() + 1);
+		values.grow(values.log_len() + 1);
```
jimpo: This is inefficient (because it initializes the new region of the vec) and points to the fact that `FieldBuffer` should not have a `grow` method. Instead, `FieldBuffer` should have a method `take_data(self) -> Data`, which returns its data. In this case it's a Vec. For now, resize the Vec (replicating the grow logic in this function), create a new FieldBuffer with log_len incremented, and proceed. Then, in a follow-up PR, rewrite this function to do this properly by splitting the loop into the range when it's below the packing width and above. In the above case, use spare_capacity_mut and set_len to bump the length of the array without duplicatively writing to the uninitialized region.

jimpo (changes requested):

crates/math/src/field_buffer.rs
benji-potamus: Removed `grow` entirely (per the `take_data` comment below) — it's no longer a `FieldBuffer` method.

crates/math/src/field_buffer.rs
benji-potamus: Made `into_boxed` `pub`. (It's removed entirely in the stacked BINIUS-319, which makes `Vec<P>` the default backing so the conversion is unnecessary.)

crates/ip-prover/src/sumcheck/switchover.rs
benji-potamus: Done — now `backing.push(P::from_scalars([F::ONE]))` and dropped the separate `set(0, F::ONE)`.

crates/math/src/multilinear/hypercube.rs
benji-potamus: Done. Removed `grow`; added `pub fn take_data(self) -> Data`. `tensor_prod_eq_ind`/`_prepend` now take the buffer by value, `take_data()` it, resize the `Vec` in place, and rewrap with the incremented `log_len`. The resize logic is factored into a private `grow_packed_backing` helper (module-local, not a `FieldBuffer` method) marked as a temporary placeholder — I left the proper rewrite (split the loop by packing width, `spare_capacity_mut` + `set_len` to skip re-initializing the overwritten region) for a follow-up PR as you suggested. Let me know if you'd prefer that follow-up as a new issue.

crates/math/src/multilinear/hypercube.rs
```diff
@@ -403,6 +408,17 @@ mod tests {
 	type P = Packed128b;
 	type F = B128;
 
+	/// A `Vec`-backed buffer of `log_len` 0 seeded with `value` in lane 0, with backing reserved
+	/// for `reserve_vars` variables — the growable seed the tensor-expansion routines start from.
+	fn seeded_vec_buffer<Q: PackedField>(
```
jimpo: This comes up in a few places. Make it a public method `FieldBuffer::scalar_with_capacity(log_capacity: usize, value: P::Scalar) -> Self`. It returns a `FieldBuffer<P, Vec<P>>` with log_len = 0 and the single value `value` but where the backing vec has capacity `1 << log_capacity.saturating_sub(P::LOG_WIDTH)`. Use it where appropriate.

jimpo (changes requested):

crates/math/src/multilinear/hypercube.rs
benji-potamus: Done — added `pub fn FieldBuffer::scalar_with_capacity(value, log_capacity) -> FieldBuffer<P, Vec<P>>` (arg order flipped to value-first per your follow-up). Now used in `switchover::new` and the two tensor-expansion tests (dropped the local `seeded_vec_buffer` helpers).

jimpo (approved):

PR #1849, opened 2026-07-19
BINIUS-320: remove FieldBuffer::zeros at the production call sites

tcoratger:
Closes [BINIUS-320](https://linear.app/jimpo/issue/BINIUS-320).

Converts the ten production `FieldBuffer::zeros` call sites to data-first construction. Pure refactor — no behavior change, the buffers hold the same zero bytes as before.

### The problem

`FieldBuffer` has two ways to come into existence: wrap data you already built (`new` / `from_values`), or ask for a zeroed buffer by log-length (`zeros` / `zeros_truncated`).

The second path hides a requirement.

- A reader at the call site cannot tell whether the zeros are *load-bearing* (read back later) or just there because the constructor happens to zero.
- The goal is one construction path — wrap existing data — with any zeroing spelled out where it happens.

### The idea

Build the backing vector first, then wrap it:

```
- let mut buf = FieldBuffer::<P>::zeros(n);
+ let packed_len = 1 << n.saturating_sub(P::LOG_WIDTH);
+ let mut buf = FieldBuffer::new(n, vec![P::zero(); packed_len].into_boxed_slice());
```

Now the allocation and its zeros are visible, and each site carries a one-line note saying *why* the zeros are (or are not) needed.

### The change

Ten sites across four crates, grouped by why they zero:

- **Zeros are load-bearing** — an accumulator (`+=`) or a sparse scatter whose untouched positions are read back:
  - `prover/src/ring_switch.rs` — the fold and reduce identities (the $\mathbb{F}_{2^{128}}$ accumulator)
  - `iop-prover/src/basefold_channel.rs` — per-oracle lift accumulator
  - `iop-prover/src/fri/fold.rs` — combined-codeword accumulator with a truncated-zip tail
  - `ip-prover/src/sumcheck/zk_mlecheck.rs` — Libra scatter whose padding is read as coefficients
  - `prover/src/and_reduction/ntt_lookup.rs` — the upper half is the NTT's zero-padded coefficients
- **Fully overwritten scratchpads** — zeroed only to construct; a fill routine writes every element before it is read:
  - `ip-prover/src/sumcheck/mle_store.rs`, `.../switchover.rs`, `.../selector_mle.rs` (two buffers)

### Scope

- **changed:** the ten call sites above, each now data-first with a `// why` note.
- **left untouched:** the `zeros` / `zeros_truncated` constructors themselves. Their remaining callers are the capacity-allocation sites deferred to [BINIUS-317](https://linear.app/jimpo/issue/BINIUS-317), a small-input fallback the audit judged not worth converting, and the test suite. Deleting the constructor is the final step once BINIUS-317/318 clear those.
- One deliberate non-optimization: the fully-overwritten scratchpads still allocate zeros rather than uninitialized memory. Realizing that win safely needs a `Pod`/`MaybeUninit` path the generic `P: PackedField` bound does not provide, and the zeroing there is once-per-thread — not worth `unsafe`. Left as a clean follow-up.

### Soundness

- Not a protocol change. Every new buffer holds the identical zero bytes the old constructor produced (`vec![P::zero(); ..]` equals `zeroed_vec` for these binary-field packed types).
- Load-bearing sites keep their zeros; overwritten sites are provably written in full before any read.

### Verification

- `cargo check` / `cargo clippy` on `binius-prover`, `binius-iop-prover`, `binius-ip-prover` — clean
- `cargo test` on the three crates — 132 passed, 0 failed

### DIFF
--- a/crates/iop-prover/src/basefold_channel.rs
+++ b/crates/iop-prover/src/basefold_channel.rs
@@ -342,7 +342,11 @@ fn prove_batch_zk_basefold<F, P, NTT, Channel>(
 
 		let eq_tensor = eq_ind_partial_eval_scalars(&outer_challenges);
 
-		let mut combined = FieldBuffer::<P>::zeros(max_n);
+		// Each oracle accumulates into only the low `2^{n_i}` of every lifted chunk.
+		// Positions that no oracle touches must read back as zero.
+		let combined_packed_len = 1 << max_n.saturating_sub(P::LOG_WIDTH);
+		let mut combined =
+			FieldBuffer::new(max_n, vec![P::zero(); combined_packed_len].into_boxed_slice());
 		let mut s_prime = F::ZERO;
 		for (fri_oracle, witness_prime, eq_i, alpha_i) in
 			izip!(fri_params.input_oracles(), witness_primes, eq_tensor, alphas)
--- a/crates/iop-prover/src/fri/fold.rs
+++ b/crates/iop-prover/src/fri/fold.rs
@@ -488,7 +488,13 @@ impl<F: Field, P: PackedField<Scalar = F>, C> BatchBrakedownFolder<P, C> {
 		let later_challenges = &challenges[max_early + log_n_oracles..];
 		let outer_tensor = eq_ind_partial_eval::<F>(outer_challenges);
 
-		let mut combined_codeword = FieldBuffer::zeros(self.log_code_len);
+		// The folders accumulate into this scalar buffer.
+		// The chunked zip stops at the shorter side, leaving a tail that no folder writes.
+		// That tail must read back as zero.
+		let mut combined_codeword = FieldBuffer::new(
+			self.log_code_len,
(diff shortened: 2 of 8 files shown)


tcoratger:
Just updated this one now that https://github.com/binius-zk/binius64/pull/1853 is merged

jimpo:
Ah, sorry, the issue description was wrong. I want to keep `FieldBuffer::zeros()` unless the zeroing is totally unnecessary. In the other cases, FieldBuffer::zeros is more convenient (and efficient!), so we should keep it. (See Linear ticket comment)

tcoratger:
> Ah, sorry, the issue description was wrong. I want to keep `FieldBuffer::zeros()` unless the zeroing is totally unnecessary. In the other cases, FieldBuffer::zeros is more convenient (and efficient!), so we should keep it. (See Linear ticket comment)

Aaah no problem I was thinking that you wanted to simplify this, let me close that for now, we can reopen later something simpler

PR #1850, opened 2026-07-19
BINIUS-312: split-eq the multilinear-evaluation MLE-check prover

tcoratger:
Closes [BINIUS-312](https://linear.app/jimpo/issue/BINIUS-312).

Makes the single-multilinear evaluation MLE-check prover split-eq, so it stops materializing and streaming the full eq-indicator buffer. Output is bit-identical — internal prover change, no protocol impact.

### Context — what this finishes

[BINIUS-313](https://linear.app/jimpo/issue/BINIUS-313) already made the AND-reduction tail (`SharedMleCheckProver`) split-eq. This applies the same idea to the one remaining full-eq streamer left on the prove hot path at $2^{21}$–$2^{22}$: the single-multilinear evaluation prover that BaseFold uses to open the PCS on the full committed-trace multilinear.

### The problem

To prove `M(z) = s`, the prover sumchecks the equivalent relation:

```
s = sum_{v in B_n} M(v) * eq(v, z)
```

The old prover built the whole `eq(v, z)` weight array — one entry per row of `M` — and each round walked the witness column and that eq array side by side, folding both.

- At $2^{22}$ rows the eq array is a 64 MB second column.
- It is streamed and folded every round, halving each time.
- The round is memory-bound, so that extra stream is close to pure tax (~1/4 of the traffic).

### The idea

`eq` factorizes over a split of the variables into a low group and the rest:

```
eq(v, z) = eq(v_lo, z_lo) * eq(v_hi, z_hi)
```

So the full array is never needed:

- keep a **small** `eq_lo` table over the low coordinates — a few thousand entries, cache-resident;
- fold the high coordinates in one at a time, which the sumcheck recursion over the rows already does.

No big eq array, no per-round eq fold.

### The change

- The evaluation prover is rebuilt on the split-eq `SharedMleCheckProver` (the machinery [BINIUS-313](https://linear.app/jimpo/issue/BINIUS-313) added) with a new degree-1 evaluator: for each row-block, dot the column's `X = 1` half against the small `eq_lo` chunk. The suffix coordinates fold through the prover's reduction step.
- `round_claim` on the shared MLE-check prover now reads the round's eq coordinate lazily (inside the post-execute branch that actually needs it), so it stays valid to call once every variable is folded — matching the evaluation prover's prior contract. Previously it indexed the point eagerly and underflowed at zero remaining variables.

### Scope

- **changed:** the single-multilinear evaluation prover (now a thin split-eq `SharedMleCheckProver` + a degree-1 evaluator) and the lazy `round_claim` read.
- **unchanged public API:** the prover's constructor and its callers (BaseFold PCS opening, IntMul re-randomization).
- **left as follow-up:** the `register_eq_tracker` path that feeds a plain `SharedSumcheckProver` (M4 re-randomization, logUp\* final layer) still streams a full eq column. It batches claims with *different* eq points through a plain-sum reduction, so the single-suffix-fold trick does not apply unchanged, and those sites are instance/layer-sized rather than the $2^{22}$ trace.

### Soundness

- The round polynomials are bit-identical to the streaming version — the split is a re-association of the same sum — so the transcript is unchanged.
- Covered by the existing conformance test (evaluation prover vs the quadratic MLE-check with the identity composition), the prove/verify roundtrip, and the BaseFold and IntMul suites.

### Verification

- `cargo build` / `cargo clippy` on `binius-ip-prover`, `binius-iop-prover`, `binius-prover` — clean
- `cargo +nightly fmt` — clean
- `cargo test` on the three crates — all pass (ip-prover 58, iop-prover 40, prover 24)

### Benchmark

Full round schedule of the evaluation prover, single core, one 64-lane packed B128, honest random instance:

| size | before (stream eq) | after (split-eq) | speedup |
|---|---|---|---|
| $2^{20}$ | 4.67 ms | 3.53 ms | **1.32×** |
| $2^{22}$ | 18.6 ms | 14.05 ms | **1.32×** |

The ~4.5 ms saved at $2^{22}$ matches the issue's estimate (~1/4 of the tail traffic).

<details>
<summary>Benchmark code (not committed — drop into <code>crates/ip-prover/benches/multilinear_eval.rs</code> + register <code>[[bench]]</code> to reproduce locally)</summary>

```rust
use binius_field::{FieldOps, Random, arch::OptimalPackedB128};
use binius_ip_prover::sumcheck::{common::SumcheckProver, multilinear_eval::MultilinearEvalProver};
use binius_math::{
	multilinear::evaluate::evaluate,
};
use criterion::{BatchSize, Criterion, Throughput, criterion_group, criterion_main};

type P = OptimalPackedB128;
type F = <P as FieldOps>::Scalar;

fn bench_multilinear_eval(c: &mut Criterion) {
	let mut group = c.benchmark_group("multilinear_eval");

	for n_vars in [20, 22] {
		group.throughput(Throughput::Elements(1 << n_vars));
		group.bench_function(format!("n_vars={n_vars}"), |b| {
			let witness = random_field_buffer::<P>(&mut rng, n_vars);
			let eval_point = random_scalars::<F>(&mut rng, n_vars);
			let eval_claim = evaluate(&witness, &eval_point);
			let challenges: Vec<F> = (0..n_vars).map(|_| F::random(&mut rng)).collect();

			b.iter_batched(
				|| witness.clone(),
				|witness| {
					let mut prover = MultilinearEvalProver::new(witness, &eval_point, eval_claim);
					for &challenge in &challenges {
						let _ = prover.execute();
						prover.fold(challenge);
					}
					prover.finish()
				},
				BatchSize::SmallInput,
			);
		});
	}
}

criterion_group!(multilinear_eval, bench_multilinear_eval);
criterion_main!(multilinear_eval);
```

The "before" numbers were taken by running the same bench against the pre-change (streaming) implementation.

</details>


crates/ip-prover/src/sumcheck/multilinear_eval.rs
```diff
@@ -3,37 +3,105 @@
...
-	last_coeffs_or_sum: RoundState<RoundCoeffs<P::Scalar>, P::Scalar>,
+/// The claim is `M(z) = s` for a multilinear `M` over the challenge field.
+/// This proves the equivalent MLE-check relation `s = sum_{v in B_n} M(v) * eq(v, z)`.
+/// Since `M` is multilinear, that relation holds if and only if `M(z) = s`.
+///
+/// The reduction runs on the split-eq [`SharedMleCheckProver`] with a degree-1 evaluator.
+/// Each round expands only a small low-coordinate prefix of the eq indicator.
+/// The higher coordinates are folded in through the prover's reduction step.
+/// The full `2^{n-1}` eq tensor is never materialized, streamed, or folded per round.
+pub struct MultilinearEvalProver<F: Field, P: PackedField<Scalar = F>> {
```
jimpo: You can remove this struct now and make a custom constructor method `multilinear_eval_prover(witness: FieldBuffer<P>, eval_point: &[F], eval_claim: F) -> ...` just like `quadratic_mlecheck_prover`.

jimpo (commented):
Thanks! Very nice

crates/ip-prover/src/sumcheck/multilinear_eval.rs
tcoratger: Just fixed

jimpo (approved):

PR #1858, opened 2026-07-20
BINIUS-90: make PackedField a Field supertrait

tcoratger:
@jimpo I read your fear on the Linear ticket, I've found this design which looks great no?

Closes [BINIUS-90](https://linear.app/jimpo/issue/BINIUS-90).

Declares in the type system what was already true: every field is a packed vector of width one. Pure refactor — no behavior change.

### The problem

"Every `Field` is also a `PackedField`" was expressed by a blanket `impl<F: Field> PackedField for F`.

- The relationship lived in an impl, not in the trait bound.
- So a `F: Field` bound alone did **not** give you the packed API — you had to add `+ PackedField` (or `+ FieldOps`) at the call site.
- And `Field` separately re-listed a long tail of bounds (arithmetic, `Divisible<Self>`, `Maskable<Self>`, `Random`, `Zeroable`, `WideMul`) that a packed field already guarantees.

### The idea

Make the relationship a **supertrait** and let each field carry its own width-one impl:

```
Field: PackedField<Scalar = Self>
```

- A `Field` bound now supplies the entire packed API for free.
- The blanket impl is gone; the `binary_field!` macro emits a per-type width-one `PackedField` impl instead — exactly how it already emits the reflexive `Divisible<Self>` / `Maskable<Self>` impls.
- The impl bodies are byte-identical to the old blanket (iterate yields the single element; `interleave` / `unzip` panic; `from_fn` calls `f(0)`).

### The change

```
// field.rs — Field's supertrait list collapses to the field-only additions
- Sized + Eq + Copy + ... + Neg + Add + ... + Divisible<Self> + Maskable<Self> + WideMul<...>
+ PackedField<Scalar = Self> + Display + Hash + SerializeBytes + DeserializeBytes + FixedSizeSerializeBytes

// packed.rs — the blanket impl is removed
- impl<F: Field> PackedField for F { ... }

// binary_field.rs — the macro now emits the width-one impl for every field type
+ impl PackedField for $name { fn from_fn(f) { f(0) } ... }
```

Everything else on the old `Field` list (all arithmetic ops, `Divisible<Self>`, `Maskable<Self>`, `Random`, `Zeroable`, `WideMul`, `Send`/`Sync`/`Copy`/`'static`/...) is now inherited transitively through `PackedField` and `FieldOps`, so listing it was redundant.

### Why this compiles — the trait-solver cycle is not real

The interesting risk: `Field` needs `PackedField`, which needs `FieldOps`, which is handed out by `impl<F: Field> FieldOps for F` — a loop back to `Field`.

It terminates because the compiler satisfies a bound by matching an impl and then checking only **that impl's own where-clauses**, not the trait's supertraits. Each field's `PackedField` impl and `Field` impl have no where-clauses, so every lookup bottoms out immediately:

```
is BinaryField1b a PackedField?  -> found `impl PackedField for BinaryField1b`, no where-clauses -> yes
is BinaryField1b a FieldOps?     -> `impl<F: Field> FieldOps for F` needs `BinaryField1b: Field` -> found impl, no where-clauses -> yes
```

Giving each field a concrete `PackedField` impl (and removing the blanket) is what keeps the first lookup from re-deriving through `Field`.

### Scope

- **changed:** `crates/field/src/{field.rs, packed.rs, binary_field.rs}` — the four field types (`BinaryField1b`, `AESTowerField8b`, `BinaryField128bGhash`, `GhashSq256b`) all satisfy the new supertrait via the macro.
- **cleanup:** dropped one now-redundant `FieldOps` import in `crates/prover/src/protocols/intmul/prove.rs`.
- **left untouched:** the `PackedPrimitiveType` and `SlicedPackedField` packed impls; the reflexive `FieldOps for F` and `ExtensionField<F> for F` blanket impls.
- No `Field` types exist outside the field crate, and there are no self-targeted `impl<P: PackedField>` / `impl<F: Field>` blankets that would newly overlap, so coherence is unaffected.

### Verification

- `cargo check --workspace --all-targets`, `clippy --workspace --all-targets`, nightly `fmt --all` — clean.
- `binius-field` tests: 208 passed, doctests pass.

### DIFF
--- a/crates/field/src/binary_field.rs
+++ b/crates/field/src/binary_field.rs
@@ -169,7 +169,7 @@ macro_rules! binary_field {
 		}
 
 		// A field element divides into exactly one element of itself. This makes the field a
-		// degenerate packed field of width one (see the blanket `PackedField for Field` impl).
+		// degenerate packed field of width one (see the `PackedField` impl below).
 		impl $crate::Divisible<$name> for $name {
 			const LOG_N: usize = 0;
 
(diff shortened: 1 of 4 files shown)


jimpo (approved):
Awesome! Much cleaner!

PR #1860, opened 2026-07-20
BINIUS-325: batch the three per-word NTT lookups in the univariate round message

tcoratger:
@jimpo I've found this to be pretty neutral locally, but I think this still is a cleanup.

Closes [BINIUS-325](https://linear.app/jimpo/issue/BINIUS-325).

Computes the AND-reduction round message exactly as before, but schedules its per-word NTT lookups to keep more independent loads in flight. Output is bit-identical — no protocol change, no new soundness assumption.

### The problem

The univariate round message extends three columns per word — $A$, $B$, and `C = A & B` — through the NTT lookup table.

Each extension is an eight-step loop: one table load per byte position, each feeding an accumulator.

The loads are the bottleneck. They hit an L1-resident table, but each load's result feeds the next accumulate, so a single `NTTLookup::ntt` keeps only **one load in flight** at a time — the loop is load-latency-bound, not throughput-bound.

Three separate `ntt` calls per word run this three times back to back, each a lone dependent chain.

### The idea

Run the same eight-position pass across all three inputs at once.

Per byte position, issue the three independent table loads before accumulating:
- three loads in flight instead of one,
- three independent accumulator chains instead of one dependent chain.

The parity select `b % 2` and the lane permute `b / 2` depend only on the byte position, not the data — so they are decided once per position and shared across the three columns.

```
// per byte position b:
table = self.0[b % 2]
load a[b], b[b], c[b]         // three independent loads
permute + xor into out[0..3]  // three independent chains
```

### The change

```diff
- let a_lde = ntt_lookup.ntt(a_i);
- let b_lde = ntt_lookup.ntt(b_i);
- let c_lde = ntt_lookup.ntt(c_i);
+ let [a_lde, b_lde, c_lde] = ntt_lookup.ntt3([a_i, b_i, c_i]);
```

### Soundness

None affected. The output is bit-identical to three `ntt` calls, so the round message and the whole transcript are unchanged. This is a scheduling change, not a protocol change.

### Scope

- **new:** `NTTLookup::ntt3([x0, x1, x2]) -> [Packed64xB8; 3]` in `crates/prover/src/and_reduction/ntt_lookup.rs`.
- **changed:** one call site in `sumcheck_round_messages.rs`.
- the single-input `ntt` stays as the reference oracle for the equivalence test.

### Verification

- nightly `fmt --all`, `check --workspace --all-targets`, `clippy -p binius-prover --all-targets` — clean.
- `ntt3` proven equal to `[ntt, ntt, ntt]` over 100 random inputs.
- round-message correctness (`test_first_round_message_matches_next_round_sum_claim`) and end-to-end prover (`test_transcript_prover_verifies`) tests pass.

### Benchmark

| arch | before | after | note |
|---|---|---|---|
| Apple M1 (`2^21`) | 59.5 ms | 59.9 ms | neutral (within noise) |
| Neoverse V3 (`2^22`) | — | — | −3.5 ms per the BINIUS-310 prototype |

Apple's wide out-of-order core already interleaves the three separate `ntt` calls, so batching is neutral there; the win shows on narrower aarch64 server cores — the target arch for the flock comparison. No regression on M1.

### DIFF
--- a/crates/prover/src/and_reduction/ntt_lookup.rs
+++ b/crates/prover/src/and_reduction/ntt_lookup.rs
@@ -47,7 +47,7 @@
 //! table footprint by 4x — to `2 * 256 * 64` field elements — adding only a cheap in-register
 //! permute to each lookup.
 
-use std::{array, marker::PhantomData};
+use std::{array, iter, marker::PhantomData};
 
 use binius_core::Word;
 use binius_field::{
@@ -122,8 +122,8 @@ impl NTTLookup {
 	/// permutation of the four 128-bit lanes, `lane[i] <- lane[i ^ (b / 2)]`. The loop is unrolled
 	/// over `b`, so the parity select and lane index are compile-time constants.
 	///
-	/// Used directly only in tests; `univariate_round_message_extension_domain` accesses the tables
-	/// inline to compute three LDE evaluations at once, which is more efficient.
+	/// The round message uses the batched [`ntt3`](Self::ntt3).
+	/// This single-input form is the reference it is checked against, and is otherwise test-only.
 	///
 	/// ## Returns
 	///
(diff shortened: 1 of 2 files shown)


jimpo:
Have you tested this on Graviton or anything? If it's not a good performance improvement (which I'm not seeing locally with AVX2 either), I'm inclined to close it. The current code seems simpler.

tcoratger:
> Have you tested this on Graviton or anything? If it's not a good performance improvement (which I'm not seeing locally with AVX2 either), I'm inclined to close it. The current code seems simpler.

No, I haven't tested it on Graviton, but if this isn't a performance improvement on both Mac and AVX, then we should close. I thought it would be better on AVX. Let's close for now.

PR #1864, opened 2026-07-20
BINIUS-316: introduce binius-compute BufferPool and pool the prover's Word buffers

benji-potamus:
## Summary

First increment of [BINIUS-316](https://linear.app/jimpo/issue/BINIUS-316), following @jimpo's plan on the issue: stand up a `BufferPool` allocation seam and thread it through the prover, starting with the `Word` buffers. This is the scaffolding step — `alloc_vec` is a pass-through to `Vec::with_capacity` for now (no recycling free list yet), establishing the API and the `'alloc` lifetime threading. `FieldBuffer` pooling is the next step (see below).

## What changed

**New crate `binius-compute` (`crates/compute`):**
- `BufferPool` with `alloc_vec<T: Pod>(&self, capacity) -> PoolVec<'_, T>` (currently `Vec::with_capacity`).
- `PoolVec<'alloc, T>` = `&'alloc BufferPool` + `Vec<T>`, exposing only the `Vec`-like surface callers actually use: `Deref`/`DerefMut` to `[T]`, `push`, `extend`, `extend_from_slice`, `resize`, `clear`, `capacity`, and `spare_capacity_mut` + `unsafe set_len` (needed for the parallel fixed-size fill pattern the prover uses). Kept generic over `T: Pod` so the crate needn't depend on `binius-core`.
- Unit tests for the buffer API.

**Threaded the pool through `crates/prover`:**
- `Prover::prove` and `ZKProver::prove` create a `BufferPool` for the call and pass `&pool` into `IOPProver::prove`, which forwards it to the three `build_*_witness` builders. The pool is a per-call **argument**, so `Prover`/`IOPProver` gain no lifetime (avoids the self-reference problem the issue flagged).
- The AND/IntMul/BinMul witness columns (`AndCheckWitness`, `MulCheckWitness`, `BinMulCheckWitness`) are now `PoolVec<'alloc, Word>`, allocated via `pool.alloc_vec::<Word>(n)`. These are all fixed-size, allocate-once buffers — a good fit for a future fixed-capacity pool.

## Design decisions

1. **`OblongZerocheckProver` made generic over its column backing** (`Data: Deref<Target = [Word]>`). It's shared: `m4-prover` (a separate prover, not reachable from `Prover`) also constructs it, with `Vec<Word>`. Rather than ripple a pool/lifetime through m4-prover's `ValueTable` hub — out of scope for "thread it through `Prover`" — the generic lets crates/prover pass `PoolVec` while m4-prover keeps `Vec` unchanged (one-token turbofish edit). This is the only change to m4-prover.

2. **Scope = the top-level `Prover` path's Word buffers.** `m4-prover`'s own Word buffers (`ValueTable`, `BatchAndCheckWitness`, its builders) stay `Vec<Word>`; pooling them would mean creating a pool at m4-prover's own entry and is a clean separate step. `ip-prover` has no `Vec<Word>`.

## Open / next steps (per the issue's plan)

- **FieldBuffers** — jimpo's "if that works, next step": convert the (non-trivial) `FieldBuffer` backings to `PoolVec`. Larger; proposed as the follow-up PR.
- **m4-prover Word buffers** — pool them too, or leave as a separate prover? Happy either way.
- **Real pool** — the recycling free list behind `alloc_vec`/`PoolVec::drop` slots in without changing these signatures.

## Verification

Green: `binius-compute`, `binius-prover`, `binius-m4-prover` tests (incl. the end-to-end `prove_verify`); `clippy -D warnings` (all-targets); nightly `fmt`; `rustdoc`; full workspace build.

### DIFF
--- a/crates/compute/Cargo.toml
+++ b/crates/compute/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "binius-compute"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+bytemuck = { workspace = true }
--- a/crates/compute/src/lib.rs
+++ b/crates/compute/src/lib.rs
@@ -1,0 +1,171 @@
+// Copyright 2026 The Binius Developers
+
+//! Buffer pooling for prover working memory.
+//!
+//! The prover allocates many large, short-lived buffers. [`BufferPool`] is the seam through which
+//! those allocations flow, so they can later be served from a recycling free list instead of the
+//! global allocator. Buffers are handed out as [`PoolVec`] handles that borrow the pool for
+//! `'alloc` and, in a future revision, return their block to it on drop.
+//!
+//! The current implementation is a thin pass-through over [`Vec`]: [`BufferPool::alloc_vec`] just
+//! calls [`Vec::with_capacity`]. This establishes the API and the `'alloc` lifetime threading
+//! without committing to a pool layout yet.
+
+use std::{
+	fmt,
+	mem::MaybeUninit,
+	ops::{Deref, DerefMut},
+};
+
+use bytemuck::Pod;
+
+/// A pool that hands out reusable buffers for prover working memory.
+///
+/// Allocation goes through [`alloc_vec`](Self::alloc_vec), which currently forwards to
+/// [`Vec::with_capacity`]. A pool is created once, above the code that uses it, and shared by
+/// borrow — every [`PoolVec`] it produces holds a `&'alloc BufferPool`.
+#[derive(Default)]
+pub struct BufferPool {
+	// No free list yet; buffers are allocated straight from the global allocator. A recycling
+	// implementation slots in behind `alloc_vec`/`PoolVec::drop` without changing their
+	// signatures.
+	_private: (),
+}
+
+impl BufferPool {
+	/// Creates a new pool.
+	pub fn new() -> Self {
+		Self::default()
+	}
+
+	/// Allocates a [`PoolVec`] with room for at least `capacity` elements.
+	///
+	/// The returned buffer is empty; fill it through the [`PoolVec`] interface.
+	pub fn alloc_vec<T: Pod>(&self, capacity: usize) -> PoolVec<'_, T> {
+		PoolVec {
+			pool: self,
+			data: Vec::with_capacity(capacity),
+		}
+	}
+}
+
+impl fmt::Debug for BufferPool {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		f.debug_struct("BufferPool").finish_non_exhaustive()
+	}
+}
+
+/// A `Vec`-like buffer borrowed from a [`BufferPool`] for `'alloc`.
+///
+/// Dereferences to `[T]`, so all slice operations are available directly. Only the growth and
+/// mutation methods actually used by callers are exposed; add more as needed rather than mirroring
+/// all of [`Vec`].
+pub struct PoolVec<'alloc, T: Pod> {
+	// Held so a future recycling pool can return the block on drop. Unused by the pass-through.
+	#[allow(dead_code)]
+	pool: &'alloc BufferPool,
+	data: Vec<T>,
+}
+
+impl<T: Pod> PoolVec<'_, T> {
+	/// Returns the number of elements the buffer can hold without reallocating.
+	pub const fn capacity(&self) -> usize {
+		self.data.capacity()
+	}
+
+	/// Appends an element to the back of the buffer.
+	pub fn push(&mut self, value: T) {
+		self.data.push(value);
+	}
+
+	/// Appends all elements of `other` to the back of the buffer.
+	pub fn extend_from_slice(&mut self, other: &[T]) {
+		self.data.extend_from_slice(other);
+	}
+
+	/// Clears the buffer, removing all elements while retaining its capacity.
+	pub fn clear(&mut self) {
+		self.data.clear();
+	}
+
+	/// Resizes the buffer to `new_len`, filling any new slots with `value`.
+	pub fn resize(&mut self, new_len: usize, value: T) {
+		self.data.resize(new_len, value);
+	}
+
+	/// Returns the spare capacity of the buffer as a slice of `MaybeUninit<T>`.
+	///
+	/// Mirrors [`Vec::spare_capacity_mut`]: used to write into a freshly allocated buffer in place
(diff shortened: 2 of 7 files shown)


jimpo:
This looks good. Now make another PR stacked on it that uses PoolVec for the FieldBuffers allocated through the protocol. I want to see feasibility of that before merging this.

benji-potamus:
On the FieldBuffer follow-up — I mapped the whole live path first, and there's a scope finding worth a quick decision before I write it, because it's much bigger and differently-shaped than the Word-buffer step.

**Mechanically it works**, with one enabling change: `PoolVec`/`alloc_vec`'s `T: Pod` bound has to relax (I used `T: Copy`). The `PackedField` trait only guarantees `Copy + Zeroable`, **not** `Pod` (only the concrete `PackedPrimitiveType` is `Pod`), so generic `P: PackedField` code can't call `alloc_vec::<P>` under `Pod`. `FieldBuffer<P, PoolVec<'alloc, P>>` is otherwise fine — `PoolVec: Deref<Target=[P]>` already satisfies `FieldBuffer`'s `Data` bound. That relaxation is done + green.

**The catch: it's all-or-nothing, not incremental.** There's no bounded slice to pool. Every *owned* protocol `FieldBuffer` flows into one of two big connected components, so `'alloc` propagates through ~15–20 structs + the 11 `Box<dyn RoundEvaluator>` sites:
- **Sumcheck web:** `FieldBuffer` → `Column`/`Gruen32` → `MleStore<'a, P>` (becomes two-lifetime `MleStore<'a, 'alloc, P>`) → `Shared{Sumcheck,MleCheck}Prover` → every evaluator → `ProdcheckProver`/`FracAddCheckProver` → `IntMulWitness` → `prove_intmul_reduction`.
- **BaseFold:** even `witness_packed` is *moved by value* into `channel.prove_oracle_relations`, so it pulls `'alloc` through `CommittedOracleData`/`QueuedRelation` → the channel.

**Walls (all surmountable, but they shape the design):**
1. **Growth** — `BinarySwitchover.tensor` grows via `tensor_prod_eq_ind_prepend` and genuinely needs a growable `Vec`; it can't be a fixed `PoolVec` block. Isolated to one struct → the realistic design is *hybrid* (pool fixed-size buffers, leave the switchover tensor `Vec`).
2. **Math-internal allocs** — `eq_ind_partial_eval` / `tensor_prod_eq_ind` / `fold_highest_var` allocate `Vec` inside `binius-math`, which can't depend on `binius-compute`. Pooling their outputs needs either pool-aware math variants or leaving those `Vec`.
3. **`!Sync`** is *not* a problem for this spike — the pass-through pool is a ZST and `PoolVec` wraps `Vec`, so it stays `Send+Sync`. It only bites once the real `RefCell` pool lands (then ~3 rayon closures share `&FieldBuffer` and ~4 sites allocate inside parallel regions).

So the feasibility answer is **yes, as a hybrid**, but the conversion is one large threading pass, not a stack of small ones. How would you like it?
- **(a)** Grind the full hybrid threading as one big stacked PR (~20 structs + the Box sites).
- **(b)** Thread just the sumcheck/`MleStore` subtree first (proves the two-lifetime hub + the `Box<dyn> + 'alloc` evaluators actually compile — the riskiest part), defer BaseFold to a second PR.
- **(c)** Treat this analysis as the go/no-go answer and decide direction before I write the bulk.

I'd lean **(b)** — it de-risks the scary part (two-lifetime GAT-ish `MleStore` + boxed evaluators) in a reviewable chunk. The `Copy`-bound enabling change is ready to push as the base of whichever you pick.

jimpo:
1. Do BINIUS-327 as a standalone PR. That will remove the last place that "grow" is used.
2. Why can't `binius-math` depend on `binius-compute`? Is it just the `BufferData` trait? If so, just move that trait to `binius-compute`. It's fine for `binius-math` to depend on `compute` for now.
3. I think I'm fine with using a Mutex for the pool, so Sync won't be a problem.

And sure, if we can do (b), then I feel confident the rest of the refactor will go through.

benji-potamus:
Sounds good, thanks. Plan:
1. BINIUS-327 standalone first (removes `tensor_prod_eq_ind_prepend` + the now-dead `grow_packed_backing` helper — the last "grow" site).
2. Then option (b): move `BufferData` into `binius-compute` and let `binius-math` depend on `compute`, pool the sumcheck/`MleStore` subtree, assume a `Mutex` pool so `Sync` is a non-issue.

Starting on 327 now.

benji-potamus:
Feasibility update on the (b) sumcheck/MleStore threading — the risky part is resolved.

**All the structurally-risky pieces now compile** (type-check clean), using a `pool: &'alloc BufferPool`-as-a-field pattern so methods pool without changing their signatures:
- `MleStore<'a, 'alloc, P>` — the two-lifetime hub, with `Column::{Owned,SplitHalf}` as `FieldBuffer<P, PoolVec<'alloc, P>>`. Kept `Gruen32`/eq-trackers on `Vec` (hybrid) so `'alloc` enters only via columns.
- `SharedSumcheckProver<'a, 'alloc, P, Ev>` and `SharedMleCheckProver<'a, 'alloc, F, P, Ev>`, incl. `into_shared_sumcheck` returning `Box<dyn SumcheckRoundEvaluator<F, P> + 'alloc>` (the boxed-evaluator + lifetime combo — no `implementation is not general enough` issues).
- `ProdcheckProver<'alloc, P>` and `FracAddCheckProver<'alloc, P>` (each with a pool field + pooled layer buffers).

Enablers landed: `PoolVec`/`alloc_vec` bound relaxed `Pod → Copy`; `BufferData` moved into `binius-compute` with `binius-math` depending on it and `PoolVec: BufferData`; pass-through pool stays `Send+Sync` so `Sync` is a non-issue (your Mutex plan keeps it that way).

**What's left is purely mechanical, zero type-system risk:** propagating the `pool` arg through the remaining caller layer — `logup_star::{prove,witness}` (constructs the GKR provers with pooled witnesses), the two selector fns, `final_layer`, then `crates/prover` (`IntMulWitness` + `prove_intmul_reduction` + the `and_reduction` bitand path) — all bottoming out at `IOPProver::prove`, which already holds `&pool`. The pattern is uniform: each store/prover-creating fn takes `pool`; `from_values`/`collect` buffers are copied into `pool.alloc_vec`. Currently ~14 caller-plumbing errors, no type errors.

Since your bar was "if (b) [the two-lifetime hub + boxed evaluators] compiles, the rest goes through" — that's demonstrated. **Is this enough feasibility signal to merge #1864, or would you like me to grind the mechanical plumbing to a fully-compiling + tested branch first?** I can do either; just checking before spending the (sizeable, but risk-free) remaining plumbing effort.

jimpo (approved):

PR #1865, opened 2026-07-20
BINIUS-96: define field structs as wrappers over PackedPrimitiveType

tcoratger:
@jimpo Let me know what you think about this. I feel like we are maybe over-using the macros in the field crate, rendering the crate pretty complex, maybe we should check in the future to limit that, but for now this looks like a good simplification.

Closes [BINIUS-96](https://linear.app/jimpo/issue/BINIUS-96).

Makes each primitive field a thin newtype over its own width-one packing and inherits the arithmetic from it. Pure refactor — no behavior change.

### The problem

Every field is already a width-one packed field, but that relationship was implicit.

- A field like `BinaryField128bGhash` was a newtype over its raw underlier `M128`.
- Its width-one `PackedField`/`Divisible`/`Maskable` plumbing was hand-rolled in the `binary_field!` macro — a near-duplicate of the generic impls already on `PackedPrimitiveType`.
- Its multiply was wired up separately from the packing that already knows how to multiply: GHASH routed `WideMul` through a `GhashWideMul1x<PackedBinaryGhash1x128b>` wrap/peel dance, AES went through `impl_arithmetic_using_packed`, and so on.

So the same field lived in two places: the scalar, and its width-one packing `PackedPrimitiveType<U, Self>`.

### The idea

Make the field *be* a newtype over that packing, and let the packing carry the arithmetic:

```
struct BinaryField128bGhash(PackedPrimitiveType<M128, Self>);
```

Then the scalar's `Mul`/`Square`/`InvertOrZero` are one-liners that defer to the inner packing — the packing is the single source of truth.

### The change

The `binary_field!` macro is split into two representation arms over a shared `@common` arm:

- `@repr_wrapped` (default) — newtype over `PackedPrimitiveType<$typ, Self>`; `Mul`/`Square`/`InvertOrZero` delegate to it.
- `@repr_direct` — newtype over the underlier itself.
- `@common` — every representation-agnostic impl (`Add`/`Sub`/`Neg`, `Sum`/`Product`, `Field`, `Divisible`, `Maskable`, `PackedField`, `Display`/`Debug`/`Hash`/`Ord`, `From`, `Distribution`), written through `val`/`from_raw` so one body serves either form.

At a call site the arithmetic collapses to a delegation:

```diff
-impl WideMul for BinaryField128bGhash {
-    type Output = <GhashWideMul1x<PackedBinaryGhash1x128b> as WideMul>::Output;
-    fn wide_mul(a, b) { /* from_underlier + GhashWideMul1x::wrap + wide_mul */ }
-    fn reduce(w)      { /* reduce + GhashWideMul1x::peel + from_underlier */ }
-}
+impl WideMul for BinaryField128bGhash {
+    type Output = <PackedBinaryGhash1x128b as WideMul>::Output;
+    fn wide_mul(a, b) { PackedBinaryGhash1x128b::wide_mul(a.0, b.0) }
+    fn reduce(w)      { Self(PackedBinaryGhash1x128b::reduce(w)) }
+}
```

This lets the now-redundant `binary_field_arithmetic` module (the `impl_arithmetic_using_packed` macro and its `*_using_packed` helpers) be deleted outright.

### Why GhashSq256b stays direct

`GhashSq256b` is a degree-two extension backed by `M256`, and `M256` has no reflexive `Divisible<M256>` (only `Divisible<M128>`, `Divisible<u128>`, …), so there is no `PackedPrimitiveType<M256, GhashSq256b>` packing to wrap. It keeps the `@repr_direct` form and its own bespoke Karatsuba arithmetic. Wrapping it would mean either adding a reflexive `Divisible` to the 256/512-bit SIMD underliers or reimplementing the plumbing anyway — neither is warranted here.

### Soundness

Not applicable in the usual sense — this is a representation refactor.

- The field types, their values, and their public API (`new`/`from`/`val`, all trait impls) are unchanged.
- Multiply tracing is preserved: the wrapped `Mul` still emits `trace_multiplication!($name)` before delegating, and the re-entrancy guard suppresses the inner packing's trace, so the emitted label and count are identical.

### Scope

- **changed:** `crates/field/src/{binary_field,ghash,aes_field,ghash_sq,lib}.rs`.
- **deleted:** `crates/field/src/binary_field_arithmetic.rs` (its arithmetic is now inherited from the packings; its helpers had no other users).
- **left untouched:** `GhashSq256b`'s representation and arithmetic; the `PackedPrimitiveType` packings themselves; the external field API.

### Verification

- `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets`, nightly `cargo fmt --all` — clean.
- `binius-field`: 208 tests + doctests pass (the field-arithmetic proptests — mul/square/invert/wide_mul, multiplicative generators, serialization — cover all four fields).


jimpo:
Seems like a good approach. I wonder though if we could instead have the fields be defined as they are now, over an Underlier, but simply always use the PackedField logic (for `PackedPrimitiveType<U, Self>`) to implement all arithmetic.

Could you try that version as well?

tcoratger:
> Seems like a good approach. I wonder though if we could instead have the fields be defined as they are now, over an Underlier, but simply always use the PackedField logic (for `PackedPrimitiveType<U, Self>`) to implement all arithmetic.
> 
> Could you try that version as well?

I've tried it, can you check the latest commit to see if this fits what you had in mind. If not I can totally revert.

jimpo (approved):
Yeah, this looks good.

I think we can go a step further and make `impl_arithmetic_via_packed!` part of the `binary_field!` macro -- basically, this is just how all binary field arithmetic is defined. Leaving as a follow-up.

jimpo:
Could you please resolve conflicts?

tcoratger:
> Could you please resolve conflicts?

Done

PR #1867, opened 2026-07-20
BINIUS-328: implement a persistent free list for BufferPool

benji-potamus:
Implements [BINIUS-328](https://linear.app/jimpo/issue/BINIUS-328/use-persistent-allocation-pool-for-bufferpool). Follows up [#1864](https://github.com/binius-zk/binius64/pull/1864), which introduced `BufferPool` as a pass-through over `Vec::with_capacity`.

## What

`BufferPool` now recycles freed allocations instead of always going to the global allocator.

Every pooled block is a `Vec<AlignedChunk>`, where `AlignedChunk` is `#[repr(align(64))] [u8; 64]`. This forces every backing allocation to a fixed 64-byte alignment — wide enough for any element type the prover uses — so the free list keys on **byte size alone**: `Mutex<HashMap<n_chunks, Vec<Vec<AlignedChunk>>>>`.

- **`alloc_vec::<T>(capacity)`** rounds the byte length (`capacity * size_of::<T>`) up to a power of two (min one 64-byte chunk), pops a block of that size from the free list or allocates a fresh `Vec<AlignedChunk>`, and hands it out as a `Vec<T>` via `Vec::from_raw_parts`. Because the alignment is uniform, a block freed by one element type backs a `PoolVec` of any other type of the same byte size.
- **`PoolVec::drop`** checks the buffer is still 64-aligned (i.e. the `Vec` didn't outgrow its block and reallocate into element-aligned storage), then rebuilds the owning `Vec<AlignedChunk>` and parks it on the free list. Every allocation and free therefore uses the same layout, and the pool frees blocks through ordinary `Vec` machinery — no manual `dealloc`, no `unsafe impl Send`.
- A `const` assertion rejects any `T` whose alignment exceeds 64 or whose size is not a power of two at compile time (both would break byte-size-keyed reuse).

## Testing

- `cargo test -p binius-compute` — 7 tests (byte-size rounding + 64-alignment, recycle, cross-type reuse, distinct-size isolation, multi-block free list) + existing.
- `cargo miri test -p binius-compute` — clean, no UB, including the cross-type reuse path (a `u64×8` block reused as `u8×64`).
- `cargo test -p binius-prover --lib` — green (exercises the pool through the parallel witness builders).
- clippy (`-D warnings`) + nightly fmt + `cargo doc` clean.

### DIFF
--- a/crates/compute/src/lib.rs
+++ b/crates/compute/src/lib.rs
@@ -3,49 +3,129 @@
 //! Buffer pooling for prover working memory.
 //!
 //! The prover allocates many large, short-lived buffers. [`BufferPool`] is the seam through which
-//! those allocations flow, so they can later be served from a recycling free list instead of the
-//! global allocator. Buffers are handed out as [`PoolVec`] handles that borrow the pool for
-//! `'alloc` and, in a future revision, return their block to it on drop.
+//! those allocations flow, recycling freed blocks instead of returning them to the global
+//! allocator. Buffers are handed out as [`PoolVec`] handles that borrow the pool for `'alloc` and
+//! return their block to it on drop.
 //!
-//! The current implementation is a thin pass-through over [`Vec`]: [`BufferPool::alloc_vec`] just
-//! calls [`Vec::with_capacity`]. This establishes the API and the `'alloc` lifetime threading
-//! without committing to a pool layout yet.
+//! Every pooled block is allocated with a fixed `BUFFER_ALIGN`-byte alignment — wide enough for any
+//! element type the prover uses — and sized to a power-of-two number of bytes. Because the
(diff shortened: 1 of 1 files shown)


crates/compute/src/lib.rs
```diff
@@ -3,50 +3,130 @@
...
 			pool: self,
-			data: Vec::with_capacity(capacity),
+			data: self.alloc_data(capacity),
 		}
 	}
+
+	fn alloc_data<T: Pod>(&self, capacity: usize) -> Vec<T> {
+		const {
+			assert!(
+				mem::align_of::<T>() <= BUFFER_ALIGN,
```
jimpo, line 81: it needs to divide BUFFER_ALIGN

jimpo (approved):

PR #1871, opened 2026-07-21
Benchmark M4 operation witness assembly

latifkasuli:
## Summary
- Replaces the circuit-fixture benchmark with a targeted one-column benchmark for `OperandColumns::build`, the current replacement for the former `build_operation_witness` helper.
- Generates deterministic random `Operand`s directly, varying the number of shifted value indices, value indices, shift amounts, and shift variants.
- Measures Criterion throughput in shifted value indices, i.e. the total non-zero entries across all generated operands.
- Uses the current pooled-buffer path via `BufferPool`, matching the prover allocation strategy.

This keeps setup minimal: a witness-only `ValueTable` provides rows for the generated operands to read, while benchmark-local `ZeroConstraint`s provide the one-operand, arity-preserving input expected by the current `OperandColumns` API. No operation circuit constraint systems are built for the benchmark cases.

## Benchmark
Short local pass on Apple M4 Pro, `RAYON_NUM_THREADS=8`, `--sample-size 10 --warm-up-time 1 --measurement-time 3`, after rebasing onto current `main`:

| case | time | throughput |
|---|---:|---:|
| `sparse_0_to_2_terms` | 1.98 ms | 496 K shifted indices/s |
| `mixed_0_to_4_terms` | 3.90 ms | 517 K shifted indices/s |
| `dense_1_to_8_terms` | 8.97 ms | 528 K shifted indices/s |

## Test plan
- `cargo +nightly-2026-01-01 fmt --all --check`
- `cargo bench -p binius-m4-prover --bench assemble_operation_witness --no-run`
- `RAYON_NUM_THREADS=8 cargo bench -p binius-m4-prover --bench assemble_operation_witness -- --sample-size 10 --warm-up-time 1 --measurement-time 3`
- `cargo test -p binius-m4-prover --lib`
- `cargo test -p binius-m4-prover --lib --features rayon`
- `cargo clippy -p binius-m4-prover --all-features --tests --benches -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-m4-prover --no-deps`

### DIFF
--- a/crates/m4-prover/benches/assemble_operation_witness.rs
+++ b/crates/m4-prover/benches/assemble_operation_witness.rs
@@ -1,94 +1,207 @@
 // Copyright 2026 The Binius Developers
-//! Benchmark for assembling a batched per-operation witness — the operand-column layout an
-//! operation reduction consumes.
+//! Benchmark for assembling one batched operation-witness column from sparse operands.
 //!
-//! Currently covers the BitAnd operation's two columns; the IntMul
-//! equivalent will be added alongside its protocol. Uses the Keccak-f1600 permutation circuit
-//! (see the `keccak_witness_gen` bench) as a realistic, AND-heavy constraint system: the circuit
-//! applies one permutation to a 25-word state, the state words are witness inputs and the permuted
-//! outputs are force-committed. Populating the batch table and preparing the constants/constraints
-//! are done once as setup; only the witness assembly is timed, over 8192 instances.
+//! This targets [`OperandColumns::build`] directly. Setup constructs a minimal witness-only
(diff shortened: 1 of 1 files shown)


jimpo:
I don't like this approach of building whole constraint systems -- it's not necessary. We should just make a more targeted benchmark for `build_operation_witness`. Have some code for generating random Operands (ideally randomizing the # of shifted indices, the shift amounts, shift variants), then building a column and timing that. Throughput should be measured in terms of the total number of shifted value indices in all operands (ie. number of non-zero entries in the sparse matrix).

jimpo (approved):

jimpo:
Sorry for the delay @latifkasuli! I was away last week. Could you please rebase & resolve conflicts?

PR #1878, opened 2026-07-21
BINIUS-113: remove test/bench-only items from binius-field (group D)

tcoratger:
Part of [BINIUS-113](https://linear.app/jimpo/issue/BINIUS-113). One of several cleanup sub-PRs — does **not** close that umbrella investigation.

Removes items in `binius-field` that live in production code but are exercised only by tests or benches. Pure removal — no behavior change.

### The problem

The audit found a cluster of `pub` items that no production code path ever reaches — they are constructed or called only inside `#[cfg(test)]` modules or a benchmark. Test-only production API is a maintenance tax: it shows up in docs and autocomplete, and it hides the fact that the machinery has no real caller.

The rule applied here: **delete anything used only by its own test/bench; keep a helper only when it is genuinely needed to test something else, and then only in test-gated form.**

### Deleted outright

- **`PackedSlice` / `PackedSliceMut`** (+ `new` / `new_with_len` and their `RandomAccessSequence[Mut]` impls) — constructed only under `#[cfg(test)]`. With them go `len_packed_slice` (only those types used it) and the tests/helpers that only exercised them (`check_packed_slice`, `check_packed_slice_mut`, `check_collection`, `check_collection_get_set`).
- **`iter_packed_slice_with_offset`** — no production caller. Its bench also timed the production `iter_slice`, so the bench is **trimmed** (keeps the `iter_begin` / `iter_slice` case, drops the offset case) rather than deleted.
- **`BinaryField128bGhash::mul_x`** + its self-test `test_mul_x`. The sibling `mul_inv_x` is used in production and stays.
- **`BinaryField1b::new_unchecked`** + its self-test `test_gf2_new_unchecked`.

### Kept, with justification

- **`single_element_mask_bits`** is *not* test-only-for-itself: it is a shared helper for the **M256 / M512 spread tests** (i.e. it tests something else), and it is already fully `#[cfg(test)]`-gated, so it never ships in a production binary. A `cfg(test) pub(crate)` fn is the right form for a helper shared across two files' test modules — deleting it would break those tests, and inlining would duplicate the mask logic into both.
- **`SlicedGhashSq1x256b` / `SlicedGhashSq4x256b`** — kept per review. These public aliases stay, and the width-1/2/4 tests use them as before.

### Scope

- **removed:** the `PackedSlice` cluster + `len_packed_slice`, `iter_packed_slice_with_offset`, `mul_x`, `BinaryField1b::new_unchecked`, plus the 4 self-tests and freed imports
- **kept:** `mul_inv_x`, all three `SlicedGhashSq{1,2,4}x256b` aliases, `single_element_mask_bits`, and full width-1/2/4 ghashsq test coverage

### Verification

- `cargo check -p binius-field --all-targets`, `cargo clippy -p binius-field --all-targets` — clean, zero warnings
- `cargo test -p binius-field` — 204 passed (208 minus the 4 deleted self-tests)
- `cargo check -p binius-field --target x86_64-unknown-linux-gnu --lib` with `-D warnings` (`+sse2,+avx2,+avx512f,+gfni`) — clean
- `cargo check` on the touched-trait consumers (`binius-math`, `binius-prover`, `binius-verifier`, `binius-hash`, `binius-iop`, `binius-ip`, `binius-transcript`, `binius-core`) — clean
- nightly `fmt` — applied


crates/field/src/packed_ghash_sq.rs
```diff
@@ -32,12 +32,8 @@ type Ghash<U> = PackedPrimitiveType<U, BinaryField128bGhash>;
 
 /// A GHASH² packing whose two GHASH coordinates pack into `PackedPrimitiveType<U, Ghash128b>`.
 pub type SlicedGhashSq256b<U> = SlicedPackedField<GhashSq256b, Ghash<U>, 2>;
-/// Packed `GhashSq256b` holding one extension scalar (the degenerate width-one packing).
-pub type SlicedGhashSq1x256b = SlicedGhashSq256b<M128>;
```
jimpo, line 36: Please leave these type aliases

tcoratger: Done

jimpo (approved):

tcoratger:
@jimpo just pushed a new commit to solve the problems with main

PR #1882, opened 2026-07-21
BINIUS-329: derive the AND-reduction C operand on the fly instead of materializing it

tcoratger:
Closes [BINIUS-329](https://linear.app/jimpo/issue/BINIUS-329/stop-materializing-the-and-reduction-c-operand-derive-it-as-in).

Stops building and streaming the BitAnd C operand column; the prover derives `C = A & B` in registers wherever it is needed. Transcript-identical — proof bytes are unchanged.

### The problem

The BitAnd zerocheck proves `A & B == C` over one row per AND constraint.
The prover materialized all three operand columns as full word buffers, 8 bytes per row each.
The monolithic prover built C by evaluating its operand — a XOR of shifted witness words — for every constraint.
The univariate fold then streamed all three buffers through memory.
The prover is memory-bandwidth-bound on Apple Silicon, so every full-buffer build and stream is wall-clock time.

### The idea

On a satisfying witness, `C = A & B` holds word-by-word — that is the constraint itself.
So the C column carries no information: any consumer holding the A and B words recreates the C word with one register AND.
The round-1 univariate message already derived C this way internally; this PR completes the picture.

```
    stream A ──┬──> fold ──> folded A
    stream B ──┼──> fold ──> folded B
               └──> A & B ──> fold ──> folded C   (no third input stream)
```

### Per proof

- **before:** build C (one operand evaluation per constraint) + hold a third column + stream it in the fold → 3 input streams
- **after:** no C build, no C buffer, 2 input streams + one register AND per word pair

→ one full build pass deleted, fold traffic cut from 3 columns to 2, peak memory down by 8 bytes per constraint row.

### The change

```diff
-		let proving_polys = [&self.first_col, &self.second_col, &self.third_col]
-			.map(|col| folder.fold::<PChallenge>(col));
+		let proving_polys =
+			folder.fold_bitand_operands::<PChallenge>(&self.first_col, &self.second_col);
```

- `fold_word.rs`: new fused three-output fold over the two stored columns.
- `and_reduction`: the prover takes only A and B; the round-1 message drops its assert-only C parameter.
- monolithic `prove.rs`: the witness builder no longer evaluates the C operand.
- `m4-prover`: the batch witness holds only A and B; the IMUL/BMUL re-randomization path (the one true C consumer) re-derives the column locally.

### Soundness

- Folding is $\mathbb{F}_2$-linear on word bits, so equal words fold to equal field elements.
- An honest prover emits the exact same transcript: verified byte-identical proofs on a deterministic SHA-256 fixture, before vs after (same digest, 35,936 bytes).
- A cheating witness is still rejected: the shift reduction binds the claimed C evaluation to the committed witness. The verifier is untouched.
- This is *not* Flock-style protocol "skipping c" (paper section 4.4), which changes the transcript — nothing about the protocol moves here.

### Scope

- **new:** fused fold kernel + a test pinning it to three independent folds across all length regimes (empty, sub-width, aligned, tail, non-power-of-two)
- **changed:** the AND-reduction prover signature, both witness builders, one test helper, the `and_reduction` bench (now measures the fused path)
- **untouched:** verifier, transcript, IntMul/BinMul reductions, shift argument

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-prover -p binius-m4-prover --all-targets`, nightly `fmt` — clean
- `binius-prover` lib (25) + `prove_verify` (9) + `shift` (1); `binius-m4-prover` (34 + 3) — all pass
- transcript pin: byte-identical proof digest against the pre-change code (stash A/B in the same tree)

### Benchmark

Isolated micro benches, M1 Pro, `--features rayon`, back-to-back stash A/B:

| bench | before | after | speedup |
|---|---|---|---|
| univariate fold, 2^21 words | 3.84 ms | 3.14 ms | 1.22× |
| m4 BitAnd witness assembly (keccak-f1600) | 5.01 ms | 2.90 ms | 1.73× |

The fold row understates the win: its baseline folds a pre-materialized C, so the deleted C build pass is on top of the 1.22×.

### DIFF
--- a/crates/m4-prover/src/operand_witness.rs
+++ b/crates/m4-prover/src/operand_witness.rs
@@ -2,11 +2,12 @@
 
 //! The batched operand-column witnesses built from a populated batch value table.
 //!
-//! Every AND, IMUL, and BMUL constraint is projected to its fixed-arity operand columns (`[A, B,
-//! C]` for AND, `[A, B, HI, LO]` for IMUL, `[A_LO, A_HI, B_LO, B_HI, C_LO, C_HI]` for BMUL), one
+//! Every AND, IMUL, and BMUL constraint is projected to its fixed-arity operand columns (`[A, B]`
+//! for AND, `[A, B, HI, LO]` for IMUL, `[A_LO, A_HI, B_LO, B_HI, C_LO, C_HI]` for BMUL), one
 //! column per operand, stacked over every instance in the batch. [`build_operation_witness`] is the
 //! shared arity-generic core: it projects one operand per constraint into one column.
-//! [`BatchAndCheckWitness`] builds the three AND columns and drives the AND-check zerocheck;
+//! [`BatchAndCheckWitness`] builds the two AND columns and drives the AND-check zerocheck.
+//! The AND check's C column is never materialized: the reduction derives `C = A & B` on the fly.
 //! [`build_intmul_witness`] builds the four IntMul columns; [`build_binmul_witness`] builds the six
 //! BinMul columns.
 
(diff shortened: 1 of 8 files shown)


jimpo (approved):
Amazing!!

PR #1890, opened 2026-07-22
BINIUS-334: fuse the RS coset replication into the first executed NTT layer

tcoratger:
Closes [BINIUS-334](https://linear.app/jimpo/issue/BINIUS-334/fuse-the-reed-solomon-coset-replication-into-the-first-executed-ntt).

Makes the Reed-Solomon encoder write the codeword once instead of twice: the coset replication and the first executed NTT layer become a single seeded pass. Byte-identical output — no transcript or protocol change.

### The problem

The encoder wrote the codeword buffer twice.
A replicate pass copied the bit-reversed message into every coset block.
Then the NTT swept the full buffer in place, with the early layers skipped.
Its first executed layer re-read exactly the bytes the replicate pass had just written.

### The idea

The early-layer skip works because a butterfly against a zero half copies the live half into both outputs.
So every coset block entering the first executed layer holds identical data — the message.
That layer's butterflies can read the single message copy and write each coset's post-layer state directly:

```
    msg low half  a ──┬──> u_c = a + t_c * b     (t_c = the layer twiddle of coset c)
    msg high half b ──┴──> v_c = u_c + b

    replicate-then-transform:  write 2^n, then read 2^n + write 2^n for the layer
    fused seed:                read 2^m once, write 2^n
```

The repeated-message intermediate is never materialized.
The transform resumes with one extra early layer skipped.

### The change

- `reed_solomon.rs`: a new seed pass runs after the message bit-reversal. Each parallel task owns two L1-resident source chunks of coset 0 and the matching destination chunks of every other coset (written through `MaybeUninit` into the vector's spare capacity); coset 0's own butterfly runs in place, last. The main transform call moves from `skip_early = log_inv_rate` to `skip_early = log_inv_rate + 1`.
- Edge shapes keep the plain replicate path: rate 1 (nothing to seed), dimension 1 (no executable layer), messages at or below the packing width.

### Soundness

- The seed performs the exact layer butterflies (`u += v * t; v += u`) with the exact per-coset twiddles the transform would have used, on the same values (the replicas were copies) — so the output bytes are unchanged.
- Pinned three ways: the existing zero-padded reference oracle, a new agreement test across the reference / single-thread / multi-thread NTT drivers, and byte-identical proofs on a deterministic SHA-256 fixture (same digest before vs after, via stash A/B).

### Scope

- **changed:** one function in the encoder plus the new private seed pass
- **new:** driver-agreement test, five boundary-shape tests
- **untouched:** the NTT implementations, FRI, the verifier

### Verification

- `cargo check --workspace --all-targets`, `clippy` (0 warnings), nightly `fmt` — clean
- `binius-math` (127 + 2), `binius-iop-prover` (40), `prove_verify` (9), `binius-m4-prover` (35 + 3) — all pass
- transcript pin: byte-identical proof digest against the pre-change code

### Benchmark

Measured with the reference harness below (not part of the diff), production NTT config (multithread, pre-expanded twiddles), M1 Pro, stash A/B:

| shape | config | before | after | delta |
|---|---|---|---|---|
| log_dim 20, rate 1/2, batch 4 (128 MB codeword) | 8 threads | 35.96 ms | 34.50 ms | +4.1% |
| same | 1 thread | 165.6 ms | 162.8 ms | +1.7% |
| log_dim 16, rate 1/4 (16 MB) | 8 threads | 1.337 ms | 1.109 ms | +20.6% |
| same | 1 thread | 4.046 ms | 4.026 ms | wash |

Honest framing: single-threaded encode is compute-bound — the seed performs the same field multiplies as the layer it replaces — so the gain is purely memory-side. It shows under multi-thread load, and most strongly when the codeword is cache-resident (the regime of smaller per-oracle encodes). The change is strictly less work and byte-identical, but a modest win at commit-sized codewords.

<details>
<summary>Reference bench harness (drop into <code>crates/math/benches/reed_solomon.rs</code> and register in <code>Cargo.toml</code>)</summary>

```rust
use binius_field::PackedBinaryGhash1x128b;
use binius_math::{
	ReedSolomonCode,
	ntt::{NeighborsLastMultiThread, domain_context::GenericPreExpanded},
	test_utils::random_field_buffer,
};

fn bench_encode_batch(c: &mut Criterion) {
	type P = PackedBinaryGhash1x128b;

	let mut rng = rand::rng();
	let mut group = c.benchmark_group("rs_encode_batch");
	group.sample_size(20);

	for (log_dim, log_inv_rate, log_batch_size) in [(20, 1, 2), (16, 2, 0)] {
		let rs_code = ReedSolomonCode::new(log_dim, log_inv_rate);
		let subspace = rs_code.subspace().clone();
		let domain_context = GenericPreExpanded::generate_from_subspace(&subspace);
		let log_num_shares = binius_utils::rayon::current_num_threads().ilog2() as usize;
		let ntt = NeighborsLastMultiThread::new(domain_context, log_num_shares);

		let message = random_field_buffer::<P>(&mut rng, log_dim + log_batch_size);

		let log_codeword_len = log_dim + log_batch_size + log_inv_rate;
		group.throughput(Throughput::Bytes(16 << log_codeword_len));
		group.bench_function(
			BenchmarkId::from_parameter(format!(
				"log_dim={log_dim}/log_inv_rate={log_inv_rate}/log_batch={log_batch_size}"
			)),
			|b| b.iter(|| rs_code.encode_batch(&ntt, message.to_ref(), log_batch_size)),
		);
	}
	group.finish();
}

criterion_group!(benches, bench_encode_batch);
criterion_main!(benches);
```

Run with `cargo bench -p binius-math --features rayon --bench reed_solomon` (drop the feature for the single-threaded numbers).

</details>

### DIFF
--- a/crates/math/src/reed_solomon.rs
+++ b/crates/math/src/reed_solomon.rs
@@ -5,7 +5,7 @@
 //!
 //! See [`ReedSolomonCode`] for details.
 
-use std::ptr;
+use std::{mem::MaybeUninit, ptr, slice::from_raw_parts_mut};
 
 use binius_field::{BinaryField, PackedField};
 use binius_utils::rayon::prelude::*;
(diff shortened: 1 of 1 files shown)


jimpo:
@tcoratger This doesn't seem like it should be that impactful, and it's quite a lot of code. I bet most of the speedup just comes from parallelizing the copying with Rayon (which now always is single-threaded)?

tcoratger:
> @tcoratger This doesn't seem like it should be that impactful, and it's quite a lot of code. I bet most of the speedup just comes from parallelizing the copying with Rayon (which now always is single-threaded)?

Yeah given the small speedup, let's close for now :)

PR #1881, opened 2026-07-21
BINIUS-218: parallelize the looker circuit build with a serial layer cutoff

tcoratger:
Part of [BINIUS-218](https://linear.app/jimpo/issue/BINIUS-218/logup-witness-inefficiencies) — this does not close it.

Builds the same logUp* fractional-addition circuits, but fans the per-looker builds across the thread pool and stops dispatching layers too small to pay for a parallel region.
Output is bit-identical — no protocol change, no new soundness assumption.

### The problem

Two dispatch problems compound in the circuit-build stage:

- The prover builds one fractional-addition circuit per looker (12 in the intmul caller), **one looker at a time**.
- Inside each build, **every** layer goes through the pool — including layers of a few dozen packed words.

One output word is three $\mathbb{F}_{2^{128}}$ multiplies, roughly 10ns.
A parallel region costs tens of microseconds to dispatch and join.
So with the `rayon` feature on, small builds were a net **loss** against single-threaded:

| `n_vars` (single circuit) | serial build | rayon build |
|---|---|---|
| 12 | 20 µs | 523 µs |
| 16 | 313 µs | 1.12 ms |
| 20 | 5.0 ms | 4.78 ms |

Only at $2^{20}$ did the pool break even.

### The idea

- Fan the circuit builds out across lookers — the circuits are independent, one task per looker.
- Below a cutoff of $2^{13}$ packed words, build the layer serially; above it, keep the parallel build.

The cutoff is where the ~10ns-per-word work overtakes the tens-of-microseconds dispatch, and the criterion sweep below confirms it.

### The change

Two spots:

```
crates/ip-prover/src/logup_star/prove.rs
-   serial: iter::zip(lookers, numerators).map(build circuit).unzip()
+   (lookers, numerators).into_par_iter().map(build circuit).unzip()   // order-preserving

crates/ip-prover/src/fracaddcheck.rs
+   const SERIAL_LAYER_BUILD_CUTOFF: usize = 1 << 13;
    per layer: below the cutoff -> serial izip build; otherwise -> the existing parallel build
```

### Soundness — preserved, for free

- Pure scheduling: the same values are computed by the same formula, only on different workers.
- The parallel collect preserves looker order, so circuit $j$ keeps its $\gamma^j$-scaled numerator.
- The fracaddcheck and logUp* round-trip tests replay the prover transcript through the real verifier, pinning bit-identity.

### Scope

[...]

PR #1891, opened 2026-07-22
BINIUS-335: write fracadd layers directly into the pooled buffers

tcoratger:
Closes [BINIUS-335](https://linear.app/jimpo/issue/BINIUS-335/write-fracadd-layers-directly-into-the-pooled-buffers).

Builds the same fractional-addition layers, but writes them once — straight into the pooled buffers — instead of collecting into temporaries and copying.

### The problem

Since the buffer pooling landed (#1869), the circuit build computes each layer into a pair of temporary heap `Vec`s and then `extend_from_slice`-copies them into pooled buffers.
Every layer pays twice for its memory: once to write the temporaries, once to memcpy them into the pool.

At the intmul regime (12 lookers, $2^{20}$ rows) that is ~200 MB of extra traffic per proof.
The warm circuit-build stage of phase 5 measures 62 ms, where it was 42 ms before pooling.

### The idea

Allocate the two pooled buffers first, then write the fractional additions directly into their spare capacity:

```
V_num[i] = a_0*b_1 + a_1*b_0
V_den[i] = b_0*b_1        // written once, into the pool allocation
```

### The shape matters — a negative result worth recording

The obvious flat implementation — one element-wise parallel zip over the uninitialized slots with `with_min_len` — **regresses badly**: +228% / +124% / +31% on the single-circuit bench at $2^{12}$/$2^{16}$/$2^{20}$, from poor codegen of the six-way element zip.

The winning shape is chunked: a parallel zip of `par_chunks` / `par_chunks_mut` at the existing 8192-word chunk bound, with a tight serial `izip!` loop inside each chunk.
This is also why #1768 (the pre-pooling ancestor of this change) measured as noise: the shape only pays for itself now that there is a real copy to delete.

- Same values, same formula, same per-element order — only where the results land changes.
- The `set_len` carries a written safety argument: the chunked zip writes every slot of both length-`out_len` spare prefixes exactly once.
- The fracaddcheck and logUp* round-trip tests replay the prover transcript through the real verifier, pinning bit-identity.

### Scope

- **changed:** the layer-build loop in `FracAddCheckProver::new`, plus two slice-trait imports
- the GKR rounds, the store, and every caller are untouched

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-ip-prover --all-targets` (with and without `rayon`), nightly `fmt` — clean
- `cargo test -p binius-ip-prover`: 58/58 pass, with and without `--features rayon`
- `cargo test -p binius-iop-prover logup_star` (committed pushforward, incl. the BaseFold round trip) — pass

### Benchmark

All numbers vs current main, medians on an M1 Pro.

**Single-circuit build** — `cargo bench -p binius-prover --features rayon --bench fracaddcheck -- new`:

| n_vars | main | this PR | change |
|---|---|---|---|
| 12 | 24.3 µs | 18.5 µs | −24% |
| 16 | 340 µs | 260 µs | −24% |
| 20 | 2.92 ms | 1.48 ms | **−49%** |

**Serial builds** (no `rayon` feature): 26.5 → 18.0 µs, 362 → 251 µs, 6.24 → 4.06 ms — the eliminated copy pays in both modes.

**intmul phase-5 spans** at $2^{20}$ multiplications (same tracing-forest harness as #1881, reproduced there):

| | main | this PR |
|---|---|---|
| Build fracadd circuits, warm | 61.8 ms | **19.5 ms (3.2×)** |
| Build fracadd circuits, cold pool | 249 ms | 181 ms |
| phase 5 total, warm | 248 ms | 204 ms |

Warm circuit build is now faster than it was even before pooling landed (42 ms).

### DIFF
--- a/crates/field/src/arch/aarch64/arithmetic/ghash.rs
+++ b/crates/field/src/arch/aarch64/arithmetic/ghash.rs
@@ -203,6 +203,9 @@ pub struct GhashClMulWideMul<T>(T);
 impl WideMul for GhashClMulWideMul<PackedPrimitiveType<M128, GhashB128>> {
 	type Output = WideGhashProduct;
 
+	// Why always inline: every packed multiply funnels through this wrapper.
+	// Left out of line, callers pay a function call and stack traffic per element.
+	#[inline(always)]
 	fn wide_mul(a: Self, b: Self) -> Self::Output {
 		WideGhashProduct::wide_mul(
 			PackedPrimitiveType::peel(Self::peel(a)),
@@ -210,6 +213,9 @@ impl WideMul for GhashClMulWideMul<PackedPrimitiveType<M128, GhashB128>> {
 		)
 	}
 
+	// Why always inline: the reduction is a handful of instructions.
+	// A call here costs more than the arithmetic it performs.
+	#[inline(always)]
 	fn reduce(wide: Self::Output) -> Self {
 		Self::wrap(PackedPrimitiveType::wrap(wide.reduce()))
 	}
--- a/crates/ip-prover/src/fracaddcheck.rs
+++ b/crates/ip-prover/src/fracaddcheck.rs
@@ -8,8 +8,9 @@ use binius_ip::{mlecheck, prodcheck::MultilinearEvalClaim, sumcheck::RoundCoeffs
 use binius_math::{
 	FieldBuffer, FieldVec, line::extrapolate_line_packed, multilinear::eq::eq_ind_partial_eval,
 };
-use binius_utils::rayon::iter::{
-	IndexedParallelIterator, IntoParallelIterator, IntoParallelRefMutIterator, ParallelIterator,
+use binius_utils::rayon::{
+	iter::{IntoParallelIterator, IntoParallelRefMutIterator, ParallelIterator},
+	slice::{ParallelSlice, ParallelSliceMut},
 };
 use itertools::izip;
 
@@ -113,19 +114,36 @@ where
 			let (num_0, num_1) = num.split_half_ref();
 			let (den_0, den_1) = den.split_half_ref();
 
-			// One packed word of the next layer from the sibling halves:
+			// One packed word of the next layer from the sibling halves, written straight into
+			// the pooled buffers:
 			//     a_0/b_0 + a_1/b_1 = (a_0*b_1 + a_1*b_0) / (b_0*b_1)
-			let (next_layer_num, next_layer_den) =
-				(num_0.as_ref(), den_0.as_ref(), num_1.as_ref(), den_1.as_ref())
-					.into_par_iter()
-					.with_min_len(LAYER_BUILD_MIN_CHUNK)
-					.map(|(&a_0, &b_0, &a_1, &b_1)| (a_0 * b_1 + a_1 * b_0, b_0 * b_1))
-					.collect::<(Vec<_>, Vec<_>)>();
-
-			let mut num_data = alloc.alloc::<P>(next_layer_num.len());
-			num_data.extend_from_slice(&next_layer_num);
-			let mut den_data = alloc.alloc::<P>(next_layer_den.len());
-			den_data.extend_from_slice(&next_layer_den);
+			// Chunks fan out across workers; each chunk is a tight serial loop.
+			let out_len = num_0.as_ref().len();
+			let mut num_data = alloc.alloc::<P>(out_len);
+			let mut den_data = alloc.alloc::<P>(out_len);
+			(
+				num_data.spare_capacity_mut()[..out_len].par_chunks_mut(LAYER_BUILD_MIN_CHUNK),
+				den_data.spare_capacity_mut()[..out_len].par_chunks_mut(LAYER_BUILD_MIN_CHUNK),
+				num_0.as_ref().par_chunks(LAYER_BUILD_MIN_CHUNK),
+				den_0.as_ref().par_chunks(LAYER_BUILD_MIN_CHUNK),
+				num_1.as_ref().par_chunks(LAYER_BUILD_MIN_CHUNK),
+				den_1.as_ref().par_chunks(LAYER_BUILD_MIN_CHUNK),
+			)
+				.into_par_iter()
+				.for_each(|(num_out, den_out, num_0, den_0, num_1, den_1)| {
+					for (num_slot, den_slot, &a_0, &b_0, &a_1, &b_1) in
+						izip!(num_out, den_out, num_0, den_0, num_1, den_1)
+					{
+						num_slot.write(a_0 * b_1 + a_1 * b_0);
+						den_slot.write(b_0 * b_1);
+					}
+				});
+			// Safety: the chunked zip writes every slot of both length-out_len spare prefixes
+			// exactly once.
+			unsafe {
+				num_data.set_len(out_len);
+				den_data.set_len(out_len);
+			}
 			let next_layer =
 				(FieldBuffer::new(num_log_len, num_data), FieldBuffer::new(den_log_len, den_data));
 
--- a/crates/ip-prover/src/fracaddcheck.rs
+++ b/crates/ip-prover/src/fracaddcheck.rs
@@ -8,9 +8,8 @@ use binius_ip::{mlecheck, prodcheck::MultilinearEvalClaim, sumcheck::RoundCoeffs
 use binius_math::{
 	FieldBuffer, FieldVec, line::extrapolate_line_packed, multilinear::eq::eq_ind_partial_eval,
 };
-use binius_utils::rayon::{
-	iter::{IntoParallelIterator, IntoParallelRefMutIterator, ParallelIterator},
-	slice::{ParallelSlice, ParallelSliceMut},
+use binius_utils::rayon::iter::{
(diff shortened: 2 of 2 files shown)


jimpo:
Sweet! The chunking is odd. What does this mean? "The obvious flat implementation — one element-wise parallel zip over the uninitialized slots with with_min_len — regresses badly: +228% / +124% / +31% on the single-circuit bench at 2 12 /$2^{16}$/$2^{20}$, from poor codegen of the six-way element zip."

Why is the generated code bad? Can we hint the inner loop into good form (with assertions outside or `assert_unchecked` inside?)

tcoratger:
> Sweet! The chunking is odd. What does this mean? "The obvious flat implementation — one element-wise parallel zip over the uninitialized slots with with_min_len — regresses badly: +228% / +124% / +31% on the single-circuit bench at 2 12 /$2^{16}$/$2^{20}$, from poor codegen of the six-way element zip."
> 
> Why is the generated code bad? Can we hint the inner loop into good form (with assertions outside or `assert_unchecked` inside?)

In this case I think that the chunking is the right approach and I had many speedups like this previously in Plonky3. The flat version makes rayon handle every single element, and that per-element bookkeeping thing is super costly. In this case (previous to the PR), rayon wrapped each word in its own hand-off, representing even a major cost when compared to the three multiples.

While with the chunking, we handle a lot of word in one go (8192), then a plain loop does the 8192 multiplies with no wrapping. For me that's why we have different speedups for each element count: +228% at 2^12, but only +31% at 2^20. When there's little real work (2^12) it's almost the entire runtime; when there's lots (2^20) the multiplies is more important.

For me the `assert_unchecked` would not improve the flat version because it only removes bounds checks in a normal loop. And we will still have the same problem. The only way I'm aware of to solve this is the chunking approach what do you think?

jimpo:
Hmm, something still doesn't make sense to me -- I want to investigate this further (not necessary because this one PR is bad, but as a pattern that will probably be replicated throughout the codebase, I want to find the cleanest one).

The "per-element overhead" shouldn't actually be a thing using `with_min_len`. I traced this through the Rayon code and tested it in https://github.com/binius-zk/binius64/pull/1886. Basically, when Rayon can't split a segment (in [bridge_producer_consumer](https://github.com/rayon-rs/rayon/blob/main/src/iter/plumbing/mod.rs#L432)) it calls `consume_iter`, which should do a normal serial iteration, which LLVM optimizes.

jimpo:
@tcoratger Locally I get equally good performance with 

```rust
			// One packed word of the next layer from the sibling halves, written straight into
			// the pooled buffers:
			//     a_0/b_0 + a_1/b_1 = (a_0*b_1 + a_1*b_0) / (b_0*b_1)
			// Chunks fan out across workers; each chunk is a tight serial loop.
			let out_len = num_0.as_ref().len();
			let mut num_data = alloc.alloc::<P>(out_len);
			let mut den_data = alloc.alloc::<P>(out_len);
			(
				num_data.spare_capacity_mut(),
				den_data.spare_capacity_mut(),
				num_0.as_ref(),
				den_0.as_ref(),
				num_1.as_ref(),
				den_1.as_ref(),
			)
				.into_par_iter()
				.with_min_len(LAYER_BUILD_MIN_CHUNK)
				.for_each(|(num_out, den_out, &num_0, &den_0, &num_1, &den_1)| {
					// {
					num_out.write(num_0 * den_1 + num_1 * den_0);
					den_out.write(den_0 * den_1);
				});
			// Safety: the chunked zip writes every slot of both length-out_len spare prefixes
			// exactly once.
			unsafe {
				num_data.set_len(out_len);
				den_data.set_len(out_len);
			}
```

tcoratger:
> @tcoratger Locally I get equally good performance with
> 
> ```rust
> 			// One packed word of the next layer from the sibling halves, written straight into
> 			// the pooled buffers:
> 			//     a_0/b_0 + a_1/b_1 = (a_0*b_1 + a_1*b_0) / (b_0*b_1)
> 			// Chunks fan out across workers; each chunk is a tight serial loop.
> 			let out_len = num_0.as_ref().len();
> 			let mut num_data = alloc.alloc::<P>(out_len);
> 			let mut den_data = alloc.alloc::<P>(out_len);
> 			(
> 				num_data.spare_capacity_mut(),
> 				den_data.spare_capacity_mut(),
> 				num_0.as_ref(),
> 				den_0.as_ref(),
> 				num_1.as_ref(),
> 				den_1.as_ref(),
> 			)
> 				.into_par_iter()
> 				.with_min_len(LAYER_BUILD_MIN_CHUNK)
> 				.for_each(|(num_out, den_out, &num_0, &den_0, &num_1, &den_1)| {
> 					// {
> 					num_out.write(num_0 * den_1 + num_1 * den_0);
> 					den_out.write(den_0 * den_1);
> 				});
> 			// Safety: the chunked zip writes every slot of both length-out_len spare prefixes
> 			// exactly once.
> 			unsafe {
> 				num_data.set_len(out_len);
> 				den_data.set_len(out_len);
> 			}
> ```

I think due to the `WIDTH` (1 for Apple hardware) the results can be different no? I've tested on my mac locally and I still have speedup for the chunked variant.

jimpo:
> I think due to the `WIDTH` (1 for Apple hardware) the results can be different no? I've tested on my mac locally and I still have speedup for the chunked variant.

Interesting... can you post the benchmark results comparing the two strategies? (with --features rayon of course)

tcoratger:
Benchmark results on my M1 (Apple, `WIDTH=1`), flat (`with_min_len`) vs chunked:

```
RUSTFLAGS="-C target-cpu=native" cargo bench -p binius-prover --features rayon \
  --bench fracaddcheck -- 'fracaddcheck/new'
```

| n_vars | flat (`with_min_len`) | chunked (this PR) |
|--------|----------------------|-------------------|
| 12     | 79.0 µs              | 18.6 µs           |
| 16     | 748 µs               | 265 µs            |
| 20     | 3.87 ms              | 1.49 ms           |

tcoratger:
Pushed the flat version + the actual fix. Recap of what we found, since the answer turned out to be somewhere neither of us was looking.

## You were right that `with_min_len` isn't the problem

Two things ruled out splitting entirely:

- Single-threaded (`RAYON_NUM_THREADS=1`), the flat/chunked gap is ~4.9x and **identical at every chunk size** from `2^6` to `2^19`. Flat sits at ~9.8 ms regardless; the knob does nothing.
- At `n_vars=12` a half-layer is 2048 words, below the 8192 floor, so nothing splits at all — and flat was still 4x slower.

Length asserts / `assert_unchecked` also changed nothing.

## It also wasn't the zip

I hand-wrote rayon's exact iterator structure — `reduce!` builds a balanced tree, then `.map(flatten)`, and the leaves delegate to `std::iter::Zip` — and timed it against `izip!`:

| shape | ns/word |
|---|---|
| `izip!` (left-linear chain) | 3.80 |
| balanced `std::iter::Zip` tree | 3.78 |
| balanced tree + `.map(flatten)` + `.for_each` | 3.78 |
| rayon's `(6-tuple).into_par_iter()` | **19.26** |

Writing rayon's own structure by hand is fast. So the arity, the nesting shape, and the flatten map are all innocent.

## The actual cause: a missing inline hint

Disassembling the real bench binary, per element:

| | instructions | `pmull` | calls |
|---|---|---|---|
| chunked | 50 | 18 | **0** |
| flat | 57 | 12 | **3** |

The three calls all went to `<GhashClMulWideMul<_> as WideMul>::reduce`, with operands spilled to `[sp]` and reloaded around each one. That impl in `crates/field/src/arch/aarch64/arithmetic/ghash.rs` is the only thing in the file **without** an inline attribute — `pmull`, `gf2_128_reduce`, `WideGhashProduct::{wide_mul, reduce}` and every operator impl already have one.

The chunked shape had enough inlining budget left that LLVM inlined it anyway. The flat shape did not. That's the whole difference — it was never about rayon.

## Benchmarks

`RUSTFLAGS="-C target-cpu=native" cargo bench -p binius-prover --features rayon --bench fracaddcheck -- 'fracaddcheck/new'`, M1 (`WIDTH=1`):

| variant | n_vars=12 | 16 | 20 |
|---|---|---|---|
| flat, as-is | 79.2 µs | 772 µs | 4.04 ms |
| **flat + `#[inline(always)]`** | **19.6 µs** | **278 µs** | **1.88 ms** |
| chunked, as-is | 18.6 µs | 265 µs | 1.49 ms |
| chunked + `#[inline(always)]` | 18.4 µs | 252 µs | 1.63 ms |

Adding the attribute cuts the flat build by **53-63%**, and helps the chunked shape too.

## What's in this PR now

The flat element-wise zip, as you wrote it, plus the two inline attributes.

One caveat worth deciding on: with the fix applied to both, chunked is still ahead by **9% / 11% / 17%** at `n_vars=12/16/20`. Its loop keeps post-increment addressing and `eor3`-fused reduction (18 `pmull` in 50 instructions) versus recomputed indexed offsets in the flat one (12 in 77). Happy to go back to chunked if that's worth keeping, but the flat version is clearly the cleaner pattern to replicate and the gap is now small.

Also worth a follow-up: the x86_64 sibling impl should be checked for the same missing attribute, since this is the entry point every packed GHASH multiply passes through.

Verified: `cargo test -p binius-ip-prover` 58/58 with and without `--features rayon`, `binius-field` 207/207, `logup_star` round-trip, clippy and nightly fmt clean.

jimpo (approved):

PR #1892, opened 2026-07-23
BINIUS-285: return batched sumcheck prover challenges in binding order

### LINKED ISSUE #1802: batch_prove reverses sumcheck challenge order but batch_verify does not (documented TODO asymmetry between batch and prove_single)
## Summary

The batched sumcheck prover reverses its challenge vector before returning, while the batched verifier returns challenges in the order produced. The two batch entry points therefore emit challenge vectors in opposite orders, and this divergence from the non-batched `prove_single` is flagged by an in-repo `TODO`. Any consumer that treats the prover's and verifier's `challenges` output as directly comparable will mismatch; today only an in-file test compensates by manually reversing.

## Affected files / symbols

**Prover reverses** — `crates/ip-prover/src/sumcheck/batch.rs`, in both `batch_prove` and `batch_prove_mle`:
```rust
// TODO: this differs from prove_single, which doesn't reverse.
// Reverse to match high-to-low folding order for evaluation points.
challenges.reverse();
```
(present in `batch_prove` and again verbatim in `batch_prove_mle`).

**Verifier does not reverse** — `crates/ip/src/sumcheck/batch.rs`, `batch_verify` / `batch_verify_mle` return the challenges from `sumcheck::verify` / `mlecheck::verify` unmodified. The high-to-low convention is documented as a *caller* responsibility on the returned field:
```rust
/// Note: reverse when folding high-to-low to obtain evaluation claim.
pub challenges: Vec<F>,
```

**Non-batched path is consistent with the verifier** — `crates/ip-prover/src/sumcheck/prove.rs` `prove_single` pushes challenges in-loop and returns them without reversing, deferring the reversal to the caller (docstring `NB: reverse when folding high-to-low ...`).

**Test compensates for the mismatch** — `crates/ip-prover/src/sumcheck/batch.rs` reverses the prover's batch challenges before asserting equality with the verifier's output:
```rust
prover_challenges.reverse();
assert_eq!(prover_challenges, sumcheck_output.challenges);
```

## Expected vs actual

- **Expected:** the batched prover and verifier agree on challenge ordering, and `batch_prove` follows the same convention as `prove_single` (reversal deferred to the caller), so downstream code has one consistent rule.
- **Actual:** `batch_prove`/`batch_prove_mle` reverse in-band; `batch_verify`/`prove_single` do not. The asymmetry is only reconciled by test-side reversal, and is explicitly marked `TODO`.

## Suggested fix

Pick one convention and apply it uniformly. Preferred: remove the in-band `challenges.reverse();` from `batch_prove`/`batch_prove_mle` so batch matches `prove_single` (reversal deferred to the caller / folding site), and drop the compensating `.reverse()` in the test. Alternatively, if in-band high-to-low ordering is desired, apply the same reversal in `batch_verify` (and `prove_single`) and update the field docs accordingly. Either way, resolve the `TODO` and remove the ordering divergence between the two batch entry points.

pauldelucia:
Closes #1802 ([BINIUS-285](https://linear.app/jimpo/issue/BINIUS-285)).

Removes the challenge-vector reversal from `batch_prove` and `batch_prove_mle`, resolving the in-repo TODO. Every sumcheck entry point now returns challenges in binding order. **Transcript bytes are unchanged** — the transcript, the folding, and the protocol are untouched; only the order of the returned `challenges` vector changes.

### The problem

The batched sumcheck provers reversed their challenge vector before returning:

```rust
// TODO: this differs from prove_single, which doesn't reverse.
// Reverse to match high-to-low folding order for evaluation points.
challenges.reverse();
```

while `prove_single` and `batch_verify` return binding order. The two batch entry points therefore emitted opposite orders, prover-side and verifier-side outputs of the same run were not directly comparable, and the in-file test compensated by manually reversing before asserting equality.

### The change

The direction the TODO points: make the batch provers match `prove_single`.

- `batch_prove` / `batch_prove_mle` no longer reverse; the `BatchSumcheckOutput::challenges` doc now states the order.
- The six prover-side consumers that want a variable-indexed (low-to-high) point reverse locally: the basefold prover channel, the logup* final layer, the fracaddcheck layer loop, intmul phases 3 and 5, and the m4 re-randomization. This mirrors the verifier side, where every `batch_verify` caller already reverses locally (`intmul/verify.rs`, `m4-verifier/verify.rs`, `ip/logup_star`, `iop/basefold`, `ip/fracaddcheck`) — the two sides now use the same idiom.
- The compensating reversals in the `batch.rs`, `round_evaluator.rs`, and `frac_add_mle.rs` tests become direct prover-vs-verifier equality assertions.

### Compatibility

This changes the meaning of `BatchSumcheckOutput::challenges` for any out-of-tree consumer of the prover API: code reading it as a low-to-high evaluation point now reverses at the call site, as the in-tree consumers do.

Prover/verifier round-trip tests pass in `binius-ip-prover`, `binius-iop-prover`, `binius-prover`, and `binius-m4-prover` — these compare prover output against verifier output, so they would catch any consumer left uncompensated.

### DIFF
--- a/crates/iop-prover/src/basefold/channel.rs
+++ b/crates/iop-prover/src/basefold/channel.rs
@@ -343,8 +343,7 @@ fn prove_batch_zk_basefold<A, F, P, NTT, Channel>(
 	};
 
 	// Reduced oracle evaluations α_i = π_i'(ρ_i) come out in arrival order; scatter them into
-	// oracle-index order to match how the verifier indexes them. `output.challenges` is already
-	// reversed to low-to-high (variable-indexed) order, so ρ_i is its first n_i coords.
+	// oracle-index order to match how the verifier indexes them.
 	//
 	// TODO: This will fail if one oracle isn't opened. For robustness, we should do a regular
 	// multilinear evaluation in that case.
@@ -358,6 +357,10 @@ fn prove_batch_zk_basefold<A, F, P, NTT, Channel>(
 	// Collapse the oracle-index variables up front at sampled batching challenges `r'`: build the
 	// combined multilinear 𝛑(X) = Σ_i e[i]·π_i^↑(X) with e = eq(·, r') into one 2^𝐧 buffer, and the
 	// combined target s' = 𝛑(r) = Σ_i e[i]·α_i·∏_{j≥n_i}(1 - r_j).
+	// `batch_prove` returns binding-order challenges; reverse to variable-indexed (low-to-high),
+	// so that ρ_i is the first n_i coords.
+	let mut challenges = challenges;
+	challenges.reverse();
 	let point = &challenges;
 	let log_n_oracles = log2_ceil_usize(n_committed);
 	let outer_challenges = channel.sample_many(log_n_oracles);
--- a/crates/ip-prover/src/fracaddcheck.rs
+++ b/crates/ip-prover/src/fracaddcheck.rs
@@ -266,7 +266,9 @@ where
 			let next_num = extrapolate_line_packed(num_0, num_1, r);
 			let next_den = extrapolate_line_packed(den_0, den_1, r);
 
+			// Sumcheck binds variables high-to-low; reverse to low-to-high for the claim point.
 			let mut next_point = output.challenges;
+			next_point.reverse();
 			next_point.push(r);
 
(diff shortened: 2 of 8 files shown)


jimpo (approved):
Thanks!

PR #1896, opened 2026-07-23
Add PackedPrimitiveType packings for GhashSq256b and derive scalar arithmetic

benji-potamus:
Part 2 of [BINIUS-336](https://linear.app/jimpo/issue/BINIUS-336), following the now-merged #1894 (the irreducible flip).

Adds the `PackedPrimitiveType` packings the issue asks for and makes the scalar field derive its arithmetic from them.

### New interleaved (array-of-structs) packings
- **`PackedGhashSq1x256b`** = `PackedPrimitiveType<M256, GhashSq256b>` (width 1)
- **`PackedGhashSq2x256b`** = `PackedPrimitiveType<M512, GhashSq256b>` (width 2)

Each stores a GHASH² scalar as one contiguous 256-bit value (the same layout as the scalar field), alongside the existing sliced (`SlicedGhashSq*`) packings, which remain.

The width-one packing carries the field arithmetic. Its widening multiply **is** the `mul_m256i_hybrid` algorithm: the diagonal Karatsuba products `[a·e, b·f]` batch into one `PackedBinaryGhash2x128b` widening multiply — a single 256-bit VPCLMUL on AVX2, two 128-bit PMULL on aarch64 via the scaled `M256` — plus a scalar cross product, reduced with `Y² = X·Y + X`. The width-two packing divides into two width-one `M256` lanes (`Divide<M256, _, 2>`).

Delegating the diagonal to the base GHASH packed coordinate type gives the per-arch algorithms (x86_64 hybrid, aarch64 NEON) for free from that type's arch specialization — no new `unsafe` intrinsics and no dependency on `binius-arith-bench` (a bench harness `binius-field` doesn't depend on; see the Linear thread).

### Scalar field derives its arithmetic
`GhashSq256b` drops the `custom_arithmetic` modifier and uses plain `binary_field!`, so its `Mul`/`Square`/`InvertOrZero`/`WideMul` derive from the width-one packing via `@arithmetic_via_packed`. The duplicated hand-written scalar arithmetic (~110 lines) is removed; behavior is identical (same `reduce∘wide_mul`).

### Reflexive `Divisible<M256>`
Added for `M256` (native on x86_64, `ScaledUnderlier` on portable/aarch64) — the width-one `PackedPrimitiveType<M256, _>` packing needs `M256: Divisible<M256>` because its scalar is itself `M256`-backed. This is the `M256` counterpart to the existing `impl_divisible_self!(M128)`; verified it does not collide with the generic `ScaledUnderlier` `Divisible` impl.

### Testing
The two new packings run through the same lane-by-lane-vs-scalar proptest suite as the sliced packings (`packed_width1`/`packed_width2`). Full matrix green: x86_64 (native + portable), aarch64 (lib + tests, cross-compiled), wasm32, clippy `-D warnings`, nightly fmt, rust-doc `-D warnings`.

### DIFF
--- a/crates/field/src/packed_ghash_sq.rs
+++ b/crates/field/src/packed_ghash_sq.rs
@@ -1,28 +1,40 @@
 // Copyright 2026 The Binius Developers
 
-//! Packed [`GhashSq256b`] in a sliced (struct-of-arrays) layout.
+//! Packed [`GhashSq256b`] in two layouts: sliced (struct-of-arrays) and interleaved
+//! (array-of-structs).
 //!
 //! [`GhashSq256b`] is the degree-two extension `a + b·Y` of the GHASH field, with `Y² = X·Y + X`.
-//! The packings here store the `a` and `b` coordinates of every lane in two separate packed GHASH
-//! registers via [`SlicedPackedField`], so a batch multiply runs as three packed GHASH multiplies
-//! over the whole batch (Karatsuba over `Y`) rather than a schoolbook product per element.
+//! Both layouts reduce a batch multiply to three packed GHASH multiplies (Karatsuba over `Y`)
+//! rather than a schoolbook product per element, and both defer the GHASH reductions and the
+//! multiply-by-`X` — all `GF(2)`-linear — so an inner product reduces once at the end.
 //!
-//! Only the field arithmetic lives here: a widening multiply ([`WideMul`]) with a deferred
-//! reduction, plus [`Square`] and [`InvertOrZero`]. `Mul` and the rest of the [`PackedField`]
-//! surface are provided generically by [`SlicedPackedField`]. The coordinate register is a
-//! [`PackedPrimitiveType`], so the reduction scales by `X` with a per-lane bit shift
-//! ([`ghash_mul_x`]) rather than a full field multiply, and the batch costs three GHASH
-//! multiplies, not four.
+//! - The **sliced** packings ([`SlicedGhashSq256b`]) store the `a` and `b` coordinates of every
+//!   lane in two separate packed GHASH registers via [`SlicedPackedField`]; `Mul` and the rest of
+//!   the [`PackedField`] surface come generically from [`SlicedPackedField`].
+//! - The **interleaved** packings ([`PackedGhashSq1x256b`] / [`PackedGhashSq2x256b`], i.e.
+//!   `PackedPrimitiveType<M256/M512, GhashSq256b>`) store each scalar as one contiguous 256-bit
+//!   value — the same layout as the scalar field. The width-one packing carries the field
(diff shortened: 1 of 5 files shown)


crates/field/src/packed_ghash_sq.rs
```diff
@@ -201,6 +212,196 @@ where
...
+		iter.fold(Self::default(), |acc, x| acc + x)
+	}
+}
+
+/// Reinterprets a width-one packed GHASH² element as its two packed GHASH coordinates `(a, b)`.
+///
+/// [`PackedGhashSq1x256b`] and [`PackedBinaryGhash2x128b`] share the `M256` underlier and lane
+/// layout (low lane = coefficient of `1`, high lane = coefficient of `Y`), so this reinterprets.
+#[inline]
+const fn to_coords_1x(x: PackedGhashSq1x256b) -> PackedBinaryGhash2x128b {
```
jimpo: Too specific. Make a function `cast_base` alongside `cast_base_mut` in `packed_extension` and use that.

crates/field/src/packed_ghash_sq.rs
```diff
@@ -201,6 +212,196 @@ where
...
+	PackedBinaryGhash2x128b::from_underlier(x.to_underlier())
+}
+
+/// Builds a width-one packed GHASH² element from its two packed GHASH coordinates `(a, b)`.
+#[inline]
+const fn from_coords_1x(coords: PackedBinaryGhash2x128b) -> PackedGhashSq1x256b {
```
jimpo: Use `cast_ext`

jimpo (changes requested):

crates/field/src/packed_ghash_sq.rs
benji-potamus: Done in 924c8e65 — added a by-value `cast_base<FSub, P>` to `packed_extension` alongside `cast_base_mut`/`cast_ext`, and replaced the ad-hoc `to_coords_1x` helper with it.

crates/field/src/packed_ghash_sq.rs
benji-potamus: Done in 924c8e65 — the reverse direction (`from_coords_1x`) now uses `cast_ext`. Both helpers removed.

benji-potamus:
@jimpo addressed both review comments in 924c8e65: added a by-value `cast_base<FSub, P>` to `packed_extension` (alongside `cast_base_mut`), and the width-one packing's reinterprets now go through `cast_base`/`cast_ext` instead of the ad-hoc `to_coords_1x`/`from_coords_1x` helpers (both removed). Full matrix re-verified green locally (x86_64 native+portable, aarch64 lib+tests, wasm32, clippy -D warnings, nightly fmt, rust-doc -D warnings). Ready for another look.

jimpo (approved):

PR #1897, opened 2026-07-23
BINIUS-337: move BufferPool onto Prover structs

benji-potamus:
## Summary

Each `prove` method (and the ZK wrapper channel's `finish`) currently creates a fresh `BufferPool` for the duration of that one call, so the recycling free list starts empty on every proof. This moves the pool onto the four top-level `Prover` structs — binius64 `Prover`/`ZKProver`, m4 `Prover`, and IronSpartan `Prover` — where it is created once in `setup` and lives until the prover is dropped. A prover reused across proofs then recycles the previous proof's working buffers instead of re-allocating (and re-faulting) them.

Linear: [BINIUS-337](https://linear.app/jimpo/issue/BINIUS-337). Follow-up to #1869 (BINIUS-316), which introduced the `Allocator`/`BufferPool` pooling.

## Changes

- Each Prover struct gains a `pool: BufferPool` field, initialized in `setup`; `prove` draws `&self.pool` in place of `let pool = BufferPool::new()`.
- `ZKWrappedProverChannel` — a per-proof transient built inside `ZKProver::prove` — is now generic over `A: Allocator` and borrows the owning prover's allocator as `alloc: &'a A`, using it in `finish`. `ZKProver::prove` passes `&self.pool` (so `A = &BufferPool`), sharing the one prover-owned pool between the inner IOP proof and the outer wrapper proof (they run sequentially, and the outer proof reuses blocks the inner one freed).
- `BufferPool` is `Mutex`-backed (`Send + Sync`), so `&self.pool` threads through the `A = &BufferPool` allocator generics exactly as the per-call pool did — no signature changes below the top level.

## Performance

Measured with the `keccak_proof_generation` examples bench, which sets up the prover once and then calls `prove` in criterion's `iter` loop. On `main` every iteration allocates a cold pool; on this branch the pool warms up over the first iterations and is reused — so the criterion steady-state mean captures exactly the cross-proof reuse. Single-threaded, AVX-512, `-C target-cpu=native`, `--release`, 10 samples, `LOG_INV_RATE=1`.

| message | `main` | branch | Δ |
|---|---|---|---|
| 16 KB | 138.01 ms | 135.27 ms | **−2.0%** |
| 64 KB | 520.54 ms | 502.32 ms | **−3.5%** |

The win grows with circuit size (a larger working set to recycle), and at 64 KB the confidence intervals are disjoint (branch 499.7–505.3 ms vs `main` 514.0–528.2 ms). A single cold proof is unchanged; the benefit is entirely from reusing a `Prover` across proofs.

## Testing

Green across `clippy --workspace --all-targets -D warnings`, `build --workspace --lib --bins --tests --examples`, the compute/math/prover/m4/spartan tests (including `test_zk_wrapped_prove_verify`, which exercises the shared inner+outer pool), doc tests, and nightly fmt.

### DIFF
--- a/crates/spartan-prover/src/wrapper/zk_wrapped_prover_channel.rs
+++ b/crates/spartan-prover/src/wrapper/zk_wrapped_prover_channel.rs
@@ -13,7 +13,7 @@
 
 use std::{iter::repeat_with, sync::Arc};
 
-use binius_compute::BufferPool;
+use binius_compute::Allocator;
 use binius_field::{BinaryField, PackedField};
 use binius_iop::channel::OracleSpec;
 use binius_iop_prover::{
@@ -40,7 +40,7 @@ use crate::{Error, IOPProver, pack_and_blind_witness, wrapper::ReplayChannel};
 /// The `ReplayFn` closure is called during [`finish`](Self::finish) with a [`ReplayChannel`] to
 /// replay the inner verification and fill the outer witness. This allows the channel to be generic
 /// over different inner verification protocols.
-pub struct ZKWrappedProverChannel<'a, P, NTT, Channel, ReplayFn>
+pub struct ZKWrappedProverChannel<'a, P, NTT, Channel, ReplayFn, A>
 where
 	P: PackedField<Scalar: BinaryField>,
 	NTT: AdditiveNTT<Field = P::Scalar> + Sync,
@@ -48,6 +48,9 @@ where
 {
 	inner_channel: BaseFoldProverChannel<'a, P::Scalar, P, NTT, Channel>,
 	outer_prover: &'a IOPProver<P::Scalar>,
+	/// Allocator for the outer proof's working buffers, borrowed from the owning prover so it
+	/// outlives this per-proof channel. Used in [`Self::finish`].
+	alloc: &'a A,
 	outer_layout: Arc<WitnessLayout<P::Scalar>>,
 	replay_fn: ReplayFn,
 	keys: Vec<P::Scalar>,
@@ -64,7 +67,7 @@ where
 	n_outer_suffix_oracles: usize,
 }
 
-impl<'a, F, P, NTT, Channel, ReplayFn> ZKWrappedProverChannel<'a, P, NTT, Channel, ReplayFn>
+impl<'a, F, P, NTT, Channel, ReplayFn, A> ZKWrappedProverChannel<'a, P, NTT, Channel, ReplayFn, A>
 where
 	F: BinaryField,
 	P: PackedField<Scalar = F>,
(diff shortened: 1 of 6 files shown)


crates/spartan-prover/src/wrapper/zk_wrapped_prover_channel.rs
```diff
@@ -48,6 +48,9 @@ where
 {
 	inner_channel: BaseFoldProverChannel<'a, P::Scalar, P, NTT, Channel>,
 	outer_prover: &'a IOPProver<P::Scalar>,
+	/// Pool for the outer proof's working buffers, borrowed from the owning prover so it outlives
+	/// this per-proof channel. Used in [`Self::finish`].
+	pool: &'a BufferPool,
```
jimpo: Make the channel generic over A: Allocator (and it holds an &'a A).

crates/spartan-prover/src/wrapper/zk_wrapped_prover_channel.rs
```diff
@@ -205,9 +212,9 @@ where
 				.expect("outer witness generation should not fail")
 		};
 
-		// Working buffers for this proof are drawn from a single pool that lives for the call.
-		let pool = BufferPool::new();
-		let alloc = &pool;
+		// Working buffers for this proof are drawn from the owning prover's pool, recycling blocks
```
jimpo: unnecessary comment now

crates/spartan-prover/tests/wrapper_integration_test.rs
```diff
@@ -146,10 +146,12 @@ fn test_zk_wrapped_prove_verify() {
 			&mut prover_transcript,
 			&mut rng,
 		);
+	let pool = BufferPool::new();
 	let mut wrapped_prover_channel = ZKWrappedProverChannel::new(
 		basefold_channel,
 		&outer_iop_prover,
 		Arc::clone(&outer_layout),
+		&pool,
```
jimpo: Use &GlobalAllocator in the test (once the channel is generic)

jimpo (changes requested):

crates/spartan-prover/src/wrapper/zk_wrapped_prover_channel.rs
benji-potamus: Done in 63b7221. `ZKWrappedProverChannel` is now generic over `A: Allocator` and holds `alloc: &'a A` (all three impl blocks + both trait impls thread the param). Only `finish` needs `A: Allocator`, so that bound lives on `finish`'s where-clause rather than the whole inherent impl (otherwise `next_key`, called from the unbounded `IPProverChannel` impl, fails to resolve). `ZKProver::prove` passes `&self.pool` (so `A = &BufferPool`); the test passes `&GlobalAllocator`.

crates/spartan-prover/src/wrapper/zk_wrapped_prover_channel.rs
benji-potamus: Removed — `finish` now uses the destructured `alloc` field directly (no local pool creation, no comment).

crates/spartan-prover/tests/wrapper_integration_test.rs
benji-potamus: Done — the test now passes `&GlobalAllocator` (dropped the local `BufferPool`), so the outer proof there uses the same non-pooling allocator as the inner prove.

jimpo (approved):

PR #1898, opened 2026-07-23
perf(circuits): derive the base64 character instead of multiplexing it

wstran:
## What

Checking one base64 character in the zklogin circuit costs 70 AND constraints, and the JWT header, payload and signature are each base64 checked. Dropping the multiplexer takes 24% of the circuit's AND constraints with it.

**Root cause:** `compute_expected_base64_char` builds a 64-entry constant lookup table and reads it with `single_wire_multiplex`, a binary tree of 63 select gates, one AND constraint each.

**Fix:** the base64url alphabet is affine on each of its ranges, so the character follows from three range checks and one addition:

```text
 0..=25 -> v + 'A'
26..=51 -> v + ('a' - 26)
52..=61 -> v + 0xfc        (byte-sized -4, masked back to a byte)
     62 -> '-'
     63 -> '_'
```

The digit offset carries out of the low byte, so the sum is masked back down to one. The two symbols are told apart by the low bit of the value, moved into the MSB that `select` reads. Measured in isolation, the mapping drops from 70 AND constraints per character to 11.

```text
before:  v -> [tree of 63 selects over 64 constants] -> char
after:   v -> 3 range checks -> offset select -> add -> mask -> char
```

## Numbers

Apple M3 Pro, `zklogin` example, timings are medians of 9 runs.

| metric | baseline | this PR | change |
| --- | ---: | ---: | ---: |
| AND constraints per character | 70 | 11 | -84% |
| AND constraints | 357,836 | 272,876 | -23.7% |
| gates | 485,515 | 404,875 | -16.6% |
| evaluation instructions | 550,362 | 474,042 | -13.9% |
| committed words | 360,378 | 275,418 | -23.6% |
| prove | 102.0 ms | 97.9 ms | -4.0% |
| shift reduction | 21.6 ms | 19.8 ms | -8.3% |

The per-character row is the marginal constraint count: a scratch circuit repeating the mapping N times, differenced across two values of N so the surrounding scaffolding cancels. It is flat from N=1 to N=200.

Both constraint counts still pad to 2^19, so the BitAnd phase is unchanged and the prove win comes from the shift reduction and the commit working over a smaller witness. The 24% headroom is what matters for a larger JWT, which previously ran closer to the next power of two. Witness generation timings were too noisy on this machine to quote, so the instruction count is given instead.

repro:

```bash
cargo run --release --example zklogin -- stat
RUST_LOG=info cargo run --release --example zklogin -- prove
```

## Correctness

Two new tests cover the whole input domain rather than a sample:

- `test_every_alphabet_index` feeds a fixture whose consecutive 6-bit groups are `0..64` in order, so its base64 encoding is the alphabet itself and the circuit checks every index once.
- `test_every_alphabet_index_rejects_wrong_character` corrupts each of the 64 encoded characters in turn and requires the circuit to reject, so no index is silently unchecked.

Both fail if the range boundaries are perturbed (moving the digit boundary from 52 to 53 breaks `test_every_alphabet_index`), so they are not vacuous.

`zklogin` is the only example whose statistics move; the other twelve snapshots are unchanged, and the zklogin proof still verifies end to end.

## Scope

Only the encoder-side character mapping changes. Group handling, the length and boundary checks, and the `is_active` masking are untouched. `single_wire_multiplex` itself is unchanged and still used elsewhere; a general "all inputs are constants" lowering in the multiplexer would be a separate change and a less clear win, since the saving here comes from the alphabet being affine rather than from the inputs being constant.


wstran:
Rebased onto main, clears the snapshot conflict from #1907. Numbers hold: zklogin 357,836 -> 272,876 AND constraints.

wstran:
#1914 stacks on this one. Alone this PR leaves zklogin at 272,876 AND constraints, which still pads to 2^19; together the two reach 258,953 and drop under 2^18, so the padded AND column halves. Measured on both, medians of 11 interleaved runs: BitAnd check 23.7ms -> 12.4ms, commit witness 11.8ms -> 5.8ms, prove 116.0ms -> 86.1ms.

benji-potamus (approved):
Approve.

Mechanical pass: rebased onto current `main` (the only conflict was numeric drift in `zklogin.snap`, reblessed), `cargo fmt`, `cargo clippy -p binius-circuits --all-features --tests --benches --examples`, and the full `binius-circuits` test suite (329 tests) are all clean.

Correctness: I traced `compute_expected_base64_char`'s arithmetic against all four ranges by hand — the digit range's `0xfc` offset does carry out of the low byte as documented (e.g. v=52 → sum=304 → masked to 48 = `'0'`), and the two symbols resolve correctly off the LSB-as-MSB trick. Beyond that, I ran `zklogin`'s `stat` (matches the committed snapshot) and `prove` (proves and verifies end-to-end) against this branch, both succeeding with the AND-constraint count the PR claims. The diff is exactly what the PR describes — the lookup table is replaced by the affine computation and two exhaustive tests are added — nothing else touched.

jimpo (approved):
This is awesome!!

jimpo:
Please re-bless snapshot

PR #1899, opened 2026-07-23
BINIUS-338: pick the packed GHASH² widening multiply per architecture

benji-potamus:
Closes [BINIUS-338](https://linear.app/jimpo/issue/BINIUS-338/special-arithmetic-for-packed-ghash2-on-aarch64-and-portable).

#1896 gave `PackedGhashSq1x256b` a single widening-multiply strategy: batch the two diagonal
Karatsuba products `[a·e, b·f]` into one `PackedBinaryGhash2x128b` widening multiply, then scale the
reduced `t₂` by `X`. That batching is a single instruction only where a 256-bit carry-less multiply
exists. Everywhere else the packed multiply decomposes back into two independent 128-bit multiplies,
each with its own reduction, so the batching buys nothing and the multiply costs three GHASH
reductions instead of two.

### Two strategies, selected per architecture

Following the `GhashWideMul1x/2x/4x` pattern, `arch` now picks the width-one GHASH² `WideMul`
through a `GhashSqWideMul1x` alias:

- **`GhashSqHybridWideMul`** (`arch/x86_64/arithmetic/ghash_sq.rs`) — the existing `mul_m256i_hybrid`
  algorithm, moved unchanged. Selected under VPCLMULQDQ + AVX2. (AVX2 is part of the condition
  because without it `PackedBinaryGhash2x128b` is backed by the scaled `M256`, whose widening
  multiply is two independent 128-bit multiplies — exactly what the batching is meant to avoid.)
- **`GhashSqSlicedWideMul`** (`arch/portable/arithmetic/ghash_sq.rs`) — the `mul_sliced` algorithm
  from `binius-arith-bench`: keep the three Karatsuba products separate and scale `t₂` by `X` while
  it is still *unreduced*, folding the multiply-by-`X` into a reduction that has to happen anyway.
  Two GHASH reductions per GHASH² product rather than three. Selected on aarch64, portable, wasm32
  and the remaining x86_64 configurations.

Both reuse `SlicedGhashSqWide` / the packing's coordinate helpers, so the GHASH²-specific algebra is
stated once.

### `MulXWide`

Scaling an unreduced product is a new trait in `arithmetic_traits`, implemented for the portable,
aarch64 and x86_64 `WideGhashProduct`s (the x86 one via a new `ClMulUnderlier::shift_left_1` for
M128/M256/M512). Each backend carries a proptest that `reduce(w.mul_x_wide())` equals
`reduce(w) · X`. The shift is exact because every limb is a carry-less product of 64-bit halves, so
its top bit is clear and XOR-accumulation preserves that.

### Benchmarks

Measured with `binius-arith-bench`'s multiply harness (16-element batch, 64 passes) on Neoverse V3,
so the numbers are directly comparable to that crate's:

| | before | after |
|---|---|---|
| aarch64, `-Ctarget-cpu=native` | 194.9 Melem/s | **210.2 Melem/s** (+7.9 %) |
| portable | 13.99 Melem/s | **14.32 Melem/s** (+2.4 %) |

For reference `binius-arith-bench`'s `ghash_sq::aarch64::mul_sliced` measures 228.8 Melem/s. The
remaining gap is not algorithmic: both compile to the same 16 PMULLs (12 for the three coordinate
products, 4 for two reductions) with the same XOR/shift mix, and the base GHASH multiply is
instruction-identical between the crates (18 instructions, 6 PMULLs). `binius-field`'s GHASH² body
is 59 instructions to arith-bench's 56 — one extra `dup` and two extra `ext`, from LLVM lowering one
reduction's `pmull2`-by-poly through a rotate/64-bit-eor sequence instead of `ext` + `pmull2`.
Holding `0x87` in the high lane so `PMULL2` reads it in place (as arith-bench does) changed neither
the instruction count nor the throughput, so that experiment is not included here.

Note that the pre-existing `ghash_widening` bench `black_box`es each 32-byte operand per multiply and
reports the opposite ordering for these two strategies; the harness above is the one to trust for
multiply cost.

### Testing

Both packings run through the existing lane-by-lane-vs-scalar suite, plus the new `MulXWide`
proptests. Green on aarch64 native + portable (tests, clippy `-D warnings`, rustdoc, nightly fmt),
`cargo check` cross-compiled for x86_64 across five target-feature combinations (baseline, PCLMUL,
VPCLMUL-without-AVX2, AVX2+VPCLMUL, AVX-512), and the wasm32 build. (`wasm32 + simd128` fails
identically on `main` — pre-existing, untouched here.)

### DIFF
--- a/crates/field/src/arch/x86_64/arithmetic/ghash.rs
+++ b/crates/field/src/arch/x86_64/arithmetic/ghash.rs
@@ -14,8 +14,10 @@ use std::{
 use bytemuck::TransparentWrapper;
 
 use crate::{
-	BinaryField128bGhash as GhashB128, Divisible, WideMul, arch::PackedPrimitiveType,
-	arithmetic_traits::Square, underlier::UnderlierType,
+	BinaryField128bGhash as GhashB128, Divisible, WideMul,
+	arch::PackedPrimitiveType,
+	arithmetic_traits::{MulXWide, Square},
+	underlier::UnderlierType,
 };
 
 /// Trait for underliers that support CLMUL operations which are needed for the
@@ -31,6 +33,13 @@ pub trait ClMulUnderlier: UnderlierType + Divisible<u128> {
 	/// For each 128-bit lane, shifts the lower 64 bits to the upper 64 bits and zeroes the lower
 	/// 64-bit.
 	fn move_64_to_hi(a: Self) -> Self;
+
+	/// For each 64-bit lane, shifts the value left by one bit.
+	fn shl_1_epi64(a: Self) -> Self;
+
+	/// For each 64-bit lane, shifts the value right by 63 bits, leaving the lane's top bit as its
+	/// low bit.
+	fn shr_63_epi64(a: Self) -> Self;
 }
 
 /// The version of the multiplication for optimized suqare operation.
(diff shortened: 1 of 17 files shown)


crates/field/src/arch/x86_64/arithmetic/ghash.rs
```diff
@@ -134,6 +140,23 @@ impl<U: ClMulUnderlier> WideGhashProduct<U> {
...
 
+impl<U: ClMulUnderlier> MulXWide for WideGhashProduct<U> {
+	/// Shifts each limb left by one bit, which scales the represented 256-bit polynomial
+	/// `lo + mid·X^64 + hi·X^128` by `X`.
+	///
+	/// Every limb is a carry-less product of 64-bit halves, so it has degree at most 126, and
+	/// XOR-accumulating such products preserves that — bit 127 is always clear and no limb
+	/// overflows into the next.
+	#[inline]
+	fn mul_x_wide(self) -> Self {
```
jimpo, line 158: This is suboptimal. See `mul_x_wide` in `binius_arith_bench::ghash::clmul` and use that algorithm.

jimpo (approved):

PR #1901, opened 2026-07-23
BINIUS-340: skip all-zero padding windows in the AND-reduction round-1 message

tcoratger:
Closes [BINIUS-340](https://linear.app/jimpo/issue/BINIUS-340).

Computes the first-round univariate message of the AND-reduction exactly as before, but skips the windows that are pure zero padding. Output is bit-identical — no protocol change, no new soundness assumption.

### The problem

The AND-reduction batches all instances together, so its columns hold one AND constraint per row.
The row count is rounded **up to a power of two**, and the slack is filled with zero words.

That padding is not free.
The round-1 message walks every 128-word window unconditionally, and each window runs its full NTT-lookup-plus-multiply inner loop — even when every word in it is zero.

For a power-of-two AND count (e.g. BLAKE3) there is no padding, so nothing is wasted.
But Keccak and SHA-256 are not power-of-two: Keccak-f[1600] is ~38400 ANDs per permutation, which rounds up to $2^{16} = 65536$ — about **40% of the work is on zero padding**.
Any non-power-of-two batch (XMSS aggregation being the motivating case) pays the same tax.

### The idea

The product column is derived, not committed: `C = A & B`.

So for any word where `A = 0` (or `B = 0`) we get `C = 0`, and the term the message accumulates is

$$ A \cdot B - C = 0. $$

A whole window that is all-zero in either operand therefore contributes exactly zero.
Detect it with a cheap scan and skip its inner loop entirely.

### The change

```diff
 (a_col_chunks, b_col_chunks).into_par_iter().map(|(a_chunk, b_chunk)| {
+    // A window all-zero in either operand contributes exactly 0: skip its NTT work.
+    if a_chunk.iter().all(|w| *w == Word::ZERO) || b_chunk.iter().all(|w| *w == Word::ZERO) {
+        return [F::ZERO; ROWS_PER_HYPERCUBE_VERTEX];
+    }
     // ... reshape, NTT lookups, weighted composition ...
 })
```

The `all` scan short-circuits on the first non-zero word, so a populated window pays only a couple of comparisons.

### Soundness

- The skip fires **only** when the contribution is provably zero, so the emitted message is bit-identical to today's.
- The transcript is unchanged, so the challenge distribution and the verifier are untouched.
- The reduced-then-weighted zero path already lands on `lookup[0] = 0` (the empty subset sum), confirming an all-zero window's contribution is the zero array.

### Scope

- **changed:** one per-window early-return in `crates/prover/src/and_reduction/sumcheck_round_messages.rs`.
- **tests:** refactored the existing round-message consistency check into a shared helper; added a padded-shape test that zeroes a whole-window-aligned suffix and pins the message against an independent verifier-side recomputation.
- **bench:** added a 40%-zero-padding case to the `and_reduction` bench.
- **left untouched:** the dense path, the arithmetization, and the padded-sumcheck decorator (which covers only the basefold channel).

### Verification

- `cargo clippy -p binius-prover --all-targets` — clean.
- nightly `cargo fmt --all` — clean.
- `cargo test -p binius-prover --lib sumcheck_round_messages` — dense and padded tests pass; the e2e AND-reduction prove/verify test still passes.

### Benchmark

Isolated round message, `and_reduction` bench at $2^{21}$ words, single-threaded:

| shape | before | after | speedup |
|---|---|---|---|
| dense (no padding) | 58.4 ms | 58.5 ms | 1.00× (no regression) |
| 40% zero padding | 59.2 ms | 37.2 ms | 1.59× |

End-to-end prove, single-threaded, `LOG_INSTANCES=10`:

| bench | before | after | speedup |
|---|---|---|---|
| `keccak_protocol` | 117.3 ms | 110.6 ms | 1.06× |
| `sha256_compressions` | 124.6 ms | 119.6 ms | 1.04× |

The dense row confirms the no-padding path is unchanged; the win concentrates in the non-power-of-two shapes, exactly where XMSS aggregation will land.

### DIFF
--- a/crates/prover/benches/and_reduction.rs
+++ b/crates/prover/benches/and_reduction.rs
@@ -61,6 +61,30 @@ fn bench(c: &mut Criterion) {
 		});
 	});
 
+	// Padded shape: rounding a non-power-of-two AND count up to a power of two leaves a zero
+	// suffix. Keccak and SHA-256 batches carry such a suffix; here it is 2/5 of the words.
+	// Keep the first 3/5 populated and zero the rest.
+	let padded_real = (1usize << log_words) * 3 / 5;
+	let mut a_words_padded = a_words.clone();
+	let mut b_words_padded = b_words.clone();
+	a_words_padded[padded_real..].fill(Word::ZERO);
+	b_words_padded[padded_real..].fill(Word::ZERO);
+
+	group.bench_function(
+		format!("univariate_round_message 2^{log_words} (40% zero padding)"),
+		|bench| {
+			bench.iter(|| {
+				univariate_round_message_extension_domain::<B128>(
+					log_words,
+					&a_words_padded,
+					&b_words_padded,
+					&big_field_zerocheck_challenges,
+					&prover_message_domain,
+				)
+			});
+		},
+	);
+
 	let urm = univariate_round_message_extension_domain::<B128>(
 		log_words,
 		&a_words,
--- a/crates/prover/src/and_reduction/sumcheck_round_messages.rs
+++ b/crates/prover/src/and_reduction/sumcheck_round_messages.rs
@@ -133,6 +133,20 @@ where
 	(a_col_chunks.as_ref(), b_col_chunks.as_ref())
 		.into_par_iter()
 		.map(|(a_chunk, b_chunk)| {
+			// Skip zero-padding windows: their chunks add nothing to the round message.
+			//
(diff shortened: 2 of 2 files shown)


jimpo (approved):

jimpo:
I generally haven't been chasing these type of <2x speedups that relate to padding. There's many of them. But this looks like an isolated enough change.

PR #1902, opened 2026-07-23
BINIUS-341: split the ring-switch suffix tensor into two factors and never materialize it

tcoratger:
Closes [BINIUS-341](https://linear.app/jimpo/issue/BINIUS-341).

Splits the ring-switch suffix equality tensor into two small factors and never materializes the full buffer.
Output is bit-identical — no protocol change, transcript-identical proofs.

### The problem

The ring-switching prover expands the full suffix equality tensor — $2^n$ field elements, 64 MB at $n = 22$ — and streams it three more times:

- one write to materialize it,
- one read to fold the witness rows against it,
- one read plus one write to transform it in place into the transparent polynomial.

That is ~256 MB of memory traffic per proof for a table that is secretly a product of two tiny ones.
On top of that, the row fold rebuilds a 16-entry subset-sum table for every 4 rows it touches.

### The idea

The equality tensor factors along any coordinate split:

```
tensor[hi << n_lo | lo] = eq_lo[lo] * eq_hi[hi]
```

Cap the low factor at $2^{10}$ entries and both consumers use the factors directly.

**Row fold.** Multiplication distributes over the sum, so the fold regroups exactly:

```
sum_x tensor[x] * row[x]  =  sum_hi eq_hi[hi] * ( sum_lo eq_lo[lo] * row[hi, lo] )
```

Each hi-block folds against `eq_lo` alone, then its 128-entry block result is scaled once by `eq_hi[hi]`.
The subset-sum tables now cover only the low factor — 64 KiB, built once, L1-resident across every block — instead of being rebuilt per 4 rows.
The price is 128 multiplies per block of 1024 rows, well under one per row.

**Transparent polynomial.** The later opening genuinely needs this buffer materialized, but not the untransformed tensor behind it.
The dense expansion already costs one multiply per entry (the doubling recursion), so building each entry as the fused product `eq_lo[lo] * eq_hi[hi]` and folding it in the same pass keeps the multiply count equal — while deleting the expand-write and transform-read passes.
One uninitialized-write pass replaces three.

### Per proof at $n = 22$

- **before:** 64 MB tensor written, read, and read+rewritten (~256 MB of traffic) + one table rebuild per 4 rows
- **after:** two factors of 16 KB and 64 KB; one 64 MB write for the transparent; tables built once

### The change

```diff
 // ring_switch::prove, same signature, callers untouched
-let suffix_tensor = eq_ind_partial_eval::<P>(eval_point_suffix);          // 2^n materialized
-let s_hat_v = fold_1b_rows_for_b128(packed_witness, &suffix_tensor);      // reads 2^n
-let rs_eq_ind = fold_b128_elems_inplace(suffix_tensor, &eq_r_double_prime); // reads+writes 2^n
+let (eq_lo, eq_hi) = (eq_ind_partial_eval(suffix_lo), eq_ind_partial_eval(suffix_hi));
+let s_hat_v = fold_1b_rows_for_b128_split(packed_witness, &eq_lo, &eq_hi); // factors only
+let rs_eq_ind = rs_eq_ind_from_factors::<P>(&eq_lo, &eq_hi, &eq_r_double_prime); // one pass
```

### How this relates

This is the Flock reference implementation's self-described main ring-switch win, ported at the caller level.
An earlier experiment porting Flock's 16-wide fold kernel regressed ~10% here; the win is not register-width mechanics but table reuse plus pass elimination, which is what this PR takes.

### Soundness

Pure regrouping of a sum and reassociation of products in $\mathbb{F}_{2^{128}}$ — both exact.
Pinned three ways:

- equality tests sweep every legal split position across three packing widths (1×/2×/4×128b);
- end-to-end proof digests match the dense route byte for byte (m4 BLAKE3 batch and monolithic SHA-256);
- the dense route survives as the fallback for suffixes too short to split, and as the reference the tests pin against.

### Scope

- **changed:** one file, `crates/prover/src/ring_switch.rs` — the two factored functions, the rewired prover, the shared table-build helpers, and the equality tests
- **unchanged:** the prover's public signature (both call sites untouched), the dense kernels (kept as fallback/reference), every other crate

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-prover --all-targets`, nightly `fmt` — clean
- prover + m4-prover suites pass, including prove→verify round trips and the new split-position sweeps

### Benchmark

Back-to-back stash A/B on an M1 Pro, 8 threads.

Isolated suffix-tensor pipeline (expand + row fold + transparent, identical outputs both routes):

| log_len | dense (before) | split (after) | speedup |
|---|---|---|---|
| 16 | 0.975 ms | 0.690 ms | 1.41× |
| 20 | 5.75 ms | 3.39 ms | 1.69× |
| 22 (production) | 20.50 ms | 11.94 ms | **1.72×** |

End to end, `blake3_compressions` at $2^{13}$ instances (witness generation + prove):

| | before | after | change |
|---|---|---|---|
| batch proof | 178.6 ms | 165.9 ms | **−7.1%** (p = 0.00) |

The e2e win slightly exceeds the phase's share of prove time because the deleted tensor passes also relieve memory-bandwidth pressure on neighboring phases.

<details>
<summary>The throwaway pipeline microbench behind the isolated rows (not part of the PR)</summary>

The e2e row comes from the existing `blake3_compressions` bench.
The isolated rows come from this group, dropped into the existing `crates/prover/benches/ring_switch.rs` and run with `--features rayon`:

```rust
/// The two suffix-tensor pipelines, whole:
///
/// - `dense`: expand the full tensor, fold the matrix against it, transform it in place.
/// - `split`: expand the two factors, fold against them, build the indicator in one pass.
///
/// Both produce the identical row fold and equality indicator, so the times compare directly.
fn bench_suffix_tensor_pipeline(c: &mut Criterion) {
	let mut group = c.benchmark_group("ring_switch/suffix_tensor_pipeline");
	group.sample_size(10);

	type P = OptimalPackedB128;
	// The split fold keeps its low-factor tables cache-resident; this mirrors the prover's cap.
	const LOG_SPLIT_BLOCK: usize = 10;

	for log_len in [16, 20, 22] {
		let elem_bytes = (1usize << log_len) * size_of::<P>();
		group.throughput(Throughput::Bytes(elem_bytes as u64));

		let mut rng = rand::rng();
		let mat = random_field_buffer::<P>(&mut rng, log_len);
		let suffix: Vec<B128> = random_scalars(&mut rng, log_len);
		let row_batching: Vec<B128> =
			random_scalars(&mut rng, <B128 as ExtensionField<B1>>::LOG_DEGREE);
		let row_batch_query = eq_ind_partial_eval::<B128>(&row_batching);

		group.bench_function(format!("dense/log_len={log_len}"), |b| {
			b.iter(|| {
				let tensor = eq_ind_partial_eval::<P>(&suffix);
				let s_hat_v = fold_1b_rows_for_b128(&mat, &tensor);
				let rs_eq_ind = fold_b128_elems_inplace(tensor, &row_batch_query);
				(s_hat_v, rs_eq_ind)
			});
		});

		group.bench_function(format!("split/log_len={log_len}"), |b| {
			let (suffix_lo, suffix_hi) = suffix.split_at(LOG_SPLIT_BLOCK.min(log_len));
			b.iter(|| {
				let eq_lo = eq_ind_partial_eval::<B128>(suffix_lo);
				let eq_hi = eq_ind_partial_eval::<B128>(suffix_hi);
				let s_hat_v = fold_1b_rows_for_b128_split(&mat, &eq_lo, &eq_hi);
				let rs_eq_ind = rs_eq_ind_from_factors::<P>(&eq_lo, &eq_hi, &row_batch_query);
				(s_hat_v, rs_eq_ind)
			});
		});
	}

	group.finish();
}
```

</details>

### DIFF
--- a/crates/prover/benches/ring_switch.rs
+++ b/crates/prover/benches/ring_switch.rs
@@ -2,9 +2,16 @@
 
 use std::mem::size_of;
 
+use binius_compute::GlobalAllocator;
 use binius_field::{BinaryField, ExtensionField, arch::OptimalPackedB128};
-use binius_math::test_utils::random_field_buffer;
-use binius_prover::ring_switch::{fold_1b_rows_for_b128, fold_elems_inplace};
+use binius_math::{
+	multilinear::eq::eq_ind_partial_eval,
+	test_utils::{random_field_buffer, random_scalars},
+};
+use binius_prover::ring_switch::{
+	LOG_SPLIT_BLOCK, fold_1b_rows_for_b128, fold_1b_rows_for_b128_split, fold_elems_inplace,
+	rs_eq_ind_from_factors,
+};
 use binius_utils::checked_arithmetics::checked_log_2;
 use binius_verifier::config::{B1, B128};
 use criterion::{BatchSize, Criterion, Throughput, criterion_group, criterion_main};
@@ -56,5 +63,65 @@ fn bench_fold_elems_inplace(c: &mut Criterion) {
 	group.finish();
 }
 
-criterion_group!(ring_switch, bench_fold_1b_rows_for_b128, bench_fold_elems_inplace);
+/// The two suffix-tensor pipelines, whole:
+///
+/// - `dense`: expand the full tensor, fold the matrix against it, transform it in place.
+/// - `split`: expand the two factors, fold against them, build the indicator in one pass.
+///
+/// Both produce the identical row fold and equality indicator, so the times compare directly.
+fn bench_suffix_tensor_pipeline(c: &mut Criterion) {
+	let mut group = c.benchmark_group("ring_switch/suffix_tensor_pipeline");
+	group.sample_size(10);
+
+	type P = OptimalPackedB128;
+
+	for log_len in [16, 20, 22] {
+		let elem_bytes = (1usize << log_len) * size_of::<P>();
+		group.throughput(Throughput::Bytes(elem_bytes as u64));
+
+		let mut rng = rand::rng();
+		let mat = random_field_buffer::<P>(&mut rng, log_len);
+		let suffix: Vec<B128> = random_scalars(&mut rng, log_len);
+		let row_batching: Vec<B128> =
+			random_scalars(&mut rng, <B128 as ExtensionField<B1>>::LOG_DEGREE);
+		let row_batch_query = eq_ind_partial_eval::<B128>(&row_batching);
+
+		group.bench_function(format!("dense/log_len={log_len}"), |b| {
+			b.iter(|| {
+				let tensor = eq_ind_partial_eval::<P>(&suffix);
+				let s_hat_v = fold_1b_rows_for_b128(&mat, &tensor);
+				let rs_eq_ind = fold_elems_inplace(tensor, &row_batch_query);
+				(s_hat_v, rs_eq_ind)
+			});
+		});
+
+		group.bench_function(format!("split/log_len={log_len}"), |b| {
+			// The prover caps the low factor here so its fold tables stay cache-resident.
+			let (suffix_lo, suffix_hi) = suffix.split_at(LOG_SPLIT_BLOCK.min(log_len));
+			b.iter(|| {
+				let eq_lo = eq_ind_partial_eval::<B128>(suffix_lo);
+				let eq_hi = eq_ind_partial_eval::<B128>(suffix_hi);
+				let s_hat_v = fold_1b_rows_for_b128_split(&mat, &eq_lo, &eq_hi);
+				// The dense route's tensor is `Vec`-backed, so the factored route allocates the
+				// same way for the two times to compare.
+				let rs_eq_ind = rs_eq_ind_from_factors::<_, P>(
+					&GlobalAllocator,
+					&eq_lo,
+					&eq_hi,
+					&row_batch_query,
+				);
+				(s_hat_v, rs_eq_ind)
+			});
+		});
+	}
+
+	group.finish();
+}
+
+criterion_group!(
+	ring_switch,
+	bench_fold_1b_rows_for_b128,
+	bench_fold_elems_inplace,
+	bench_suffix_tensor_pipeline
+);
 criterion_main!(ring_switch);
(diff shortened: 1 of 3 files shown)


jimpo:
Please add criterion benchmarks for these new functions to the `ring_switch` benchmark. It's hard to compare otherwise.

jimpo:
This is a lot of code -- my instinct is that there's a simpler way to get the perf gain and this is hiding some assembly-level optimization. But let's see the criterion benchmarks.

jimpo:
the `fold_1b_rows...` algorithm is very similar to `WordFolder`, except that that is specialized for words. Maybe we can generalize that to other underliers and use it for ring switching.

tcoratger:
@jimpo I've added the bench and try to unify things as per your latest comments. Let me know what you think about this? Maybe we should handle some followup stuffs in followup PRs to avoid inflating this one no?

jimpo:
Sorry, will review tomorrow!

jimpo (approved):

crates/prover/benches/ring_switch.rs
```diff
@@ -56,5 +63,101 @@ fn bench_fold_elems_inplace(c: &mut Criterion) {
...
+			// The prover caps the low factor here so its fold tables stay cache-resident.
+			let (suffix_lo, suffix_hi) = suffix.split_at(LOG_SPLIT_BLOCK.min(log_len));
+			b.iter(|| {
+				let eq_lo = eq_ind_partial_eval::<B128>(suffix_lo);
+				let eq_hi = eq_ind_partial_eval::<B128>(suffix_hi);
+				let s_hat_v = fold_1b_rows_for_b128_split(&mat, &eq_lo, &eq_hi);
+				// The dense route's tensor is `Vec`-backed, so the factored route allocates the
+				// same way for the two times to compare.
+				let rs_eq_ind = rs_eq_ind_from_factors::<_, P>(
+					&GlobalAllocator,
```
jimpo: Better to use BufferPool in benchmarks. See how the sumcheck benchmarks do it for example.

tcoratger: Fixed

PR #1909, opened 2026-07-25
Remove the unreachable mask opcodes from the bytecode interpreter

tcoratger:
Deletes two bytecode handlers that nothing can reach. Pure removal — no behaviour change.

### The problem

The witness interpreter dispatches an opcode byte to a handler:

```
0x40 => self.exec_iadd32_cin_cout(ctx),
0x46 => self.exec_iadd32_cout(ctx),
0x50 => self.exec_mask_low(ctx),      // nothing emits 0x50
0x51 => self.exec_mask_high(ctx),     // nothing emits 0x51
```

Every other opcode has a matching emitter on the bytecode builder, and a gate that calls it.
These two have neither.

Grepping the whole workspace for the emitter side comes back empty, so no bytecode can contain either byte.

They are dead handlers sitting in the middle of the witness-generation path, which is the one place a second implementation of an operation can quietly drift away from the real one.

### The latent bug in the high-bit handler

The mask is built as:

```
Word::from_u64(!((1u64 << (64 - n_bits)) - 1))
```

reached whenever `n_bits < 64`. At `n_bits == 0` that shift is `1u64 << 64`, which overflows.

Confirmed both ways with a standalone probe, taking the shift amount through `black_box` so it is a run-time value as it would be when read from bytecode:

| build | result | intended |
|---|---|---|
| overflow checks on | panics, `attempt to shift left with overflow` | — |
| overflow checks off | `0xffffffffffffffff` | `0x0000000000000000` |

So selecting the top zero bits would have returned every bit instead of none.

Unreachable today, but it was waiting for whoever emitted the opcode first.

### Scope

- **removed:** two dispatch arms and their two handlers
- **unchanged:** everything else — no emitter, gate, constraint or test referenced them

The `0x50` and `0x51` bytes are simply free again.

### Verification

- `binius-frontend` 158 tests, `binius-circuits` 320 tests — pass
- circuit statistics snapshots unchanged, spot-checked on blake3, sha256, keccak and zklogin

### DIFF
--- a/crates/frontend/src/compiler/eval_form/exec.rs
+++ b/crates/frontend/src/compiler/eval_form/exec.rs
@@ -112,10 +112,6 @@ impl<'a> Executor<'a> {
 				EvalOpcode::Iadd32CinCout => self.exec_iadd32_cin_cout(ctx),
 				EvalOpcode::Iadd32Cout => self.exec_iadd32_cout(ctx),
 
-				// Masks
-				EvalOpcode::MaskLow => self.exec_mask_low(ctx),
-				EvalOpcode::MaskHigh => self.exec_mask_high(ctx),
-
 				// Assertions
 				EvalOpcode::AssertEq => self.exec_assert_eq(ctx),
 				EvalOpcode::AssertEqCond => self.exec_assert_eq_cond(ctx),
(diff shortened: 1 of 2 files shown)


jimpo (approved):

jimpo:
Please resolve conflicts

tcoratger:
> Please resolve conflicts

Done

PR #1912, opened 2026-07-25
BINIUS-344: thread the prover's allocator through IOPProverChannel and pack_witness

benji-potamus:
Closes [BINIUS-344](https://linear.app/jimpo/issue/BINIUS-344/pool-the-packed-trace-thread-an-allocator-through-pack-witness).

The packed trace was the last large per-proof buffer still on the global allocator. At keccak-f1600 / 2^13 instances it's 64 MiB, allocated and freed on every `prove` — above glibc's `DEFAULT_MMAP_THRESHOLD_MAX`, so it is always `mmap`/`munmap`ed and re-faults every proof. This draws it from the prover's `BufferPool` instead.

> This PR previously held a deliberately not-for-merge spike, pushed so the cost of the change could be judged before committing to it. That lineage is gone; the branch is now the real change. The history and the reasoning are on the Linear issue.

## The change

`pack_witness` takes an allocator and returns a pool-drawn buffer, and the channel trait gains the allocator parameter so that buffer can actually reach the channel instead of being copied back out:

```rust
pub trait IOPProverChannel<P, A: Allocator> { ... }

fn pack_witness<P, A: Allocator>(
    alloc: &A,
    log_witness_elems: usize,
    witness: &[Word],
) -> Result<FieldVec<P, A>, Error>
```

1. **`bea13e8c`** — `ring_switch::prove` takes the packed witness as a `FieldSlice` (per review feedback), dropping its `Data` type parameter.
2. **`524ca724`** — same treatment for the logUp\* pushforward in `prove_reduction` / `prove_final_layer` / `split_halves`.
3. **`8b3c4e31`** — the allocator parameter, propagated into `BaseFoldProverChannel` and `QueuedRelation`; `pack_witness` and `ValueTable::pack` take the allocator; both provers pass their pool.

18 files across 5 crates. Commits 1 and 2 are pure widenings that stand alone; commit 3 is atomic by necessity — the moment the trait requires a pool-drawn message, every producer must already be allocator-drawn, so any finer split would have to reintroduce a copy at the boundary.

**Only the message slot is generalized.** The relation item is:

```rust
type Item = (Self::Oracle, FieldBuffer<P, A::Vec<P>>, FieldBuffer<P>, P::Scalar);
```

The transparent multilinear stays `Vec`-backed, because every transparent argument at every call site already is. That keeps `binius-math` entirely out of the change — `rs_eq_ind` and `eq_ind_partial_eval` are untouched.

Two things fell out of it:

- `finish` used to take its own allocator type parameter — a *different* allocator from the one backing the queued messages. It is now the channel's own, so they match by construction. Nothing was unsound before (a `PoolVec` carries its own pool reference), but the incoherent combination is no longer expressible.
- `spartan-prover` needed real work rather than just bounds: its three messages were all `Vec`-backed, so `pack_and_blind_witness` became allocator-drawn and the ZK transcript mask is now pooled too. This isn't optional — there's no version of the change where those stay `Vec`-backed short of copying them at the boundary.

## Measurement

keccak-f1600 @ 2^13, `--release`, `RUSTFLAGS="-C target-cpu=native"`, 6-core Linux VM, 15 proofs per run on one warm `Prover`, 3 runs on `main` and 5 on the branch.

| | `main` | branch |
|---|---|---|
| `Prepare trace` span, median | 22.78 / 24.22 / 24.40 ms | 6.07 / 6.21 / 7.26 / 6.10 / 6.36 ms |
| prove wall, median per run | 919.29 / 923.74 / 926.01 ms | 903.08 / 905.23 / 913.18 / 940.29 / 1035.77 ms |

**The span goes ~23.5 ms → ~6.2 ms, −74%**, and that reproduces on every run. That's the result.

**End-to-end is a different story and I'd rather state it plainly than sell it.** −1.8% is a modelled ceiling — span delta over total — not something observed above noise. The branch's own run-to-run spread on medians is 903–1036 ms, so a single contaminated run exceeds the entire effect. The clean-run clusters (`main` 919–926, branch 903–913) don't overlap, which is the strongest honest claim available and is weaker than statistical significance. A defensible end-to-end number needs a quiet dedicated machine and many more paired runs.

For context on why: `Commit trace` is ~306 ms of a ~910 ms proof and is untouched. Packing was never where the proof time lives. On the `binius-prover` side there's even less — `Prepare witness` is 0.35% of the proof there.

## Correctness

Proof bytes are unchanged. Fingerprinted with the blinding RNG seeded so bytes are reproducible, on paths that reach the changed code:

| path | exercises | `main` | branch |
|---|---|---|---|
| ZK sha256 | `pack_and_blind_witness`, the mask, `masks_buffer` | `0xab0a04be2e6e43f6` | `0xab0a04be2e6e43f6` |
| plain sha256 | `pack_witness` | `0xc8ac80dcc67f6982` | `0xc8ac80dcc67f6982` |
| m4 keccak | `ValueTable::pack` | `0x44c84c91dee11f04` | `0x44c84c91dee11f04` |

Each proof is verified from those exact bytes before being fingerprinted, and each is proved twice — cold pool, then warm — with an equality assert, so a proof that depended on pool state would fail rather than pass quietly. That last check is the one that matters here: pooled blocks are recycled dirty, so the fills have to leave no stale bytes from a prior proof.

This is differential evidence of behavioural identity on the shapes exercised, not exhaustive input coverage.

Full `cargo test --workspace` green; each of the three commits builds and tests green on its own. No cross-compilation matrix run — the diff contains no `target_arch`, `target_feature` or intrinsic hunks.

### DIFF
--- a/crates/m4-prover/src/prove.rs
+++ b/crates/m4-prover/src/prove.rs
@@ -353,7 +353,7 @@ impl IOPProver {
 			// The point is `r_j || r_rho || r_y`.
 			// Its instance coordinates fold the trace at `r_rho`.
 			let trace_point = [r_j, r_rho.as_slice(), r_y].concat();
-			ring_switch::prove(&trace_packed, &trace_point, channel)
+			ring_switch::prove(trace_packed.to_ref(), &trace_point, channel)
 		};
 
 		// Queue the trace opening against the ring-switch's transparent multilinear.
--- a/crates/m4-prover/src/prove.rs
+++ b/crates/m4-prover/src/prove.rs
@@ -112,7 +112,7 @@ impl IOPProver {
 	pub fn prove<P, Channel, A>(&self, table: &ValueTable, channel: &mut Channel, alloc: &A)
 	where
 		P: PackedField<Scalar = B128>,
-		Channel: IOPProverChannel<P>,
+		Channel: IOPProverChannel<P, A>,
 		A: Allocator,
 	{
 		let cs = &self.cs;
@@ -120,7 +120,7 @@ impl IOPProver {
 		// Pack the 2-D table into one multilinear and commit it as the trace oracle.
 		let trace_packed = {
 			let _scope = tracing::debug_span!("Prepare trace").entered();
-			table.pack::<P>()
+			table.pack::<P, _>(alloc)
 		};
 		let trace_oracle = {
 			let _scope = tracing::debug_span!("Commit trace").entered();
@@ -429,7 +429,9 @@ where
 	{
 		let mut channel = self
 			.basefold_compiler
-			.create_channel_without_zk_from_transcript::<StdHashSuite, Challenger_, _>(transcript);
+			.create_channel_without_zk_from_transcript::<StdHashSuite, Challenger_, _, _>(
+				transcript,
+			);
 
 		// Working buffers for this proof are drawn from the prover's pool, recycling blocks freed
 		// by earlier proofs.
@@ -644,7 +646,7 @@ impl RerandomizedOperations<'_> {
 	) -> (Vec<B128>, OperatorData<B128>, OperatorData<B128>, OperatorData<B128>)
 	where
 		P: PackedField<Scalar = B128>,
-		Channel: IOPProverChannel<P>,
+		Channel: IOPProverChannel<P, A>,
 		A: Allocator,
 	{
 		let _scope = tracing::debug_span!("Re-randomize instances").entered();
--- a/crates/prover/src/ring_switch.rs
+++ b/crates/prover/src/ring_switch.rs
@@ -13,7 +13,7 @@ use binius_field::{
 };
 use binius_ip_prover::channel::IPProverChannel;
 use binius_math::{
-	FieldBuffer, inner_product::inner_product, multilinear::eq::eq_ind_partial_eval,
+	FieldBuffer, FieldSlice, inner_product::inner_product, multilinear::eq::eq_ind_partial_eval,
 	tensor_algebra::TensorAlgebra,
 };
 use binius_utils::{
@@ -303,7 +303,7 @@ pub struct RingSwitchOutput<P: PackedField> {
 /// * `packed_witness.log_len() + log_packing == eval_point.len()` where log_packing is the base-2
 ///   log of the extension degree of B128 over B1 (= 7)
 pub fn prove<P, Channel>(
-	packed_witness: &FieldBuffer<P>,
+	packed_witness: FieldSlice<P>,
 	eval_point: &[B128],
 	channel: &mut Channel,
 ) -> RingSwitchOutput<P>
@@ -321,7 +321,7 @@ where
 
 	// Ring-switching partial evaluations (Method of Four Russians)
 	let s_hat_v = tracing::debug_span!("Compute ring-switching partial evaluations")
-		.in_scope(|| fold_1b_rows_for_b128(packed_witness, &suffix_tensor));
+		.in_scope(|| fold_1b_rows_for_b128(&packed_witness, &suffix_tensor));
 	channel.send_many(s_hat_v.as_ref());
 
 	// Basis transpose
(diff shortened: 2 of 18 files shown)


crates/prover/src/ring_switch.rs
```diff
@@ -302,13 +302,14 @@ pub struct RingSwitchOutput<P: PackedField> {
 ///
 /// * `packed_witness.log_len() + log_packing == eval_point.len()` where log_packing is the base-2
 ///   log of the extension degree of B128 over B1 (= 7)
-pub fn prove<P, Channel>(
-	packed_witness: &FieldBuffer<P>,
+pub fn prove<P, Data, Channel>(
+	packed_witness: &FieldBuffer<P, Data>,
```
jimpo: Slightly cleaner to have `packed_witness: FieldSlice<P>` to avoid the `Data` type parameter.

crates/m4-prover/src/prove.rs
```diff
@@ -356,9 +356,42 @@ impl IOPProver {
...
+		// boundary, and this copy exists only to keep the spike compiling.
+		//
+		// The copy negates the win the change is for. Measured on keccak-f1600 at 2^13 instances,
+		// where the buffer is 64 MiB: packing into a fresh `Vec` takes 26.6 ms and into a pooled
+		// block 7.5 ms, but copying the same 64 MiB back out costs about that 26.6 ms again, so
+		// the spike as it stands is *slower* than main, not faster. Even with the copy removed the
+		// span is 2.34% of a 1.16 s proof, so the ceiling is -1.8% end-to-end, under this box's
+		// +/-3% noise floor.
+		//
+		// Removing the copy is wide but mechanical, *not* type-system-blocked. The shape that
```
jimpo: Do this change

jimpo (commented):
I want to proceed with this. Prepare the PR for merge & take out of draft.

crates/prover/src/ring_switch.rs
benji-potamus: Done in `bea13e8c` — `packed_witness: FieldSlice<P>` by value, `Data` parameter gone, callers pass `.to_ref()`.

Worth noting it paid off beyond the tidier signature: the same `&FieldBuffer<P>` shape was in the logUp\* reduction too, so `524ca724` gives `prove_reduction` / `prove_final_layer` / `split_halves` the same treatment. Had those stayed as they were, both would have forced a `Data` parameter into the allocator commit — so your cleanup is what let commit 3 stay about the allocator and nothing else.

crates/m4-prover/src/prove.rs
benji-potamus: Done — this comment and the copy it described are both gone, replaced by `8b3c4e31`. `IOPProverChannel<P, A>` propagated into `BaseFoldProverChannel` and `QueuedRelation`; the pooled buffer now flows straight into `prove_oracle_relations`.

Three things I'd flag from doing it, since the estimate in the comment you were reading turned out to be off in both directions:

**Only the message slot is generalized.** `transparent` stays `Vec`-backed — every transparent argument at every call site already was — which keeps `binius-math` entirely out of the change. The comment's claim that this reaches into `binius-math` was a choice presented as a requirement.

**`spartan-prover` needed real work, not just bounds.** Its three messages were all `Vec`-backed, so `pack_and_blind_witness` became allocator-drawn and the ZK transcript mask is now pooled too. Not splittable and not optional — the alternative is copying at the boundary, which is the thing being deleted.

**One genuine wart, which then deleted itself.** `A` appears in no method signature except `prove_oracle_relations`, so it's uninferrable for the blanket `NaiveProverChannel` impl, and my first cut needed fully-qualified `IOPProverChannel::<_, GlobalAllocator>::send_oracle` at two test sites. Pinning that channel to `GlobalAllocator` instead — it's test-only, and `GlobalAllocator::Vec<T> = Vec<T>` so its signature reverts to exactly what `main` had — made `wiring.rs` drop out of the diff entirely. Worth knowing that wart exists in the design even though it costs nothing here.

Final footprint 18 files / 5 crates, against the 10/4 the comment claimed.

jimpo (approved):

PR #1914, opened 2026-07-25
perf(circuits): rotate the slice input once instead of per word

wstran:
## What

Extracting an 85-word slice out of a 135-word input costs 12,996 AND constraints. The zklogin circuit slices twice while concatenating the JWT signing payload, and those two calls alone hold 4% of its AND constraints.

**Root cause:** `slice()` builds a fresh `single_wire_multiplex` over the whole input for every output word, so an M-word slice out of N words spends `M * (N - 1)` select gates, one AND constraint each.

**Fix:** every output word reads `input[word_offset + i]` and its successor, so one dynamic rotate of the input serves all of them. `rotate_left_dynamic` runs one barrel stage per bit of the rotate amount, largest stage first so the working window shrinks as stages are applied, costing one select per position per stage.

```text
before:  M independent multiplexer trees, M * (N - 1) selects
after:   ceil(log2 N) barrel stages over a shrinking window
```

## Numbers

AND constraint counts are deterministic (`CircuitStat::collect`).

| gadget | baseline | this PR | change |
| --- | ---: | ---: | ---: |
| `slice(input=135, slice=85)` | 12,996 | 1,465 | -88.7% |
| `slice(input=64, slice=32)` | 2,584 | 504 | -80.5% |
| `slice(input=32, slice=16)` | 775 | 247 | -68.1% |
| `slice(input=16, slice=8)` | 262 | 126 | -51.9% |
| `zklogin` example | 272,492 | 258,569 | -5.1% |

With #1898 merged, this takes zklogin from 272,492 down to 258,569, under 2^18. The AND column pads to the next power of two, so the padded work halves from 524,288 to 262,144 and the prover feels it directly. Apple M3 Pro, medians of 11 runs interleaved between the two binaries so drift hits both equally:

| span | main | this PR | change |
| --- | ---: | ---: | ---: |
| BitAnd check | 16.4 ms | 11.7 ms | -28.7% |
| Commit witness | 10.2 ms | 5.2 ms | -48.7% |
| Prove | 92.3 ms | 79.3 ms | -14.1% |

Core samples do not overlap (Prove: main 91.1 to 100.0, this PR 77.6 to 87.1).

repro:

```bash
cargo run --release --example zklogin -- stat
RUST_LOG=info cargo run --release --example zklogin -- prove
```

## Correctness

`rotate_left_dynamic` is checked against `words[(shift + i) % n]` for every shift it can express, at sizes on both sides of a power of two (7, 8, 9, 16, 5, 2, 1) and with a prefix shorter than the array. Perturbing the stage order, the stage amount, or the select arms all fail that test, so it is not vacuous.

The rotate wraps where the multiplexer previously returned its out-of-range value. Both land past `len_slice`, which `slice()` already documents as unconstrained, and both callers (`jwt_claims` and the dynamic-offset branch of `concat`) pass the result straight into `assert_slice_eq`, which forces both sides to zero past the length. The output stays a fully constrained function of the input and the offset, so a prover gains no freedom.

zklogin is the only example whose statistics move. Its proof still verifies, as do `bitcoin_p2pkh` and `ethsign` end to end.

## Scope

Only the dynamic word extraction changes. The bounds checks, the byte-level alignment path, and `assert_slice_eq` are untouched. `single_wire_multiplex` keeps its other callers in `sha256`, `sha512`, `keccak`, and the ECDSA scalar multiply; a caller that needs one element rather than a run of them is still better served by the multiplexer.

### DIFF
--- a/crates/circuits/src/multiplexer.rs
+++ b/crates/circuits/src/multiplexer.rs
@@ -115,6 +115,57 @@ pub fn single_wire_multiplex(b: &CircuitBuilder, inputs: &[Wire], sel: Wire) -> 
 	current_level[0]
 }
 
+/// Rotates `words` left by a dynamic word count, returning the first `n_out` positions.
+///
+/// `out[i]` is `words[(shift + i) % words.len()]`. A caller that only needs a prefix reads the
+/// wrap-around at positions past its own validity bound, so it must treat those as unconstrained.
+///
+/// # Arguments
+/// * `b` - Circuit builder
+/// * `words` - The array to rotate
+/// * `shift` - Rotate amount in words (only `ceil(log2(len))` LSB bits are used)
+/// * `n_out` - Number of leading positions to return
+///
+/// # Implementation Details
+///
+/// One barrel stage per bit of `shift`, largest stage first so the working window shrinks as the
+/// stages are applied. That costs one select per position per stage, where a multiplexer per
+/// output position would instead cost `words.len() - 1` selects each.
+///
+/// # Panics
+/// * If `words` is empty
+/// * If `n_out` exceeds `words.len()`
+pub fn rotate_left_dynamic(
+	b: &CircuitBuilder,
+	words: &[Wire],
+	shift: Wire,
+	n_out: usize,
+) -> Vec<Wire> {
+	let n = words.len();
+	assert!(n > 0, "words must not be empty");
+	assert!(n_out <= n, "n_out ({n_out}) must not exceed words.len() ({n})");
+
+	let n_bits = log2_ceil_usize(n);
+
+	// Width the window must have before each stage, derived back from `n_out`: a stage rotating by
+	// `2^bit` reads positions `i` and `i + 2^bit`, and indices wrap, so the width caps at `n`.
+	let mut widths = vec![n_out];
+	for bit in 0..n_bits {
+		widths.push(n.min(widths[widths.len() - 1] + (1 << bit)));
+	}
+
+	let mut current = words[..widths[n_bits]].to_vec();
+	for bit in (0..n_bits).rev() {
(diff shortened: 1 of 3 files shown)


benji-potamus (approved):
Approve.

Mechanical pass: rebased onto current `main` (the only conflict was numeric drift in `zklogin.snap`, reblessed), `cargo fmt`, `cargo clippy -p binius-circuits --all-features --tests --benches --examples`, and the full `binius-circuits` test suite (330 tests) are all clean.

Correctness: beyond the suite, I ran `prove` (which proves and verifies end-to-end) for `zklogin`, `bitcoin_p2pkh`, and `ethsign` against this branch — all three succeeded, matching the PR's claim that only zklogin's statistics move and the others are unaffected. I also traced through `rotate_left_dynamic`'s shrinking-window barrel shifter by hand against a couple of the exhaustive test cases (e.g. n=8/n_out=3) and the capped-window case that `slice()` hits when `max_n_words + 1 > input.len()` (exercised by `concat.rs`'s dynamic-offset path) — the modular indexing holds up in both.

Nice writeup — the wraparound equivalence argument (multiplexer's old out-of-range behavior vs. the rotate's wraparound, both landing past `len_slice` where `slice()` already leaves things unconstrained) is exactly the right thing to call out and made this fast to verify.

jimpo (approved):
Very nice idea @wstran!! Thanks!

jimpo:
Could you please re-bless the snap.

wstran:
Rebased onto main now that #1898 is in. This PR alone now carries zklogin under 2^18: 272,492 -> 258,569 AND constraints, so the padded AND column halves. Re-measured against current main, medians of 11 interleaved runs: prove 92.3ms -> 79.3ms, BitAnd check 16.4ms -> 11.7ms. Body updated with the new numbers.

PR #1917, opened 2026-07-25
BINIUS-351: stop copying and hashing what gate fusion can index

tcoratger:
Closes [BINIUS-351](https://linear.app/jimpo/issue/BINIUS-351).

Four copies and hashes the fusion and deduplication passes do not need. Circuit compilation gets 3 to 6% faster; the compiled constraint system is unchanged.

### The graph copied every linear operand

Building the linear-expression graph cloned each linear constraint's right-hand side into its node — one allocation per fused wire, and a large circuit has hundreds of thousands.

The constraint builder already owns that operand and outlives the graph, so there is nothing to copy:

```diff
-leg.add_lin_def(lin.dst, lin.rhs.clone(), index);
+leg.add_lin_def(lin.dst, index);
```

The node keeps the position instead, and the builder is threaded to the handful of places that read an operand back.

### A dense key was hashed

A wire's node was found through a hash map, though wire identifiers are consecutive integers:

```diff
-pub wire_to_node: FxHashMap<Wire, NodeIndex>,
+pub wire_to_node: SecondaryMap<Wire, Option<NodeIndex>>,
```

### Applying patches hashed per constraint

A patch names the constraints it replaces by position. Those went into a hash set, so every surviving constraint was then looked up by hashing a key — across all four constraint lists.

Positions index a bitmap directly:

```diff
-let mut subsumes: FxHashSet<ConstraintRef> = FxHashSet::default();
-... .filter_map(|(index, c)| if subsumes.contains(&ConstraintRef::And { index }) { None } else { Some(c) })
+let mut subsumed_and = vec![false; cb.and_constraints.len()];
+retain_unsubsumed(&mut cb.and_constraints, &subsumed_and);
```

That also collapses four copy-pasted filter blocks into one helper.

### Deduplication keyed on a hash, then re-checked

Structurally identical gates were found through a map keyed by a *hash* of the structure, which then needed a full field-by-field comparison to reject collisions.

Keying on the structure itself removes both the collision path and the second comparison — and with them a wart worth naming: on a collision, `canonical.entry(hash).or_insert(gate)` kept the older entry, so the colliding gate never became canonical for its own structure and its later duplicates never collapsed.

```diff
-let mut canonical: FxHashMap<u64, Gate> = FxHashMap::default();
+let mut canonical: FxHashMap<GateStructure, Gate> = FxHashMap::default();
```

### Correctness

Every one of these is a representation change behind an unchanged decision, so the compiled constraint system is unchanged. The existing fusion and deduplication tests cover the behaviour, and the circuit statistics snapshots — which record constraint counts, distinct value indices and the value-vector layout — are untouched.

### Verification

- `binius-frontend` 158 tests — pass
- circuit statistics snapshots unchanged, so no re-blessing was needed

### Benchmark

Timing `CircuitBuilder::build()` end to end:

| circuit | before | after | change |
|---|---|---|---|
| sha256 | 17.877 ms | 16.743 ms | **-6.3%** |
| keccak | 14.947 ms | 14.519 ms | **-2.9%** |

Both at p = 0.00.


tcoratger:
Merged `main` twice to clear the conflict and land on the current base.

**The conflict.** One hunk, in `patch.rs`, on the signature line of `build_committed_lin_def_patch`. Both sides had rewritten it for unrelated reasons:

- `main` (#1965, BINIUS-386) dropped the AND-vs-Zero choice from the doc comment, since a committed linear definition now always lowers to a Zero constraint, and left the builder parameter unused as `_cb`.
- this branch reads the operand back out of the builder instead of storing a copy in the node, so `cb` is used.

Resolved as `main`'s doc wording over this branch's used `cb` parameter. Nothing else in the hunk differed.

**The near-miss.** `main` also reworked `commit_set.rs` (#1974, BINIUS-393: single-term definitions exempt from the depth cap; shift distances now accumulate along a path). That pass decides on graph edges, not on operands, so it reads nothing this branch changed and needed no adjustment.

**Verification on the merged base**

- `binius-frontend`: 191 tests pass
- `binius-examples`: 22 tests pass, and the snapshots are byte-identical to `main` — so fusion and deduplication still compile exactly the constraint system `main` does, under `main`'s new commit-set rules
- `clippy -p binius-frontend --all-targets --all-features -D warnings` clean; nightly `fmt --all --check` clean

**Re-measured after the merge**, since #1974 changed which definitions get committed and so could have moved the cost around. Timing `Sha256Example::build` + `CircuitBuilder::build()` on a 64 KiB message, 25 iterations, two binaries run interleaved:

| circuit | main | this branch | change |
|---|---|---|---|
| sha256 | 2017 ms | 1573 ms | **-22%** |
| keccak | 1743 ms | 1432 ms | **-18%** |

Medians, and the two rounds agree to within 3%. This is a larger circuit than the table in the description and it includes circuit construction, so it is a different measurement rather than a replacement for that one — the point is only that the win survives the new base.

PR #1918, opened 2026-07-25
BINIUS-352: let the comparison gates ask for a carry without computing a sum

tcoratger:
Closes [BINIUS-352](https://linear.app/jimpo/issue/BINIUS-352).

Three gates need only the carries of an addition, but the only instruction available also produces the sum — so each stored a word nothing reads, once per instance, into a slot it had to reserve. Cuts up to **12% of the scratch segment**. Constraint counts and the committed layout are unchanged.

### The problem

Unsigned less-than adds the complement of one operand to the other and reads the carry out of the top. Equality does the same with the difference of its operands. Non-zero does it with the value itself.

In all three the sum is irrelevant. Only where the addition carried matters.

But the addition instruction produces both, so each gate had to name a destination for the sum:

```rust
builder.emit_iadd_cin_cout(
    wire_to_reg(*scratch_sum_unused), // sum (unused)
    wire_to_reg(*cout),
    ...
);
```

That `scratch_sum_unused` is a real value-vector slot and a real store, paid once per instance on every witness fill. Two of the three also declared a zero constant that existed only to be passed as the carry in.

### The change

A carry-only instruction:

```diff
-builder.emit_iadd_cin_cout(sum_slot, cout, all_one, x, zero);
+builder.emit_iadd_carry(cout, all_one, x);
```

Per gate that removes a scratch slot, a store per instance, and — for two of them — a constant.

The full addition instruction had exactly one caller among the three, so nothing emits it any more. It is deleted rather than left behind, that being precisely the situation that left a pair of mask instructions as dead weight until [#1909](https://github.com/binius-zk/binius64/pull/1909).

### The saving

Freed slots are in the scratch segment, which the batched witness buffer pays for once per instance:

| circuit | scratch before | after | saved |
|---|---|---|---|
| ec_msm | 191,067 | 167,801 | 23,266 (**12.2%**) |
| bitcoin_p2pkh | 141,609 | 124,490 | 17,119 (**12.1%**) |
| ethsign | 195,586 | 172,189 | 23,397 (**12.0%**) |
| zklogin | 399,081 | 362,646 | 36,435 (**9.1%**) |
| hashsign | 262,441 | 261,793 | 648 (0.2%) |
| bitcoin_headers | 39,247 | 39,167 | 80 (0.2%) |

The seven circuits with no comparisons show no change, which is the other half of the check.

Instruction counts are unchanged throughout — one instruction is swapped for another, not removed. The per-instance store it no longer performs is not separately timed: the circuits where it would show are the expensive ones to benchmark, and the scratch figures above are the concrete measurement.

### Soundness

The scratch segment holds values no constraint operand references, so nothing the proof sees can move.

The snapshot diff bears that out: across all six changed files the **only** line that differs is the scratch count. Constraint counts, constants, and the public and committed sections are byte-identical.

### Verification

- `binius-frontend` 158 tests, `binius-circuits` 320 tests — pass
- all 13 circuit statistics snapshots re-blessed and re-checked, 13/13 matching

The comparison gates are covered end to end by the existing property tests, which build a circuit, fill a witness and verify the constraints; a wrong carry would fail them immediately.


jimpo (approved):

jimpo:
Please resolve conflicts

tcoratger:
> Please resolve conflicts

Done

PR #1919, opened 2026-07-25
BINIUS-353: drop the redundant scratch mask from the conditional-equality gate

tcoratger:
Closes [BINIUS-353](https://linear.app/jimpo/issue/BINIUS-353).

Removes a scratch wire whose value the interpreter never actually used. Pure cleanup — no behavior change, no constraint change.

### The problem

`assert_eq_cond` enforces `x = y` whenever the condition wire is MSB-bool-true.

Its witness path allocated a scratch wire, filled it with `cond ~>> 63`, and handed *that* to the interpreter.

The arithmetic shift broadcasts bit 63 across the word, so the mask is all-ones or all-zeros.

But the interpreter never reads the whole word. It asks exactly one question:

```
fn exec_assert_eq_cond(...) {
    let cond = self.read_reg();
    ...
    if ctx.load(cond, i).is_msb_true() { ... }
}
```

Broadcasting bit 63 leaves bit 63 where it was. So the mask and the condition give `is_msb_true()` the same answer, always.

Every `assert_eq_cond` was paying one instruction and one value-vector slot to compute a value tested only for a bit it already had.

### The idea

Hand the interpreter the condition itself. The mask had no other reader.

### The change

```diff
 pub const fn shape() -> OpcodeShape {
-        n_scratch: 1,
+        n_scratch: 0,
 }

-let [mask] = scratch else { unreachable!() };
-
-// Broadcast MSB: mask = cond >> 63 (arithmetic)
-builder.emit_shift(wire_to_reg(*mask), wire_to_reg(*cond), ShiftVariant::Sar, 63);
-
 builder.emit_assert_eq_cond(
-        wire_to_reg(*mask),
+        wire_to_reg(*cond),
         wire_to_reg(*x),
         wire_to_reg(*y),
         assertion_path.as_u32(),
 );
```

### What this does not touch

The AND constraint is unchanged. `constrain` still builds

```
(x ^ y) & (cond ~>> 63) == 0
```

with the shift as a *shifted operand*, which the constraint system encodes natively at no extra cost.

So the constraint side of the gate is byte-for-byte identical. Only the witness side shrinks.

### Soundness

- The emitted constraint system is unchanged, so the prover and verifier see exactly what they saw before.
- The witness path changes only which register is read, and the two registers agree on bit 63 by construction.
- No circuit's satisfiability moves, and no constraint count moves.

### Scope

- **changed:** `crates/frontend/src/compiler/gate/assert_eq_cond.rs` (one gate, ~10 lines)
- **re-blessed:** four snapshots whose circuits reach the gate transitively — through `base64` (zklogin) and the pseudo-Mersenne reduction in `bignum` (the secp256k1 circuits)
- the other nine snapshotted circuits do not use the gate and are byte-identical

### Verification

- `cargo test -p binius-frontend` — 165 passed
- all 13 example snapshots run through `check-snapshot`; the nine unaffected ones pass unchanged

### Measurements

Instruction and scratch-slot counts, from the committed snapshots:

| circuit | insns before | insns after | scratch before | scratch after | Δ each |
|---|---|---|---|---|---|
| zklogin | 550,362 | 549,282 | 399,081 | 398,001 | −1,080 |
| bitcoin_p2pkh | 178,391 | 176,775 | 141,609 | 139,993 | −1,616 |
| ec_msm | 246,270 | 244,030 | 191,067 | 188,827 | −2,240 |
| ethsign | 252,550 | 250,294 | 195,586 | 193,330 | −2,256 |

The two columns move by the same amount, which is the gate's instance count in each circuit — one instruction and one slot per instance, exactly as expected.

AND, IMUL and BMUL counts are unchanged in all four, confirming the constraint system did not move.

Peak-live scratch is also unchanged (1,543 / 109 / 190 / 192), so this is a saving in the per-instance store rather than in the pooled working set — it composes with BINIUS-346 rather than overlapping it.


jimpo (approved):

tcoratger:
@jimpo Just fixed conflicts

tcoratger:
@jimpo Just fixed the conflicts

PR #1922, opened 2026-07-25
BINIUS-356: replace the fluent constraint builders with methods that cannot lose an operand

tcoratger:
@jimpo Not sure you like this one, just a proposal, feel free to close if this is not the semantic you envision :) 

Closes [BINIUS-356](https://linear.app/jimpo/issue/BINIUS-356).

Deletes four near-identical fluent builders and turns a silent-zero footgun into a compile error. Pure refactor — no behavior change.

### The problem

`constraint_builder/builder.rs` is 225 lines of the same shape, four times: a struct holding a `&mut ConstraintBuilder` plus one `WireOperand` per slot, a setter per slot, and a `build` that pushes the record.

The repetition is the smaller issue. The real one is what an *unset* operand means.

Three of the four initialise every slot to `WireOperand::new()` — the empty XOR, i.e. **zero**:

```
fn new(builder: &mut ConstraintBuilder) -> Self {
    Self { builder, a: WireOperand::new(), b: WireOperand::new(), c: WireOperand::new() }
}
```

So a gate author who writes `.a(x).c(z)` and forgets `.b(y)` gets a perfectly well-formed constraint:

```
A & 0 == C
```

That is not a no-op. It forces `C = 0` — quietly, in a system nobody reads by hand.

`LinearConstraintBuilder` already got this right: its `dst` is an `Option<Wire>` and `build` expects it. The other three did not.

### The idea

Take every operand as a *parameter* rather than a setter.

Then the compiler counts them for you, and a deliberate zero has to say so out loud.

### The change

```diff
-pub const fn and(&mut self) -> AndConstraintBuilder<'_> { ... }
+pub fn and(&mut self, a: impl Into<WireExpr>, b: impl Into<WireExpr>, c: impl Into<WireExpr>) {
+        self.and_constraints.push(WireAndConstraint {
+                a: a.into().into_operand(),
+                b: b.into().into_operand(),
+                c: c.into().into_operand(),
+        });
+}
```

At the call sites (233 of them, across the 21 gate modules, the fusion patch pass and the tests):

```diff
-builder
-        .and()
-        .a(expr::xor3(*x, *all_one, expr::sll(*bout, 1)))
-        .b(expr::xor2(*y, expr::sll(*bout, 1)))
-        .c(expr::xor2(*bout, expr::sll(*bout, 1)))
-        .build();
+builder.and(
+        expr::xor3(*x, *all_one, expr::sll(*bout, 1)),
+        expr::xor2(*y, expr::sll(*bout, 1)),
+        expr::xor2(*bout, expr::sll(*bout, 1)),
+);
```

```diff
-builder
-        .bmul()
-        .a_lo(*a_lo).a_hi(*a_hi)
-        .b_lo(*b_lo).b_hi(*b_hi)
-        .c_lo(*c_lo).c_hi(*c_hi)
-        .build();
+builder.bmul(*a_lo, *a_hi, *b_lo, *b_hi, *c_lo, *c_hi);
```

### Was the footgun ever fired?

No — and that is worth stating plainly rather than implying.

I checked every one of the 233 converted chains: **every existing call site set every operand.** The four gates that genuinely want a zero operand (`assert_eq`, `assert_eq_cond`, `assert_zero`, `assert_false`) already wrote `expr::empty()`.

So this closes a hole before anything fell through it. `expr::empty()` is now the *only* way to express a zero operand, which is the point.

### Scope

- **deleted:** `constraint_builder/builder.rs` (225 lines, all four fluent types)
- **changed:** `constraint_builder/mod.rs` gains the four methods; 21 gate modules, `gate_fusion/patch.rs`, and the tests convert their call sites
- 28 files, **−297 net lines**
- no crate outside `binius-frontend` touched — the fluent types were re-exported from `constraint_builder`, but only within the crate

### Verification

- `cargo test -p binius-frontend` — 165 passed

The strongest evidence that nothing moved is already in the suite: `gate_fusion/tests/awhole.rs` carries **30 `insta` snapshots that pin the stringified constraint system**, operand by operand, for circuits spanning AND, IMUL, BMUL and linear constraints. All 30 pass byte-identical.

That is a tighter check than a circuit-statistics snapshot (which only compares counts), so no example snapshot run was needed.

### Merge order

Best landed **before** the gate-dispatch-table work — that restructures the same 21 gate modules, and doing this first avoids rebasing them twice.


jimpo (approved):
Yeah. It's a little harder to read, but it's fine and a super low-level API.

PR #1923, opened 2026-07-25
BINIUS-357: give the evaluation bytecode a shared opcode type instead of loose bytes

tcoratger:
Closes [BINIUS-357](https://linear.app/jimpo/issue/BINIUS-357).

Replaces the loose hex opcode bytes with one enum, so the emitter and the executor agree at compile time. Pure refactor — no behavior change, emitted bytecode is byte-identical.

### The problem

The evaluation bytecode has two ends, and they agreed only by hand.

The emitter wrote a literal:

```
pub fn emit_band(&mut self, dst: u32, src1: u32, src2: u32) {
    self.emit_u8(0x01);
    ...
}
```

The executor read one back and matched on it:

```
match opcode {
    0x01 => self.exec_band(ctx),
    ...
    _ => panic!("Unknown opcode: {:#x} at pc={}", opcode, self.pc - 1),
}
```

Nothing connected the two. The only thing between "the emitter can write this" and "the executor can run this" was that `panic!`, reached only once some circuit actually used the gate.

The wildcard arm is what made it invisible. A `match` on `u8` can never be exhaustive, so the compiler had nothing to say.

### The idea

Name the opcodes, and let dispatch match on the name.

The wildcard disappears, the match becomes exhaustive over a real type, and the compiler starts checking what a comment used to promise.

### The change

A new `eval_form/opcode.rs`:

```
#[repr(u8)]
pub enum EvalOpcode {
    Band = 0x01,
    Bor  = 0x02,
    ...
    Hint = 0x80,
}
```

The emitter writes variants:

```diff
 pub fn emit_band(&mut self, dst: u32, src1: u32, src2: u32) {
-        self.emit_u8(0x01);
+        self.emit_opcode(EvalOpcode::Band);
```

The executor decodes once, then matches on the enum:

```diff
-let opcode = self.read_u8();
-match opcode {
-        0x01 => self.exec_band(ctx),
+let byte = self.read_u8();
+let opcode = EvalOpcode::from_byte(byte)
+        .unwrap_or_else(|| panic!("Unknown opcode: {byte:#x} at pc={}", self.pc - 1));
+match opcode {
+        EvalOpcode::Band => self.exec_band(ctx),
         ...
-        _ => panic!("Unknown opcode: ..."),
 }
```

The panic survives, but its job shrank. It now only means *this byte is not an opcode* — a malformed program. It can no longer mean *the emitter knows an instruction the executor does not*.

### The guarantee, checked

Adding a variant with no executor arm:

```
error[E0004]: non-exhaustive patterns: `EvalOpcode::ZzzProbe` not covered
   --> crates/frontend/src/compiler/eval_form/exec.rs:91:10
```

### No macro — and why

An earlier draft generated the enum and its decoder from one list, so each byte appeared once. It read worse than what it replaced, and the guarantee above does not depend on it — that comes from the enum existing and dispatch matching on it.

So `from_byte` is a plain `match`, and each byte appears twice: once as a discriminant, once as a pattern. Two tests hold the pair in agreement:

- `every_opcode_decodes_back_to_itself` — every variant decodes from its own discriminant
- `no_byte_decodes_to_an_opcode_it_does_not_belong_to` — decoding is injective, so no two opcodes claim one byte

The residual gap, stated plainly: a variant added to the enum but not to `from_byte` *and* not to the tests' `ALL` list slips past. That failure is loud and immediate — `from_byte` returns `None` and the executor panics on first use. The silent failure mode, a wrong byte in one of the two places, is fully covered.

### Two unrelated tidies in `gate/mod.rs`

Both stand on their own, neither involves the opcode work:

- the `Opcode::Hint` arm was a twelve-line block inline in the match, and becomes a named `emit_hint`;
- the six assertion arms each opened with `let assertion_path = graph.assertion_names[gate];`. That hoists above the match, collapsing each arm from three lines to one.

### What this deliberately does not do

The three gate dispatches (`Opcode::shape`, `gate::constrain`, `gate::emit_gate_bytecode`) keep their explicit matches.

They are **already** exhaustive over `Opcode` — none of them has a wildcard — so a new gate variant already fails to compile in all three today. I verified that against unmodified `main` before deciding:

```
error[E0004]: non-exhaustive patterns: `Opcode::ZzzProbe` not covered
  --> crates/frontend/src/compiler/gate/mod.rs:37:8
  --> crates/frontend/src/compiler/gate/mod.rs:73:8
  --> crates/frontend/src/compiler/gate/opcode.rs:89:9
```

Generating them from a table would deduplicate a module name across four sites and cost readability, without buying a guarantee that is missing. So it is left alone.

### On the byte gaps

`0x04`, `0x22` and `0x41`–`0x45` are unassigned, and stay that way. An opcode is never reassigned, so an old bytecode dump keeps meaning one thing. The enum documents this rather than quietly filling the holes.

### Scope

- **new:** `eval_form/opcode.rs` (~150 lines, half of it the enum, half the tests)
- **changed:** `builder.rs` (21 emitters), `exec.rs` (the dispatch), `eval_form/mod.rs` (one `mod` line), `gate/mod.rs` (the two tidies)
- no crate outside `binius-frontend` touched; `EvalOpcode` is private to the module

### Verification

- `cargo test -p binius-frontend` — 168 passed (165 existing, plus the 3 new opcode tests)

No snapshot run: opcode bytes and gate dispatch are unchanged in value, so no circuit's statistics can move.

### Merge order

Touches `exec.rs` and `builder.rs`, which #1909 and #1918 also touch. Whichever lands second needs a small rebase — mechanical, since the conflict is over which opcode rows exist, not over the mechanism.

### DIFF
--- a/crates/frontend/src/compiler/eval_form/builder.rs
+++ b/crates/frontend/src/compiler/eval_form/builder.rs
@@ -4,6 +4,8 @@
 
 use binius_core::constraint_system::ShiftVariant;
 
+use super::opcode::EvalOpcode;
+
 /// Builder for constructing bytecode during circuit compilation
 pub struct BytecodeBuilder {
 	bytecode: Vec<u8>,
@@ -21,7 +23,7 @@ impl BytecodeBuilder {
 	// Bitwise operations
 	pub fn emit_band(&mut self, dst: u32, src1: u32, src2: u32) {
 		self.n_eval_insn += 1;
-		self.emit_u8(0x01);
+		self.emit_opcode(EvalOpcode::Band);
 		self.emit_reg(dst);
 		self.emit_reg(src1);
 		self.emit_reg(src2);
@@ -29,7 +31,7 @@ impl BytecodeBuilder {
 
 	pub fn emit_bor(&mut self, dst: u32, src1: u32, src2: u32) {
 		self.n_eval_insn += 1;
-		self.emit_u8(0x02);
+		self.emit_opcode(EvalOpcode::Bor);
 		self.emit_reg(dst);
 		self.emit_reg(src1);
 		self.emit_reg(src2);
@@ -37,7 +39,7 @@ impl BytecodeBuilder {
 
 	pub fn emit_bxor(&mut self, dst: u32, src1: u32, src2: u32) {
 		self.n_eval_insn += 1;
-		self.emit_u8(0x03);
+		self.emit_opcode(EvalOpcode::Bxor);
 		self.emit_reg(dst);
 		self.emit_reg(src1);
 		self.emit_reg(src2);
@@ -45,7 +47,7 @@ impl BytecodeBuilder {
 
 	pub fn emit_bxor_multi(&mut self, dst: u32, srcs: &[u32]) {
 		self.n_eval_insn += 1;
-		self.emit_u8(0x06); // Opcode for multi-way XOR
+		self.emit_opcode(EvalOpcode::BxorMulti);
 		self.emit_reg(dst);
 		self.emit_u32(srcs.len() as u32);
 		for &src in srcs {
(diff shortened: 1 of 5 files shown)


jimpo (approved):

jimpo:
Please resolve conflicts

tcoratger:
> Please resolve conflicts

Fixed

PR #1938, opened 2026-07-27
BINIUS-362: move the allocator-backed FieldBuffer constructors off the Data-parameterised impl

tcoratger:
Closes [BINIUS-362](https://linear.app/jimpo/issue/BINIUS-362).

Lets callers build an allocator-backed field buffer without naming a generic parameter the constructor never uses.

### The problem

A `FieldBuffer` is parameterised by its backing store, and two of its constructors draw that store from an allocator.
Both always return `FieldVec<P, A>` — a buffer backed by `A`'s memory — so the impl block's own store parameter appears nowhere in either signature.
It was still generic, so inference had nothing to pin it with, and every caller had to name it anyway:

```
FieldVec::<P, A>::zeros_in(alloc, n_vars)
FieldBuffer::<_, PoolVec<_>>::clone_from_slice(&alloc, src.to_ref())
```

The turbofish is pure noise: `FieldVec<P, A>` *is* `FieldBuffer<P, A::Vec<P>>`, so it restates what the allocator already determines.

### The idea

Move both constructors into the impl block whose backing store is already fixed.
The return type is then the only thing left to infer, and it follows from the allocator alone:

```
FieldBuffer::zeros_in(alloc, n_vars)
FieldBuffer::clone_from_slice(&alloc, src.to_ref())
```

### Why not the `Self`-returning shape

The neighbouring `from_values_in` solves this differently — it returns `Self` and ties the store to the allocator with `where A: Allocator<Vec<P> = Data>`.
That shape does not work for these two, and the reason is worth recording.

With the store reachable only through a projection, rustc will not resolve the inference variable eagerly, so two ordinary uses break:

- a method call on the result — `buffer.set(..)` in `expand_libra_eval`
- an argument-position use — the tuple passed to `FracAddCheckProver::new` in `logup_star/prove.rs`

Both fail with `E0283: cannot infer type of the type parameter Data`, at 3 sites in `binius-ip-prover` alone.
It would have traded a turbofish for a `let x: FieldVec<P, A> = ...` annotation, which is not a win.
So `from_values_in` is left exactly as it is, and the two constructors move one block up.

### The change

```
crates/math/src/field_buffer.rs

- impl<P: PackedField, Data: VecLike<P>> FieldBuffer<P, Data> {   // store generic -> turbofish
-     pub fn zeros_in<A: Allocator>(..) -> FieldVec<P, A>
-     pub fn clone_from_slice<A: Allocator>(..) -> FieldVec<P, A>

+ impl<P: PackedField> FieldBuffer<P> {                           // store pinned -> inferable
+     pub fn zeros_in<A: Allocator>(..) -> FieldVec<P, A>
+     pub fn clone_from_slice<A: Allocator>(..) -> FieldVec<P, A>
```

Signatures, bodies, and behavior are untouched — only the enclosing impl block changes.

### Scope

- **changed:** 7 library call sites (`binius-ip-prover` x3, `binius-prover` x2, `binius-iop-prover`, `binius-m4-prover`) and 9 bench call sites, all dropping the turbofish
- **changed:** six benches drop a now-unused `PoolVec` import
- **new:** first test coverage for either constructor — length, `log_len` carry-over, scalar equality against the source, the sub-packing-width case, and in-place writability, run against both a plain heap allocator and a recycling pool
- **untouched:** `from_values_in`, for the reason above

### Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo doc -p binius-math --document-private-items` — clean
- tests pass for `binius-math` (127, including the two new ones), `binius-ip-prover`, `binius-iop-prover`, `binius-m4-prover`, `binius-prover`, plus math doctests

### DIFF
--- a/crates/iop-prover/src/basefold/channel.rs
+++ b/crates/iop-prover/src/basefold/channel.rs
@@ -379,7 +379,7 @@ fn prove_batch_zk_basefold<A, F, P, NTT, Channel>(
 
 		let eq_tensor = eq_ind_partial_eval_scalars(&outer_challenges);
 
-		let mut combined = FieldBuffer::<P>::zeros_in(alloc, max_n);
+		let mut combined = FieldBuffer::zeros_in(alloc, max_n);
 		let mut s_prime = F::ZERO;
 		for (fri_oracle, witness_prime, eq_i, alpha_i) in
 			izip!(fri_params.input_oracles(), witness_primes, eq_tensor, alphas)
--- a/crates/ip-prover/benches/bivariate_product.rs
+++ b/crates/ip-prover/benches/bivariate_product.rs
@@ -5,7 +5,7 @@
 //! the hypercube), and a [`SharedMleCheckProver`] driving one [`QuadraticMleEvaluator`] (the
 //! evaluation claim on the product's multilinear extension at a random point).
 
-use binius_compute::{BufferPool, PoolVec};
+use binius_compute::BufferPool;
 use binius_field::{FieldOps, PackedField, arch::OptimalPackedB128};
 use binius_ip_prover::sumcheck::{
 	self,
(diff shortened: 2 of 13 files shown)


jimpo (approved):
Thanks! I'm gonna follow up with a small change.

PR #1940, opened 2026-07-27
BINIUS-363: align build_mulcheck_witness on the [..len] spare-capacity convention

tcoratger:
Closes [BINIUS-363](https://linear.app/jimpo/issue/BINIUS-363).

Makes `build_mulcheck_witness` write into its spare capacity the same way `fold_constraints` does, twenty lines above it.
Pure cleanup — no behavior change.

### The problem

`crates/spartan-prover/src/wiring.rs` has two functions that fill an allocator-backed buffer in parallel before committing its length.
BINIUS-360 (#1927) rewrote `fold_constraints` to slice the spare capacity to `[..len]`.
`build_mulcheck_witness` zips the three raw `spare_capacity_mut()` slices instead, so the same idiom reads two different ways in one file.

The `[..len]` form is not just the tidier of the two — it is the one that stays honest when capacity exceeds `len`. And it does:

```
// crates/compute/src/buffer_pool.rs
let byte_len = (capacity * size_of::<T>()).next_power_of_two().max(BUFFER_ALIGN);
let elem_cap = byte_len / size_of::<T>();
```

`BUFFER_ALIGN` is 64 bytes, so a `PoolVec` always hands out at least `64 / size_of::<P>()` words of spare capacity.
The bare-slice loop writes all of them; `set_len(len)` then commits only the first `len`.

### The change

```
 	let mut a = alloc.alloc::<P>(len);
 	let mut b = alloc.alloc::<P>(len);
 	let mut c = alloc.alloc::<P>(len);

-	(a.spare_capacity_mut(), b.spare_capacity_mut(), c.spare_capacity_mut())
+	(
+		&mut a.spare_capacity_mut()[..len],
+		&mut b.spare_capacity_mut()[..len],
+		&mut c.spare_capacity_mut()[..len],
+	)
 		.into_par_iter()
```

Plus the matching lead-in comment and the same `// SAFETY:` wording `fold_constraints` uses.

I picked the `[..len]` shape over the bare one because it is what the rest of the repo already does — roughly fifteen call sites, including `crates/prover/src/fold_word.rs:218-220`, which is this exact three-buffers-in-one-rayon-tuple case.

### Soundness

Nothing here touches the protocol.
The committed contents of `a`, `b`, `c` are unchanged: the loop writes the same value into each of the `len` slots that `set_len` exposes.

The dropped writes were the slots past `len`, and they were dead in both directions:

- **safe** — they sat inside the allocation, and `P::from_fn` returns `F::ZERO` for every `constraint_idx >= n_constraints`, so no out-of-bounds read fed them;
- **unread** — `set_len(len)` truncates below them, so no consumer ever saw them.

### Scope

- **changed:** one function in `crates/spartan-prover/src/wiring.rs`
- `fold_constraints` is untouched — it already had the `[..len]` form; this brings the other function to it
- no call sites, signatures, or tests changed

### Verification

- `cargo clippy -p binius-spartan-prover --all-targets` — no warnings
- `cargo +nightly fmt --all` — no diff beyond the change itself
- `cargo test -p binius-spartan-prover` — 7 passed, 0 failed (4 unit incl. `test_wiring_prove_verify`, 3 integration)

One caveat worth stating: those tests all run through `GlobalAllocator`, where `Vec::with_capacity(len)` yields capacity `== len` in practice, so they do not distinguish the two shapes. The over-allocating path is `PoolVec`, exercised by the prover's end-to-end tests.

### DIFF
--- a/crates/spartan-prover/src/wiring.rs
+++ b/crates/spartan-prover/src/wiring.rs
@@ -203,7 +203,13 @@ pub fn build_mulcheck_witness<A: Allocator, F: Field, P: PackedField<Scalar = F>
 	let mut b = alloc.alloc::<P>(len);
 	let mut c = alloc.alloc::<P>(len);
 
-	(a.spare_capacity_mut(), b.spare_capacity_mut(), c.spare_capacity_mut())
+	// Fill the packed words in parallel through the allocated buffers' spare capacity, then commit
+	// the lengths once every word has been written.
+	(
+		&mut a.spare_capacity_mut()[..len],
+		&mut b.spare_capacity_mut()[..len],
+		&mut c.spare_capacity_mut()[..len],
+	)
 		.into_par_iter()
 		.enumerate()
 		.for_each(|(i, (a_i, b_i, c_i))| {
@@ -231,7 +237,7 @@ pub fn build_mulcheck_witness<A: Allocator, F: Field, P: PackedField<Scalar = F>
 			}
 		});
 
-	// Safety: all entries in a, b, c are initialized in the parallel loop above.
+	// SAFETY: the loop above initialized every one of the `len` spare words of each buffer.
 	unsafe {
 		a.set_len(len);
 		b.set_len(len);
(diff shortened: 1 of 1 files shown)


jimpo:
~Ahhh, interesting. I'd actually rather just fix this at the PoolVec level so that these `[..len]` aren't necessary (it's very very not obvious why they are). We can merge this and I'll file a follow-up ticket.~

Hmm, I looked at the Vec::with_capacity documentation and it also promises *at least* that capacity. So strike what I said above ^. I'm gonna try a slightly different approach though with zipping.

jimpo (approved):

PR #1944, opened 2026-07-27
BINIUS-364: don't require a power-of-two number of constraints

benji-potamus:
Closes [BINIUS-364](https://linear.app/jimpo/issue/BINIUS-364/dont-require-power-of-two-of-constraints).

The constraint system padded its AND, IMUL, and BMUL constraint vectors up to the next power of two in `validate_and_prepare`, so every consumer saw a power-of-two count and the true count was lost. This retains the true counts and moves the rounding to the two places that actually need it.

Rebased onto current `main`, which now includes #1901 (BINIUS-340). That landed first and needs nothing from this: its padding-window skipping detects all-zero windows from the column data, and the columns this produces are byte-for-byte what the padded constraint vectors produced before. The true count is simply available now for whatever wants it.

## Commits

1. **`don't require a power-of-two number of constraints`** — the substance.
2. **`retire validate_and_prepare in favour of validate`** — with the padding gone, that method was a `&mut self` wrapper whose whole body was `self.validate()`.
3. **`name and document row counts as row counts`** — naming and comments only, no behavioral change.

## Design

The prover's operand-column builders (`build_operation_columns`, in both the single-instance and M4 provers) round the constraint axis up to a power of two and zero-fill the tail. Zero is the correct padding value: an empty operand evaluates to `Word::ZERO`, and `0 & 0 ^ 0 = 0` and `0 * 0 = 0 || 0` hold, so padded rows satisfy every constraint type.

Three new accessors on `ConstraintSystem` — `log_and_constraints`, `log_imul_constraints`, `log_bmul_constraints` — give the sumcheck variable count as `log2_ceil(n_constraints)`, and both verifier stacks derive their counts from them. `log2_ceil(0) = 0` keeps the empty-AND case at a single row with no special case.

The shift reduction's monster evaluation previously iterated its `r_x'` tensor and indexed the constraints by that position. The tensor is now the longer of the two, so `call` and `call_native` zip the constraints against it instead. The extra tensor entries correspond to padding and contribute zero, so skipping them is an identity — and it is the truncation direction that is safe: `split_at` panics rather than short-slicing, so the tensor is provably never shorter than the constraint slice.

## The proof is unchanged

For any circuit this produces a **byte-identical** proof. Padding constraints were `Default`, whose operands are empty, so the operand columns already carried zero tail rows and the key collection never saw them.

Verified by proving each of the following both ways — once from the constraint system as-is, once from one whose vectors were manually resized exactly as the deleted padding block did — and comparing finalized transcript bytes:

| Stack | True counts | Old padded | Result |
|---|---|---|---|
| prover/verifier, mixed AND+IMUL+BMUL | 6/3/3 | 8/4/4 | byte-identical, both verify |
| prover/verifier, SHA-256 preimage | 918/0/0 | 1024/0/0 | byte-identical, both verify |
| m4-prover/m4-verifier, mixed, K=8 | 6/3/3 | 8/4/4 | byte-identical, both verify |

The M4 row matters independently: `m4_prover::operand_witness` is a separate padding implementation (`ptr::write_bytes` over a `MaybeUninit` tail rather than `fill`), so the single-instance results would not have covered it.

This is strong evidence rather than exhaustive coverage — three configurations plus the argument above, not a proof over all circuit shapes. A property test over randomized constraint counts would be the natural way to broaden it.

## Serialization

`SERIALIZATION_VERSION` stays at **6**. The wire format is byte-for-byte unchanged (three `Vec`s, same order). An existing serialized system with padded constraints still deserializes, still validates — padding constraints have empty operands — and yields the same proof, since a padded count is its own power of two so every `log2_ceil` derivation agrees. `test_deserialize_from_reference_binary_file` passes against the existing v6 fixture.

## Tests

`test_prove_verify_non_power_of_two_constraint_counts` is the point of the change: a circuit with 6 AND / 3 IMUL / 3 BMUL constraints — asserted non-power-of-two in the test itself, so it cannot silently rot into a power-of-two circuit — run through both `prove_verify` and `prove_verify_zk`.

Also added: `test_validate_keeps_true_constraint_counts` and `test_log_constraint_counts_round_up` (including the `n == 0` edge) in `binius-core`; `evaluate_monster_ignores_zero_padding_constraints` and a non-power-of-two case (n=37) in the existing `call`/`call_native` consistency test; and `build_zero_pads_a_non_power_of_two_constraint_count` in `m4-prover`, replacing a `#[should_panic("Not a power of two")]` test whose invariant no longer exists.

Three `debug_assert_eq!`s in `shift::verify::operation_inputs` tie each caller's `r_x'` length to the corresponding accessor. Debug assertions are on under `cargo test`, so these execute across the suite — including all six M4 `protocol_round_trips*` tests — rather than only compiling.

Workspace totals are **1338 passed / 0 failed / 16 ignored**. This branch contributes +4 (5 added, 1 removed — `build_rejects_non_power_of_two_constraint_count`, whose invariant no longer exists); the other +9 comes from the eleven commits that landed on `main` underneath.

One visible side effect worth calling out: the tracing spans reporting constraint counts now report the **true** count rather than the padded one, so a Keccak circuit logs 38400 AND constraints where it previously logged 65536. More accurate, but it will move dashboard numbers.

## Not in scope

- **Optimizing any subprotocol** with the now-available true count. None is changed here.
- **End-to-end coverage of `n_and == 0`.** That path is newly reachable, since the constraint vector is no longer padded up to one dummy constraint. Both sides still agree on zero variables and the reduction runs over a single all-zero row — bit-identical to the previous padded-to-one behaviour — but it's verified by analysis rather than by a test. `test_prove_verify_non_power_of_two_constraint_counts` asserts a non-zero count. Worth its own ticket.
- **Value-vector shape rules.** `validate_shape`'s `PublicInputPowerOfTwo` check and the hidden-segment padding are a separate, still-required invariant.
- **`ConstraintSystem::log_and_constraints()` and its callers' local names.** Commit 3 renames the AND reduction's parameter to `log_row_count`, but constraint-framed locals that feed it survive at `verifier/src/verify.rs:211`, `prover/tests/shift.rs:235`, and `prover/benches/shift_reduction.rs:98`. They mirror the public accessor, which is the actual root of the naming; its doc is accurate, so nothing misleads, but the accessor is unevenly named. Renaming a public `binius-core` accessor is a real API change and didn't belong in a comments-only commit — worth its own ticket if we want the class closed at the root.

### DIFF
--- a/crates/core/src/constraint_system/system.rs
+++ b/crates/core/src/constraint_system/system.rs
@@ -249,15 +249,6 @@ impl ConstraintSystem {
 		Ok(())
 	}
 
-	/// [Validates][`Self::validate`] and prepares this constraint system for proving/verifying.
-	///
-	/// This function performs the following:
-	/// 1. Validates the value vector layout (including public input checks)
-	/// 2. Validates the constraints.
-	pub fn validate_and_prepare(&mut self) -> Result<(), ConstraintSystemError> {
-		self.validate()
-	}
-
 	/// Returns the number of AND constraints in the system.
 	pub const fn n_and_constraints(&self) -> usize {
 		self.and_constraints.len()
@@ -650,7 +641,7 @@ mod tests {
 			vec![ValueIndex(8)], // valid private value
 		));
 
-		let result = cs.validate_and_prepare();
+		let result = cs.validate();
 		assert!(result.is_err(), "Should reject constraint referencing padding");
 
 		match result.unwrap_err() {
@@ -681,7 +672,7 @@ mod tests {
 			vec![ShiftedValueIndex::plain(ValueIndex(13))], // hi
 		]));
 
-		let result = cs.validate_and_prepare();
+		let result = cs.validate();
 		assert!(
 			result.is_ok(),
 			"Should accept constraints with only valid references: {:?}",
@@ -690,11 +681,11 @@ mod tests {
 	}
 
 	#[test]
-	fn test_validate_and_prepare_keeps_true_constraint_counts() {
-		let mut cs = create_test_constraint_system();
+	fn test_validate_keeps_true_constraint_counts() {
+		let cs = create_test_constraint_system();
 		let (n_and, n_imul, n_bmul) =
 			(cs.n_and_constraints(), cs.n_imul_constraints(), cs.n_bmul_constraints());
-		cs.validate_and_prepare().unwrap();
+		cs.validate().unwrap();
 		assert_eq!(cs.n_and_constraints(), n_and);
 		assert_eq!(cs.n_imul_constraints(), n_imul);
 		assert_eq!(cs.n_bmul_constraints(), n_bmul);
@@ -726,7 +717,7 @@ mod tests {
 			vec![ValueIndex(8)],  // valid private value
 		));
 
-		let result = cs.validate_and_prepare();
+		let result = cs.validate();
 		assert!(result.is_err(), "Should reject constraint with out-of-range index");
 
 		match result.unwrap_err() {
@@ -758,7 +749,7 @@ mod tests {
 			vec![ShiftedValueIndex::plain(ValueIndex(100))], // hi: WAY out of range!
 		]));
 
-		let result = cs.validate_and_prepare();
+		let result = cs.validate();
 		assert!(result.is_err(), "Should reject IMUL constraint with out-of-range index");
 
 		match result.unwrap_err() {
@@ -792,7 +783,7 @@ mod tests {
 			vec![ShiftedValueIndex::plain(ValueIndex(100))], // c_hi: WAY out of range!
 		]));
 
-		let result = cs.validate_and_prepare();
+		let result = cs.validate();
 		assert!(result.is_err(), "Should reject BMUL constraint with out-of-range index");
 
 		match result.unwrap_err() {
@@ -827,7 +818,7 @@ mod tests {
 			vec![ValueIndex(8)],
 		));
 
-		let result = cs.validate_and_prepare();
+		let result = cs.validate();
 		assert!(result.is_err());
 
 		// Should get OutOfRangeValueIndex, not PaddingValueIndex
@@ -858,7 +849,7 @@ mod tests {
 			vec![ShiftedValueIndex::plain(ValueIndex(8))],
 		));
 
-		match cs.validate_and_prepare().unwrap_err() {
+		match cs.validate().unwrap_err() {
 			ConstraintSystemError::ShiftAmountTooLarge {
 				constraint_type,
 				constraint_index,
--- a/crates/core/src/constraint_system/system.rs
+++ b/crates/core/src/constraint_system/system.rs
@@ -30,6 +30,15 @@ use crate::{error::ConstraintSystemError, word::Word};
 ///
 /// A constraint may reference any value word, but not a padding word.
 ///
+/// # Constraint counts
+///
+/// The AND, IMUL and BMUL constraint counts are the true counts; none of them is rounded up to a
+/// power of two. The reductions run over power-of-two-sized operand columns, so the prover rounds
+/// each count up to a power of two and zero-fills the tail when it materializes the column;
+/// [`Self::log_and_constraints`] and its IMUL and BMUL siblings report the resulting variable
+/// count. Zero is a valid padding row for every constraint type: an empty operand evaluates to
+/// [`Word::ZERO`], and `0 & 0 ^ 0 = 0` and `0 * 0 = 0 || 0` both hold.
+///
 /// # Clone
 ///
 /// While this type is cloneable it may be expensive to do so since the constraint systems often
@@ -245,34 +254,8 @@ impl ConstraintSystem {
 	/// This function performs the following:
 	/// 1. Validates the value vector layout (including public input checks)
 	/// 2. Validates the constraints.
-	/// 3. Pads the AND, IMUL, and BMUL constraints to the next po2 size
 	pub fn validate_and_prepare(&mut self) -> Result<(), ConstraintSystemError> {
-		self.validate()?;
-
-		// Require all constraint types to have a power-of-two count. An empty IMUL (resp. BMUL)
-		// constraint set is kept at zero (rather than padded to a single dummy constraint) so the
-		// prover and verifier can skip the IntMul (resp. BinMul) reduction entirely — see
-		// `IOPProver::prove` / `IOPVerifier::verify`.
-		let and_target_size = self.and_constraints.len().next_power_of_two();
-		let imul_target_size = if self.imul_constraints.is_empty() {
-			0
-		} else {
-			self.imul_constraints.len().next_power_of_two()
-		};
-		let bmul_target_size = if self.bmul_constraints.is_empty() {
-			0
-		} else {
-			self.bmul_constraints.len().next_power_of_two()
-		};
-
-		self.and_constraints
-			.resize_with(and_target_size, AndConstraint::default);
-		self.imul_constraints
-			.resize_with(imul_target_size, ImulConstraint::default);
-		self.bmul_constraints
-			.resize_with(bmul_target_size, BmulConstraint::default);
-
-		Ok(())
+		self.validate()
 	}
 
 	/// Returns the number of AND constraints in the system.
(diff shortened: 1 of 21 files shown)


crates/m4-prover/src/operand_witness.rs
```diff
@@ -105,29 +105,36 @@ where
...
-	let mut out = alloc.alloc::<Word>(1 << (log_instances + log_constraints));
+	let total = 1 << (log_instances + log_constraints);
+	let mut out = alloc.alloc::<Word>(total);
+	// The allocator may hand back more capacity than requested, so bound the spare slice to the row
+	// count before splitting it into the constraint rows and the zero padding.
+	let (constraint_rows, padding_rows) =
+		out.spare_capacity_mut()[..total].split_at_mut(n_constraints << log_instances);
+	// Safety: `padding_rows` is a valid slice of `MaybeUninit<Word>`; `Word` has no invalid bit
+	// patterns, so writing zero bytes over it initializes every element to `Word::ZERO`.
+	unsafe { ptr::write_bytes(padding_rows.as_mut_ptr(), 0, padding_rows.len()) };
```
jimpo: Use `bytemuck::fill_zeroes`, which is safe.

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -407,6 +407,14 @@ impl<F: BinaryField> MonsterEvalFn<'_, F> {
 	where
 		E: FieldOps<Scalar = F> + From<F>,
 	{
+		// Each operation's `r_x'` section is as long as its reduction has constraint variables, and
+		// [`OperationEvalFn::split_input`] re-derives that length from the constraint count alone.
+		// The two derivations must agree or both sides mis-split the same input. An absent IntMul
```
jimpo, line 412: typos flags "mis-split". Add an appropriate exception to the Cargo.toml (anything that's mis- with a dash should be OK).

crates/verifier/src/verify.rs
```diff
@@ -493,7 +492,7 @@ where
 ///
 /// Returns an error if any sumcheck round message or the final consistency check fails.
 pub fn verify_bitand_reduction<F, C>(
-	log_constraint_count: usize,
+	log_row_count: usize,
```
jimpo: undo the variable renaming. `log_constraint_count` is fine.

crates/core/src/constraint_system/system.rs
```diff
@@ -290,6 +263,34 @@ impl ConstraintSystem {
 		self.bmul_constraints.len()
 	}
 
+	/// Returns the number of variables the BitAnd reduction runs over.
+	///
+	/// The reduction operates on operand columns with one row per AND constraint, zero-padded up
+	/// to a power of two, so it has `ceil(log2(n_and_constraints))` variables. A system with no
+	/// AND constraints still gets a single all-zero row, which every constraint type satisfies.
+	pub const fn log_and_constraints(&self) -> usize {
```
jimpo: These should return `Option<usize>` -- None if the # of constraints is 0.

crates/m4-verifier/src/verify.rs
```diff
@@ -157,7 +156,7 @@ impl IOPVerifier {
 		// The IntMul columns span every instance's constraints.
 		// So the check runs over `log_instances + log_n_imul` row variables.
 		let intmul_output = if cs.n_imul_constraints() > 0 {
-			let log_n_imul = checked_log_2(cs.n_imul_constraints());
+			let log_n_imul = cs.log_imul_constraints();
```
jimpo: log_imul_constraints should return an Option, then the if above becomes `if let Some(log_n_imul) = cs.log_imul_constraints() {`...

jimpo (changes requested):

crates/m4-prover/src/operand_witness.rs
benji-potamus: Done — now `bytemuck::fill_zeroes(padding_rows)`, `unsafe` block gone.

Two things worth flagging since they're not local to this line:

- `bytemuck` wasn't a dependency of `binius-m4-prover`, so it's been added (`bytemuck.workspace = true`). `fill_zeroes` landed in bytemuck 1.14, below the workspace floor of 1.18.0, so no version change.
- The slice here is `&mut [MaybeUninit<Word>]`, and bytemuck gates `Zeroable for MaybeUninit<T>` behind its `zeroable_maybe_uninit` feature, which the workspace didn't enable — the build failed without it. I've enabled it on the workspace dependency with a comment saying why. The feature is empty (`zeroable_maybe_uninit = []`) and gates exactly one `unsafe impl`, so it adds no deps and no codegen. Happy to scope it differently if you'd rather it not be workspace-wide.

`crates/prover/src/prove.rs:583` does the equivalent fill for the single-instance prover but already used the safe `fill(MaybeUninit::new(Word::ZERO))`, so it needed no change; I left it rather than churn it for symmetry.

I also left the pre-existing `unsafe ptr::write_bytes` at `operand_witness.rs:167` (the empty-operand stripe) alone, since it predates this PR and isn't what you commented on — happy to convert it here or in a follow-up if you'd like it to go.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done. Added to `default.extend-ignore-re` in `[workspace.metadata.typos]`:

```toml
# The hyphenated English prefix "mis-", as in "mis-split"; `typos` reads the bare "mis" as a
# misspelling of "miss". Only the prefix is exempted, so the word it attaches to is still
# checked.
"\\bmis-",
```

Exempts the prefix generally rather than special-casing `mis-split`, as you asked.

Verified against the version prek pins (typos-cli 1.32.0 from the prek cache), not the newer one on `PATH` — reproduced the CI failure with it first, confirmed exit 0 after, and checked the regex is narrow enough that `mis-recieve` still reports `recieve`. Full sweep over the tracked text files is clean.

crates/verifier/src/verify.rs
benji-potamus: Reverted to `log_constraint_count`, on both the verifier parameter here and the prover local in `crates/prover/src/and_reduction/mod.rs` — you commented on the verifier, but the two were renamed together and leaving the prover on the other name would have been incoherent. `grep log_row_count` is clean.

I kept the doc/comment text that explains what each of the two callers passes (single-instance passes `ceil(log2(n))`, M4 adds its log instance count) and that the column is zero-padded, since that's what keeps the retained name unambiguous. Only the identifier moved back. Let me know if you'd rather that prose go too.

crates/m4-verifier/src/verify.rs
benji-potamus: Done, both here and in `crates/verifier/src/verify.rs` — the `n_*_constraints() > 0` guard plus separate `log_*` call collapse into the single `if let Some(..)` binding for IntMul and BinMul, exactly as you wrote it. The tracing spans still read `n_imul_constraints()` for their field since they want the true count rather than the log.

One thing I'd like your call on, raised in the sibling thread on `system.rs` — the AND accessor doesn't have an "absent" case to signal, so `Option` there ends up unwrapped at every call site. Details there.

crates/core/src/constraint_system/system.rs
benji-potamus: Done — all three now return `Option<usize>`, `None` at zero, still `const fn`. IntMul and BinMul collapse to `if let Some(..)` in both verifier stacks, which is a clear improvement.

**On the AND one I'd like your call, because it landed awkwardly and I don't want to guess at your intent.**

BitAnd has no skip branch — an empty AND set still reduces, over the single all-zero padding row. So `None` has no `if let` to feed, and every AND call site immediately does `.unwrap_or(0)`.

The sharper way to put it: **for AND, `None` and `Some(0)` are behaviourally indistinguishable.** `n_and = 0` → `None` → 0 variables; `n_and = 1` → `Some(log2_ceil(1))` = `Some(0)` → 0 variables. The two states the `Option` exists to distinguish collapse to the same downstream value, which is why every site discards it. For IntMul/BinMul the `Option` genuinely *is* the skip signal and earns its keep; for AND it carries no information.

So: would you rather `log_and_constraints()` keep returning plain `usize`, with `Option` only on `log_imul_constraints` / `log_bmul_constraints`? Your comment landed on a line covering all three, and the sibling comment was specifically about `log_imul_constraints`, so I implemented all three as written — but I suspect the AND asymmetry wasn't in view.

Current state is behaviour-preserving either way, and I checked that rather than assuming: previously `log2_ceil_usize(0) == 0`, now `None.unwrap_or(0) == 0`, and the provers derive their row count independently from the materialised column length (`checked_log_2(n.next_power_of_two()) == log2_ceil_usize(n)` for all `n ≥ 0`, including 0 → 1 row). Both sides agree at every `n`, so no transcript moves. Tests hold at 1329/0/16.

For scale: 15 `unwrap_or(0)` sites, but only 2 are production (`verifier/verify.rs:207`, `m4-verifier/verify.rs:179`); the rest are `debug_assert_eq!`, tests, benches and fixtures.

Happy either way — say the word and I'll switch AND back to `usize`, or leave it as-is if you want the three accessors uniform.

jimpo: Leave it as is with the `unwrap_or(0)`.

jimpo (approved):

PR #1949, opened 2026-07-28
BINIUS-368: Fuse the subspace Lagrange evaluation into one allocation

tcoratger:
Closes [BINIUS-368](https://linear.app/jimpo/issue/BINIUS-368).

Fuses `lagrange_evals_scalars` — the engine under both subspace Lagrange entry points — from four allocations and ~5n multiplications into one allocation and ~4n.
Same identity, same results; 1.4–1.9× measured across verifier-realistic domain sizes.

### The problem

The routine computes the barycentric form

```
L_i(z) = w * prod_{j != i} (z - d_j)
```

where an additive subgroup admits a single shared weight `w`.
It materialized four vectors: the converted domain, prefix products, suffix products, and the combined output.
The combine step spent two multiplies per element, one of them a clone of `w`.
The same file already has the fused shape (`EvaluationDomain::lagrange_evals`); the subspace variant — 42 call sites across the workspace — never got it.

### The idea

One output vector, two sweeps:

```
result[i] = z - d_i                        // seed with the linear terms t_i
backward:  result[i] = w * prod_{j>i} t_j  // suffix sweep, accumulator seeded with w
forward:   result[i] *= prod_{j<i} t_j     // prefix sweep, terms recomputed on the fly
```

- Seeding the suffix accumulator with `w` absorbs the whole multiply-by-weight pass.
- Recomputing `t_i` in the forward sweep is free: iterating a binary subspace is an XOR walk.

### Per domain point

- **before:** 1 prefix mul + 1 suffix mul + 2 combine muls (plus a `w` clone), across 4 vectors
- **after:** 1 suffix mul + 2 forward muls, in place

→ ~20% fewer $\mathbb{F}_{2^{128}}$ multiplies and 3 fewer allocations; the measured win is larger because the dropped allocations, their zero-init, and the clones go with them.

### Soundness

- The computed value is the same barycentric formula, term for term.
- `invert_or_zero` boundary behavior and the dim-0 result (`[1]`) are unchanged.
- The existing suite pins it: the delta property `L_i(x_j)` at every domain point, partition of unity, interpolation accuracy against Horner evaluation, and a cross-check against the independent fold-based `extrapolate_over_subspace` for dims 0–6.

### Scope

- `crates/math/src/univariate.rs`, one function body (+27/−19); both entry points inherit it.
- `EvaluationDomain` and `extrapolate_over_subspace` untouched.

### Verification

- `cargo check --workspace --all-targets`, `cargo clippy -p binius-math --all-targets`, nightly `fmt` — clean.
- `cargo nextest run -p binius-math` — 128/128.

### Benchmark

Probe crates against the old and new code, two order-swapped runs each, idle M1 Pro:

| domain | old | new | speedup |
|---|---|---|---|
| dim 3 (n=8) | 354–555 ns | 315–320 ns | 1.1–1.8× |
| dim 5 (n=32) | 1.16–1.56 µs | 0.83–0.84 µs | 1.4–1.9× |
| dim 7 (n=128) | 4.66–5.40 µs | 2.80–2.97 µs | 1.6–1.9× |
| dim 8 (n=256) | 9.09–9.82 µs | 5.56–6.08 µs | 1.6× |

### DIFF
--- a/crates/math/src/univariate.rs
+++ b/crates/math/src/univariate.rs
@@ -1,6 +1,8 @@
 // Copyright 2024-2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
+use std::{iter, mem};
+
 use binius_field::{BinaryField, Field, field::FieldOps};
 use itertools::izip;
 
@@ -36,8 +38,8 @@ pub fn evaluate_univariate<F: FieldOps>(coeffs: &[F], x: F) -> F {
 /// 3. Replace inversions with multiplications for better performance
 ///
 /// # Complexity
-/// - Time: O(n) where n = subspace size, using 4n - 2 multiplications and 1 inversion
-/// - Space: O(n) for prefix/suffix arrays
+/// - Time: O(n) where n = subspace size, using about 4n multiplications and 1 inversion
+/// - Space: O(n) — the output vector is the only allocation
 ///
 /// # Parameters
 /// - `subspace`: The binary subspace defining the evaluation domain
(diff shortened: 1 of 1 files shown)


benji-potamus (approved):
Approve.

**Checks:** fmt, `clippy -p binius-math --all-targets`, and `cargo test -p binius-math` (129+2) all clean; rebases onto current `main` without conflicts.

**Correctness:** I checked the rewrite against the old implementation directly rather than relying on the existing suite — a throwaway differential test running both over dims 0–8, at 20 random points each plus every domain point (where individual linear terms vanish). Bit-identical output everywhere.

**One doc nit, non-blocking:** `crates/math/src/univariate.rs:41` goes from `4n - 2 multiplications` to `about 4n multiplications`, but those two numbers count different things. The old `4n - 2` excluded the weight fold (prefix `n-1` + suffix `n-1` + combine `2n`); the new `~4n` only holds if you include it. The sweeps themselves went `4n - 2` → `3n`, so the doc line reads as a slight regression when the code actually improved ~25%. The PR body gets the accounting right — it's just this line that swaps basis relative to the one it replaces. Suggest: "using about 3n multiplications for the sweeps plus n - 1 for the shared weight, and 1 inversion".

jimpo (approved):

PR #1951, opened 2026-07-28
BINIUS-333: derive parallel minimum task lengths from a cost model

tcoratger:
Closes [BINIUS-333](https://linear.app/jimpo/issue/BINIUS-333).

Replaces hardcoded parallel-loop minimum lengths with floors derived from a cost model, and adds floors to the hot loops that had none.
No protocol change, no behavior change — only how work is divided across workers.

### The problem

Rayon splits a loop whenever another worker is free, however small the pieces get.
Handing a slice of work to another worker costs on the order of a microsecond.
A task that runs for much less than that loses more to the handoff than it gains.

Three loops in the tree guarded against this, and each did it with a hardcoded item count:

```
crates/math/src/multilinear/fold.rs       FOLD_MIN_CHUNK        = 1 << 14
crates/ip-prover/src/fracaddcheck.rs      LAYER_BUILD_MIN_CHUNK = 1 << 13
crates/math/src/multilinear/hypercube.rs  ROUND_MIN_CHUNK       = 1 << 14
```

Those numbers are the real problem, not the loops.
A count of items is only meaningful for one packing width.
Widen `P: PackedField` and each item carries more scalars, so the same count now guards twice the work — the floor silently drifts, and nothing fails to tell you.

The remaining ~70 parallel loops had no floor at all.
Some were extreme: one swapped a single pair of elements per rayon item, nested inside another parallel region.

### The idea

Never state a length. State what one item costs, and divide a budget by it:

```
min items per task = budget per task / estimated cost per item
```

Two cost models cover every loop in this codebase.

**Loops bound by memory traffic** charge one item by the bytes it moves.
This is what makes the floor robust to the packing parameter, which was the open question on the issue:

```
min items = byte budget / size_of::<item>()
```

Double the scalars in a packed element and `size_of` doubles, so the item count halves — automatically, with no constant to revisit.
A loop zipping several streams counts all of them by naming an array type.

**Loops bound by arithmetic** charge one item by a coarse work class.
A floor only has to land within an order of magnitude, so three classes suffice.

### Calibration

The per-item costs are measured, not guessed — sequential loops over buffers far larger than cache, on an Apple M1 Pro:

| work class | measured | used |
|---|---|---|
| 1 to 3 packed binary-field multiplies per word | 1.4 to 3.9 ns | 4 ns |
| one two-to-one SHA-256 compression | 30 ns | 30 ns |
| one scalar inversion | 140 ns | 150 ns |

The proposal on the issue also had an extension-field class.
I dropped it: in this codebase the packed multiply and the extension multiply land in the same bucket, so it would have been a distinction without a difference.
Inversion is the one class that genuinely sits an order of magnitude away.

Defaults are a 100 microsecond budget per task and a 1 MiB byte budget, both overridable via `BINIUS_TASK_TARGET_NS` and `BINIUS_MIN_TASK_BYTES`.

### The change

At a call site, the count disappears and the cost appears:

```
- .with_min_len(FOLD_MIN_CHUNK)                 // 1 << 14, tuned for one width
+ .with_min_task(WorkPerItem::FieldMuls)        // one item is a few multiplies

- .into_par_iter()                              // no floor: one pair swapped per item
+ .into_par_iter().with_min_task_bytes::<T>()   // one item moves one element
```

### Scope

- **new:** `crates/utils/src/rayon/task_size.rs` — the budgets, the cost classes, and the iterator adapters
- **changed:** all three hardcoded constants are now derived
- **changed:** floors added to 10 previously unfloored hot loops across math, ip-prover, iop-prover, prover, and m4-prover
- **untouched:** loops already coarse by construction — NTT shares sized by thread count, Merkle layers, per-instance fan-outs

The adapters work under both the rayon build and the sequential shim, so a no-rayon build still compiles and behaves identically.

### Verification

- `cargo clippy` on all six touched crates, `--all-targets` — clean
- tests pass for every touched crate, under both `--features rayon` and `--no-default-features`

One pre-existing unrelated failure is worth flagging honestly: `binius-ip-prover`'s `logup_star::prove::tests::test_out_of_range_index_panics` fails on `main` as well, in code this PR does not touch.

### Benchmark

Comparing separate builds was useless here — this machine drifts about 10% run to run, enough to invent a 12% regression that vanishes on re-measurement.
The environment overrides give a clean method instead: setting both budgets to 1 floors every loop at a single item, reproducing an unfloored loop in the *same binary*, so both sides run back to back with no recompile.

Anyone can reproduce this on the existing benches:

```
BINIUS_TASK_TARGET_NS=1 BINIUS_MIN_TASK_BYTES=1 \
    cargo bench -p binius-math -- --save-baseline flooroff
cargo bench -p binius-math -- --baseline flooroff
```

The numbers below come from a sweep over sizes from 2^10 to 2^20, since the existing benches all start at 2^16 — above the floor, where it cannot bind.

| loop | size | floor off | floor on | change |
|---|---|---|---|---|
| `bit_reverse_indices` | 2^20 | 5.39 ms | 5.33 ms | −16.1% |
| `inner_product_par` | 2^10 | 884 ns | 849 ns | −7.3% |
| `inner_product_par` | 2^20 | 914 us | 899 us | −4.4% |
| `inner_product_par` | 2^12 | 3.44 us | 3.38 us | −3.8% |
| `eq_ind_partial_eval` | 2^18 | 675 us | 656 us | −2.4% |
| `fold_highest_var_inplace` | 2^10 to 2^20 | — | — | neutral |
| `fracaddcheck` / `prodcheck` | 2^12 to 2^20 | — | — | neutral (±2%) |

The last row is the important negative result: this triples the fractional-addition floor from 8192 to ~25000 items, and that costs nothing.

### Does the derived floor beat the hand-tuned one?

Rebasing picked up #1948, which tuned `ROUND_MIN_CHUNK = 1 << 14` for the two tensor-expansion loops by measurement.
This PR deletes that constant, so the derived floor has to earn its place against a number that was chosen carefully.

The overrides can reproduce `1 << 14` exactly on both loops — 65536 ns over 4 ns per word, and 524288 bytes over 32 bytes per item — so the two policies run back to back in one binary:

| expansion | hand-tuned `1 << 14` | derived | change |
|---|---|---|---|
| one cube, 2^16 | 162.1 us | 162.9 us | no change |
| inf cube, 2^16 | 78.6 us | 77.9 us | −0.7% |
| one cube, 2^20 | 2.82 ms | 2.73 ms | −3.2% |
| inf cube, 2^20 | 1.250 ms | 1.255 ms | +1.4% |
| one cube, 2^24 | 45.1 ms | 45.2 ms | no change |
| inf cube, 2^24 | 20.9 ms | 20.2 ms | −3.7% |

Neutral or better on five of six points, with one +1.4% outlier.
So nothing from #1948 is given up by replacing its constant with a derived floor.

I would not oversell the speedup — rayon's splitting is already adaptive, so it only subdivides when a worker is genuinely free, which absorbs some of what a static floor would otherwise recover.
The durable win is that no floor in the tree is tied to a packing width any more.


jimpo (approved):
Love it! 💯

PR #1952, opened 2026-07-28
BINIUS-369: cut the redundant bindings and maskings around BLAKE3's paired core

tcoratger:
Closes [BINIUS-369](https://linear.app/jimpo/issue/BINIUS-369).

Cuts the redundant bindings and maskings around BLAKE3's paired compression core. Same digests — the 7-round core is untouched.

> **Split per review:** the frontend folds that were originally here now live in #1958, reworked as a gate-graph pass. This PR is circuits only.

### The problem

A BLAKE3 compression compiles to **626** AND constraints.

Its 336 modular additions account for only 336 of them.

The rest are linear wires the fusion pass has to commit, at 1 AND each:

- **224 are structural.** The adder emits its sum as `X ^ Y ^ sll32(cout, 1)`, and BLAKE3 immediately consumes that sum under `rotr32`. A left shift and a rotate do not compose, so 4 of the 6 sums per mixing function must be committed.
- **~57** come from the fusion pass's depth guard.

I tried five re-encodings to remove that left-shifted carry — committing the carry-in word, using a rotate so the kinds match, committing the sum instead, and two ways of splitting the shift. Each is either incomplete (rejects overflow), unsound (leaves a free carry-in bit), or hits the same wall.

So the paired core is already at its floor. The slack is in the overhead *around* it, and three places waste it.

### The idea

**1. Pin each hinted chaining-value word once, not twice.**

The paired sequential compression bound each of its 8 words with two assertions plus two maskings. The two halves are disjoint, so one 64-bit equality says the same thing:

```
hinted word:  [ high lane = input cv | low lane = C1 output ]
                     must equal             must equal
                input cv << 32       ^     result >> 32
```

Shifting up discards the input's high bits and shifting down discards the result's low bits, so neither side needs a mask.

**2. Thread chunk chaining without masking the high half.**

A compression never lets a carry or a rotate cross bit 32, and the paired core reads an input chaining value's low half only (through a left shift). So the pair's first output can ride along in the high half untouched.

Only the value *leaving* the chunk needs a clean high half, because callers do read those bits: a parent node merges two children with a shift and an exclusive-or, and a digest is compared over all 64 bits. So the mask happens once at the end rather than after every pair.

**3. Run a lone trailing block on one lane.**

```
6 blocks: [0,1] [2,3] [4,5]              -> 3 paired cores
7 blocks: [0,1] [2,3] [4,5] then 6 alone -> 3 paired cores + 1 one-lane core
```

Padding to an even count with a dummy block would spend a chaining-value binding and a second lane's parameter packing on a result nothing reads.

**4. Interleave the chain step's two values directly.**

The Winternitz chain step split each 32-byte chain value into single-lane 32-bit words and packed them afterwards, masking the same halves twice. Going straight to the two-lane layout masks each half once.

### Soundness

- The merged binding is logically **equivalent** to the two assertions it replaces, not weaker. Shifting up pins the high half, shifting down pins the low half, and together they pin all 64 bits.
- Dropping the intermediate mask relies on lane independence, which the constraint gates give: carries and rotates stay inside their own 32-bit half.
- The defensive masking on the paired compression's block words **stays**. Message words are prover-supplied in the witness case, so dirty high bits must not be able to steer a lane. Dropping it was worth another ~16 AND per pair and is not safe.

### Scope

- **changed:** the paired sequential binding, the chunk's chaining and trailing block, the chain-step lane transpose.
- **left untouched:** the 7-round core, the message schedule, the hint itself.

### Verification

- `cargo check --workspace --all-targets`, clippy on the circuits crate, nightly `fmt` — clean.
- Correctness is pinned against the **specification**, not only the reference crate: the compression trace, the parent-node trace and both digests from Appendix B of `draft-aumasson-blake3-00` reproduce word for word. (Its B.2 trace is in fact unkeyed despite the section title — the chaining value is the standard initial value and the flags carry no keyed-hash bit.)
- Proptests: compression vs reference, lane independence, the paired pair vs two chained references, and the full hash vs the reference crate over random lengths spanning both block parities.
- 327 circuits tests pass.

### Benchmark

Compiled counts, from the committed snapshots:

| circuit | AND before | AND after | change | committed before | committed after |
|---|---|---|---|---|---|
| blake3 | 5,451 | 5,147 | **-5.6%** | 5,660 | 5,411 |
| hashsign | 297,984 | 295,392 | -0.9% | 297,094 | 294,504 |
| every other example | — | — | unchanged | — | — |

Only the two circuits that use these gadgets move, which is the point of the split: nothing here depends on a compiler change.

Stacked with #1958 the two compose — that PR's zero propagation is what removes the padding-word carry chains, and this one removes the bindings around them.

### DIFF
--- a/crates/circuits/src/blake3/compress.rs
+++ b/crates/circuits/src/blake3/compress.rs
@@ -198,6 +198,9 @@ fn round(builder: &CircuitBuilder, state: &mut [Wire; 16], msg: &[Wire; 16], rou
 ///
 /// All wires carry 32-bit values in their low 32 bits, matching [`blake3_compress`].
 ///
+/// - A wire feeding a lane's low half is masked here, not assumed clean.
+/// - So a caller leaving the high 32 bits dirty cannot steer either compression.
+///
 /// - `cv`: input chaining value for the first compression (8 words).
 /// - `blocks`: the two message blocks (`blocks[0]` for C1, `blocks[1]` for C2), 16 words each.
 /// - `counter`: the 64-bit block counter, shared by both compressions. Sequential chaining only
@@ -236,12 +239,6 @@ pub fn blake3_compress_2x_seq(
 	let pack = |lo: Wire, hi: Wire| builder.bxor(lo, builder.shl(hi, 32));
 	let clear = |w: Wire| clear_high_bits(builder, w, 32);
 
-	// Bind the hint's claimed first-compression input (high 32 bits of each merged CV word) to the
-	// genuine input `cv` (low 32 bits), so the high lane provably compresses the real `cv`.
-	for (merged, cv_word) in iter::zip(merged_cv, cv) {
-		builder.assert_eq("blake3_compress_2x_seq.cv_in", builder.shr(merged, 32), clear(cv_word));
-	}
-
 	let merged_block: [Wire; 16] = array::from_fn(|i| pack(clear(blocks[1][i]), blocks[0][i]));
 
 	// Both compressions share the same block counter. Sequential chaining (C2 takes C1's output
@@ -264,14 +261,26 @@ pub fn blake3_compress_2x_seq(
 		merged_flags,
 	);
 
-	// Bind the hint: the first compression's in-circuit output (high lane of `out`) must equal the
-	// chaining value the hint fed into the second compression's input (low 32 bits of `merged_cv`).
-	for (merged, out_word) in iter::zip(merged_cv, out) {
-		builder.assert_eq(
-			"blake3_compress_2x_seq.c1_out",
-			clear(merged),
-			builder.shr(out_word, 32),
-		);
+	// Bind the hinted chaining value, one 64-bit equality per word.
+	// A single equality pins both lanes, since the two halves never overlap.
+	//
+	//     bits 32..64 <- the first compression's declared input, shifted up
+	//     bits  0..32 <- the first compression's in-circuit output, shifted down
+	//
+	//     hinted word:  [ high lane = input cv | low lane = C1 output ]
+	//                          must equal              must equal
+	//                     input cv << 32     ^     result >> 32
+	//
+	// Consequences, which together leave the hint no freedom:
+	// - The high lane provably compresses the caller's own chaining value.
+	// - The low lane provably chains from what the first compression really produced.
+	//
+	// Each side is one shift of a committed word, so no masking is needed:
+	// - Shifting up discards the input's own high bits.
+	// - Shifting down discards the result's low bits.
+	for (merged, (cv_word, out_word)) in iter::zip(merged_cv, iter::zip(cv, out)) {
+		let expected = builder.bxor(builder.shl(cv_word, 32), builder.shr(out_word, 32));
+		builder.assert_eq("blake3_compress_2x_seq.merged_cv", merged, expected);
 	}
 
 	out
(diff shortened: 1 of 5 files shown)


crates/circuits/src/blake3/compress.rs
```diff
@@ -264,14 +261,26 @@ pub fn blake3_compress_2x_seq(
...
+	//                     input cv << 32     ^     result >> 32
+	//
+	// Consequences, which together leave the hint no freedom:
+	// - The high lane provably compresses the caller's own chaining value.
+	// - The low lane provably chains from what the first compression really produced.
+	//
+	// Each side is one shift of a committed word, so no masking is needed:
+	// - Shifting up discards the input's own high bits.
+	// - Shifting down discards the result's low bits.
+	for (merged, (cv_word, out_word)) in iter::zip(merged_cv, iter::zip(cv, out)) {
```
jimpo, line 281: Nice :-)

crates/circuits/src/blake3/mod.rs
```diff
@@ -134,17 +123,30 @@ pub fn blake3_chunk(
...
-		let last_odd = odd && pair + 1 == n_pairs;
-		cv = std::array::from_fn(|i| {
-			if last_odd {
-				builder.shr(out[i], 32)
-			} else {
-				clear_high_bits(builder, out[i], 32)
-			}
-		});
+		// The chaining value after the pair is the second compression's output, in the low 32 bits.
+		cv = std::array::from_fn(|i| clear_high_bits(builder, out[i], 32));
```
jimpo: Do we actually have to clear high bits? I thoughts blake3_compress_2x_seq was tolerant to garbage in the high bits. (and blake3_compress should be as well, or at least blake3_compress_2x)

crates/frontend/src/compiler/mod.rs
```diff
@@ -768,6 +784,25 @@ impl CircuitBuilder {
...
+		//     arithmetic : fills with the sign bit, which is clear
+		//     rotate     : carries zeros around
+		//
+		// Why only zero:
+		// - Returning the operand keeps the result recognisable as zero.
+		// - The zero identities on the exclusive-or and addition gates then still fire.
+		// - That is how a zero-padded message word survives being packed into a lane.
+		// - Folding any known word would instead intern one constant per shifted value.
+		// - A circuit with many distinct round constants then commits more than it saves.
+		if self.const_of(x) == Some(Word::ZERO) {
```
jimpo: Can you split the frontend changes off into a separate PR from the circuit optimizations please?

Also, is there a way to put this optimization logic in the compiler rather than the CircuitBuilder? I like the architecture where we handle this sort of thing with optimization passes on the gate graph, rather than doing all the optimizations at circuit build time. If it's too much of a pain though, we can let this through.

jimpo:
> I tried five re-encodings to remove that left-shifted carry — committing the carry-in word, using a rotate so the kinds match, committing the sum instead, and two ways of splitting the shift. Each is either incomplete (rejects overflow), unsound (leaves a free carry-in bit), or hits the same wall.

That's a good idea, lemme try something here.

crates/circuits/src/blake3/mod.rs
tcoratger: Fixed thanks!

crates/frontend/src/compiler/mod.rs
tcoratger: Done with https://github.com/binius-zk/binius64/pull/1958

tcoratger:
@jimpo Just fixed the conflicts and review comments :)

jimpo (approved):

tcoratger:
@jimpo Just fixed the remaining conflicts

PR #1953, opened 2026-07-29
BINIUS-372: add Zero constraints to the ConstraintSystem

benji-potamus:
Linear: [BINIUS-372](https://linear.app/jimpo/issue/BINIUS-372/frontend-add-zero-constraints-to-the-constraintsystem) (sub-issue of [BINIUS-370](https://linear.app/jimpo/issue/BINIUS-370/add-zero-constraints-to-the-constraintsystem)) · Spec: [binius-zk/binius.xyz#120](https://github.com/binius-zk/binius.xyz/pull/120)

Adds **Zero** constraints, the fourth native constraint form, to `binius-core`. A Zero constraint carries a single operand array and asserts that it vanishes:

```
d = ⊕_{(y, op, s) ∈ L} op(w[y], s) = 0
```

which is an arbitrary $\mathbb{F}_2$-linear relation among shifted prover words — a two-term operand equates two shifted values, a three-term operand `{x, y, z}` makes `z = x ⊕ y` native.

This is the first PR of the BINIUS-370 stack. Nothing produces or consumes Zero constraints yet: the frontend emits none (BINIUS-373 adds the compiler option), and neither protocol reduces them (BINIUS-374/375). The field exists, validates, serializes, and is checked by `verify_constraints`.

### Changes

- **`ZeroConstraint`** — an operand-array wrapper of arity 1 (operand name `d`), matching `AndConstraint`/`ImulConstraint`/`BmulConstraint`. It implements `AsRef<[Operand; 1]>`, so the arity-generic machinery downstream (`OperationEvalFn`, `build_operation_columns`, `build_key_collection`) applies to it unchanged.
- **`ConstraintSystem::zero_constraints`** — placed ahead of the AND constraints, per the spec's ordering. Wired through `validate` (operands labelled `zero`/`d`), `n_zero_constraints`, `log_zero_constraints` (`None` is the skip signal, as for IMUL/BMUL), and `ValueVecLayout::constraint_system_shape`.
- **`verify_zero_constraint`** in `binius_core::verify`, run by `verify_constraints`.
- **`SERIALIZATION_VERSION` 6 → 7**; reference binary regenerated as `constraint_system_v7.bin` and the `test_data` README refreshed.


crates/core/src/constraint_system/constraint.rs
```diff
@@ -50,6 +53,60 @@ fn deserialize_operands<const ARITY: usize>(
...
+///
+/// The operands are stored in the order given by [`ZeroConstraint::OPERAND_NAMES`].
+#[derive(Debug, Clone, Default)]
+pub struct ZeroConstraint(pub [Operand; ZERO_ARITY]);
+
+impl ZeroConstraint {
+	/// Number of operands.
+	pub const ARITY: usize = ZERO_ARITY;
+	/// Names of the operands, in storage order.
+	pub const OPERAND_NAMES: [&'static str; ZERO_ARITY] = ["d"];
```
jimpo: That's a weird name. Just call it "val"

jimpo (changes requested):

crates/core/src/constraint_system/constraint.rs
benji-potamus: Renamed to `val` — `OPERAND_NAMES`, the `val()` accessor, the constructor parameters, and the validation error label all use it now.

jimpo (approved):

PR #1955, opened 2026-07-29
BINIUS-188: add Spark timestamp buffers over the GHASH multiplicative group

benji-potamus:
Linear: [BINIUS-188](https://linear.app/jimpo/issue/BINIUS-188/spark-1-timestamp-buffers-over-the-ghash-multiplicative-group) (first subissue of [BINIUS-187](https://linear.app/jimpo/issue/BINIUS-187/spark-compiler-for-the-native-binius64-wiring-matrices), the Spark compiler epic).

Adds `binius_ip::spark::timestamps`, a pure, deterministic routine that turns a memory access pattern into the timestamp buffers for Spark's offline memory-checking argument. It lives in `binius-ip` so prover and verifier can both precompute the buffers from the public address sequences; nothing calls it yet.

`MemoryTimestamps::new(addrs, mem_size)` returns:

* `read_ts[k]` — the timestamp the cell at `addrs[k]` held when operation `k` read it,
* `final_ts[a]` — the timestamp cell `a` was last written at.

Timestamps are powers of `Field::MULTIPLICATIVE_GENERATOR`, since the integer counter `0, 1, 2, …` collapses in characteristic 2. `INIT_TS` is `g^0 = 1` for every cell, so the `Init` column is a constant and is not materialized. `write_ts()` derives the write timestamps as `read_ts × g` (Spartan optimization #3), so they never need to be committed either.

**One interpretation worth flagging.** The issue describes both a global operation counter walking `g^1, …, g^n` and `write_ts[k] = read_ts[k] × g`; those are two different conventions and only the second is implementable as stated. This takes the second: each cell counts its own accesses, which is the read-only-memory (Lasso-style) variant where multiset equality alone is sound. The `g^1, …, g^n` range is still what the precomputed powers table spans, since one cell may absorb every operation.

## Tests

* A proptest over random `(mem_size, addrs)` checks the multiset identity `H(Init)·H(WS) = H(RS)·H(Audit)` directly in-field, with the address-dropped tuple hash `h_γ1(v, t) = v·γ1 + t` that Spark actually uses. The generated cases cover empty and single-operation sequences, repeated addresses, never-accessed cells and single-cell memories.
* An exact-value test pins the per-cell access counting, including a never-accessed cell staying at `INIT_TS`.
* The multiset test was mutation-checked: dropping the `× g` from `write_ts` makes it fail.

### DIFF
--- a/crates/ip/Cargo.toml
+++ b/crates/ip/Cargo.toml
@@ -16,6 +16,7 @@ thiserror.workspace = true
 
 [dev-dependencies]
 binius-math = { path = "../math", features = ["test-utils"] }
+proptest.workspace = true
 rand.workspace = true
 sha2.workspace = true
 
--- a/crates/ip/src/lib.rs
+++ b/crates/ip/src/lib.rs
@@ -17,6 +17,7 @@
 //! - [`MultilinearEvalClaim`] - A claim that a multilinear polynomial evaluates to a value
 //! - [`sumcheck`] - Sumcheck protocol verification
 //! - [`prodcheck`] - Product check protocol verification
+//! - [`spark`] - Spark compiler for sparse multilinear polynomials
 //! - [`channel`] - IP verifier channel traits
 //!
 //! # Related crates
@@ -31,6 +32,7 @@ pub mod fracaddcheck;
 pub mod logup_star;
 pub mod mlecheck;
 pub mod prodcheck;
+pub mod spark;
 pub mod sumcheck;
 
 /// A claim that a multilinear polynomial evaluates to a specific value at a point.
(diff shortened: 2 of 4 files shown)


jimpo:
Design change in BINIUS-187.

benji-potamus:
Closing: BINIUS-188 was cancelled in favour of the logup\* redesign ([BINIUS-379](https://linear.app/jimpo/issue/BINIUS-379/spark-compiler-logup-variant-for-the-native-binius64-wiring-matrices)), which replaces the offline memory-checking argument these timestamp buffers exist to feed. The branch is left in place, so this is reopenable if the multiset-hash route comes back.

PR #1956, opened 2026-07-29
BINIUS-375: prove and verify Zero constraints in Binius64

benji-potamus:
Linear: [BINIUS-375](https://linear.app/jimpo/issue/BINIUS-375/binius64-protocol-support-for-zero-constraints) (sub-issue of [BINIUS-370](https://linear.app/jimpo/issue/BINIUS-370/add-zero-constraints-to-the-constraintsystem)) · Spec: [binius-zk/binius.xyz#120](https://github.com/binius-zk/binius.xyz/pull/120) §4.6

Teaches the Binius64 prover and verifier to discharge ZERO constraints. Follows #1953 and #1954, both merged; M4 (BINIUS-374) is next.

The reduction is free. A ZERO constraint is linear, so the polynomial whose vanishing is at stake is the constraint array's oblong-multilinearization `d_hat` itself rather than a product of witness multilinears. A multilinear that vanishes on the cube is the zero polynomial, so **one evaluation at a point the prover cannot predict certifies the whole set** — no prover message, no sumcheck round. The only cost is one more constraint array in the shift reduction's batch.

### Commit 1 — thread a Zero operand claim through the shift reduction

The shift reduction batches one claim per constraint type; this adds a fourth of arity 1.

- `Operation::Zero` in the prover's `KeyCollection` (tag `3`), and `build_key_collection` harvests `cs.zero_constraints`.
- A `zero_data` claim threaded through `shift::prove` → `phase_1` / `phase_2` / `monster`, and through the verifier's `shift::verify` / `check_eval` / `MonsterEvalFn`. `OperationEvalFn` was already arity-generic, so the monster evaluation needed no new machinery.
- Callers pass an empty claim at this commit. M4 samples the new batching lambda so its transcript stays in step with the shared `shift::verify`.

### Commit 2 — reduce Zero constraints against the BitAnd challenges

- New `binius_verifier::protocols::zero`, documenting the reduction and deriving its constraint point: ρ is the length-ℓ_zero prefix of the BitAnd zerocheck challenges, extended by fresh challenges when the ZERO set has more rows than the AND set. Both sides derive it identically and draw the extension at the same transcript position.
- `AndCheckOutput` now carries the `zerocheck_challenges` the reduction ran over — drawn *before* the reduction, unlike `eval_point` — so both sides can build ρ from them. Reuse is sound despite the correlation: the Zero bound appeals only to the marginal distribution of the challenges it consumes.
- The claimed value is the constant `0`. Nothing verifies it in isolation; the shift reduction, running against the committed witness, is what certifies it.

### Tests

`prove_verify.rs` gains a hand-built constraint system (the compiler's Zero lowering is off by default) with five ZERO constraints — XOR, shifts both directions, NOT against the all-ones constant, and one mixing in a public word — plus three AND constraints. Neither count is a power of two.

- ℓ_zero (3) > ℓ_and (2), so the reduction draws a fresh challenge past the BitAnd point; proves and verifies, plain and ZK.
- The same system with the ZERO set truncated below the AND set, so ρ is a strict prefix and nothing extra is drawn.
- **Negative:** flipping a bit of `p3 = p0 << 5` — a word no AND constraint reads — must not verify. A ZERO constraint is the only thing standing between that witness and a valid proof, so this pins that the reduction actually binds rather than passing vacuously.

### DIFF
--- a/crates/verifier/src/protocols/shift/verify.rs
+++ b/crates/verifier/src/protocols/shift/verify.rs
@@ -20,7 +20,7 @@ use binius_math::{
 use getset::Getters;
 
 use super::{
-	BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, OperationEvalFn, SHIFT_VARIANT_COUNT,
+	BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, OperationEvalFn, SHIFT_VARIANT_COUNT, ZERO_ARITY,
 	encode_operation_input, error::Error, evaluate_h_op,
 };
 
@@ -95,6 +95,8 @@ impl<F: FieldOps, const ARITY: usize> OperatorData<F, ARITY> {
 /// verification steps including PCS verification.
 #[derive(Debug, Getters)]
 pub struct VerifyOutput<F> {
+	/// Random coefficient for batching ZERO constraint evaluations.
+	zero_lambda: F,
 	/// Random coefficient for batching AND constraint evaluations.
 	bitand_lambda: F,
 	/// Random coefficient for batching IMUL constraint evaluations.
@@ -171,6 +173,7 @@ impl<F> VerifyOutput<F> {
 /// - Propagates sumcheck verification errors
 pub fn verify<F, C>(
 	constraint_system: &ConstraintSystem,
+	zero_data: &OperatorData<C::Elem, ZERO_ARITY>,
 	bitand_data: &OperatorData<C::Elem, BITAND_ARITY>,
 	intmul_data: &OperatorData<C::Elem, INTMUL_ARITY>,
 	binmul_data: &OperatorData<C::Elem, BINMUL_ARITY>,
@@ -180,11 +183,13 @@ where
 	F: BinaryField,
 	C: IPVerifierChannel<F>,
 {
+	let zero_lambda = channel.sample();
 	let bitand_lambda = channel.sample();
 	let intmul_lambda = channel.sample();
 	let binmul_lambda = channel.sample();
 
-	let eval = bitand_data.batched_eval(bitand_lambda.clone())
+	let eval = zero_data.batched_eval(zero_lambda.clone())
+		+ bitand_data.batched_eval(bitand_lambda.clone())
 		+ intmul_data.batched_eval(intmul_lambda.clone())
 		+ binmul_data.batched_eval(binmul_lambda.clone());
 
@@ -216,6 +221,7 @@ where
 	let witness_eval = channel.recv_one()?;
 
 	Ok(VerifyOutput {
+		zero_lambda,
 		bitand_lambda,
 		intmul_lambda,
 		binmul_lambda,
@@ -259,6 +265,7 @@ where
 pub fn check_eval<F, C>(
 	constraint_system: &ConstraintSystem,
 	public: &[Word],
+	zero_data: &OperatorData<C::Elem, ZERO_ARITY>,
 	bitand_data: &OperatorData<C::Elem, BITAND_ARITY>,
 	intmul_data: &OperatorData<C::Elem, INTMUL_ARITY>,
 	binmul_data: &OperatorData<C::Elem, BINMUL_ARITY>,
@@ -273,6 +280,7 @@ where
 	C::Elem: FieldOps<Scalar = F> + From<F>,
 {
 	let VerifyOutput {
+		zero_lambda,
 		bitand_lambda,
 		intmul_lambda,
 		binmul_lambda,
@@ -291,6 +299,7 @@ where
 	// materialize the result as a single inout wire instead of building the entire
 	// sub-circuit in wrapper channels.
 	let monster_eval = {
+		let zero_r_x_prime_len = zero_data.r_x_prime.len();
 		let bitand_r_x_prime_len = bitand_data.r_x_prime.len();
 		let intmul_r_x_prime_len = intmul_data.r_x_prime.len();
 		let binmul_r_x_prime_len = binmul_data.r_x_prime.len();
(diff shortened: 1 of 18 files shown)


jimpo:
Oops, I realized we should use the BitAnd output evaluation point, not its input zerocheck challenges. I updated the spec here: https://github.com/binius-zk/binius.xyz/pull/120/changes/6dd6f77f8a754aa30efb0f3b458ac46f448e474d. See commit message.

Update this PR accordingly.

benji-potamus:
Updated to close at the BitAnd sumcheck's output point. This turned out to *remove* code rather than add it: ρ now comes from `bitand_claim.r_x_prime`, which the shift reduction already had in hand, so the `AndCheckOutput::zerocheck_challenges` field the previous version needed is gone, along with the prover-side reconstruction of the deterministic prefix and the four destructuring sites in M4 that had to spell it out. Commit 2 is down from 12 changed files to 8.

`protocols::zero`'s docs now carry the plain Schwartz–Zippel argument and note why the input challenges would have needed the extra F_2-independence step. Also dropped the stale "since ℓ_zero ≥ 3" reasoning — nothing in the implementation ever enforced it, and with the output point it isn't wanted.

Behaviour is otherwise unchanged, and the tests still cover all three cases: ℓ_zero > ℓ_and (draws a fresh challenge past the BitAnd point), ℓ_zero < ℓ_and (strict prefix, draws nothing), and the negative case where corrupting a word only a ZERO constraint reads must fail to verify.

I have not made the matching change to M4 yet — that is BINIUS-374, which stacks on this. Worth noting for that one: M4's AND-check output point splits as `r_rho_and || r_x_and`, so "the BitAnd output point" there needs the constraint half specifically, and the instance half interacts with the `RerandomizedOperations` unification onto the shared `r_rho`. I will work that through against the spec and flag it on the issue if it needs a call from you.

crates/verifier/src/verify.rs
```diff
@@ -201,7 +202,7 @@ impl IOPVerifier {
 			n_constraints = self.constraint_system.n_and_constraints()
 		)
 		.entered();
-		let (r_zhat_prime, bitand_claim) = {
+		let (r_zhat_prime, bitand_claim, log_n_and) = {
 			// The BitAnd reduction has no skip branch: an empty AND set still reduces, over the
 			// single all-zero padding row, so `None` is zero variables here.
 			let log_n_constraints = self.constraint_system.log_and_constraints().unwrap_or(0);
```
jimpo: style: Pull this variable assignment above the block and use it instead of `log_n_and`.

crates/verifier/src/verify.rs
```diff
@@ -279,6 +284,22 @@ impl IOPVerifier {
...
 		};
 
+		// [phase] Verify Zero Reduction - linear constraint verification
+		//
+		// The reduction reads nothing from the transcript and runs no sumcheck: a ZERO constraint
+		// is linear, so its oblong multilinearization vanishing at one unpredictable point
+		// certifies it. The point is the one the BitAnd sumcheck just output, extended when the
+		// ZERO set has more rows; the prover draws the same extension at the same place. An empty
+		// ZERO set yields an empty claim, which contributes zero to the shift reduction.
+		let zero_claim = match self.constraint_system.log_zero_constraints() {
```
jimpo: Get rid of the match and use `log_n_zero = self.constraint_system.log_zero_constraints().unwrap_or(0);`

crates/verifier/src/verify.rs
```diff
@@ -279,6 +284,22 @@ impl IOPVerifier {
...
+			Some(log_n_zero) => {
+				let extra = channel.sample_many(zero::n_extra_challenges(log_n_zero, log_n_and));
+				let point = zero::reduction_point(&bitand_claim.r_x_prime, &extra, log_n_zero);
```
jimpo: don't sample extra here, just pass the channel to `zero::reduction_point` and sample there. This removes the `n_extra_challenges` function.

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -488,6 +519,9 @@ impl<F: BinaryField> MonsterEvalFn<'_, F> {
 
 		// Re-encode the shared tensors with each operation's `r_x'` and `lambda`. IntMul and BinMul
 		// are skipped (no input built) when they have no constraints.
+		let zero_input = (!self.constraint_system.zero_constraints.is_empty()).then(|| {
```
jimpo: Make zero constraints always-present, like bitand. (So no empty check).

jimpo (changes requested):

crates/verifier/src/verify.rs
benji-potamus: Done — `log_n_and` is now bound above the block and used inside it, so the block returns just the challenge and the claim.

crates/verifier/src/verify.rs
benji-potamus: Done, `unwrap_or(0)` with no match. Same on the prover side and in M4 (#1959).

I also updated `log_zero_constraints`'s doc comment in `binius-core`, which still claimed `None` was a skip signal "as for IMUL/BMUL" — with this change it behaves like `log_and_constraints`, so the doc now says so.

crates/verifier/src/verify.rs
benji-potamus: `n_extra_challenges` is gone, and `reduction_point` now draws the extension itself.

One deviation to flag: it takes a `impl FnMut() -> F` sampler rather than the channel. `IPProverChannel` and `IPVerifierChannel` are unrelated traits with no shared supertrait, and the prover and verifier both call this function, so a channel parameter would need either a new sampling trait (which can't be blanket-impl'd for the prover channel from `binius-verifier` without tripping the orphan rule) or a duplicated copy of the derivation on each side. A closure keeps one implementation of a soundness-critical derivation; the call sites read `zero::reduction_point(&bitand_claim.r_x_prime, log_n_zero, || channel.sample())`.

Happy to switch to a small `ChallengeSampler` trait in `binius-ip` if you'd rather have the channel passed literally — that's the version that would let both traits implement it without orphan trouble.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done — the Zero input is built unconditionally like BitAnd's, so `OperationInputs.zero` is a plain `Vec` and both `call` and `call_native` lost their match. An empty ZERO set reduces over the single all-zero padding row: `split_input` derives zero variables from the empty constraint slice, and the evaluation loop iterates nothing and contributes zero.

jimpo (approved):

PR #1959, opened 2026-07-29
BINIUS-374: prove and verify Zero constraints in M4

benji-potamus:
Linear: [BINIUS-374](https://linear.app/jimpo/issue/BINIUS-374/m4-protocol-support-for-zero-constraints) (sub-issue of [BINIUS-370](https://linear.app/jimpo/issue/BINIUS-370/add-zero-constraints-to-the-constraintsystem)) · Spec: [binius-zk/binius.xyz#120](https://github.com/binius-zk/binius.xyz/pull/120) §4.6

Last of the BINIUS-370 stack, now rebased onto main since #1956 merged — the diff is one commit over three files.

M4 batches instances, so its AND-check runs over `log_instances + log_n_and` row variables and splits its output point into an instance half and a constraint half, `r_rho_and || r_x_and`. The Zero reduction takes **the constraint half**, exactly as the single-instance reduction takes the whole point, and extends it with fresh challenges when the ZERO set has more rows than the AND set.

### Why the Zero claim skips the instance re-randomization

IntMul and BinMul close at their own instance points, so `RerandomizedOperations` transports every operand claim onto one shared `r_rho` before the witness is folded. The Zero claim does not join that step, and does not need to: a ZERO constraint array vanishes identically, so its fold over the instance axis at *any* point is still identically zero. The claim stays the constant `0` at the shared instance point without being transported there, and the reduction stays free in M4 too — no prover message, no sumcheck round, and nothing added to the re-randomization's batched sumcheck.

Concretely this is why the claim only carries a constraint point: `OperatorData` holds the constraint axis, and the instance axis is common to all operations, handled by the `r_rho` fold of the witness.

Carries the review simplifications from #1956: `log_zero_constraints().unwrap_or(0)` with no match, and `zero::reduction_point` drawing its own extension challenges.

### Tests

The frontend's Zero lowering is crate-private, and M4 builds its `ValueTable` from a `Circuit`, so the tests rewrite a compiled constraint system instead: every AND constraint of the form `X & all-1 = Y` — the default lowering of a linear constraint — becomes the equivalent ZERO constraint `X ^ Y = 0`. The two encodings constrain the same witnesses, so a table filled for the original satisfies the rewritten system unchanged.

- **Round trip with ZERO alongside AND, IMUL and BMUL.** This is the case M4 has that the single-instance protocol does not: with IMUL and BMUL present the re-randomization actually runs, so the test pins that the Zero claim coexists with it.
- **Negative:** dropping the last term of a satisfied ZERO constraint (`x ^ y ^ z = 0` → `x ^ y = 0`) must fail to verify over 64 random instances.

### DIFF
--- a/crates/m4-prover/src/prove.rs
+++ b/crates/m4-prover/src/prove.rs
@@ -44,6 +44,7 @@ use binius_verifier::{
 		bitand::AndCheckOutput,
 		intmul::IntMulOutput,
 		shift::{BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY},
+		zero,
 	},
 };
 
@@ -241,6 +242,15 @@ impl IOPProver {
 		// coordinates, the constraint index on the high coordinates.
 		let (r_rho_and, r_x_and) = eval_point.split_at(table.log_instances());
 
+		// The Zero reduction's claim, at the constraint half of the AND-check output point. See
+		// `IOPVerifier::verify` for why it skips the re-randomization below.
+		let log_n_zero = cs.log_zero_constraints().unwrap_or(0);
+		let zero_data = OperatorData {
+			evals: vec![B128::ZERO],
+			r_zhat_prime: z_challenge,
+			r_x_prime: zero::reduction_point(r_x_and, log_n_zero, || channel.sample()),
+		};
+
 		// Reduce to one shared instance point `r_rho` and the operand claims at that point.
 		//
 		// The re-randomization runs whenever IntMul or BinMul is present: BitAnd always enters,
@@ -332,6 +342,7 @@ impl IOPProver {
 				&self.key_collection,
 				&public_words,
 				&folded_witness,
+				zero_data,
 				bitand_data,
 				intmul_data,
 				binmul_data,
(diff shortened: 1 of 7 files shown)


benji-potamus:
Rebased onto main now that #1956 has merged — the conflict was this branch still carrying #1956's two commits against the squash-merged main. Down to one commit over three files (+180/-15), and the description no longer claims to be stacked.

Also carries the review simplifications from #1956 so the two sides stay symmetric: `log_zero_constraints().unwrap_or(0)` with no match, and `zero::reduction_point` drawing its own extension challenges.

M4 test suites pass on the rebase, and the format hooks are clean.

crates/m4-prover/src/prove.rs
```diff
@@ -918,6 +937,145 @@ mod tests {
...
 
+	// Rewrites every AND constraint of the form `X & all-1 = Y` into the equivalent ZERO constraint
+	// `X ^ Y = 0`.
+	//
+	// That shape is how the frontend lowers a linear constraint by default, and the two encodings
+	// constrain the same witnesses, so a table filled for the original system satisfies the
+	// rewritten one unchanged. The frontend can emit ZERO constraints directly
+	// (`Options::enable_zero_constraints`), but that knob is crate-private, so the M4 tests reach
+	// the same constraint systems this way.
+	fn lower_linear_and_constraints_to_zero(cs: &ConstraintSystem) -> ConstraintSystem {
```
jimpo: Get rid of this. Just use the compiler option to enable zero constraints (in the test). If this same pattern is in the binius-prover, remove it there too.

crates/m4-prover/src/prove.rs
```diff
@@ -241,6 +242,23 @@ impl IOPProver {
 		// coordinates, the constraint index on the high coordinates.
 		let (r_rho_and, r_x_and) = eval_point.split_at(table.log_instances());
 
+		// The Zero reduction writes nothing to the transcript and runs no sumcheck: a ZERO
```
jimpo: This comment is on the verifier side, you don't need it on the prover side as well.

jimpo (changes requested):

crates/m4-prover/src/prove.rs
benji-potamus: Gone. Both M4 tests and the three `binius-prover` ones now build circuits with `Options { enable_zero_constraints: true, .. }`; the rewrite helper and the hand-built `ConstraintSystem` in `prove_verify.rs` are both deleted.

That needed the knob to be reachable, so `Options` and `CircuitBuilder::with_opts` are now `pub`, with every field documented and a doc-example on the struct. Shout if you'd rather expose something narrower than the whole pass set.

Two things the switch turned up:

- **Gate fusion has to be off for the fixtures to emit anything.** It inlines a linear definition into the gate that consumes it, so with fusion on these circuits produced *zero* ZERO constraints even with the option enabled — `force_commit` doesn't hold one back. The tests set `enable_gate_fusion: false` with a comment saying why.
- The hand-built system had ZERO constraints over *shifted* operands, which a plain `bxor` chain doesn't reach. `rotr` gates emit their own linear constraints, so the helper takes separate `n_xor`/`n_rotr` counts and asserts at least one ZERO constraint carries a shifted term.

The negative test no longer hardcodes a witness index: it picks a word that appears in some ZERO constraint and in no AND constraint, so it still fails for the right reason.

crates/m4-prover/src/prove.rs
benji-potamus: Trimmed to a one-line pointer at `IOPVerifier::verify`. Did the same to the binius64 prover's copy in `crates/prover/src/prove.rs`, which had the identical duplication from #1956.

jimpo (approved):

PR #1960, opened 2026-07-29
BINIUS-380: cut the doubled hint binding and per-pair masking around SHA-256's paired core

tcoratger:
Closes [BINIUS-380](https://linear.app/jimpo/issue/BINIUS-380).

Cuts the redundant bindings and maskings around SHA-256's paired compression core. Same digests — the 64-round core is untouched.

Applies the two patterns reviewed and merged for BLAKE3 in #1952 to SHA-256, which was never updated.

### The core is already at its floor

A single compression compiles to **912** AND constraints:

- **600** modular additions — 144 schedule, 448 rounds, 8 feed-forward.
- **128** for choose and majority, both already in their one-constraint form (`z ^ (x & (y ^ z))` and `((x ^ z) & (y ^ z)) ^ z`).
- **~176** for sums the fusion pass must commit. A sum's carry term is left-shifted, the next round consumes it under a rotate, and those two shift kinds do not compose. 128 of these are `a` and `e` feeding the big sigmas; 46 are schedule words feeding the small ones.

I checked the add trees against the bit-equation bound: a k-operand 32-bit add needs exactly k-1 AND constraints, and carry-save gives no advantage because a majority costs the same as a full add. Nothing to win there.

So the slack is in the overhead *around* the core.

### 1. The paired core bound its hinted state twice

Each of the 8 words cost two assertions plus two maskings. The two halves are disjoint, so one 64-bit equality says the same thing:

```
hinted word:  [ high lane = input state | low lane = S1 output ]
                     must equal                must equal
                input state << 32       ^     result >> 32
```

Shifting up discards the input's own high bits; shifting down discards the result's low bits. Neither side needs a mask.

### 2. The fixed-length gadget masked the chaining state every pair

The paired core tolerates a dirty high half:

- Its hint truncates each input word to 32 bits.
- Its binding shifts the input state up, discarding the high bits anyway.
- A compression never lets a carry or a rotate cross bit 32, so the lanes stay apart.

So the paired output is threaded straight into the next pair, and only the escaping digest is masked — a caller compares all 64 bits.

A single-block message skips the mask entirely: it never enters the paired core, and the one-lane core maps an empty high half to an empty high half. Without that guard the change would have *added* 8 gates to every single-block circuit.

### Soundness

- The merged binding is logically **equivalent** to the two assertions it replaces, not weaker. Shifting up pins the high half, shifting down pins the low half, and together they pin all 64 bits.
- Dropping the intermediate mask rests on lane independence, which the gates give: `iadd_32`, `rotr32` and `srl32` all stay inside their own 32-bit half.
- The defensive masking on the paired core's **block** words stays. Message words are prover-supplied, so dirty high bits must not be able to steer a lane.

### Scope

- **changed:** the paired sequential binding, and the fixed-length gadget's chaining.
- **left untouched:** the 64-round core, the message schedule, the hint itself.
- **left untouched on purpose:** the variable-length gadget keeps its per-block mask, because its masked state feeds the digest packer, which does need a clean low half. It still inherits the binding fix.

### Verification

- `cargo check --workspace --all-targets`, clippy on the circuits crate, nightly `fmt` — clean.
- 327 circuits tests pass.
- The two-block known-answer test now pins the **chaining value handed between the blocks** (`85e655d6 417a1795 3363376a 624cde5c 76e09589 cac5f811 cc4b32c1 f20e533a`), not only the final digest. That boundary is exactly what the merged binding constrains, and it was previously checked only against our own reference.

### Benchmark

Compiled counts, from the committed snapshots and gadget-level counts:

| gadget / circuit | AND before | AND after | change | committed before | committed after |
|---|---|---|---|---|---|
| paired sequential compression | 998 | 974 | -2.4% | — | — |
| `sha256_fixed` (1 KiB) | 8,856 | 8,608 | **-2.8%** | — | — |
| `sha256_fixed` (4 KiB) | 32,808 | 31,792 | **-3.1%** | — | — |
| sha256 example | 8,856 | 8,608 | **-2.8%** | 9,189 | 9,005 |
| bitcoin_headers example | 20,036 | 19,868 | -0.8% | 20,167 | 20,079 |
| zklogin example | 356,617 | 356,369 | -0.07% | 359,159 | 358,999 |
| every other example | — | — | unchanged | — | — |

Gate count and distinct shifted value indices fall alongside AND in all three, so nothing is traded. The zklogin gain is small because it uses the variable-length gadget, which only picks up the binding fix.

### DIFF
--- a/crates/circuits/src/sha256/mod.rs
+++ b/crates/circuits/src/sha256/mod.rs
@@ -126,15 +126,20 @@ pub fn sha256_fixed(builder: &CircuitBuilder, message: &[Wire], len_bytes: usize
 	let mask32 = builder.add_constant(Word::MASK_32);
 	let mut state = State::iv(builder);
 	let mut block_idx = 0;
+	// The threaded state carries the pair's first compression in its high half.
+	// That half is left as it is rather than masked off, since nothing downstream reads it:
+	//
+	// - A compression never lets a carry or a rotate cross bit 32, so the halves stay apart.
+	// - The paired core takes an input state's low half only, through a left shift.
+	//
+	// So the low half of a result depends on the low halves of its inputs alone.
 	while block_idx + 1 < n_blocks {
-		let out = sha256_compress_2x_seq(
+		// The chaining state after the pair is the second compression's output, in the low half.
+		state = sha256_compress_2x_seq(
 			&builder.subcircuit(format!("sha256_fixed_compress[{block_idx}..{}]", block_idx + 2)),
 			state,
 			[blocks[block_idx], blocks[block_idx + 1]],
 		);
-		// The chaining state after the pair is the second compression's output.
-		// It sits in the low 32 bits of each word; the mask restores the empty high half.
-		state = State::new(std::array::from_fn(|i| builder.band(out.0[i], mask32)));
 		block_idx += 2;
 	}
 	if block_idx < n_blocks {
@@ -145,8 +150,16 @@ pub fn sha256_fixed(builder: &CircuitBuilder, message: &[Wire], len_bytes: usize
 		);
 	}
 
-	// Return the final state as 8 32-bit words
-	state.0
+	// The escaping digest is the one place a clean high half is required.
+	// So mask once here rather than after every pair, since a caller compares all 64 bits.
+	//
+	// Why a single-block message skips it:
+	// - Such a message never enters the paired core.
+	// - The one-lane core maps an empty high half to an empty high half.
+	if n_blocks < 2 {
+		return state.0;
+	}
+	std::array::from_fn(|i| builder.band(state.0[i], mask32))
 }
 
 /// Computes the SHA-256 hash of a variable-length message.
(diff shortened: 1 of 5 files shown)


crates/circuits/src/sha256/mod.rs
```diff
@@ -145,8 +150,16 @@ pub fn sha256_fixed(builder: &CircuitBuilder, message: &[Wire], len_bytes: usize
...
+	// The escaping digest is the one place a clean high half is required.
+	// So mask once here rather than after every pair, since a caller compares all 64 bits.
+	//
+	// Why a single-block message skips it:
+	// - Such a message never enters the paired core.
+	// - The one-lane core maps an empty high half to an empty high half.
+	if n_blocks < 2 {
+		return state.0;
+	}
+	std::array::from_fn(|i| builder.band(state.0[i], mask32))
```
jimpo, line 162: nit: better to use clear_high_bits. It doesn't create an AND constraint.

jimpo (approved):

PR #1961, opened 2026-07-29
Check BMUL constraints in verify_constraints

pauldelucia:
`verify_constraints` walks the constants section, the AND constraints, and the IMUL constraints, then returns `Ok` — it never looks at `bmul_constraints`. Any test that checks a bmul gadget through this function therefore under-checks: a witness carrying a wrong GHASH product passes.

Found while writing negative tests for a bmul-based gadget — the corrupted witnesses were accepted.

## The change

`verify_bmul_constraint` checks `A * B = C` in the GHASH field, with each element carried by its `(lo, hi)` word pair, and `verify_constraints` gains the third loop.

The multiply is a self-contained shift-and-reduce over `u128` rather than a call into `binius-field`: `binius-core` doesn't depend on the field crate, and an independent restatement of the reduction is what makes the check worth having. Its convention is pinned against `BinaryField128bGhash` in the tests.

## Testing

`ghash_mul_matches_field_arithmetic` pins the products (`X * X`, the `X^63`/`X^127` boundary cases, and one random pair) against values computed with `BinaryField128bGhash`. `bmul_constraint_accepts_correct_product` and `bmul_constraint_rejects_wrong_product` run a one-constraint system through `verify_constraints`, the second flipping a single product bit that passed before.

### DIFF
--- a/crates/core/src/verify.rs
+++ b/crates/core/src/verify.rs
@@ -6,7 +6,8 @@
 
 use crate::{
 	constraint_system::{
-		AndConstraint, ConstraintSystem, ImulConstraint, ValueIndex, ValueVec, ZeroConstraint,
+		AndConstraint, BmulConstraint, ConstraintSystem, ImulConstraint, ValueIndex, ValueVec,
+		ZeroConstraint,
 	},
 	word::Word,
 };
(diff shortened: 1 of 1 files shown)


jimpo (approved):
Oof major gap! Thanks!

I think this probably got dropped hung on the decision of whether to make core depend on binius-field. I think it's time.

@benji-potamus Follow up to do that

benji-potamus:
Filed [BINIUS-383](https://linear.app/jimpo/issue/BINIUS-383/let-binius-core-depend-on-binius-field) to track making `binius-core` depend on `binius-field` and swapping `ghash_mul` here for the real `BinaryField128bGhash` multiply.

PR #1962, opened 2026-07-29
BINIUS-382: mask low 32 bits with clear_high_bits instead of a bitand

benji-potamus:
Closes [BINIUS-382](https://linear.app/jimpo/issue/BINIUS-382).

Replaces every `band` against a 32-bit low mask with `clear_high_bits(builder, w, 32)`, so masking
to a word's low half is spelled one way across the circuit library.

> [!IMPORTANT]
> **This does not pay off yet, and costs a little today.** The premise on the issue was that
> `clear_high_bits` avoids an AND constraint. It does not — not under the default options, and not
> under `enable_zero_constraints` either. Measurements and the reason are below. Opening it anyway
> per @jimpo's call that the win lands once linear constraints lower to Zero constraints; the
> compiler is not there yet.

### Sites converted

| file | sites |
|---|---|
| `sha256/mod.rs` | 3 — the fixed-length escaping digest (added by #1960), the var-length BE split, the var-length paired chaining |
| `hash_based_sig/hashing/chain_blake3.rs` | 3 — `split_u32_words`, `interleave_cv_2x`, `deinterleave_cv_2x` |
| `bitcoin/double_sha256.rs`, `bitcoin/p2pkh_signature.rs`, `blake2s/mod.rs`, `blake3/compress.rs`, `popcount.rs`, `examples/circuits/bip32.rs` | 1 each |

`binius_circuits::util` becomes `pub` so `bip32`, in the examples crate, can reach the gadget.
In `chain_blake3.rs` the only surviving use of `LOW32_MASK` was `!LOW32_MASK`, so the constant is
now spelled `HIGH32_MASK` directly.

### Why it costs constraints today

`clear_high_bits` is `shr(shl(w, n), n)`. Both shifts are linear constraints, but `Sll(32)` and
`Srl(32)` do not compose into a single shifted operand (`constraint_builder/shift.rs:229`), so the
intermediate has to be committed. Gate fusion then lowers that committed linear definition to an
**AND constraint** against all-ones — `build_committed_lin_def_patch` builds a
`NonLinearConstraint::And` unconditionally (`gate_fusion/patch.rs:214-229`).

So the shift pair costs the same one AND constraint the `band` did. The regression on top is that
`band(x, mask)` could absorb the surrounding shift/XOR chain into its own operands, whereas the
committed `sll` intermediate cannot.

### Benchmark: compiled circuit stats (default options)

Baseline is the committed snapshots at `main`. Blessed snapshots are in this PR.

| example | AND before | AND after | change | committed before | committed after |
|---|---|---|---|---|---|
| hashsign | 274,964 | 275,951 | **+987** (+0.36%) | 274,075 | 275,062 |
| bitcoin_headers | 19,868 | 20,038 | +170 (+0.86%) | 20,079 | 20,248 |
| zklogin | 356,369 | 356,445 | +76 (+0.02%) | 358,999 | 359,074 |
| bitcoin_p2pkh | 113,474 | 113,485 | +11 | 135,880 | 135,890 |
| sha256 | 8,608 | 8,616 | +8 | 9,005 | 9,012 |
| blake2s | 13,656 | 13,656 | unchanged | 13,829 | 13,829 |
| ethsign, sha512, keccak, blake2b, blake3, blake3_compress, ec_msm | — | — | unchanged | — | — |

`blake2s` is the shape the issue expected: +128 gates, but AND and committed both unchanged — there
the shift pair fuses away entirely.

### The Zero-constraint option does not flip it

I rebuilt with `enable_zero_constraints: true` (`compiler/mod.rs:77`) to check where the payoff is:

| example | ZERO | AND (flag on) | AND (flag off) |
|---|---|---|---|
| sha256 | 0 | 8,616 | 8,616 |
| hashsign | 0 | 275,951 | 275,951 |
| bitcoin_headers | 0 | 20,038 | 20,038 |

Zero constraints stay at **0** and nothing moves, because the flag is read in
`ConstraintBuilder::build`, which runs *after* gate fusion has already promoted every committed
linear definition to an AND constraint. No linear constraint survives for the flag to act on.

Realising the win therefore needs `build_committed_lin_def_patch` (and the `commit_set` decision
that feeds it) to emit a Zero constraint when the option is on — a compiler change, not a flag
flip. Happy to open that as a follow-up issue if you want this landed as the call-site half.

### Scope

- **converted:** every `band` against a compile-time 32-bit low mask in library and example code.
- **left alone:** low masks of other widths — `(1 << k) - 1` in `chain_blake3::zeroed_u32_words`,
  `ripemd`, `concat`, `blake2b`, `keccak`, and the uniform 32/16/8/4/2 ladder in `float64/utils.rs`,
  where converting only the 32 stage would break the pattern. The same transform generalises to all
  of them, but that is a wider change than this issue asked for.
- **left alone:** high-half masks, which want a `clear_low_bits` counterpart rather than this one.
- **left alone:** the two test-only sites (`sha256/compress.rs:479`, `prover/tests/prove_verify.rs:111`).

### Verification

- 327 circuits tests pass; the new `clear_high_bits` doctest passes.
- `cargo clippy -p binius-circuits -p binius-examples --tests --benches --examples -- -D warnings` clean.
- `cargo +nightly-2026-07-01 fmt --check` clean.
- All 13 snapshots re-blessed; 6 changed.
- The stale claim on `clear_high_bits` that the shift pair "is cheaper than masking with a `band`"
  is corrected to describe what the compiler actually does.

### DIFF
--- a/crates/circuits/src/bitcoin/double_sha256.rs
+++ b/crates/circuits/src/bitcoin/double_sha256.rs
@@ -1,11 +1,11 @@
+// Copyright 2026 The Binius Developers
 // Copyright 2025 Irreducible Inc.
 //! The Bitcoin double-SHA256 hash function.
 
-use binius_core::word::Word;
 use binius_frontend::{CircuitBuilder, Wire, WitnessFiller};
 use sha2::Digest;
 
-use crate::{bytes::swap_bytes_32, sha256::sha256_fixed};
+use crate::{bytes::swap_bytes_32, sha256::sha256_fixed, util::clear_high_bits};
 
 /// Retained only so callers can obtain the reference digest via [`Self::populate_inner`].
 ///
@@ -27,8 +27,6 @@ impl DoubleSha256 {
 		message: &[Wire],
 		digest: [Wire; 4],
 	) -> Self {
-		let mask32 = builder.add_constant(Word::MASK_32);
-
 		// First SHA-256. `message` is little-endian 8-byte wires; `sha256_fixed` consumes
 		// big-endian 32-bit schedule words, so byte-swap within each 32-bit half and split each
 		// 64-bit wire into its two schedule words (mirrors `sha256::sha256_varlen`'s input
@@ -36,7 +34,7 @@ impl DoubleSha256 {
 		let mut message_be: Vec<Wire> = Vec::with_capacity(message.len() * 2);
 		for &w in message {
 			let swapped = swap_bytes_32(builder, w);
-			message_be.push(builder.band(swapped, mask32));
+			message_be.push(clear_high_bits(builder, swapped, 32));
 			message_be.push(builder.shr(swapped, 32));
 		}
 		let digest_0_be = sha256_fixed(builder, &message_be, message.len() * 8); // [Wire; 8] BE
--- a/crates/circuits/src/bitcoin/p2pkh_signature.rs
+++ b/crates/circuits/src/bitcoin/p2pkh_signature.rs
@@ -11,6 +11,7 @@ use crate::{
 	ripemd::ripemd160_fixed,
 	secp256k1::{Secp256k1, Secp256k1Affine},
 	sha256::sha256_fixed,
+	util::clear_high_bits,
 };
 
 /// Convert 20-byte address payload into five little-endian u32 words (circuit witness layout).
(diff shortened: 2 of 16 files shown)


jimpo:
> Realising the win therefore needs build_committed_lin_def_patch (and the commit_set decision
that feeds it) to emit a Zero constraint when the option is on — a compiler change, not a flag
flip. Happy to open that as a follow-up issue if you want this landed as the call-site half.

@benji-potamus File that as a follow-up ticket

benji-potamus:
Filed [BINIUS-384](https://linear.app/jimpo/issue/BINIUS-384/lower-committed-linear-definitions-to-zero-constraints-under-enable) to thread `enable_zero_constraints` into `build_committed_lin_def_patch` so the compiler side actually realizes the win.

jimpo (approved):

PR #1966, opened 2026-07-30
BINIUS-387: pass the non-padding AND constraint count into the AND-reduction prover

benji-potamus:
Linear: [BINIUS-387](https://linear.app/jimpo/issue/BINIUS-387/pass-the-non-padding-and-constraint-count-into-the-and-reduction)

Follow-up to [BINIUS-340](https://linear.app/jimpo/issue/BINIUS-340/skip-all-zero-padding-windows-in-the-and-reduction-round-1-message) and [BINIUS-364](https://linear.app/jimpo/issue/BINIUS-364/dont-require-power-of-two-of-constraints). Pure performance change — no arithmetization or protocol change; output stays bit-identical.

### The problem

`and_reduction::prove` only ever saw the already-padded column length (`checked_log_2(a.len())`). BINIUS-364 added `ConstraintSystem::n_and_constraints()`, the true non-padded AND count, but it was never threaded into the reduction — so:

- The round-1 message (`univariate_round_message_extension_domain`) had to *rediscover* padding windows at runtime via a per-window `.all() == Word::ZERO` scan (BINIUS-340).
- The fold step (`fold_bitand_operands`) had no skip logic at all — it unconditionally ran the Method-of-Four-Russians lookup on every chunk, real or padding.

### The change

Thread an explicit `n_real_rows: usize` (the exact row index where real rows end and zero padding begins) from the two `and_reduction::prove` call sites down into `OblongZerocheckProver::new`, and from there into both consumers:

- **Round-1 message**: the runtime all-zero scan is gone, replaced by a direct index check (`window_start_row >= n_real_rows`) — a whole window strictly past the boundary is skipped without reading its words. A window that straddles the boundary still holds real rows and runs the full computation, unchanged.
- **Fold step**: the same boundary check now skips whole `P::WIDTH` chunks (and the short tail) entirely past `n_real_rows`, writing the zero element directly instead of paying the bytewise lookup on words already known to be zero.

Call sites:
- `crates/prover/src/prove.rs` (single-instance): `cs.n_and_constraints()`.
- `crates/m4-prover/src/prove.rs` (M4 batch): `cs.n_and_constraints() << table.log_instances()`, since M4 lays columns out constraint-major (`row = constraint_index * n_instances + instance`) and pads only the flattened tail — confirmed against `operand_witness.rs`'s `build_operation_witness`.

### Why output stays identical

The padding contract already guarantees every row from `n_real_rows` onward is `Word::ZERO` in both operands. Folding a zero word (via the bytewise lookup transform) always produces the zero field element regardless of the weight vector, and `C = A & B` derived from two zero words is also zero — so writing the zero element directly for a known-padding chunk is exactly what the unskipped computation would have produced anyway. `two_proofs_from_one_prover_are_byte_identical` and `test_prove_verify_non_power_of_two_constraint_counts` (both pre-existing) pass unchanged, and the new tests below pin this explicitly.

### Tests

- `sumcheck_round_messages.rs`: added `test_first_round_message_with_boundary_inside_a_window`, a boundary landing 40 words into a window (not aligned to the 128-word window size) — the shape that matters for real Keccak/SHA-256-sized batches. The two pre-existing tests now pass `n_real_rows` explicitly (dense / window-aligned boundary).
- `fold_word.rs`: added `test_fold_bitand_operands_skips_padding_without_changing_output`, pinning the skip path (boundary inside a chunk, not aligned to `P::WIDTH`) against running the same input with no boundary at all — the two must agree exactly, since the "padding" words are truly zero either way.
- `crates/m4-prover/src/operand_witness.rs`: the three existing `and_reduction::prove` test call sites now pass the real row count (`and_constraints.len() << table.log_instances()`).

### Perf impact

Criterion bench (`crates/prover/benches/and_reduction.rs`), added a `(40% zero padding)` variant of the fold-step bench matching the existing round-message padding bench (`log_words = 21`, 3/5 real, 2/5 zero — the Keccak/SHA-256-shaped case the issue calls out):

| bench | time |
| --- | ---: |
| `univariate fold 2^21` (dense, this branch) | 102.98 ms |
| `univariate fold 2^21 (40% zero padding)` (this branch) | 88.90 ms |

Comparing dense vs. 40%-padded **within this branch** isolates the skip's effect cleanly: **~13.7% faster** on this shape. I also ran the same dense bench against a `main`-based scratch workspace baseline (121.51 ms) — the ~15% gap there is *not* attributable to this change (the dense case never takes the skip path, so it should be identical to `main`); it's noise between separate `cargo bench` invocations on this machine, not a real regression-then-fix. The reliable number is the same-branch dense-vs-padded comparison above.

I did not measure end-to-end example prove time; the fold step is one part of the BitAnd phase, and its share of total prove time varies by circuit padding fraction (see BINIUS-386's per-circuit padding table for how much this varies — `hashsign` pads ~44%, several others ~6-20%).

### DIFF
--- a/crates/m4-prover/src/operand_witness.rs
+++ b/crates/m4-prover/src/operand_witness.rs
@@ -866,11 +866,13 @@ mod tests {
 		let [a, b] =
 			build_operation_columns(&table, constants(&c), &and_constraints, &GlobalAllocator);
 		let log_total = checked_log_2(a.len());
+		let n_real_rows = and_constraints.len() << table.log_instances();
 
 		// Prover and verifier agree on the reduced claim over the batched columns.
 		let mut prover_transcript = ProverTranscript::new(StdChallenger::default());
 		let prove_output = and_reduction::prove::<_, B128, P, _, _>(
 			[a, b],
+			n_real_rows,
 			&mut prover_transcript,
 			&GlobalAllocator,
 		);
@@ -902,11 +904,13 @@ mod tests {
 		let [a, b] =
 			build_operation_columns(&table, constants(&c), &and_constraints, &GlobalAllocator);
 		let log_total = checked_log_2(a.len());
+		let n_real_rows = and_constraints.len() << table.log_instances();
 
 		// Produce a faithful proof.
 		let mut prover_transcript = ProverTranscript::new(StdChallenger::default());
 		let _ = and_reduction::prove::<_, B128, P, _, _>(
 			[a, b],
+			n_real_rows,
 			&mut prover_transcript,
 			&GlobalAllocator,
 		);
@@ -961,9 +965,10 @@ mod tests {
 				.map(|(&a, &b)| a & b)
 				.collect();
 			let log_total = checked_log_2(a.len());
+			let n_real_rows = and_constraints.len() << table.log_instances();
 
 			let mut prover_transcript = ProverTranscript::new(StdChallenger::default());
-			let prove_output = and_reduction::prove::<_, B128, P, _, _>([a, b], &mut prover_transcript, &GlobalAllocator);
+			let prove_output = and_reduction::prove::<_, B128, P, _, _>([a, b], n_real_rows, &mut prover_transcript, &GlobalAllocator);
 
 			let mut verifier_transcript = prover_transcript.into_verifier();
 			let verify_output =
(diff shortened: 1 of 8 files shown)


benji-potamus:
Closing without merging: [BINIUS-387](https://linear.app/jimpo/issue/BINIUS-387/pass-the-non-padding-and-constraint-count-into-the-and-reduction) was cancelled — jimpo decided the design here isn't right and opened [BINIUS-388](https://linear.app/jimpo/issue/BINIUS-388/m4-dont-pad-the-operation-witness-word-vecs) to remove the operand-witness padding at the source instead of skipping over it at compute time. That work supersedes this PR's approach.

PR #1967, opened 2026-07-30
BINIUS-389: let the AND reduction consume unpadded operand columns

benji-potamus:
Sub-issue 1 of [BINIUS-388](https://linear.app/jimpo/issue/BINIUS-388/m4-dont-pad-the-operation-witness-word-vecs) — [BINIUS-389](https://linear.app/jimpo/issue/BINIUS-389/let-the-and-reduction-consume-unpadded-operand-columns).

Relaxes the AND reduction's power-of-two precondition on its operand columns, so BINIUS-392 can later hand it columns of length `n_and_constraints << log_instances` instead of the next power of two. **No caller changes**: both `crates/prover/src/prove.rs` and `crates/m4-prover/src/prove.rs` still pass padded columns, and every existing proof is byte-identical.

### Commits

1. **Window the round-1 message over a zero-extended operand column.** `univariate_round_message_extension_domain` required both columns to be exactly `1 << log_words` words and windowed them through the power-of-two-only `duplicate_to_fixed_chunks`. It now accepts a shorter column and reads the rows up to the axis length as zero, via a new module-private `operand_window` helper.

2. **Let `and_reduction::prove` accept unpadded columns.** `checked_log_2(a.len())` becomes `log2_ceil_usize(a.len())`, plus an equal-length assertion.

### Why the padding rows are free

An all-zero operand row forces the derived `C = A & B` to zero as well, so `A * B - C` vanishes on it at *every* point of the univariate domain — not just on the hypercube. A window lying entirely past the columns' last row is therefore skipped without touching a word, where before it was discovered by a runtime all-zero scan.

The fold step needed no change at all: `fold_bitand_operands_with_transform` already rounds the axis up to a power of two, folds only the real chunks plus a short tail, and zero-pads its *output* buffers. So a shorter column already skips the padding chunks rather than running them through the Method-of-Four-Russians lookup. That is the fold-step speedup [BINIUS-387](https://linear.app/jimpo/issue/BINIUS-387) was chasing, obtained from the shape of the input instead of an extra parameter — which is why that issue was cancelled in favour of this approach.

### The one subtlety: repeat vs. zero-fill

Two different padding rules apply, and getting them backwards is silently wrong rather than loud:

- Rows between `words.len()` and `1 << log_words` are **zero**. They are real padding rows of the constraint axis, and the fold produces zero there, so the round message must too.
- When the whole axis is shorter than one 128-word window, the remaining window slots **repeat** the zero-extended axis. Those slot index bits are weighted by the fixed *small* zerocheck challenges, which are non-zero, and summing a non-zero eq challenge over a duplicated coordinate returns exactly one copy — whereas zeroing them would scale the axis by `1 + r`. This is the same "variable padding over the boolean hypercube" argument `duplicate_to_fixed_chunks` already relied on; it is spelled out on `operand_window` now.

`test_first_round_message_with_unpadded_columns` has teeth here: replacing the repetition with a zero-fill makes it fail.

### Tests

- `unpadded_column_matches_explicit_zero_padding` — the round message of a short column is bit-identical to that of the same column zero-padded by the caller.
- `test_first_round_message_with_unpadded_columns` — the short column's message agrees with the verifier's own independent fold, at every windowing shape. (The equality above would also hold if both sides were wrong in the same way; this pins the absolute value.)
- `unpadded_columns_reduce_identically_to_padded_ones` — the full reduction over `and_reduction::prove` yields a byte-identical transcript and identical reduced claim, and that claim still verifies against the padded row count the verifier derives from the constraint system.
- The shared shape list covers a sub-window axis, a sub-window axis with a real zero tail, exactly one window, one window plus a straddling one, whole windows, and whole windows plus a straddling one.

### Note for review

`operand_window` is deliberately a new module-private helper rather than a generalization of `duplicate_to_fixed_chunks`, which still serves `fold_across_words` and `WordFolder::fold` under a power-of-two precondition. BINIUS-390 (BinMul) and BINIUS-391 (IntMul) will need the same short-input tolerance for those two, and that is the point at which promoting a shared helper into `fold_word.rs` pays for itself. Happy to hoist it now instead if you'd rather not have the two coexist.

### DIFF
--- a/crates/prover/src/and_reduction/sumcheck_round_messages.rs
+++ b/crates/prover/src/and_reduction/sumcheck_round_messages.rs
@@ -1,22 +1,24 @@
 // Copyright 2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::{array, iter};
+use std::{array, borrow::Cow, iter};
 
 use binius_core::word::Word;
 use binius_field::{
-	AESTowerField8b as B8, BinaryField, BinaryField1b as B1, ExtensionField,
+	AESTowerField8b as B8, BinaryField, BinaryField1b as B1, ExtensionField, Field,
 	PackedAESBinaryField64x8b as Packed64xB8, PackedField, WideMul, util::expand_subset_sums_array,
 };
 use binius_math::{BinarySubspace, multilinear::eq::eq_ind_partial_eval};
-use binius_utils::rayon::prelude::*;
+use binius_utils::rayon::{self, iter::Either, prelude::*};
 use binius_verifier::{
 	config::PROVER_SMALL_FIELD_ZEROCHECK_CHALLENGES, protocols::bitand::ROWS_PER_HYPERCUBE_VERTEX,
 };
 use itertools::izip;
 
 use super::ntt_lookup::NTTLookup;
-use crate::fold_word::duplicate_to_fixed_chunks;
+
+/// Number of big field zerocheck challenges whose equality indicator is expanded per window.
+const N_FIXED_LARGE_CHALLENGES: usize = 4;
 
 /// Generates a univariate polynomial for the sumcheck protocol in AND constraint reduction.
 ///
@@ -49,12 +51,19 @@ use crate::fold_word::duplicate_to_fixed_chunks;
 ///
 /// # Arguments
 ///
-/// * `log_words` - Base-2 logarithm of the number of words in each column
+/// * `log_words` - Base-2 logarithm of the constraint axis's length
 /// * `a_words` - First multiplicand (a) as a one-bit oblong multilinear polynomial
 /// * `b_words` - Second multiplicand (b) as a one-bit oblong multilinear polynomial
 /// * `eq_ind_big_field_challenges` - Partial equality indicator evaluations for big field variables
 /// * `prover_message_domain` - The NTT domain subspace (dimension `SKIPPED_VARS + 1`) from which
 ///   the low-degree-extension lookup table is built internally
+///
+/// # Preconditions
+///
+/// * The two columns have equal length, at most `1 << log_words`. They need not fill the constraint
+///   axis: a shorter column has its remaining rows read as zero. Such a row forces the derived `C =
+///   A & B` to zero as well, so `A * B - C` vanishes on it at every point of the univariate domain
+///   and it adds nothing to the message.
 ///
 /// # Returns
 ///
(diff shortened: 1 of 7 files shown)


crates/prover/src/and_reduction/sumcheck_round_messages.rs
```diff
@@ -142,13 +156,15 @@ where
 			//
 			// A populated chunk stops the scan at its first non-zero word.
 			// Dense inputs therefore pay only a couple of word comparisons.
+			// Windows the padding covers whole are already gone above; this catches a real window
```
jimpo: What? No, delete this now.

jimpo (changes requested):
Oh my god, this is so complicated. Simplify it. Make the code beautiful.

crates/prover/src/and_reduction/sumcheck_round_messages.rs
benji-potamus: Deleted, along with its test (`test_first_round_message_with_zero_padding_windows`).

You're right that it's dead weight now, and it's worth spelling out why: BINIUS-340 added that scan to *rediscover* which windows were padding, by comparing every word to zero. This PR removes the padding from the input, so there is nothing left for it to find — all it did was cost dense inputs a couple of word comparisons per window. The structural bound (`n_windows = a_words.len().div_ceil(WINDOW)`) subsumes it entirely.

That also removed the awkward "this catches a real window that happens to be all-zero, and the one straddling the boundary" caveat, which was the tell that the scan no longer had a job.

benji-potamus:
Fair. Rewritten — the new-code half of the diff is now ~30 lines instead of ~90.

**What went.** `n_operand_windows` and `operand_window` are both gone, and with them the `Cow`, the `Option`, the let-else, the two-branch dispatch and about 25 lines of doc comment. The runtime all-zero scan went too, per your other comment — it existed to rediscover padding, and there is no padding to find any more.

**What's left.** The loop states its own bound:

```rust
let n_windows = a_words.len().div_ceil(WINDOW);

(0..n_windows).into_par_iter().map(|window| {
    let (a_rest, b_rest) = (&a_words[window * WINDOW..], &b_words[window * WINDOW..]);
    match (a_rest.first_chunk::<WINDOW>(), b_rest.first_chunk::<WINDOW>()) {
        // A window the columns fill, read in place.
        (Some(a_chunk), Some(b_chunk)) => window_term(a_chunk, b_chunk),
        // The window the columns end in, completed with the padding rows.
        _ => window_term(&tail_window(a_rest, log_words), &tail_window(b_rest, log_words)),
    }
})
.zip(eq_ind_extra.as_ref())
```

The existing per-window body moved into a `window_term` closure unchanged, so the two arms share it. `tail_window` is the only new helper, fifteen lines:

```rust
fn tail_window<const N: usize>(words: &[Word], log_words: usize) -> [Word; N] {
    let axis_rows = (1usize << log_words).min(N);

    let mut window = [Word::ZERO; N];
    window[..words.len()].copy_from_slice(words);

    let (axis, rest) = window.split_at_mut(axis_rows);
    for copy in rest.chunks_exact_mut(axis_rows) {
        copy.copy_from_slice(axis);
    }
    window
}
```

The thing I'd previously written as two separate padding regimes collapses into that one `split_at_mut` + `chunks_exact_mut`: it is a plain no-op whenever the axis is at least one window wide, and only a circuit with fewer than 128 rows ever executes the repeat. Worth stating precisely, since it's the one non-obvious line — the repeat is only *load-bearing* below 8 rows, where a window index bit past `log_words` still carries one of the three fixed small challenges, which are non-zero. Above that the extra bits carry zero-filled challenges, whose eq weight is `[1, 0]`, so repeat and zero-fill agree.

**Tests.** Also dropped one I'd added: pinning the unpadded message byte-for-byte against the padded one was redundant next to `test_first_round_message_with_unpadded_columns`, which pins it against the verifier's own independent fold, and next to the end-to-end transcript equality in `mod.rs`. Two tests where I had three.

I re-ran the mutation check after the rewrite — deleting the repeat loop still fails `test_first_round_message_with_unpadded_columns` and `unpadded_columns_reduce_identically_to_padded_ones`, so the invariant is still pinned rather than just documented.

Local gate green: tests, clippy, rustdoc, fmt, typos, for both `binius-prover` and `binius-m4-prover`, and each commit verified green on its own.

jimpo (approved):

PR #1968, opened 2026-07-30
BINIUS-390: let the BinMul check consume unpadded operand columns

benji-potamus:
Sub-issue 2 of [BINIUS-388](https://linear.app/jimpo/issue/BINIUS-388/m4-dont-pad-the-operation-witness-word-vecs) — [BINIUS-390](https://linear.app/jimpo/issue/BINIUS-390/let-the-binmul-check-consume-unpadded-operand-columns).

The `fold_word.rs` half of BINIUS-390 was split out into #1971 at your request and has now merged, so this PR is rebased onto `main` and is just `binmul.rs`.

Relaxes the BinMul check's power-of-two precondition on its six operand columns. **No caller changes**: the callers keep passing padded columns, and every existing proof is byte-identical.

### What this does and does not buy

BinMul's multilinears are inherently power-of-two — the mlecheck runs over the whole `2^ℓ` hypercube — so this does **not** make BinMul do less arithmetic. What it buys is that BinMul now carries the padding in the packed GHASH-field tables it already allocates at hypercube size, instead of requiring the caller to materialize a padded `Vec`. That's what lets BINIUS-392 drop the padded word columns. I've scoped it that way deliberately so it isn't reviewed against a speedup it can't deliver.

### The change

Two commits.

**1. Relax the precondition.** `prove` derives the hypercube's variable count with `log2_ceil_usize` rather than `strict_log_2`, and asserts only that the six columns have equal length. `build_table` reads a row at or past the columns' end as the zero GHASH element, so the padding lives in the packed table it already allocates at hypercube size. A zero row satisfies `0 · 0 = 0`, so the zerocheck claim is unchanged.

**2. Build the table with slice iterators.** `build_table` zips the two columns' `chunks_exact(P::WIDTH)` iterators for the packed elements the columns fill whole, and takes the trailing words from the iterators' `remainder`:

```rust
let (lo_chunks, hi_chunks) = (lo.chunks_exact(P::WIDTH), hi.chunks_exact(P::WIDTH));
let (lo_tail, hi_tail) = (lo_chunks.remainder(), hi_chunks.remainder());
values.extend(
    iter::zip(lo_chunks, hi_chunks)
        .map(|(lo_chunk, hi_chunk)| P::from_scalars(iter::zip(lo_chunk, hi_chunk).map(elem))),
);

if !lo_tail.is_empty() {
    values.push(P::from_scalars(iter::zip(lo_tail, hi_tail).map(elem)));
}

values.resize(packed_len, P::zero());
```

The chunk bound is the iterator's own, so neither phase computes an index. The lanes `from_scalars` is not given for the tail default to zero, which is exactly the start of the padding, and `resize` fills the rest of the hypercube.

`remainder()` is read before the iterators are consumed — sound, because `chunks_exact` splits the slice at construction, so it doesn't depend on iteration position. The two tails are equal-length by the `assert_eq!(lo.len(), hi.len())` at the top, so the zip over them cannot silently truncate.

### Tests

`unpadded_columns_prove_identically_to_padded_ones` — a non-power-of-two BinMul witness yields a byte-identical transcript and identical claims to the zero-padded one, and still verifies against the padded variable count, at n = 1, 3, 65, 100 (covering a sub-`P::WIDTH` column, a straddling one, and `remainder == 0`).

`random_witness` now takes a constraint count rather than its log, so non-power-of-two witnesses are expressible; the two existing callers pass `1 &lt;&lt; LOG_N`.

### DIFF
--- a/crates/prover/src/protocols/binmul.rs
+++ b/crates/prover/src/protocols/binmul.rs
@@ -1,5 +1,7 @@
 // Copyright 2026 The Binius Developers
 
+use std::iter;
+
 use binius_compute::{Allocator, VecLike};
 use binius_core::word::Word;
 use binius_field::{BinaryField, PackedField};
@@ -37,22 +39,23 @@ where
 	let n_vars = log2_ceil_usize(lo.len());
 	let packed_len = 1 << n_vars.saturating_sub(P::LOG_WIDTH);
 	let elem =
-		|index: usize| F::from((lo[index].as_u64() as u128) | ((hi[index].as_u64() as u128) << 64));
+		|(&lo, &hi): (&Word, &Word)| F::from((lo.as_u64() as u128) | ((hi.as_u64() as u128) << 64));
 
 	let mut values = alloc.alloc::<P>(packed_len);
 
-	// The packed elements the columns fill whole. Rounding down to a multiple of `P::WIDTH` keeps
-	// every lane of this loop in range, so it packs without a per-lane bounds check.
-	let n_full = lo.len() / P::WIDTH;
+	// The packed elements the columns fill whole. `chunks_exact` stops at the last of them, so
+	// every lane this loop packs is a real row.
+	let (lo_chunks, hi_chunks) = (lo.chunks_exact(P::WIDTH), hi.chunks_exact(P::WIDTH));
+	let (lo_tail, hi_tail) = (lo_chunks.remainder(), hi_chunks.remainder());
 	values.extend(
-		(0..n_full).map(|i| P::from_scalars((0..P::WIDTH).map(|j| elem(i * P::WIDTH + j)))),
+		iter::zip(lo_chunks, hi_chunks)
+			.map(|(lo_chunk, hi_chunk)| P::from_scalars(iter::zip(lo_chunk, hi_chunk).map(elem))),
 	);
 
 	// The columns' trailing words share one packed element with the start of the padding. The lanes
 	// `from_scalars` is not given default to zero, which is exactly that padding.
-	let remainder = lo.len() - n_full * P::WIDTH;
-	if remainder > 0 {
-		values.push(P::from_scalars((0..remainder).map(|j| elem(n_full * P::WIDTH + j))));
+	if !lo_tail.is_empty() {
+		values.push(P::from_scalars(iter::zip(lo_tail, hi_tail).map(elem)));
 	}
 
 	// The rest of the hypercube is zero padding. A zero row satisfies `0 * 0 = 0`, so it
--- a/crates/prover/src/protocols/binmul.rs
+++ b/crates/prover/src/protocols/binmul.rs
@@ -1,6 +1,6 @@
 // Copyright 2026 The Binius Developers
 
-use binius_compute::Allocator;
+use binius_compute::{Allocator, VecLike};
 use binius_core::word::Word;
 use binius_field::{BinaryField, PackedField};
 use binius_ip_prover::{
@@ -8,7 +8,7 @@ use binius_ip_prover::{
 	sumcheck::{ProveSingleOutput, prove_single_mlecheck, quadratic_mlecheck_prover},
 };
 use binius_math::{FieldVec, field_buffer::FieldBuffer};
-use binius_utils::checked_arithmetics::strict_log_2;
+use binius_utils::checked_arithmetics::log2_ceil_usize;
 pub use binius_verifier::protocols::binmul::BinMulOutput;
 
 use crate::fold_word::WordFolder;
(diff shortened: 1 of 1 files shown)


crates/prover/src/protocols/binmul.rs
```diff
@@ -17,21 +17,38 @@ use crate::fold_word::WordFolder;
...
+	};
+
 	let p_width = P::WIDTH.min(1 << n_vars);
 	let packed_len = 1 << n_vars.saturating_sub(P::LOG_WIDTH);
 	let mut values = alloc.alloc::<P>(packed_len);
 	values.extend((0..packed_len).map(|i| {
 		P::from_scalars((0..p_width).map(|j| {
 			let index = i << P::LOG_WIDTH | j;
-			F::from((lo[index].as_u64() as u128) | ((hi[index].as_u64() as u128) << 64))
+			F::from(word_at(lo, index) | (word_at(hi, index) << 64))
```
jimpo: Do better. Two extends on values -- in the first one, the over the indices that are less than n_constraints. In fact, round down to the nearest multiple of P::WIDTH so this loop always packs full packed field elements. Then one conditional push to the Vec for the remaining words past the P::WIDTH alignment. Then resize to power-of-two with P::zero().

crates/prover/src/fold_word.rs
```diff
@@ -555,22 +557,25 @@ impl<F: BinaryField> WordFolder<F> {
...
-		for (chunk, &suffix_weight) in iter::zip(chunks.as_ref(), self.suffix_weights.as_ref()) {
-			accumulate_word_chunk(chunk, &self.lookups, suffix_weight, &mut folded);
-		}
-
-		folded
+		assert!(words.len() <= self.n_words, "words.len() must not exceed 2^point.len()");
+
+		// Accumulate each chunk's contribution, scaled by that chunk's suffix weight. The zip stops
+		// at the shorter side, dropping the weights of the chunks the list does not reach.
+		iter::zip(chunks_zero_extended::<CHUNK_SIZE>(words), self.suffix_weights.as_ref()).fold(
```
jimpo: Split the fold_words changes off to a separate PR.

jimpo (changes requested):

crates/prover/src/protocols/binmul.rs
benji-potamus: Done — restructured exactly as described. `build_table` is now three phases:

1. `n_full = lo.len() / P::WIDTH` full packed elements, extended in one pass. Every lane is in range by construction, so the per-lane `column.get(index).map_or(0, ..)` is gone.
2. One conditional `push` for the `lo.len() - n_full * P::WIDTH` trailing words. The lanes `from_scalars` isn't given default to zero, which is exactly the start of the padding, so nothing extra is needed there.
3. `values.resize(packed_len, P::zero())` for the rest of the hypercube.

`p_width` and the `word_at` closure are both gone. The capacity from `alloc.alloc(packed_len)` still bounds the whole thing — `n_full + 1 <= packed_len` whenever there's a remainder, since `1 << n_vars` is a multiple of `P::WIDTH` and at least `ceil(lo.len() / P::WIDTH) * P::WIDTH` — so the `push` never reallocates off the pool.

The test's shape list covers all three phases: n = 1 (sub-`P::WIDTH`, phase 2 only), 3, 65, 100 (`remainder == 0`, phases 1 and 3 only).

crates/prover/src/fold_word.rs
benji-potamus: Split out to #1971. It was already its own commit, so this PR is now rebased onto that branch and its diff is just `binmul.rs`.

Kept as a stack rather than two independent PRs because `binmul::prove` genuinely needs the relaxation — it calls `folder.fold(a_lo)` on a short column — so #1968 can't land before #1971. Once #1971 merges I'll rebase this onto main and re-target it.

jimpo (changes requested):
In a new commit, refactor build_table to use slice iterators, chunks_exact, and zipping to build the table instead of indexing. chunks_exact gives a remainder method on the iterator that you can use.

benji-potamus:
Done in a new commit (`bb5ffc9e`, "build the BinMul table with slice iterators"). No indexing left in `build_table`:

```rust
let (lo_chunks, hi_chunks) = (lo.chunks_exact(P::WIDTH), hi.chunks_exact(P::WIDTH));
let (lo_tail, hi_tail) = (lo_chunks.remainder(), hi_chunks.remainder());
values.extend(
    iter::zip(lo_chunks, hi_chunks)
        .map(|(lo_chunk, hi_chunk)| P::from_scalars(iter::zip(lo_chunk, hi_chunk).map(elem))),
);

if !lo_tail.is_empty() {
    values.push(P::from_scalars(iter::zip(lo_tail, hi_tail).map(elem)));
}
```

`elem` goes from `|index: usize|` closing over both slices to `|(&lo, &hi): (&Word, &Word)|`, so it's now a plain zip mapper and both phases share it. The `n_full` / `i * P::WIDTH + j` arithmetic and the `remainder > 0` test are all gone — the chunk bound is the iterator's own.

Two things worth noting:

- `remainder()` is read *before* the iterators are consumed. That's sound: `chunks_exact` splits the slice at construction, so `remainder()` doesn't depend on iteration position — which is what lets both tails be bound in one line next to the chunk iterators rather than after the `extend`.
- The two tails are equal-length by the `assert_eq!(lo.len(), hi.len())` at the top, so the `iter::zip` over them can't silently truncate.

Rebased onto `main` (`a5401f39`, which now includes #2022). Both commits are independently green — tests, clippy, and rustfmt.

jimpo (approved):

PR #1969, opened 2026-07-30
BINIUS-391: let the IntMul check consume unpadded operand columns

benji-potamus:
Sub-issue 3 of [BINIUS-388](https://linear.app/jimpo/issue/BINIUS-388/m4-dont-pad-the-operation-witness-word-vecs) — [BINIUS-391](https://linear.app/jimpo/issue/BINIUS-391/let-the-intmul-check-consume-unpadded-operand-columns). Independent of #1967 and #1968; all three unblock the payoff sub-issue.

`intmul::prove` now accepts four operand columns of any equal length, zero-extending them to the constraint axis of `2^⌈log₂ n⌉` rows once, at the entry point, before building the witness. **No caller changes**, and every existing proof is byte-identical.

### This deviates from the plan in the issue — deliberately

BINIUS-391 as written called for threading the row bound down through `Witness::new`, `limb_leaves`, `compute_b_leaves`, `compute_b_leaves_parallel`, `two_valued_field_buffer`, `fold_across_words` and the phase-5 bitmask cast. I built the entry-point extension instead, for two reasons that only became clear once I read those sites:

**1. There is nothing to save.** IntMul's structures all span the full axis: four constant-base GKR trees, the variable-base `b` tree, the per-limb power tables, the logup\* lookup. Their leaf buffers hold `2^LOG_N_LIMBS` (= 8) or `Word::BITS` (= 64) *field elements* per row — 128 to 1024 bytes against the word column's 8. So the padded word column is well under 1% of IntMul's live footprint, and teaching seven call sites to stop early buys nothing measurable. Unlike the AND path (#1967), IntMul genuinely cannot skip the padding work; it can only skip the caller's allocation.

**2. Threading it is the more dangerous option.** A padding row is *not* uniformly zero once it reaches those buffers. The product-check leaves read `tables[limb][limb_index(word, limb)]`, and `power_table` row 0 holds `base^0 = 1` (`witness.rs:239`). So a padding row must contribute the multiplicative **identity** to every product-check tree — a zero fill there would collapse the entire product. Zero-extending at the *word* level encodes that fact once and lets all four buffer-construction sites derive it; hand-written fills would each have to independently get ONE-not-ZERO right, in the most intricate protocol in the tree, for no gain.

The interface goal of BINIUS-388 is fully met either way: BINIUS-392 can hand `intmul::prove` an unpadded column and `build_operation_columns` stops padding. Happy to do the invasive version if you disagree with the tradeoff — say so on [BINIUS-391](https://linear.app/jimpo/issue/BINIUS-391) and I'll switch.

### Also worth knowing

- A power-of-two count copies nothing: `extend_to_constraint_axis` returns `Cow::Borrowed`. Only a genuinely non-power-of-two count pays one `Vec` of at most `2n` words.
- `Witness::new`'s existing power-of-two check and `Error::ExponentsPowerOfTwoLengthRequired` are left in place — `prove` now guarantees the precondition, but `Witness::new` is `pub` and reached directly by `benches/intmul.rs`, so the variant still has a producer. The issue's acceptance item about deleting it therefore doesn't apply. `prove`'s own error surface narrows to `ExponentLengthMismatch`, checked up front so the four columns are known equal before extension.

### Tests

- `unpadded_columns_prove_identically_to_padded_ones` — a non-power-of-two IntMul witness yields a byte-identical transcript and identical claims to the explicitly zero-padded one, and still verifies against the axis's variable count, at n = 3, 5, 33.

  Byte-identity is the whole story for this approach, and that is the point: the padded side *is* the pre-existing tested path, so the two agreeing byte for byte leaves no room for a phase to have received the wrong padding value. The invasive version would have needed per-site value tests instead.
- `random_products(rng, n)` extracted from `prove_and_verify` so non-power-of-two witnesses are expressible; the existing test passes `1 << LOG_EXPONENTS` and is otherwise unchanged.

### DIFF
--- a/crates/prover/src/protocols/intmul/error.rs
+++ b/crates/prover/src/protocols/intmul/error.rs
@@ -1,9 +1,8 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-	#[error("Exponent length should be a power of two")]
-	ExponentsPowerOfTwoLengthRequired,
 	#[error("All exponent slices must have the same length")]
 	ExponentLengthMismatch,
 	#[error("transcript error")]
--- a/crates/prover/src/protocols/intmul/prove.rs
+++ b/crates/prover/src/protocols/intmul/prove.rs
@@ -1,7 +1,7 @@
 // Copyright 2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::marker::PhantomData;
+use std::{iter, marker::PhantomData};
 
 use binius_compute::Allocator;
 use binius_core::word::Word;
@@ -49,15 +49,22 @@ use crate::fold_word::{fold_across_words, fold_words};
 /// Proves the integer multiplication (IntMul) reduction over the four operand columns.
 ///
 /// The four `columns` are the multiplicand `a`, the multiplicand `b`, and the product's low and
-/// high words, in the order `[a, b, lo, hi]`, each of power-of-two length. This builds the
-/// [`Witness`], drives an [`IntMulProver`], and reduces the multiplication relation to per-bit
-/// evaluation claims on the four columns at a common point. See [`IntMulProver::prove`] for the
-/// protocol description and [`IntMulOutput`] for the output shape.
+/// high words, in the order `[a, b, lo, hi]`, all of equal length. This builds the [`Witness`],
+/// drives an [`IntMulProver`], and reduces the multiplication relation to per-bit evaluation claims
+/// on the four columns at a common point. See [`IntMulProver::prove`] for the protocol description
+/// and [`IntMulOutput`] for the output shape.
+///
+/// # Padding
+///
+/// The columns' length need not be a power of two. The reduction runs over the constraint axis of
+/// `2^ceil(log2(n))` rows, with the rows past the columns' end read as `Word::ZERO`; a zero row
+/// satisfies `0 * 0 = 0 || 0`, so no claim moves. The padding is never materialized as words:
+/// every buffer built from the columns spans the whole axis and derives its own padding value,
+/// which is the multiplicative identity in the product-check trees and zero elsewhere.
 ///
 /// # Errors
 ///
-/// Returns an error when the operand columns do not have a power-of-two length or their lengths
-/// disagree.
+/// Returns an error when the operand columns' lengths disagree.
 pub fn prove<A, F, P, Channel>(
 	columns: [&[Word]; 4],
 	channel: &mut Channel,
@@ -254,9 +261,14 @@ where
 			.into_par_iter()
 			.map(|j| {
 				let (tree, limb) = (j / N_LIMBS, j % N_LIMBS);
+				// The columns' padding rows are `Word::ZERO`, whose every limb index is 0 — the
+				// shared table's row 0. So a padding row looks up `base^0 = 1`, as it does in the
+				// product-check trees.
+				let n_padding = (1 << n_vars) - exponents[tree].len();
 				exponents[tree]
 					.iter()
 					.map(|&word| limb_index(word, limb))
+					.chain(iter::repeat_n(0, n_padding))
 					.collect::<Vec<_>>()
 			})
 			.collect::<Vec<_>>();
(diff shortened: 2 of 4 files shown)


jimpo (changes requested):
Go a step further. Change intmul::Witness::new so that it handles the columns directly. Change limb_leaves and compute_b_leaves accordingly. (though still return po2-sized buffers).

benji-potamus:
Understood — your call, and I'll build it. Working on it now; this touches more sites than the three you named, so a note on what I'm finding as I go.

`Witness::new`, `limb_leaves` and `compute_b_leaves` (plus `compute_b_leaves_parallel`) are straightforward. The padding value there is the interesting bit and it falls out correctly: a zero exponent word indexes row 0 of every power table, which holds `base^0 = 1`, so `limb_leaves` can chain `repeat_n(table.get(0), ..)` after the real rows and each product-check tree gets the multiplicative identity for free. `compute_b_leaves` likewise — all bits clear ⇒ `F::ONE` per leaf.

The rest of the columns' readers need the same treatment, since `Witness` keeps the raw slices and phases 3–5 re-read them:

- `two_valued_field_buffer` — currently `get_unchecked(index)` over the full power-of-two extent. I'll restructure it the way you asked for `build_table` in #1968: full packed elements, then a remainder, then fill.
- `fold_across_words` — asserts an exact length. That's a `fold_word.rs` change, so per your split instruction it goes in #1971 alongside `WordFolder::fold` rather than here, and this PR stacks on it.
- the two `assert_eq!(b_exponents.len(), 1 << n_vars)` in phases 3 and 5.

One I'd like your call on, because it leaves the crate: phase 5 does `bytemuck::cast_slice(b_exponents)` and hands the result to `SelectorMlecheckProver::new`, which asserts `bitmasks.len() == selected.len()` (`crates/ip-prover/src/sumcheck/selector_mle.rs:73-77`) and passes them into `BinarySwitchover`. Relaxing that is a `binius-ip-prover` change, not an `intmul` one. Two options:

1. Relax `SelectorMlecheckProver`/`BinarySwitchover` to accept a short bitmask slice, reading missing rows as zero.
2. Keep one padded copy of `b` at that single call site.

I'll push with (2) so the PR is complete and reviewable, marked with a TODO pointing here — say the word and I'll switch it to (1). I didn't want to expand into another crate's invariants without asking.

benji-potamus:
Done — the entry-point copy is gone and the columns go into `Witness::new` unpadded. Rebased onto #1971, since `fold_across_words` needed the same relaxation and that's a `fold_word.rs` change.

**The three you named**, plus the padding value each derives for itself:

- `Witness::new` — takes the four columns at any equal length; `n_vars` is `log2_ceil_usize`.
- `limb_leaves` — chains `repeat_n(table.get(0), n_padding)` after the real rows. This is the load-bearing detail: `table.get(0)` is `base^0 = 1`, so a padding row contributes the multiplicative **identity** to the product check. A zero there would collapse the tree's product outright.
- `compute_b_leaves` and `compute_b_leaves_parallel` — a missing row's exponent bits read as clear, so every leaf is `F::ONE`, the same identity. The parallel path needed the most care: its 3-way zip against `exponents.par_chunks(P::WIDTH)` silently truncated, which would have left `out_vec` partly uninitialized under a `set_len`. It now enumerates over the packed elements and slices the chunk itself.

**Four more the columns reach**, since `Witness` keeps the raw slices:

- `two_valued_field_buffer` — restructured as you asked for `build_table`: full packed elements, one partial, then fill. The per-lane `get_unchecked` is gone. A padding row selects `elements[0]`.
- the phase-4 limb regather — same `table.get(0)` chain as `limb_leaves`.
- the phase-5 lookup index columns — a padding row's every limb index is 0, the shared table's row 0.
- both `assert_eq!(b_exponents.len(), 1 << n_vars)`.

`Error::ExponentsPowerOfTwoLengthRequired` now has no producer, so it's deleted — which does close your original acceptance item after all.

**The one thing I did not do.** Phase 5 hands `bytemuck::cast_slice(b_exponents)` to `SelectorMlecheckProver`, which asserts `bitmasks.len() == selected.len()` and passes them to `BinarySwitchover` (`crates/ip-prover/src/sumcheck/selector_mle.rs:73`). That's `binius-ip-prover`'s invariant, not `intmul`'s, so I left a padded copy of `b` at that single call site behind a `TODO(BINIUS-391)` rather than reaching into another crate uninvited. Say the word and I'll relax `SelectorMlecheckProver`/`BinarySwitchover` to read a missing row's bitmask as zero and drop it.

Worth noting these were all found by the tests rather than by reading: each of `SelectorMlecheckProver`, the phase-4 regather and the logup\* index column failed loudly in turn on a non-power-of-two witness. `unpadded_columns_prove_identically_to_padded_ones` still asserts a byte-identical transcript against the zero-padded witness at n = 3, 5, 33, so the whole rework is pinned to producing exactly today's proof.

Local gate green on `binius-prover` and `binius-m4-prover`: tests, clippy, rustdoc, fmt, typos.

jimpo (approved):

PR #1971, opened 2026-07-30
BINIUS-390: let the word folds accept a short word list

benji-potamus:
Split out of #1968 at your request. That PR is now rebased onto this one and carries only the `binmul.rs` change.

[BINIUS-390](https://linear.app/jimpo/issue/BINIUS-390/let-the-binmul-check-consume-unpadded-operand-columns), sub-issue 2 of [BINIUS-388](https://linear.app/jimpo/issue/BINIUS-388/m4-dont-pad-the-operation-witness-word-vecs).

### The change

`WordFolder::fold` asserted `words.len() == 1 << point.len()` and windowed through the power-of-two-only `duplicate_to_fixed_chunks`. It now accepts `words.len() <= 1 << point.len()`, reading the missing high rows as zero: an absent row's weight multiplies nothing, so chunks lying entirely past the list's end are never visited at all.

The accumulator loop also becomes an `iter::zip(..).fold(..)`.

### Two padding rules, not one

The new `chunks_zero_extended` sits next to `duplicate_to_fixed_chunks` with the **opposite** rule, and each doc comment points at the other:

- **zero-extend** (new) — for rows whose weights are *real*, so an absent row has to read as zero to contribute nothing. That is `WordFolder`'s case: its point is fully real over the word axis.
- **repeat** (existing) — for padding coordinates whose weights must sum to one over a duplicated row. That is what `fold_across_words` relies on, via the fixed challenge set.

Picking the wrong one is silently wrong rather than loud, which is why they are two named functions rather than one with a flag. Worth knowing that #1967 needs *both at once* (zero-extend to the constraint axis, then repeat that axis to fill the round-1 window) — its `operand_window` is private to the AND reduction for now, and a three-way consolidation may be worth doing once these all land.

### Tests

`word_folder_folds_a_short_list_as_if_zero_padded` — a short fold equals both the same list zero-padded and the independent naive reference, over sub-chunk, straddling, whole-chunk and empty shapes.

### DIFF
--- a/crates/prover/src/fold_word.rs
+++ b/crates/prover/src/fold_word.rs
@@ -450,9 +450,12 @@ fn accumulate_word_chunk<F: BinaryField>(
 /// suffix: the prefix tensor expansion is folded into each chunk of `CHUNK_SIZE` words by lookup,
 /// and the suffix tensor expansion scales each chunk's contribution before the chunks are summed.
 ///
+/// A list shorter than the word axis reads its missing high rows as zero, exactly as
+/// [`WordFolder::fold`] does.
+///
 /// ## Preconditions
 ///
-/// * `words.len() == 1 << point.len()`
+/// * `words.len() <= 1 << point.len()`
 ///
 /// [Method of Four Russians]: <https://en.wikipedia.org/wiki/Method_of_Four_Russians>
 pub fn fold_across_words<F, P>(words: &[Word], point: &[F]) -> [F; Word::BITS]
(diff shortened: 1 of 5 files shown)


benji-potamus:
Pushed an unrequested simplification before you get to this — applying the lesson from #1967, where the same `Cow`-returning windowing helper drew "this is so complicated".

The `chunks_zero_extended` iterator is gone. `WordFolder::fold` now splits the list itself, which is both shorter and closer to how `fold_words_with_transform` already does it a few hundred lines up:

```rust
let (chunks, tail) = words.as_chunks::<CHUNK_SIZE>();
let mut folded = [F::ZERO; Word::BITS];

for (chunk, &suffix_weight) in iter::zip(chunks, self.suffix_weights.as_ref()) {
    accumulate_word_chunk(chunk, &self.lookups, suffix_weight, &mut folded);
}

if !tail.is_empty() {
    let mut chunk = [Word::ZERO; CHUNK_SIZE];
    chunk[..tail.len()].copy_from_slice(tail);
    accumulate_word_chunk(&chunk, &self.lookups, self.suffix_weights.get(chunks.len()), &mut folded);
}

folded
```

Net: one deleted helper, no `Cow`, no `impl Iterator` return, and the whole diff is now confined to `fold`'s body. I did revert to a `for` loop rather than the `iter::zip(..).fold(..)` I had — accumulating into a fixed `[F; Word::BITS]` reads better as the loop it is, and it's what the function did before.

The padding contract is unchanged and still the interesting part: rows past the list's end read as **zero**, not repeated. `WordFolder`'s point is fully real over the word axis, so an absent row's weight multiplies nothing. That is the opposite of `duplicate_to_fixed_chunks`, whose callers have padding coordinates whose weights must sum to one over a duplicated row — I've left a note on each pointing at the other, since picking the wrong rule is silently wrong rather than loud.

Local gate green (tests, clippy, rustdoc, fmt, typos), and #1968 on top of it is green too.

jimpo (approved):

PR #1975, opened 2026-07-31
BINIUS-394: hand-written NEON kernel for the multi-lane Blake3 block compression

tcoratger:
Closes [BINIUS-394](https://linear.app/jimpo/issue/BINIUS-394).

Replaces the arithmetic inside the multi-lane Blake3 block compression with a hand-written NEON kernel on aarch64.

### The problem

The Blake3 Merkle suite hashes every leaf and every inner node through the portable multi-lane kernel.

That kernel is plain scalar lane loops, written so LLVM turns each `0..N` loop into one SIMD op.
It is a bet on the auto-vectorizer, and on aarch64 the bet only partly pays off.

Measured on an Apple M1 Pro, hashing 256-byte leaves (the shape the prover commits at):

- portable kernel: 1.038 GiB/s
- the `blake3` crate's own hand-written NEON kernel, best case: 1.317 GiB/s

So roughly 25% is unclaimed, and there was no hand-written kernel for this path at all.

### The idea

The multi-lane state is *already* in the layout NEON wants.

Each of the 16 compression words is held as one array of per-lane values:

```
row 0:  [ lane_0 lane_1 lane_2 lane_3 | lane_4 lane_5 lane_6 lane_7 | ... ]   word 0 of every lane
row 1:  [ lane_0 lane_1 lane_2 lane_3 | lane_4 lane_5 lane_6 lane_7 | ... ]   word 1 of every lane
            \________ group 0 ______/   \________ group 1 ______/
```

When the lane count is a multiple of four, a row is just $N/4$ four-lane vectors side by side.

So nothing about the layout changes.
Only the arithmetic is swapped, and the words it produces are bit-identical.

### Why interleaving several groups is the win

Blake3 mixing is a dependency chain: add, xor, rotate, then add again on the same word.
A single four-lane vector leaves the pipeline stalled waiting on that chain.
Advancing several independent groups in lockstep fills the idle slots.

Concretely, each mixing step is its own loop over the groups rather than one loop over the steps:

```
grouped by step:  a0 a1 a2 | d0 d1 d2 | c0 c1 c2   <- independent, issues freely
grouped by group: a0 d0 c0 | a1 d1 c1 | a2 d2 c2   <- serial within a group
```

Throughput on 256-byte leaves, M1 Pro:

| lanes | vector groups | throughput |
|---|---|---|
| 4 | 1 | 1.16 GiB/s |
| 8 | 2 | 1.77 GiB/s |
| 12 | 3 | 1.95 GiB/s |
| 16 | 4 | 1.93 GiB/s |
| 20 | 5 | 1.53 GiB/s |

Three groups is the peak, four is within noise, five spills enough registers to lose the gains.
The suite's existing width of 16 lanes is kept, so no batching or tail logic moves.

### The change

```
compress_block:
+   if the lane count has a vector kernel -> run it, return
    fall through to the scalar lane loops
```

The kernel lives in a private child module, so it is reachable from the compression but not part of any public surface.

### Soundness

- Block compression is a bit-exact function, so this is a pure implementation swap.
- No protocol value, transcript, digest, or commitment changes.
- The two kernels are pinned equal by tests rather than argued.

### How this relates to the Flock challenge — and what it is *not*

The mechanism comes from the Layr-Labs Flock prover challenge, where the same trick on their Blake3 Merkle tree was worth about 26%.

Their current variant is hand-written assembly, generated with `-mcpu=apple-m3` and pinned to a fixed 1024-byte chunk contract.
That is deliberately not followed here: this stays intrinsics, stays correct for any leaf length, and keeps every non-aarch64 target on the scalar loops.

Two neighbouring mechanisms from that challenge were checked and are **not** worth porting, both measured on M1 Pro:

- three-PMULL Karatsuba for the 128-bit carryless product is *slower* than schoolbook here (3.41 µs vs 3.16 µs)
- the same interleaving applied to SHA-256 gains nothing — that path runs on the ARMv8 crypto unit, which one lane already saturates (25.4 ns/block flat from 1 to 4 lanes, regressing past 5)

### Scope

- **new:** the vector kernel module, as a private child of the portable kernel
- **changed:** one dispatch in the block compression; a Blake3 arm added to the Merkle-tree benchmark, which previously covered only SHA-256
- **untouched:** the scalar lane loops, the chunk/flag/tail handling, the suite's lane count, and every non-aarch64 target

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-hash -p binius-iop-prover --all-targets --all-features`, nightly `fmt --all` — clean
- `cargo test -p binius-hash -p binius-iop-prover` — 61 passed, 0 failed
- new tests: a proptest over arbitrary state, plus eight fixed boundary cases (every flag combination, empty and partial blocks, both counter halves), each checked at 4, 8, 12 and 16 lanes against the scalar loops
- the new tests were mutation-checked: changing one rotation amount in the kernel fails them, along with six existing reference tests

### Benchmark

Leaf hashing kernel, by leaf size:

| leaf size | portable | NEON | speedup |
|---|---|---|---|
| 64 B | 966 MiB/s | 1.152 GiB/s | 1.22x |
| 128 B | 1024 MiB/s | 1.252 GiB/s | 1.25x |
| 256 B | 1.038 GiB/s | 1.303 GiB/s | 1.25x |
| 512 B | 1.061 GiB/s | 1.343 GiB/s | 1.27x |
| 1024 B | 1.075 GiB/s | 1.340 GiB/s | 1.25x |

Full Merkle commit, $2^{17}$ leaves of 16 field elements (32 MiB), ten threads:

| packing | portable | NEON | speedup |
|---|---|---|---|
| `128b` | 6.434 ms | 5.026 ms | 1.28x |
| `1x128b` | 6.154 ms | 5.129 ms | 1.20x |

Both Merkle numbers are criterion A/B runs at p < 0.05.
The commit gains slightly more than the leaf kernel alone because inner-node compression goes through the same core.

### DIFF
--- a/crates/hash/src/blake3_neon.rs
+++ b/crates/hash/src/blake3_neon.rs
@@ -1,0 +1,316 @@
+// Copyright 2026 The Binius Developers
+
+//! AArch64 NEON block compression for the multi-lane Blake3 kernel.
+
+use std::arch::aarch64::{
+	uint32x4_t, vaddq_u32, vdupq_n_u32, veorq_u32, vld1q_u32, vreinterpretq_u16_u32,
+	vreinterpretq_u32_u16, vrev32q_u16, vshlq_n_u32, vsriq_n_u32, vst1q_u32,
+};
+
+use super::{IV, MSG_PERMUTATION, N_ROUNDS};
+
+/// Widest interleave this module implements, counted in four-lane groups.
+///
+/// # Why this value
+///
+/// - A group holds 16 state vectors plus 16 message vectors, against 32 architectural registers.
+/// - Interleaving groups hides the latency of each one's add, xor, rotate chain.
+/// - Past four groups the spills cost more than the added parallelism returns.
+///
+/// Throughput hashing 256-byte leaves on an Apple M1 Pro:
+///
+/// ```text
+///      4 lanes (1 group):  1.16 GiB/s
+///      8 lanes (2 groups): 1.77 GiB/s
+///     12 lanes (3 groups): 1.95 GiB/s   <- peak
+///     16 lanes (4 groups): 1.93 GiB/s
+///     20 lanes (5 groups): 1.53 GiB/s
+/// ```
+const MAX_GROUPS: usize = 4;
+
+/// Rotates every 32-bit lane right by 16 bits.
+///
+/// A 16-bit rotate of a 32-bit word is exactly a swap of its two halfwords.
+/// That is a single reverse instruction, cheaper than the shift-insert pair the other amounts need.
+#[inline(always)]
+fn rotr16(x: uint32x4_t) -> uint32x4_t {
+	// SAFETY: this module is only reachable on aarch64 with `neon` statically enabled.
+	unsafe { vreinterpretq_u32_u16(vrev32q_u16(vreinterpretq_u16_u32(x))) }
+}
+
+/// Rotates every 32-bit lane right by `R` bits, where `L` is the complementary left shift.
+///
+/// # Arguments
+///
+/// * `R` - bits to rotate right, in `0 < R < 32`.
+/// * `L` - the value `32 - R`, supplied separately because a const parameter cannot be used in
+///   arithmetic at an intrinsic's immediate operand.
+///
+/// # Performance
+///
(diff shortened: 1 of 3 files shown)


jimpo:
Yay, thanks

PR #2017, opened 2026-08-06
Split the value-vector segment format into a borrowed writer and an owned reader

tcoratger:
Reshapes the on-disk format for one segment of a value vector. Wire bytes are unchanged — only the API around them moves.

### The problem

A value vector splits in two, and each half is written as its own file:

```
segment     | holds                        | who reads it
------------+------------------------------+------------------
public      | constants, inputs, outputs   | prover, verifier
non-public  | witness and internal values  | prover only
```

Those two files plus a serialized constraint system are how the `prover` and `verifier` binaries hand a proving job between hosts.

One `Cow`-backed type served both directions of that format. The `Cow` bought nothing:

- The borrowed variant only ever existed to serialize.
- The owned variant only ever came out of deserialize.
- No value was ever sometimes-borrowed-sometimes-owned.

What it did cost was a lifetime parameter on every signature, and eleven API items that no caller outside the module used.

### The idea

Split by direction, the way a string slice relates to an owned string.

- **Writing** borrows a slice. A non-public segment runs to tens of megabytes, and none of it is copied to reach the buffer.
- **Reading** owns its words, because the byte buffer they were decoded from need not outlive the call.

Owning on the read side is also what lets the version check live on a type at all. The blanket impls in `binius-utils` encode and decode word vectors with **no version tag**, so a newtype is the only place to hang a tagged impl:

```
&[Word]     --(borrowed type)--> [ version | count | words ]   nothing copied
Vec<Word>  <--(owned type)-----  [ version | count | words ]   version checked
```

### The change

```
- ValuesData::from(witness.public())        // build a Cow, then serialize the instance
- write_serialized(&data, path)?
+ write_serialized(&ValuesRef::new(witness.public()), path)?
```

Both directions now go through the same serialization trait, so the CLI's generic write helper covers segments too and the bespoke one it needed is gone.

Deleted, none of it used outside the module: `borrowed`, `owned`, `as_slice`, `len`, `is_empty`, `to_owned`, `into_owned`, `AsRef`, and three `From` impls. `to_owned` needed to go regardless — it silently shadowed `ToOwned::to_owned` on a `Clone` type.

### Two fixes that fall out

**The reference-binary helpers wrote to a dead path.** Both regeneration helpers targeted `verifier/core/test_data/`, a layout this repo no longer has, while the tests read from `test_data/`. Regenerating either fixture created a stray directory instead of updating the file. Both are `#[ignore]`d, so nobody noticed.

**The pipeline had no test.** Nothing exercised save → prove → verify across files, which is why the stale naming and the broken paths survived. Two tests now compile a one-gate circuit, write the three artifacts, and run the real binaries:

- The happy path proves and verifies across process boundaries.
- Corrupting the version tag of the public segment — after a *valid* proof exists, so the tag is the only thing left to fail on — is rejected.

### Scope

- **new:** `crates/examples/tests/cli_artifacts.rs`
- **changed:** the segment format, one round-trip test, two CLI call sites, one verifier call site
- **renamed:** the fixture, `witness_v1.bin` to `values_data_v1.bin`, matching its siblings
- **fixed:** stale `PublicWitness` naming in the fixture README, for a type that never existed
- **untouched:** the byte format, and the version constant at 1

### On performance

No measurable difference. Five write shapes over a 32 MiB segment, measured in one harness with byte-identical output asserted:

| shape | write, 32 MiB |
|---|---|
| borrowed newtype (this PR) | 8.49–8.71 ms |
| newtype + local binding | 8.34–8.85 ms |
| newtype + `#[inline]` | 8.36–8.79 ms |
| free function taking the slice | 8.47–9.18 ms |
| the original `Cow` | 8.54–9.01 ms |

The write is bound by memory traffic, so the shape does not matter. The point of the reshape is the API, not speed — and nothing regressed.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-core -p binius-examples --all-targets`, nightly `fmt --all --check` — all clean, run on this commit in a fresh worktree.
- 86 core tests, 24 examples tests including the two new ones.
- Byte compatibility proven the hard way: regenerating the fixture through the fixed path reproduces the committed file exactly, and git records the rename at 100% similarity.


jimpo (commented):
Can you remove the constants from the serialized public segment? The constants are in the ConstraintSystem. The fact that they end up getting merged into one "public" segment looks increasingly like an implementation detail of the cryptographic protocol.

tcoratger:
> Can you remove the constants from the serialized public segment? The constants are in the ConstraintSystem. The fact that they end up getting merged into one "public" segment looks increasingly like an implementation detail of the cryptographic protocol.

@jimpo Done

jimpo:
Ah, sorry, this needs another rebase. The handling of constants & inout in ValueVec changed quite a lot.

PR #2024, opened 2026-08-07
BINIUS-397: dense shift encoding for the phase-1 g accumulator

benji-potamus:
Closes [BINIUS-397](https://linear.app/jimpo/issue/BINIUS-397/dense-shift-encoding-for-the-phase-1-g-accumulator).

`build_g_parts` accumulated into a buffer covering the whole `(shift variant, shift amount)` space — 8 variants × 64 amounts = 512 rows of 64 scalars, or 512 KiB over `B128` — allocated and zeroed once per rayon fold partial, then summed again in `reduce_with`. A constraint system uses a few dozen of those rows, so nearly all of that traffic was over zeros.

Each key segment now records the pairs it actually uses, and its keys index that dense encoding. The phase-1 accumulator holds one row per pair used; a scatter through the encoding restores the sparse layout callers expect. Phase 2's per-operation scalar tables shrink the same way, and become per segment since the encoding is.

The returned `g` parts are bit-identical, so proofs are byte-identical — the transcript is unaffected.

### Commits

1. **Give the shift-reduction bench a real ZERO claim** — a pre-existing bug found while measuring, unrelated to the rest: `benches/shift_reduction.rs` hands the ZERO operator `OperatorData::zero_claim`, whose contract is *an operation the constraint system does not use*, while the SHA256 circuit it benchmarks asserts its digest and those assertions lower to ZERO constraints. Both bench functions panic on `main` (`index out of bounds` in `Key::accumulate_wide`, indexing a length-1 tensor). It sits at the bottom of the stack because the issue's verification plan runs this bench. Happy to split it out if you'd rather.
2. **Add a dense shift encoding to each key segment** — the `DenseShiftEncoding` type, a `KeySegment` field, serialization, `KeyCollection::VERSION` 2 → 3. Purely additive; keys still carry the sparse id.
3. **Index shift keys by a dense shift index** — `Key::id` → `Key::dense_shift_idx`, and the three consumers move over. The rename is deliberate: `key.id * stride` is textually valid under either meaning, so every use had to be a compile error until audited.

### Sizes, on the SHA256 circuit over a 16 KiB message

| segment | pairs used (of 512) | phase-1 accumulator, per fold partial |
|---|---|---|
| public | 8 | 512 KiB → 8 KiB |
| hidden | 16 | 512 KiB → 16 KiB |

Both now fit comfortably in L2 instead of evicting it. The phase-2 scalar tables shrink by the same factor.

### Compatibility

The `VERSION` bump rejects prebuilt key-collection files and serialized ZK bundles (`crates/examples/src/cli.rs`, `crates/prover/src/zk_config.rs`) — they must be regenerated. That is an explicit deserialization error, not silent corruption.

### Verification

- `cargo test --workspace` green, including the transcript round-trips (`crates/prover/tests/prove_verify.rs`) and the m4 cross-check of `build_g_parts` against the folded builder.
- Four new unit tests in `key_collection.rs`: the encoding covers exactly the pairs its segment uses, keys index their own segment's encoding, the encoding survives serialization, and a non-ascending encoding is rejected on deserialization.
- `cargo clippy --all --tests --benches --examples -- -D warnings`, `cargo +nightly-2026-07-01 fmt`, `cargo doc` clean.
- Benchmarks in a comment below.

### DIFF
--- a/crates/m4-prover/src/shift.rs
+++ b/crates/m4-prover/src/shift.rs
@@ -154,13 +154,8 @@ where
 ///
 /// `SHIFT_VARIANT_COUNT` multilinears of [`LOG_LEN`] variables, one per shift variant.
 ///
-/// A key's id splits into the variant and the shift amount:
-///
-/// ```text
-/// id = (variant << Word::LOG_BITS) | amount
-/// ```
-///
-/// The variant picks the multilinear, and the amount picks the run of bit slots within it:
+/// The segment's dense shift encoding decodes a key's shift index into the variant and the shift
+/// amount. The variant picks the multilinear, and the amount picks the run of bit slots within it:
 ///
 /// ```text
 /// g[variant][amount * Word::BITS + bit] += folded_bit * acc(key)
@@ -195,11 +190,12 @@ pub fn build_g_parts_from_folded_words<F: BinaryField, A: Allocator>(
 				&operator_data.lambda_powers,
 			);
 
-			// The key id is `(variant << Word::LOG_BITS) | amount`, so its variant selects the
-			// multilinear and its shift amount the bit slots within that multilinear.
-			let variant = key.id as usize >> Word::LOG_BITS;
-			let amount_base = (key.id as usize & (Word::BITS - 1)) * Word::BITS;
-			let slots = &mut multilinears[variant].as_mut()[amount_base..amount_base + Word::BITS];
+			// The key's shift variant selects the multilinear, and its shift amount the bit slots
+			// within that multilinear.
+			let (variant, amount) = segment.dense_shift_enc.decode(key.dense_shift_idx as usize);
+			let amount_base = amount as usize * Word::BITS;
+			let slots =
+				&mut multilinears[variant as usize].as_mut()[amount_base..amount_base + Word::BITS];
 
 			// Scatter the accumulator across this key's bit slots, scaling each by the folded bit.
 			for (slot, &folded_bit) in iter::zip(slots, word) {
--- a/crates/prover/benches/shift_reduction.rs
+++ b/crates/prover/benches/shift_reduction.rs
@@ -101,6 +101,11 @@ fn bench_prove_and_verify(c: &mut Criterion) {
 				.map(F::new)
 				.collect::<Vec<_>>()
 		};
+		// The circuit's digest assertions lower to ZERO constraints, so the Zero reduction runs
+		// over its own constraint point, as wide as the ZERO set.
+		let r_x_prime_zero = (0..cs.log_zero_constraints().unwrap_or(0) as u128)
+			.map(F::new)
+			.collect::<Vec<_>>();
 		// SHA256 has no IMUL constraints, so the IntMul operator is the zero claim (four zero evals
 		// at an empty point), exactly as the real prover/verifier synthesize it (`prove.rs` /
 		// `verify.rs` `None` branch). Its `r_x_prime` is therefore empty.
@@ -109,6 +114,7 @@ fn bench_prove_and_verify(c: &mut Criterion) {
 		// Sample univariate eval point — shared across bitand and intmul operators.
 		let r_zhat_prime = F::random(&mut rng);
 
+		let zero_evals = [F::random(&mut rng)];
 		let bitand_evals = [F::random(&mut rng); 3];
 		let intmul_evals = [F::ZERO; 4];
 		let key_collection = build_key_collection(&cs);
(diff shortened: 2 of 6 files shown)


benji-potamus:
### Performance

**What was measured:** `cargo bench -p binius-prover --bench shift_reduction -- shift_reduction_phases`, criterion medians, on the SHA256 circuit over a 16 KiB message. Baseline is commit 1 of this stack (the bench fix alone, no behaviour change) in a scratch workspace; branch is the tip. Both `--release` with `RUSTFLAGS="-C target-cpu=native"`. Machine: 4-core Xeon Platinum 8175M @ 2.5 GHz.

`sample_size` is pinned to 10 in the bench, so each run's interval is wide; I ran the two changed phases 4× per side and report the spread of medians.

| bench | baseline (median of medians) | branch | delta |
|---|---|---|---|
| `phase1_build_g_parts` | 32.1 ms (31.1–32.7) | 30.0 ms (29.0–32.8) | ≈ −6% |
| `phase2_build_monster_segments` | 14.9 ms (12.5–15.6) | 15.1 ms (8.9–15.4) | no signal |

**Being honest about this: the wall-clock result is inconclusive on this box.** Run-to-run variance is ±10% on both sides — the untouched `phase1_run_sumcheck` and `phase2_run_sumcheck` moved by +7% and +3% between the same two runs — so the ~6% on `build_g_parts` sits at the noise floor rather than above it. `build_monster_segments` shows nothing; its scalar tables were never the bottleneck (they are built once per segment, not per word).

What *is* exact is the structural change, which the table in the PR body gives: the per-thread accumulator drops 64× (public) and 32× (hidden), from 512 KiB to 8/16 KiB. Two reasons that should read better than it does here:

- The saving is **per rayon worker**: allocating, zeroing and cross-thread-reducing 512 KiB is paid once per fold partial. This machine has 4 cores; on a 64-core prover the zeroing and reduction work saved is 16× larger.
- The zeroed rows also evict the accumulator's live rows from L2 during the scatter, which a 4-core VM under sampling noise is a poor instrument for.

I'd read this as "no regression, and a structurally smaller working set," not as a measured speedup. If a firm number matters before merging, the honest way to get it is a larger-core machine or a longer criterion sample; say the word and I'll rerun.

jimpo (commented):
Seems like we can simplify by removing the u16 encoding of sparse IDs. Their byte size isn't important anymore, we can just keep the separate ShiftVariant and amount fields.

benji-potamus:
Good call — done in `9253681`.

The `u16` packing was load-bearing only while `key.id` *was* the accumulator address; once the key carries a dense index, the encoding is just a list of shifts, so it now stores them as they come:

```rust
pub struct DenseShiftEncoding {
    /// The shift each dense index encodes, ordered by variant and then by amount.
    shifts: Vec<(ShiftVariant, u8)>,
}
```

`decode` and `iter` yield `(ShiftVariant, u8)` instead of a pair of `usize`s, and `BuilderKey` carries the `(ShiftVariant, u8)` straight off the `ShiftedValueIndex`. That deletes the `match shift_variant { Sll => 0, Slr => 1, ... }` block in `update_with_operand` entirely — it was a second source of truth for an ordering the enum discriminants already fix. Serialization writes variant + amount per entry rather than a packed `u16`.

The packed form survives only as `shift_index(shift) = variant as usize * Word::BITS + amount`, a private helper for the two fixed-size lookup tables inside construction (`new`'s occupancy sweep and `inverse`). Nothing stores it.

Two follow-on notes:

- The three consumers now write `variant as usize` to index their per-variant arrays. That leans on `ShiftVariant`'s explicit `#[repr(u8)]` discriminants matching the order of `build_h_parts`'s `[sll, srl, sra, rotr, sll32, srl32, sra32, rotr32]` and `evaluate_h_op`'s return — which is exactly the mapping the deleted `match` spelled out by hand, so this removes a duplicate rather than adding a dependency. Say the word if you'd rather it stayed explicit.
- **I squashed commits 2 and 3.** The split was "add the encoding, then switch the keys over," but with the sparse id gone the first half can only stay compiling by packing an id inline for the `Key` field the second half deletes — scaffolding that exists for one commit. One commit reads better. The bench fix is still separate at the bottom.

Also added a test that deserialization rejects an out-of-range shift amount, alongside the existing unordered-encoding one. Full gate green: `cargo test -p binius-prover -p binius-m4-prover`, workspace clippy, `cargo doc --document-private-items` with `-D warnings`, fmt.

jimpo (approved):

PR #2030, opened 2026-08-07
BINIUS-402: commit field buffers through a binius-math chunker

tcoratger:
Closes [BINIUS-402](https://linear.app/jimpo/issue/BINIUS-402).

Replaces two private chunking helpers in `binius-iop-prover` with a single primitive in `binius-math`, and fixes a latent commitment bug that lived in the copy.

### The problem

`commit_field_buffer` splits a `FieldSlice` into Merkle leaves.

A leaf either spans whole packed words or sits inside one, so the split needs two cases.
It implemented both itself, in `to_par_scalar_big_chunks` and `to_par_scalar_small_chunks`.

`binius-math` already has that exact split — `FieldBuffer::chunks_par`.
So the prover carried a second copy of logic that belongs to `FieldBuffer`, and the copy got a `FieldBuffer` invariant wrong.

### The invariant it got wrong

A buffer shorter than one packed word still occupies a whole word.
The lanes past `1 << log_len` are dead — `FieldBuffer`'s own docs say they are not logical elements and may hold unrelated data.

The private helper counted leaves from the backing store, not the logical length:

```
n_leaves = packed_slice.len() * P::WIDTH        // words, including dead lanes
```

For a 2-scalar buffer in a 4-lane packing that is **4 leaves, not 2**.
Two dead lanes were hashed as committed data and the tree came out one level too deep.

No panic, no error — a silently wrong commitment.
Reachable through `MerkleIPProverChannel::send_merkle_commitment`; no in-tree call site hits it today.

### The stated precondition was also the wrong one

The doc comment asked that the leaf count be a power of two.
In the log domain the leaf count is `2^(log_len - log_leaf_len)`, which is always a power of two.

So the documented precondition was vacuous, and the real one — `log_leaf_len <= buffer.log_len()` — was neither documented nor checked.

### The idea

Give `binius-math` the one piece it was missing, then delete the copy.

`chunks_par` cannot be reused directly: it yields `FieldSlice` by value, and a sub-width chunk owns its packed word, so a borrowing scalar iterator cannot outlive it.

`FieldBuffer::par_chunk_scalars` addresses chunks in scalar space and hands out scalar iterators directly:

```
chunk i  ->  scalars [i * 2^log_chunk_size, (i+1) * 2^log_chunk_size)
```

The two chunk shapes are dispatched once, before any scalar is read:

```
chunk >= one packed word  ->  a run of whole words        (par_chunks + iter_slice)
chunk <  one packed word  ->  a lane range inside a word  (map over the lanes of one word)
```

Both sides wrap into `Either`, mirroring what `chunks_par` already does one method above.
Choosing per scalar instead is what costs: a whole-word chunk pays a branch for a `skip` and a `take` that are both no-ops, and a sub-word chunk walks a `flat_map` over a one-word slice it never has to advance.

Summing every scalar of a 2^20-scalar buffer, chunk by chunk:

| | nested `skip`/`take` | `Either` |
|---|---|---|
| 1x128b, 16-scalar chunk | 526 µs | **496 µs** (−6%) |
| 4x128b, 16-scalar chunk | 1.143 ms | **998 µs** (−13%) |
| 4x128b, 2-scalar chunk | 1.449 ms | **860 µs** (−41%) |
| 4x128b, 1-scalar chunk | 1.791 ms | **1.324 ms** (−26%) |

### The `Either` did not survive the no-rayon shim

`binius-utils` ships a `rayon` subset used when the `rayon` feature is off, which is the default.
Its `ParallelIterator for Either` panics in `as_inner`:

```
not implemented: as_inner not supported for Either
```

`as_inner` is how the trait's own `opt_len` and `IndexedParallelIterator::len` read a length without consuming the iterator, and Merkle tree construction asks for that length.
So on the default feature set, every commitment through this path panicked.

This is a pre-existing hole rather than one this branch opened: `chunks_par` already returns an `Either`, and any caller asking it for a length hits the same panic.

Both methods now dispatch on the side the `Either` holds instead of borrowing the inner iterator.
`as_inner` still cannot be implemented — there is no `Either<L::Inner, R::Inner>` inside an `&Either<L, R>` to point at — so its comment now names the methods that cover for it.

### The change

`commit_field_buffer` stops being a free function beside the trait and becomes a provided method on `MerkleTreeProver`, next to `commit` and `commit_iterated`:

```
- if log_batch_size >= P::LOG_WIDTH { ...big chunks... } else { ...small chunks... }
+ self.commit_iterated(buffer.par_chunk_scalars(log_leaf_len), 1 << log_leaf_len)
```

Both preconditions now hold by construction.
`par_chunk_scalars` asserts `log_chunk_size <= log_len`, and the leaf count is then a power of two for free.

### Tests

The old `test_parallel_iterator` only pinned the two private helpers to each other.
Both were wrong in the same way, so it passed.

It is replaced by a comparison against `MerkleTreeProver::commit`, which reads a flat scalar slice and never sees the packing — an independent reference.
The sweep covers buffer and leaf sizes on both sides of the packing width, on two packing widths.

### Scope

- **new:** `FieldBuffer::par_chunk_scalars` in `binius-math`, with a proptest cross-checking every chunk against the serial `chunk` accessor
- **deleted:** `to_par_scalar_big_chunks`, `to_par_scalar_small_chunks`, and the free `commit_field_buffer`
- **changed:** `commit_field_buffer` is now a `MerkleTreeProver` method; two call sites updated
- **fixed:** `len` and `opt_len` on the no-rayon shim's `Either`, with a test pinning the length of each side
- **tidied:** a `ProverDigest<T, M>` alias retires all three `#[allow(clippy::type_complexity)]` in the module, and the redundant `F` parameter is gone
- `chunks_par` is untouched — it still serves consumers that want `FieldSlice` chunks

Behaviour changes only where the old code was wrong: a buffer narrower than one packed word.
Everywhere else the committed scalar sequence is unchanged.

### Verification

- `cargo check --workspace --all-targets`, `clippy -p binius-utils -p binius-math -p binius-iop-prover --all-targets`, nightly `fmt --check` — all clean
- `cargo test -p binius-utils -p binius-math -p binius-iop-prover` — pass, with and without the `rayon` feature


crates/math/src/field_buffer.rs
```diff
@@ -353,6 +353,54 @@ impl<P: PackedField, Data: Deref<Target = [P]>> FieldBuffer<P, Data> {
...
+			// Scalars `[start, start + chunk_len)` span these words:
+			//
+			//     word_range = start / WIDTH .. ceil((start + chunk_len) / WIDTH)
+			//
+			// Below the packing width that is one word, and `lane_offset` picks the chunk in it.
+			let start = chunk_index << log_chunk_size;
+			let word_range = (start >> P::LOG_WIDTH)..(start + chunk_len).div_ceil(P::WIDTH);
+			let lane_offset = start & (P::WIDTH - 1);
+
+			P::iter_slice(&words[word_range])
```
jimpo: This nested iterator pattern doesn't look very efficient. Can we return an Either-based iterator, depending whether chunk size is less than packing width or not?

jimpo (changes requested):

crates/math/src/field_buffer.rs
tcoratger: Fixed

tcoratger: Done in 68a70c7d — the branch is now taken once, outside the iterator:

```
chunk >= one packed word  ->  a run of whole words        (par_chunks + iter_slice)
chunk <  one packed word  ->  a lane range inside a word  (map over the lanes of one word)
```

Both sides wrap into `Either`, so the outer parallel iterator and the per-chunk scalar iterator are each a two-variant union chosen before any scalar is read. The nested version paid for that choice per scalar: for a whole-word chunk the `skip` and `take` are no-ops that still cost a branch each, and for a sub-word chunk the `flat_map` walked a one-word slice it never had to advance.

Summing every scalar of a 2^20-scalar buffer, chunk by chunk:

| | old | new |
|---|---|---|
| 1x128b, 16-scalar chunk | 526 µs | **496 µs** (−6%) |
| 4x128b, 16-scalar chunk | 1.143 ms | **998 µs** (−13%) |
| 4x128b, 2-scalar chunk | 1.449 ms | **860 µs** (−41%) |
| 4x128b, 1-scalar chunk | 1.791 ms | **1.324 ms** (−26%) |

One thing this turned up, worth flagging: the `Either` did not survive the no-rayon shim. Its `ParallelIterator` impl panics in `as_inner`, which is exactly how the trait's own `opt_len` and `IndexedParallelIterator::len` read a length without consuming the iterator. Merkle tree construction asks for that length, so with `--no-default-features` every commitment through this path panicked with `not implemented: as_inner not supported for Either`.

That is a pre-existing hole rather than something this branch introduced — `FieldBuffer::chunks_par` already returns an `Either`, and any caller that asks it for a length hits the same panic. Both methods now dispatch on the side the `Either` holds instead of borrowing the inner iterator, and a shim test pins the length of each side. `as_inner` itself still cannot be implemented (there is no `Either<L::Inner, R::Inner>` inside an `&Either<L, R>` to point at), so its comment now says which methods cover for it.

tcoratger:
@jimpo Just fixed the comment :)

jimpo (approved):

PR #2040, opened 2026-08-08
BINIUS-207: one-padding MLE-check prover for a product-check layer

benji-potamus:
First of two PRs for [BINIUS-207](https://linear.app/jimpo/issue/BINIUS-207/batch-prodcheck-proving-with-variable-padding-unequal-tree-depths). This one adds the per-layer machinery; #2041 wires it into a `batch_prove_unequal_depths` entry point.

Batching product checks over trees of unequal depths pads each shallow tree up to the deepest one with constant-1 leaves. That leaves its product unchanged and makes the batch's layer loop uniform, so the verifier's control flow depends only on the maximum depth. The substance is that the prover never has to materialize a padded layer: each padded round polynomial is a cheap correction of the unpadded one.

The protocol and the derivation are the *Batched Product Checks of Unequal Depths* appendix of the whitepaper (`app:padded-prodcheck`).

### The prover

`OnePadMleCheckProver` decorates the unpadded layer's own MLE-check, over a claim point split into a padding segment below a real one. MLE-check binds high-to-low, so the real rounds come first:

- *real rounds* — delegate to the inner prover, then apply the affine correction `R'(X) = 1 + q·(R(X) − 1)`, for `q` the equality weight of the claim point's padding segment;
- *padding rounds* — by now every real variable is bound, so the inner prover's two child evaluations are **scalars** and the round polynomial is the closed form `(1 − e) + e·sel(E(X), g₀)·sel(E(X), g₁)`. No multilinear is touched.

Padding costs O(1) per round. With no padding the inner prover runs every round untouched.

Restricting to full trees — one scalar product each, no content variables — is what makes the second phase scalar-only: there is no partially-folded table to carry through it, no third phase to resume on transformed columns, and no enabling changes needed elsewhere in `sumcheck`.

### Testing

The prover is checked against an ordinary bivariate-product MLE-check run over an explicitly materialized one-padded layer, round polynomial by round polynomial and then on the child evaluations, across real rounds {0, 1, 3} × padding {0, 1, 3}. A second case pins the degenerate shape a shallow tree's purely-padding layers take, where one child is identically one.

### DIFF
--- a/crates/ip-prover/src/prodcheck/mod.rs
+++ b/crates/ip-prover/src/prodcheck/mod.rs
@@ -23,6 +23,8 @@ use crate::{
 	},
 };
 
+pub mod one_pad_mle;
+
 /// Witness-based prover for the product check protocol.
 ///
 /// This prover reduces the claim that a multilinear polynomial evaluates to a product over a
--- a/crates/ip-prover/src/prodcheck/one_pad_mle.rs
+++ b/crates/ip-prover/src/prodcheck/one_pad_mle.rs
@@ -1,0 +1,360 @@
+// Copyright 2026 The Binius Developers
+
+//! MLE-check prover for one layer of a one-padded product check.
+//!
+//! One-padding lifts a product tree of depth $k$ to depth $n \ge k$ by filling the extra leaves
+//! with ones, which leaves the tree's product unchanged. Batching product checks of unequal depths
+//! pads each shallow tree up to the deepest one, so the batch's layer loop runs a single uniform
+//! schedule and the verifier never learns the individual depths. See the *Batched Product Checks of
+//! Unequal Depths* appendix of the Binius64 whitepaper for the protocol and the derivation behind
+//! the round polynomials below.
+//!
+//! The point of this module is that the prover never materializes a padded layer:
+//! [`OnePadMleCheckProver`] wraps the unpadded layer's own MLE-check and corrects its messages at a
+//! cost of $O(1)$ per round.
+
+use std::{iter, mem};
+
+use binius_compute::Allocator;
+use binius_field::{Field, PackedField};
+use binius_ip::sumcheck::RoundCoeffs;
+use binius_math::{FieldVec, multilinear::eq::eq_one_var};
+
+use crate::sumcheck::{
+	bivariate_product_mle,
+	common::{MleCheckProver, SumcheckProver},
+};
+
+/// The one-padding selector $\textsf{sel}(s, v) = 1 + (v - 1) s$.
+///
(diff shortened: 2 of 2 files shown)


jimpo (commented):
Move one_pad_mle to prodcheck. It's specifically for that.

benji-potamus:
Moved — `sumcheck::one_pad_mle` is now `prodcheck::one_pad_mle`. `prodcheck.rs` becomes `prodcheck/mod.rs` to host it (the repo has no `foo.rs` + `foo/` submodule precedent), and the module itself is private with `OnePadMleCheckProver` re-exported, so the public path is `prodcheck::OnePadMleCheckProver`. Force-pushed; CI re-running.

Also pushed the rest of the stack for the bigger picture: **#2041** adds `batch_prove_unequal_depths` and the verifier-side `unpad_leaf_claim`, and is based on this branch.

benji-potamus:
Done — the whole stack is now restricted to full trees (no content variables), and it collapses a lot more than I expected. Force-pushed both PRs.

**The enabling commit is gone.** With no content dimension, the inner prover has *zero* variables left when the padding rounds begin, so its two child evaluations come straight out of `finish()`. That removed the need to stop it early (`SharedMleCheckProver::into_store`) and, with it, the need to name its type (`BivariateProductMleEvaluator`) — `sumcheck/bivariate_product_mle.rs` and `sumcheck/round_evaluator.rs` are untouched now. `OnePadMleCheckProver<F, Inner>` is generic over the inner prover instead, like `PaddedSumcheckDecorator`.

**Three phases became two.** There is no content phase: nothing to transform, no second prover to build, no sub-packing-width dead-lane handling. And the padding-round polynomial dropped its `(z, S, Q)` scalars — with scalar children the bracket factors back into the one-padding selector itself:

```
R'(X) = (1 - e) + e * sel(E(X), g_0) * sel(E(X), g_1)
```

which is the definition of `OnePad` read off directly, rather than its expanded moments. The running inner claim no longer has to be tracked at all, since `Q` was only ever `(g_0 - 1)(g_1 - 1)`.

**The driver lost its wrapper struct.** `UnequalDepthInstance` is gone: a shallow tree's padding layers are one-paddings of its scalar product, which is already `claimed_products[i]`. So the signature is `batch_prove`'s minus `content_point`, and `unpad_leaf_claim` loses its `content_len`.

Net: 1090 → 760 lines changed, across 3 files instead of 6, and 3 commits instead of 2. Everything still verifies the same way — the differential test against a materialized padded layer, and the end-to-end tests against unchanged `prodcheck::verify` plus transcript parity with `batch_prove` at equal depths.

jimpo (approved):

PR #2041, opened 2026-08-08
BINIUS-207: batched product-check proving for trees of unequal depths

benji-potamus:
Second and last PR for [BINIUS-207](https://linear.app/jimpo/issue/BINIUS-207/batch-prodcheck-proving-with-variable-padding-unequal-tree-depths). #2040 has merged, so this is now a standalone one-commit PR against `main`.

`batch_prove` requires every prover to have the same number of product layers. `batch_prove_unequal_depths` drops that requirement: each shallower tree is proved as a product check over the one-padding of its witness — the same witness with constant-1 leaves filling the extra depth, which leaves its product unchanged. The transcript is then exactly that of an equal-depth batch of the maximum depth, so `binius_ip::prodcheck::verify` runs unchanged and never learns the individual depths.

Every tree must reduce over *all* of its witness variables, so each product is a scalar and there is no content point. Every layer runs through the `OnePadMleCheckProver` from #2040, so no padded layer is materialized. Like `batch_prove_until_final_layer`, the driver stops one layer short and hands back the final-layer provers, so a caller can batch the widest (most expensive) layer with other sumchecks.

### Shape

- The signature is `batch_prove`'s, minus `content_point`. There is no per-instance wrapper struct: a shallow tree's padding layers are one-paddings of its scalar product, which is already `claimed_products[i]`.
- The layer loop is the one `batch_prove` already used, extracted as `prove_layer_rounds` rather than duplicated. Its existing call sites are unchanged, and a transcript-parity test pins the extraction.
- `binius_ip::prodcheck::unpad_leaf_claim` is the verifier's last step: the batch's per-tree leaf claims are on the *padded* witnesses, and this divides out the padding coordinates' equality weight and drops them from the point. It asserts on the negligible zero-weight event — those coordinates are the verifier's own challenges, so no prover can induce it.

The one branch in the whole driver is which buffer a tree hands over: `[product, 1]` while the batch is still above it, its own next layer once the batch reaches it. Its purely-padding layers are not a special case of the round polynomials — they are the general case with a child that happens to be one.

### Testing

End to end against unchanged `prodcheck::verify(n_max)`: mixed depths, a single tree, a power-of-two count with a >1 depth gap, all-depth-1, and a depth-1 tree beside a depth-6 one (so all but its last reduction is padding). Each test unpads every tree's leaf claim and checks it against that tree's own witness. A further test asserts byte-for-byte transcript equality with `batch_prove` when the depths happen to be equal.

### DIFF
--- a/crates/ip/src/prodcheck.rs
+++ b/crates/ip/src/prodcheck.rs
@@ -17,7 +17,7 @@
 //! ; X) \text{eq}(z ; Z) p_{i+1}(z, 0, x) p_{i+1}(z, 1, x) $$
 
 use binius_field::Field;
-use binius_math::line::extrapolate_line;
+use binius_math::{line::extrapolate_line, multilinear::eq::eq_one_var};
 use binius_transcript::Error as TranscriptError;
 
 // Re-export MultilinearEvalClaim from crate root for backward compatibility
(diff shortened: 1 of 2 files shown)


crates/ip/src/prodcheck.rs
```diff
@@ -71,6 +71,56 @@ where
...
+) -> Result<MultilinearEvalClaim<F>, Error> {
+	assert!(point.len() >= content_len + n_pad_vars); // precondition
+
+	let pad_eq = point[content_len..content_len + n_pad_vars]
+		.iter()
+		.map(|&coord| eq_one_var(F::ZERO, coord))
+		.product::<F>();
+	// The weight vanishes only if a padding coordinate is one, which the verifier's challenges are
+	// with probability at most `n_pad_vars / |K|`.
+	if pad_eq == F::ZERO {
```
jimpo: assert, not VerificationError. This happens with negligible probability.

jimpo (changes requested):

crates/ip/src/prodcheck.rs
benji-potamus: Done. `unpad_leaf_claim` returns `MultilinearEvalClaim<F>` directly now, with `assert!(pad_eq != F::ZERO)`, and the `ZeroPaddingEquality` variant is gone from `VerificationError`. The `# Panics` note records why it's safe to assert: the padding coordinates are the verifier's own challenges, so no prover can induce the zero. That also makes it consistent with the prover side, which already asserted on the same event.

jimpo (approved):

PR #2042, opened 2026-08-08
BINIUS-403: MLE-check prover for zero-fraction-padded fracadd layers

benji-potamus:
Bottom of a two-PR stack for [BINIUS-403](https://linear.app/jimpo/issue/BINIUS-403/batch-fracaddcheck-proving-for-trees-of-unequal-depth). This PR adds the per-layer prover; #2043 (stacked on top) adds the batch driver and the verifier-side unpadding.

## Why

Padding a fractional-addition tree up to a greater depth fills the extra leaves with the **zero fraction** `0/1` — numerators with zero, denominators with one — which leaves the tree's fractional sum unchanged, since `a/b + 0/1 = a/b`. That is what lets trees of unequal depths batch: pad each shallow tree up to the deepest one and the batch's layer loop runs a single uniform schedule, so the verifier's control flow depends only on the maximum depth.

This is the fractional-addition analog of the one-padding that BINIUS-207 built for prodcheck (#2040). The whitepaper's *Batched Product Checks of Unequal Depths* appendix derives the multiplicative case; the denominators here carry that one-padding verbatim, and the numerators carry the zero-padding that falls out of the same lemma.

## What

**`fracaddcheck::zero_pad_mle::ZeroPadMleCheckProver`** reduces one padded layer without materializing it, at `O(1)` extra work per round. The claim point is node coordinates only, padding lowest:

```
[ padding (nu) | real (m) ]
```

MLE-check binds high-to-low, so the real rounds come first:

- **Real rounds** delegate to the unpadded layer's own MLE-check and correct its two round polynomials. Off the all-zeros padding slab every leaf is the zero fraction, whose numerator composition vanishes and whose denominator composition is one, so with `q = eq(0^nu, rho_pad)`:

  ```
  R_num(X) = q * R(X)                R_den(X) = 1 + q * (R(X) - 1)
  ```

- **Padding rounds** touch no multilinear. Every real variable is bound, so the inner prover's four child evaluations are scalars and both polynomials are closed forms in them, quadratic through `E(X)`, the equality weight of the bound padding coordinates together with this round's.

**`frac_add_mle::new_split_half`** (first commit) is the layer-prover construction lifted out of `FracAddCheckProver::layer_prover`, which consumes the prover and so can't serve a driver that pops layers itself. It now sits beside the evaluators it wires up, mirroring `bivariate_product_mle::new_split_half`. The `LayerProver` alias moves too and is re-exported from `fracaddcheck`, so its public path is unchanged. (`pop_layer`, the other half of that split, lives in #2043 beside its only caller.)

## Testing

Both round-polynomial phases are checked differentially: the padded prover and an ordinary fracadd MLE-check over an **explicitly materialized** padded layer run in lockstep and must agree on every round polynomial and every child evaluation. Covered across real-round counts `{0, 1, 3}` × padding lengths `{0, 1, 3}`, plus the degenerate layer whose high child is the zero fraction — the shape the driver feeds in for a tree the batch has not yet reached.

## Note for review

`zero_pad_mle` and `prodcheck::one_pad_mle` are structurally parallel — same phase machine, same prefix table, same affine correction on the denominator. I did not factor out a shared decorator: the compositions differ (a bivariate product over two columns vs. fractional addition over four), so unifying them would need the decorator to be generic over the composition, which reads worse than two ~250-line siblings that each state their own closed form. Happy to reconsider.

### DIFF
--- a/crates/ip-prover/src/fracaddcheck/mod.rs
+++ b/crates/ip-prover/src/fracaddcheck/mod.rs
@@ -30,6 +30,8 @@ use crate::{
 /// Both claims share the same evaluation point, that of the layer they describe.
 pub type FracEvalClaim<F> = (MultilinearEvalClaim<F>, MultilinearEvalClaim<F>);
 
+pub mod zero_pad_mle;
+
 pub use crate::sumcheck::frac_add_mle::LayerProver;
 
 /// A numerator/denominator pair of pooled column buffers.
--- a/crates/ip-prover/src/fracaddcheck/zero_pad_mle.rs
+++ b/crates/ip-prover/src/fracaddcheck/zero_pad_mle.rs
@@ -1,0 +1,397 @@
+// Copyright 2026 The Binius Developers
+
+//! MLE-check prover for one layer of a zero-fraction-padded fractional-addition check.
+//!
+//! Zero-fraction padding lifts a fractional-addition tree of depth $k$ to depth $n \ge k$ by
+//! filling the extra leaves with the zero fraction $0/1$, which leaves the tree's fractional sum
+//! unchanged. The numerators are therefore zero-padded and the denominators one-padded. Batching
+//! fracadd checks of unequal depths pads each shallow tree up to the deepest one, so the batch's
+//! layer loop runs a single uniform schedule and the verifier never learns the individual depths.
+//!
+//! This is the fractional-addition analog of [`crate::prodcheck::one_pad_mle`]; the *Batched
+//! Product Checks of Unequal Depths* appendix of the Binius64 whitepaper derives the multiplicative
+//! case, whose one-padding is exactly the padding these denominators carry.
+//!
+//! The point of this module is that the prover never materializes a padded layer:
+//! [`ZeroPadMleCheckProver`] wraps the unpadded layer's own MLE-check and corrects its messages at
+//! a cost of $O(1)$ per round.
+
+use std::{iter, mem};
+
+use binius_compute::Allocator;
+use binius_field::{Field, PackedField};
+use binius_ip::sumcheck::RoundCoeffs;
+use binius_math::{FieldVec, multilinear::eq::eq_one_var};
+
+use crate::sumcheck::{
+	common::MleCheckProver,
+	frac_add_mle::{self, LayerProver},
+};
+
+/// The one-padding selector $\textsf{sel}(s, v) = 1 + (v - 1) s$.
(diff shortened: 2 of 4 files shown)


crates/ip-prover/src/fracaddcheck/mod.rs
```diff
@@ -171,6 +167,16 @@ where
 		self.layers.len()
 	}
 
+	/// Pops the widest remaining layer and returns its numerator and denominator buffers.
+	///
+	/// # Preconditions
+	/// * `self.n_layers() >= 1`
+	pub fn pop_layer(&mut self) -> PooledFractionalBuffer<A, P> {
```
jimpo: This is unused. remove it

crates/ip-prover/src/fracaddcheck/zero_pad_mle.rs
```diff
@@ -0,0 +1,400 @@
...
+	let pad_eq = pad_eq_prefixes[pad_len];
+	assert!(pad_eq != F::ZERO, "a padding coordinate of the claim point equals one");
+
+	// The padded claims are the images of the unpadded ones under the padding, so the inner prover
+	// starts from the preimages.
+	let pad_eq_inv = pad_eq.invert_or_zero();
+	let [num_claim, den_claim] = claims;
+	let inner_claims = [
+		num_claim * pad_eq_inv,
+		F::ONE + (den_claim - F::ONE) * pad_eq_inv,
```
jimpo: nit: use `select` helper

jimpo (changes requested):

crates/ip-prover/src/fracaddcheck/mod.rs
benji-potamus: Right — `pop_layer` has no caller until the batch driver in #2043. Moved it there rather than deleting it, so it lands in the commit that uses it and this PR adds no dead code.

crates/ip-prover/src/fracaddcheck/zero_pad_mle.rs
benji-potamus: Done — that expression was `select` spelled out. Now `[num_claim * pad_eq_inv, select(pad_eq_inv, den_claim)]`, which also makes the asymmetry between the two claims read directly: the numerator is scaled, the denominator goes through the padding selector.

jimpo (approved):

PR #2043, opened 2026-08-08
BINIUS-403: batched fracadd proving for trees of unequal depths

benji-potamus:
Second and final PR for [BINIUS-403](https://linear.app/jimpo/issue/BINIUS-403/batch-fracaddcheck-proving-for-trees-of-unequal-depth). #2042 (the per-layer padded prover this drives) has merged, so this is now based directly on `main` and is the whole remaining change.

## What

**`fracaddcheck::batch_prove_unequal_depths`** is `batch_prove` without the requirement that every tree have the same depth. Each shallower tree is proved as a fracadd check over its zero-fraction-padded witness, so the transcript is exactly that of an equal-depth batch of the maximum depth: the verifier runs the ordinary `binius_ip::fracaddcheck::verify` over `n_layers` layers and never learns the individual depths. The prover never materializes a padded witness — each layer goes through `zero_pad_mle`.

The layers a tree spends while the batch is still above it are **not a special case**: they are paddings of the tree's own fractional sum, whose two children are that fraction and the zero fraction `0/1`. Feeding those two children to the same prover yields the degree-1 messages the protocol wants, so there is one uniform construction and no separate code path.

As with prodcheck, every tree must reduce over all of its witness variables, so the fractional sums are scalars and there is no content point. The padding is only worth its bookkeeping on full trees, and dropping the content dimension keeps that bookkeeping to four scalars per layer.

`FracAddCheckProver::pop_layer` comes with the driver — it's the accessor that lets the driver take a tree's widest layer without consuming the prover, and this is its only caller.

**`binius_ip::fracaddcheck::unpad_leaf_claim`** is the verifier-side counterpart. `verify` is oblivious to the padding, so for a padded tree it outputs claims on the padded witness; this divides out the padding coordinates' equality weight and drops them from the point:

```
num = num' / e        den = 1 + (den' - 1) / e        e = eq(0^nu, rho_<nu)
```

`e = 0` requires a verifier challenge to equal one, which no prover can induce, so it is an assertion rather than a verification error (matching `prodcheck::unpad_leaf_claim`).

## Structure

Like `batch_prove_until_final_layer`, the driver stops one layer short and hands back each tree's reduced fraction beside the prover for its final (widest) layer — the layer whose reduction dominates the cost, which a caller can therefore batch with other sumchecks.

## Testing

Five end-to-end tests prove a batch against the depth-oblivious verifier and then unpad each tree's leaf claims and check them against that tree's own witness: mixed depths `[2,4,5]`, a single tree, a power-of-two batch where the shallowest is padded by four layers and the deepest not at all, all-depth-1, and `[1,6]` where all but one of the shallow tree's reductions are padding.

One more pins the transcript: at equal depths every tree is padded by nothing, so `batch_prove_unequal_depths` must emit **byte-for-byte** what `batch_prove` does. It does.

### DIFF
--- a/crates/ip-prover/src/fracaddcheck/mod.rs
+++ b/crates/ip-prover/src/fracaddcheck/mod.rs
@@ -6,12 +6,16 @@ use binius_compute::{Allocator, VecLike};
 use binius_field::{Field, PackedField};
 use binius_ip::{mlecheck, prodcheck::MultilinearEvalClaim, sumcheck::RoundCoeffs};
 use binius_math::{
-	FieldBuffer, FieldVec, line::extrapolate_line_packed, multilinear::eq::eq_ind_partial_eval,
+	FieldBuffer, FieldVec,
+	batch_invert::BatchInversion,
+	line::extrapolate_line_packed,
+	multilinear::eq::{eq_ind_partial_eval, eq_one_var},
 };
 use binius_utils::rayon::{
 	iter::{IntoParallelIterator, IntoParallelRefMutIterator, ParallelIterator},
 	task_size::{IndexedParallelIteratorExt, WorkPerItem},
 };
+use either::Either;
 use itertools::izip;
 
 use crate::{
@@ -25,14 +29,16 @@ use crate::{
 	},
 };
 
+pub mod zero_pad_mle;
+
+use zero_pad_mle::{ConstantFraction, ZeroPadMleCheckProver};
+
+pub use crate::sumcheck::frac_add_mle::LayerProver;
+
 /// The numerator and denominator evaluation claims of one fractional-addition layer.
 ///
 /// Both claims share the same evaluation point, that of the layer they describe.
 pub type FracEvalClaim<F> = (MultilinearEvalClaim<F>, MultilinearEvalClaim<F>);
-
-pub mod zero_pad_mle;
-
-pub use crate::sumcheck::frac_add_mle::LayerProver;
 
 /// A numerator/denominator pair of pooled column buffers.
 type PooledFractionalBuffer<A, P> = (FieldVec<P, A>, FieldVec<P, A>);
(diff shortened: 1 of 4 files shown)


crates/ip-prover/src/fracaddcheck/mod.rs
```diff
@@ -750,9 +762,164 @@ where
...
+
+	izip!(provers, pad_lens, roots, claims)
+		.map(|(prover, &pad_len, &(root_num, root_den), &(num, den))| {
+			let alloc = prover.alloc;
+			// While the batch is still above this tree, the layer is a padding of the tree's own
+			// fractional sum, whose two children are that fraction and the zero fraction 0/1. Only
+			// once the batch reaches the tree does it start consuming its layers.
+			let (layer_num, layer_den) = if node_len < pad_len {
+				(
+					FieldBuffer::from_values_in(alloc, &[root_num, F::ZERO]),
```
jimpo: Why is this not simply `&[root_num]` and `&[root_den]` (below)? Basically, why two values instead of 1?

crates/ip-prover/src/fracaddcheck/mod.rs
```diff
@@ -750,9 +762,164 @@ where
...
+					FieldBuffer::from_values_in(alloc, &[root_den, F::ONE]),
+				)
+			} else {
+				let layer = prover.pop_layer();
```
jimpo: Why do you need this? Can't you use `FracAddCheckProver` and wrap its LayerProver with a ZeroPadMleCheckProver?

crates/ip-prover/src/fracaddcheck/mod.rs
benji-potamus: Because a *layer* is the thing whose two halves get combined, so it needs at least one variable to split on. `[root_num, 0]` / `[root_den, 1]` is a 1-variable buffer whose low half is the tree's own fraction and whose high half is the zero fraction `0/1`. A single value would be a 0-variable buffer with no sibling to add it to, and `zero_pad_mle::new` rejects it (`num.log_len() >= 1`).

The second slot isn't filler to satisfy that precondition — it's the actual `b = 1` child. A padding layer is `ZeroPad/OnePad` of the root fraction, and its high child is identically `0/1`, so writing the pair out explicitly is what lets the ordinary fracadd prover handle these layers with no special path. It's also what makes both round polynomials come out degree 1 by themselves: with children `(N, D)` and `(0, 1)` you get `S = N`, `T = 0`, `Q = 0`, killing the quadratic terms.

crates/ip-prover/src/fracaddcheck/mod.rs
benji-potamus: Two things get in the way of `layer_prover` here.

**It consumes the prover, but half the trees must not pop anything this round.** While the batch is still above a tree (`node_len < pad_len`), that tree's layer is the synthesized `[root, identity]` pair — there is no layer to take, and its `FracAddCheckProver` has to come out of the round untouched for later. So the loop needs `&mut` access, whereas `layer_prover(self, ..)` takes ownership and hands back `Option<Self>`.

**The inner prover has to be seeded with the unpadded claim.** `ZeroPadMleCheckProver` wraps an MLE-check over the *unpadded* layer at the *unpadded* point — `eval_point[pad_len..]` with claims `num/q` and `sel(1/q, den)`. `zero_pad_mle::new` derives those; `layer_prover` would have to be handed them, which means computing the padding preimages at the call site.

And even setting both aside, the synthesized branch can't use `layer_prover` at all, so it would still go through `new_split_half` — mixing the two constructors for the two branches is more code than one `pop_layer` plus one uniform call.

That said, if what you're after is `zero_pad_mle` not depending on `frac_add_mle`, there's a clean version: make `new` take an inner-builder `FnOnce(Vec<F>, [F; 2]) -> Inner` instead of the buffers. The decorator computes the truncated point and preimage claims and hands them over, and the caller builds the layer prover however it likes — including via `layer_prover`. That drops `alloc`/`FieldVec`/`Allocator` from the module and makes it a pure padding decorator. Happy to do that if you'd prefer it; I left it out because the current signature keeps the call site to one construction path.

crates/ip-prover/src/fracaddcheck/mod.rs
jimpo: Hmm, but can't we just wrap a SumcheckProver for a constant fraction in a ZeroPadMleCheckProver instead of wrapping a real LayerProver?

crates/ip-prover/src/fracaddcheck/mod.rs
jimpo: I don't see the problem. We can consume the FracAddCheckProver in the closure, and simply pass it through in the case when we don't pop the layer. And we know that when we do pop the layer, that the returned Option is Some because the padded depths equal.

> And even setting both aside, the synthesized branch can't use layer_prover at all, so it would still go through new_split_half — mixing the two constructors for the two branches is more code than one pop_layer plus one uniform call.

No, I don't care about the uniform call -- two separate constructors is cleaner.

I also don't understand the second point, we can just pull that code out of zero_pad_mle::new.

jimpo (changes requested):

crates/ip-prover/src/fracaddcheck/mod.rs
benji-potamus: You're right on both counts — done.

`zero_pad_mle::new` now takes an inner-builder `FnOnce(Vec<F>, [F; 2]) -> Inner`: it owns the padding arithmetic, hands the builder the real segment of the claim point and the claims with the padding divided back out, and lets the caller decide what to construct. `pop_layer` is gone; the driver consumes each `FracAddCheckProver` in the closure and passes it straight through in the no-pop case, exactly as you described. The module no longer mentions `Allocator`, `FieldVec` or `frac_add_mle` at all — it's a pure padding decorator now.

I also needed `impl MleCheckProver for Either`, since the two branches now produce different inner types. BINIUS-65 left that adaptor on `SumcheckProver` alone when it split the traits; added it in its own commit.

crates/ip-prover/src/fracaddcheck/mod.rs
benji-potamus: Yes — that's better than what I was defending, and it's what I've done. The synthesized buffer is gone; the branch now builds `ConstantFraction`, a zero-variable `MleCheckProver` holding `[num, 0, den, 1]` whose only real method is `finish`. Every variable of such a layer is a padding variable, so the decorator goes straight to its padding rounds and never asks for anything else — no allocation, no `MleStore`, no evaluators.

My "why two values" answer was explaining why the *materialized* layer needs two slots, which was beside the point once the layer needn't be materialized at all.

The differential test moved with it: `constant_fraction_matches_padded_reference` now drives `ConstantFraction` against an ordinary fracadd MLE-check over the explicitly padded `[root, identity]` layer and requires identical round polynomials and child evaluations, so the stand-in is checked against the thing it stands in for.

crates/ip-prover/src/fracaddcheck/mod.rs
```diff
@@ -750,9 +753,186 @@ where
...
+	let mut next_provers = Vec::with_capacity(provers.len());
+
+	let layer_provers = izip!(provers, pad_lens, roots, claims)
+		.map(|(prover, &pad_len, &(root_num, root_den), &(num, den))| {
+			let mut remaining = None;
+			let layer_prover = zero_pad_mle::new(
+				pad_len.min(node_len),
+				node_point.to_vec(),
+				[num, den],
+				|point, [num_claim, den_claim]| {
```
jimpo: Get rid of the closure. Create the inner provers here and pass them to the ZeroPadMleCheckProver constructor. Compute all the pad_eq_prefixes here, outside the loop because they depend on the same node_point. Then batch invert to get the factor for computing the unpadded claims. You don't even need to take `roots` as an argument -- in the ConstantFraction case, the root is the de-padded claim.

jimpo (changes requested):

crates/ip-prover/src/fracaddcheck/mod.rs
benji-potamus: All four, done.

The closure is gone: `zero_pad_mle::new(pad_eq_prefixes, eval_point, inner)` takes the built inner prover. It no longer takes the claims at all — de-padding moved out to `unpad_claims(pad_eq_inv, claims)`, which the caller applies before seeding.

`layer_provers` now computes one prefix table for the whole batch (every tree's padding segment is a prefix of the same node point), collects each tree's `q` from it, and runs a single `BatchInversion` over them instead of an `invert_or_zero` per tree.

`roots` is gone — you're right that the de-padded claim *is* the root fraction for an all-padding layer, so `ConstantFraction::new` takes it straight from `unpad_claims`. That also removes the `claimed_fractions.clone()` the driver was carrying for it. I added an assertion of that identity to `constant_fraction_matches_padded_reference` so the equivalence is pinned rather than assumed.

Rebased onto `main` (f1caa201) along the way. All three commits verified independently, and the equal-depth test still shows byte-identical transcripts against `batch_prove`.

jimpo (approved):

PR #2046, opened 2026-08-08
Charge one Merkle multi-proof per input oracle in the FRI proof size

tcoratger:
> **Stacked on #2045.** This branch contains that commit too, so review #2045 first and read only
> the second commit here. Once #2045 merges this diff reduces to it alone.

`size_estimation::proof_size` modelled a batch of input oracles as a single combined oracle.
The prover does not. No protocol change; a reported number moves.

### The problem

`BatchBrakedownOracleProver::open_queries` loops over the input oracles and sends each one's
openings against its own commitment:

```
for oracle in &self.oracles {
    oracle.open_queries(indices, channel);   // its own Merkle multi-proof
}
```

A batch of $N$ therefore sends $N$ multi-proofs. `proof_size` charged one.

### What was miscounted

- **Merkle proofs** — one charged, $N$ sent.
- **Commitment digests** — `2 + n_arities` charged; a batch carries `N + 1 + n_arities`, one per
  input oracle, one per fold round, one for the terminal codeword.
- **Coset values** — charged against `params.log_batch_size()`, which folds in the
  `log_n_oracles` outer challenges. Those recombine values the input oracles already opened, so
  they cost no proof bytes. This one was an *over*count.

### The change

```
- arities = once(params.log_batch_size()).chain(fold_arities)   // one combined initial fold
- log_n_cosets = params.log_len()
+ for spec in params.input_oracles() {                          // one per committed oracle
+     open(log_dim - spec.log_lift + log_inv_rate, spec.log_batch_size());
+ }
+ let mut log_n_cosets = params.index_bits();                   // then the fold rounds
```

The two loops share one `open` closure, so the layer depth is always chosen for the tree it
indexes — the coupling that caused a panic when only half of it was corrected in #2045.

### Ground truth

The downstream `spartan-verifier` size test moved, so I printed the parameters it actually uses
rather than assume my number was the right one:

```
n_input_oracles = 3     all ZK, log_batch_size = 1, log_lift = 0
fold_arities    = []
log_dim = 7, log_inv_rate = 3, index_bits = 10
```

Three committed oracles and no fold rounds. The proof carries three multi-proofs over $2^{10}$-leaf
trees and four commitment digests; the old formula charged one and two.

| | charged before | actually sent |
|---|---|---|
| Merkle multi-proofs | 1 | 3 |
| commitment digests | 2 | 4 |

So `46784` was an underestimate, and `73600` is the honest figure. The direction and rough size are
what the structure predicts.

### The test now covers what it claimed to

`optimizer_estimate_matches_exact_proof_size` was scoped to a single oracle in #2045, because the
two computations diverged past that. That restriction is gone.

It now runs seven batch shapes — a lone oracle, equal lengths, lifted oracles, ZK mixed with
flexible non-ZK, and a non-power-of-two count — across 3 rates and 3 query counts, 63
configurations in all:

```
estimate + (N + 1 + n_arities) * digest_size == proof_size(params)
```

Together with #2045, the arity cost model and the exact byte count are pinned to each other for
every parameter shape I could construct.

### Scope

- **changed:** `size_estimation::proof_size`
- **changed:** the agreement test, from single-oracle to arbitrary batches
- **changed:** the pinned figure in `spartan-verifier`'s `proof_size_test`
- **left untouched:** the prover and verifier — this function only reports

### What this affects

Anything *reporting* a batched proof size was understating it: `channel::size_tracking` and the
spartan test. The benches are unaffected — they measure real `proof_bytes.len()`.

### Verification

- `cargo clippy -p binius-iop -p binius-spartan-verifier --all-targets`, `cargo +nightly fmt --all` — clean
- `cargo test -p binius-iop --lib fri::` — 6 passed
- `cargo test -p binius-spartan-verifier --test proof_size_test` — passes


jimpo (approved):

jimpo:
Created https://linear.app/jimpo/issue/BINIUS-405/rewrite-sizetrackingchannel-to-implement-merkleipverifierchannel as a follow-up issue

PR #2048, opened 2026-08-09
BINIUS-404: make ValueIndex segment-aware

benji-potamus:
Closes [BINIUS-404](https://linear.app/jimpo/issue/BINIUS-404/switch-valueindex-to-segment-aware-witnessindex).

## What changes

`ValueIndex(pub u32)` — an absolute position in the `ValueVec` — keeps its name but stops being an absolute position. It now names a word by its **segment** and its position *within* that segment:

```rust
pub enum ValueSegment { Constant = 0, InOut = 1, Private = 2, Scratch = 3 }
```

The pair is packed into one `u32`: the 2-bit tag on top, a 30-bit index below. Constraint systems hold millions of these, so packing matters — measured, an unpacked `{ segment, index }` struct would take `ShiftedValueIndex` from 8 bytes to 12.

Per the investigation on the issue, `Scratch` is the fourth segment rather than a separate flat space, and `ConstraintSystem::validate` now rejects any operand term naming it — so scratch words exist only in the circuit's wire mapping and its evaluation form, never in a constraint system.

## Why

The flat index conflated two things: *which value a constraint names* and *where the proving protocol stores that word*. Padding was the visible symptom — the constraint system carried `n_const_pad` / `n_inout_pad` / `n_private_pad` and an `is_padding` check whose only job was policing a flat address space whose shape is a BaseFold/shift-protocol requirement.

With segment-relative indices an index simply cannot name a padding word, so `is_padding` is deleted and the bounds check becomes "index < that segment's value count". Where each segment sits in the vector is now the layout's business.

## Commits

1. **Make `ValueIndex` segment-aware** — the substance.
2. **Rename back to `ValueIndex`** — this first landed as `WitnessIndex`; renaming back collapses most of the churn, since the ~170 sites that merely mention the type are unchanged in the net diff.
3. **Drop the reserved `INVALID`** — it was referenced nowhere and only existed to give `Default` a value. Its removal also takes the reserved last index, the off-by-one `SEGMENT_CAPACITY`, and a `Debug` special case.
4. **Resolve through `word_offset`** — replaces a `segment_starts: [usize; 4]` array threaded through the prover and verifier.
5. **Index the monster word tensor by segment** — the verifier's `FieldFn` input now carries one run per segment and reads a term at `r_y_tensor[segment][index]`.

## Notable points for review

- **`ValueVec` gained a flat `word`/`word_mut` accessor.** The eval-form interpreter addresses the vector as a register file — one `u32` register per position, scratch tail included — and the batched path indexes rows of a transposed array by flat position, so absolute addressing is unavoidable there. `EvalForm::build` resolves each wire's segment+index to a position once at bytecode-emission time via `ValueVecLayout::word_offset`, keeping the hot interpreter loop a single array index with no branch. This is the one place absolute addressing survives.
- **Two `word_offset` implementations, plus `ValueVec`'s own.** The prover never sees a `ValueVecLayout` — it's the compiler's view and carries `n_witness`/`n_internal`/`n_scratch` that a serialized `ConstraintSystem` lacks. So `ValueVecLayout` serves the frontend, `ConstraintSystem` serves the prover's `build_key_collection`, and `ValueVec` resolves against the offsets it stores (needed because `new_from_data` rebuilds from serialized segments with no layout to consult). After commit 5 the verifier needs none of them.
- **A stray `Scratch` index panics rather than reading a committed word** — it resolves past the end for both `ConstraintSystem` and a `new_from_data` vector.
- **The word-index tensor no longer carries padding.** Commit 5 cuts it into one run per segment, so only nameable words reach the encoded `FieldFn` input.
- **M4.** `ValueWords::term_words` turns `if value_index < self.witness_offset` into an exhaustive `match`. The `InOut` and `Scratch` arms panic with a comment naming why neither can occur — documenting the M4 inout gap the investigation identified, ahead of the separate protocol work.
- **Serialization bumps to v8**, and `test_data/constraint_system_v7.bin` is replaced by a regenerated `constraint_system_v8.bin`. Old files name words by absolute position and are not compatible.
- **Insta snapshots in `gate_fusion/tests/awhole.rs` are regenerated.** Private indices now count from zero, so `v[2]` becomes `v[0]` and so on; the formatter also renders inout wires as `io[N]`. The diffs are pure renumbering, which is itself a decent sanity check on the change.
- **Circuit stat snapshots are unchanged.** `stat.rs` counts distinct `ShiftedValueIndex` values, and flat ↔ (segment, index) is a bijection over the indices actually used, so the distinct counts are identical. Verified with `cargo run --example sha256 -- check-snapshot`.
- **No spartan changes.** Those crates have their own unrelated `WitnessIndex`/`WitnessSegment`; an over-broad rename touched them briefly and has been fully reverted.

## Follow-up (stacked PR, not in this one)

Now that indices are segment-relative, the padding fields and the power-of-two / hidden-≥-public shape rules can move out of `ConstraintSystem` into the prover/verifier layout. That is the actual decoupling payoff; keeping it separate keeps this diff reviewable.

Worth noting from the investigation: the frontend never emitted const padding in the first place (`offset_inout == n_const` always), so `n_const_pad` is dead weight today.

## Verification

Rebased onto `main` at "BINIUS-389: let the AND reduction consume unpadded operand columns". All checks re-run on the rebased tip with `RUSTFLAGS="-C target-cpu=native"`:

| check | result |
|---|---|
| `binius-core` | ✅ 99 passed |
| `binius-prover` | ✅ 37 + 5 passed (incl. `prove_verify`) |
| `binius-examples` | ✅ 22 passed |
| `sha256 check-snapshot` | ✅ matches |

The previous head was 7/8 green on CI (`rust-doc` failed on a private intra-doc link, fixed in the tip); `rust-checks-x86_64`, `-portable`, `arm64`, `wasm-tests`, `wasm-vanilla-build`, `format` and `circuits-snapshot` all passed.


crates/core/src/constraint_system/witness_index.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// The pair is packed into a single `u32`: the [`WitnessSegment`] in the top two bits and the
+/// index in the bottom [`Self::INDEX_BITS`]. Constraint systems hold millions of these, so the
+/// packing keeps a [`ShiftedValueIndex`](super::ShiftedValueIndex) at eight bytes rather than
+/// twelve.
+///
+/// The packing also makes the derived [`Ord`] order the words by segment and then by index, which
+/// is the order they occupy in the value vector.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct WitnessIndex(u32);
```
jimpo: In a new commit, rename this to `ValueIndex`. It will reduce the overall churn of the PR.

crates/core/src/constraint_system/witness_index.rs
```diff
@@ -0,0 +1,236 @@
...
+
+	/// The bits of the packed word holding the index.
+	const INDEX_MASK: u32 = (1 << Self::INDEX_BITS) - 1;
+
+	/// The witness index that is not considered to be valid.
+	///
+	/// This is the last index of the scratch segment, whose words no constraint may reference
+	/// anyway. Reserving it keeps the all-ones word invalid, so a zeroed-then-defaulted map reads
+	/// back as unassigned.
+	pub const INVALID: WitnessIndex = WitnessIndex(u32::MAX);
```
jimpo: I don't get this, do we need it?

crates/core/src/constraint_system/layout.rs
```diff
@@ -52,6 +52,26 @@ impl ValueVecLayout {
 		self.offset_witness + self.n_hidden_words
 	}
 
+	/// Returns the word each segment starts at, indexed by
+	/// [`WitnessSegment`](super::WitnessSegment).
+	///
+	/// This is the one place the section order is written down; both the value vector and the
+	/// evaluation form resolve a [`WitnessIndex`] through it. The witness and internal values
+	/// share the private segment, in that order, so it starts where the witness values do.
+	pub const fn segment_starts(&self) -> [usize; 4] {
```
jimpo: I don't like this. Instead of passing "segment_starts" around, pass around layout: &ValueVecLayout itself. And remove `WitnessIndex::offset_within` and put its logic in `word_offset`.

jimpo (changes requested):

crates/core/src/constraint_system/witness_index.rs
benji-potamus: Done in a new commit — `WitnessIndex` → `ValueIndex`. You're right that it collapses most of the churn: the type keeps its old name, so the ~170 call sites that just mention it are unchanged in the net diff, leaving only the sites that actually gain a segment.

I also renamed the companion enum `WitnessSegment` → `ValueSegment` and put the file back at `value_index.rs`, on the grounds that the container is `ValueVec`/`ValueVecLayout` so the vocabulary should match. That part is my inference, not something you asked for — say the word if you'd rather the segment enum keep the `Witness` name.

This also settles the naming question I raised in the PR description: `ShiftedValueIndex { value_index: ValueIndex }` now reads consistently, so no follow-up rename is needed there.

crates/core/src/constraint_system/witness_index.rs
benji-potamus: You're right, we don't. Removed in a new commit.

I checked before touching it: `INVALID` is referenced nowhere outside its own file. It existed only so `Default` had something to return, and `Default` is required because `SecondaryMap::new()` is bounded on it (`with_default` is the alternative, but that just moves the arbitrary choice to the call site).

So the sentinel is gone along with everything it dragged in — the reserved last index, `SEGMENT_CAPACITY` being one short of the index space, and the `Debug` special case. `SEGMENT_CAPACITY` is now simply `1 << INDEX_BITS`, and `Default` names the first constant with a comment saying it exists only as `SecondaryMap` filler and that the compiler assigns every wire an index before anything reads one.

Note this is inherited rather than new — the old `ValueIndex` had the same unused `INVALID = u32::MAX` and I carried it over without questioning it. Should have.

crates/core/src/constraint_system/layout.rs
benji-potamus: Done in a new commit. `segment_starts()` and `ValueIndex::offset_within` are both gone; the resolution logic now lives directly in `word_offset`, and callers pass the owning type rather than an array.

One wrinkle worth naming, since it shapes what "the layout" means here: the prover and verifier never see a `ValueVecLayout`. It's the compiler's view — it carries `n_witness`/`n_internal`/`n_scratch`, which a serialized `ConstraintSystem` doesn't have. So what actually gets passed depends on who has what:

- `ValueVecLayout::word_offset` — used by the frontend, i.e. `EvalForm::build` lowering wires to bytecode registers.
- `ConstraintSystem::word_offset` — used by `build_key_collection` and `OperationEvalFn`, which now take `&ConstraintSystem` instead of `[usize; 4]`. Both already had one in scope, so this is strictly less plumbing than before.
- `ValueVec` resolves against the three offsets it stores, since `new_from_data` rebuilds a vector from serialized segments and has no layout to consult.

That's three small four-arm matches instead of one shared helper. I read that as the right trade under "avoid premature generalization" — each match sits next to the offsets it reads and needs no indirection to follow — but it is duplication, so flagging it in case you'd rather they were unified some other way.

Verified: `Scratch` resolves past the end for both `ConstraintSystem` (which `validate` rejects outright) and a `new_from_data` vector (which has no scratch tail), so a stray scratch index panics rather than silently reading a committed word. Noted in both doc comments.

crates/spartan-frontend/src/circuit_builder.rs
```diff
@@ -8,8 +8,8 @@ use bytemuck::zeroed_vec;
 use smallvec::{SmallVec, smallvec};
 
 use crate::constraint_system::{
-	ConstraintSystem, ConstraintWire, MulConstraint, Operand, WireKind, Witness, WitnessIndex,
-	WitnessLayout, WitnessSegment,
+	ConstraintSystem, ConstraintWire, MulConstraint, Operand, ValueIndex, ValueSegment, WireKind,
```
jimpo: Ah, no. No changes to the spartan crates. Leave the naming here alone.

crates/verifier/src/protocols/shift/monster.rs
```diff
@@ -111,12 +114,18 @@ pub struct OperationEvalFn<'a, C, const ARITY: usize> {
 	/// The operation's constraints, each exposing its `ARITY` operands as an array in storage
 	/// order.
 	constraints: &'a [C],
+	/// The system the constraints belong to, which places an operand term's segment-relative
+	/// index in the word-index tensor.
+	constraint_system: &'a ConstraintSystem,
```
jimpo: I don't like storing this here. Instead, simply store the fields `n_constants`, `n_inout`, `n_hidden`. Then, in `split_input` read three different r_y_tensor segments `r_y_tensor_constants`, `r_y_tensor_inout`, `r_y_tensor_hidden`. Put them into an array `r_y_tensor: [&'i [E]; 3]`. Then index into it with the segment of the value index and the index, as a 2D array instead of using `word_offset`. The calling side that assembles the inputs should write them in the right way of course.

jimpo (changes requested):
After these changes,take the PR out of draft.

crates/spartan-frontend/src/circuit_builder.rs
benji-potamus: Good catch, and sorry — this was a straight mistake, not a judgement call.

When I did the rename I ran a repo-wide `sed` for `\bWitnessIndex\b` / `\bWitnessSegment\b`. The spartan crates have their **own** independent types by those names, so the sed clobbered them: 7 files across `spartan-frontend`, `spartan-prover` and `spartan-verifier`. Nothing to do with binius64's `ValueIndex`; I just didn't scope the pattern.

Reverted. I fixed it in the commits that introduced it rather than adding a revert on top — commit 2 (the rename) and commit 4, where a workspace-wide `cargo fmt` had also reformatted the files commit 2 touched. `jj diff -r 'main..@' --stat` now lists no spartan files at all, and the workspace builds.

The lesson for the rest of this PR: the sed was doing the renaming, so it was only ever as good as its scope. I've re-checked the remaining rename for the same failure mode — everything else it touched is in `core`/`frontend`/`prover`/`verifier`/`m4-prover`/`examples`, which are the crates that genuinely consume `binius_core::ValueIndex`.

crates/verifier/src/protocols/shift/monster.rs
benji-potamus: Done as described, in a new commit. `OperationEvalFn` now stores `n_constants` / `n_inout` / `n_hidden`; `split_input` cuts the tail of the input into `[&'i [E]; 3]`; a term reads as

```rust
r_y_tensor[svi.value_index.segment() as usize][svi.value_index.index() as usize]
```

which is just the term's own `(segment, index)` pair — no address arithmetic in between. `encode_operation_input` takes the three runs and writes them in `ValueSegment` order, and the assembly side in `verify.rs` slices the two indicators it already builds:

```rust
let r_y_tensor = [
    &public_tensor[..cs.n_const()],
    &public_tensor[cs.offset_inout()..cs.offset_inout() + cs.n_inout],
    &hidden_tensor[..cs.n_private],
];
```

Two things that fell out, both improvements:

- **The padding words no longer reach the input.** The tensor used to carry one entry per value-vector word, padding included; now it carries only words a constraint can name, so the encoded input is strictly smaller. The `n_public_words` / `value_vec_len` juggling in `operation_inputs` is gone with it.
- **This was the verifier's last use of `ConstraintSystem::word_offset`.** The flatten-then-re-resolve step is fully gone from this path. `word_offset` survives only for the prover's `build_key_collection`, which genuinely indexes a per-word array over the whole value vector.

The tests got simpler too — they no longer need a synthetic `ConstraintSystem`, just `(0, 0, n_words)` with empty constant and inout runs, since every term they generate is private.

Verified with `RUSTFLAGS="-C target-cpu=native"`: workspace clippy `-D warnings` clean, fmt clean, and the 5 `prove_verify` end-to-end tests pass — which is the check that matters here, since prover and verifier both evaluate this `FieldFn` and would diverge if the encode and split sides disagreed.

jimpo (approved):

PR #2051, opened 2026-08-09
BINIUS-405: rewrite SizeTrackingChannel to implement MerkleIPVerifierChannel

tcoratger:
Closes [BINIUS-405](https://linear.app/jimpo/issue/BINIUS-405/rewrite-sizetrackingchannel-to-implement-merkleipverifierchannel).

Moves the proof-size channel below the BaseFold reduction instead of standing in for it, so the size it reports is the size a prover sends rather than an estimate of it.

### The problem

The channel implemented the oracle-level interface, which is the level BaseFold *produces*. Standing at that level, it had to answer for openings it never performed:

```
before   IOP verifier -> SizeTrackingChannel        oracle openings replaced by a formula
after    IOP verifier -> BaseFold -> SizeTracking   oracle openings actually executed
```

So `verify_oracle_relations` short-circuited and added `fri::proof_size` instead. That formula is a second description of the protocol, kept in step with the real one by hand, and its own comment conceded it fell short:

> It is a slight underestimate because it does not account for some smaller BaseFold components (e.g. sumcheck coefficients within BaseFold, blinding elements for ZK).

### The idea

Implement [`MerkleIPVerifierChannel`] instead. BaseFold already compiles an oracle protocol down onto that interface, so the real reduction runs on top of the counter and every byte is counted by the code that would send it.

Each operation is priced the way the wire prices it:

- a commitment is one digest;
- a multi-opening is one internal layer plus one branch per index, which the scheme itself prices, then the opened leaf values;
- a committed vector is its elements, and needs no branch.

The oracle specs and FRI parameters the channel used to carry are gone. It takes only the scheme whose openings it prices.

### Why zeros are enough

Every receive returns zero. Zero is a fixed point of the folds and equality checks a verifier performs, so a verifier fed zeros reaches the same receives, in the same order, as it would on a real proof.

That is a claim about the protocol, not a convenience, so it is tested rather than asserted.

### Verification

`proof_size_exactness_test` proves an honest proof, takes its transcript length, and asserts the counted size equals it, over `log_inv_rate` in `{1, 2, 3}`. The rate moves the FRI arities, and with them the split between digests and values, so the three cases do not price the same way.

A transcript holds only sent bytes — observed values are absorbed without being written — so its length is exactly what a verifier reads.

| | bytes |
|---|---|
| previous estimate | 73600 |
| counted, and equal to a real proof | 73936 |

The 336 bytes are 21 field elements: the BaseFold sumcheck coefficients and the blinding elements the old formula omitted.

### Scope

- **changed:** `size_tracking.rs`, rewritten against the Merkle interface
- **changed:** finishing a BaseFold channel returns the Merkle channel it wrapped, without which the count would be dropped with it; one call site updated
- **new:** `proof_size_exactness_test`, pinning the count to a real transcript
- **left untouched:** `fri::proof_size`, still used by the arity search, which needs a size before any transcript exists

The count assumes a non-hiding scheme, whose leaves carry no salt. The trait exposes no salt width, the same limitation `fri::proof_size` has.

- `cargo clippy --workspace --all-targets` — clean
- `cargo doc --document-private-items` — clean

### DIFF
--- a/crates/iop/src/channel/size_tracking.rs
+++ b/crates/iop/src/channel/size_tracking.rs
@@ -1,99 +1,83 @@
 // Copyright 2026 The Binius Developers
 
-//! An [`IOPVerifierChannel`] that counts proof bytes without verifying, for proof-size estimation.
+//! A Merkle channel that counts proof bytes instead of verifying them.
 
-use binius_field::{BinaryField, util::FieldFn};
+use std::{marker::PhantomData, mem::size_of};
+
+use binius_field::{Field, util::FieldFn};
 use binius_ip::channel::IPVerifierChannel;
+use binius_utils::serialization::FixedSizeSerializeBytes;
 
 use crate::{
-	channel::{Error, IOPVerifierChannel, OracleLinearRelation, OracleSpec},
-	fri::{self, FRIParams},
+	merkle_channel::{Error, MerkleIPVerifierChannel},
 	merkle_tree::MerkleTreeScheme,
 };
 
-/// Default size in bytes for a single field element.
-const DEFAULT_ELEMENT_SIZE: usize = 16;
-
-/// Default size in bytes for a single oracle commitment.
-const DEFAULT_ORACLE_SIZE: usize = 32;
-
-/// An [`IOPVerifierChannel`] that tracks proof size without doing verification.
+/// What a commitment fixes about the tree behind it.
 ///
-/// All `recv_*` methods return dummy zero values and accumulate the expected byte count.
-/// Sampling and observation methods are no-ops.
-///
-/// After verification completes, call [`proof_size()`](Self::proof_size) to read the
-/// accumulated proof size.
-pub struct SizeTrackingChannel<'a, F: BinaryField, MerkleScheme_: MerkleTreeScheme<F>> {
-	element_size: usize,
-	oracle_size: usize,
-	oracle_specs: Vec<OracleSpec>,
-	fri_params: &'a [FRIParams<F>],
-	merkle_scheme: &'a MerkleScheme_,
-	next_oracle_index: usize,
-	proof_size: usize,
+/// A commitment carries no data here, only the two numbers that decide what opening it costs.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CommittedShape {
+	/// Field elements in each leaf.
+	leaf_size: usize,
+	/// Base-2 logarithm of the number of leaves.
+	depth: usize,
 }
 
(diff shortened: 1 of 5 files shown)


crates/iop/src/channel/size_tracking.rs
```diff
@@ -1,99 +1,83 @@
...
 }
 
-impl<F: BinaryField, MerkleScheme_: MerkleTreeScheme<F>> IPVerifierChannel<F>
-	for SizeTrackingChannel<'_, F, MerkleScheme_>
+impl<F, MerkleScheme_> IPVerifierChannel<F> for SizeTrackingChannel<'_, F, MerkleScheme_>
+where
+	F: Field + FixedSizeSerializeBytes,
+	MerkleScheme_: MerkleTreeScheme<F>,
 {
 	type Elem = F;
```
jimpo, line 70: You might be able to make this `type Elem = DummyElem<F>;`

tcoratger: Tried it — it compiles for the channel alone, but it breaks the composition this PR is built on.

`BaseFoldVerifierChannel` is bound on `MerkleIPVerifierChannel<F, Elem = F>`, so with a dummy element the channel can no longer sit underneath the reduction:

```
error[E0271]: type mismatch resolving `<SizeTrackingChannel<..>>::Elem == BinaryField128bGhash`
   |         .create_channel(SizeTrackingChannel::new(&merkle_scheme));
   |          expected `BinaryField128bGhash`, found `DummyElem<BinaryField128bGhash>`
```

That bound is not incidental. `recv_openings` and `recv_committed_vector` hand back `Vec<F>`, and FRI folds those opened values together with sampled challenges — `fold_coset(.., self.final_challenges, coset.to_vec())` in `fri/verify.rs`. Challenges and openings therefore have to be the same type.

`OracleSetupChannel` gets away with the dummy because it only does a structural dry run and never reaches that arithmetic. This channel has to, since running the real reduction is what makes the count exact.

I have left a comment at the `type Elem` line so the next reader does not have to redo the experiment.

There is a real follow-up hiding here, though: the Merkle trait hardwires `Vec<F>` while the IP trait has `Elem`. If the Merkle layer returned `Self::Elem` too, a dummy-element channel could run at that layer and the zeroed vectors would cost nothing. That is a wider change than this PR, happy to file it if you think it is worth having.

tcoratger: I checked with the agents locally and looks like it breaks things! I've left a comment in the code to explain for the future, let me know if this is what you had in mind or if i should try something else

PR #2052, opened 2026-08-09
BINIUS-406: run the full table fracaddcheck for the pushforward

benji-potamus:
[BINIUS-406](https://linear.app/jimpo/issue/BINIUS-406/logup-star-run-the-full-fracaddcheck-for-the-pushforward). A follow-up PR removes the API this orphans.

## The gotcha this removes

The table-side fracaddcheck stopped one layer short so its final layer could be batched with the `<T, Y>` product sumcheck — both end in an evaluation of `Y`, so sharing one sumcheck collapsed them into a single point. But a GKR layer works on columns split over the highest variable, so the product claim had to be **split to match**: `<T, Y> = e` became `e = e_0 + e_1` with `e_0` on the transcript, four batched claims over six half-columns `[Y_0, Y_1, D_0, D_1, T_0, T_1]`, and a closing line-fold to put the halves back together.

## What replaces it

Run the fracaddcheck to its leaf. The leaf claims `Y` at a point `z`, and its denominator half is the transparent `D = c - J`, which the verifier now evaluates itself rather than taking from the prover. That leaves two claims that both read `Y`:

```
S_1 = sum_x eq(x; z) * Y(x) = Y(z)      the leaf claim, re-randomized
S_2 = sum_x (Y * T)(x)      = e         the product claim
```

One batched `m`-variable sumcheck over the **unsplit** pair `[Y, T]` reduces both to a single evaluation point. `S_1` carries the `eq` factor so it starts as an MLE-check evaluator and converts to a plain sumcheck to batch with the eq-free `S_2` — which is what `into_shared_sumcheck` was built for.

Nothing splits, nothing folds at the end, and the batched degree drops from **3 to 2**. The transcript loses `e_0` and the line-fold challenge, and carries two column evaluations instead of four:

```
  before                                    after
  5. table GKR, first m-1 layers            5. table GKR, all m layers
  6a. recv e_0                              6a. sample batch_coeff
  6b. sample batch_coeff                    6b. m rounds of degree-2 sumcheck
  6c. m-1 rounds of degree-3 sumcheck       6c. recv [Y, T]
  6d. recv [Y_0, Y_1, T_0, T_1]
  6e. sample r  (line-fold)
```

Both `final_layer.rs` modules are deleted; `pushforward.rs` replaces them on each side. Neither multilinear is copied: `Y` and `T` are pushed as borrowed store columns, so the store's first fold is the only place either is written, at half width.

## Testing

The existing logUp* round-trip tests carry over unchanged and all pass — the shape spread `(6,2) (5,3) (4,4) (3,5) (7,1)`, `m = 1`, `n = 0`, the multi-looker case, and the soundness test that a wrong `eval_claim` is rejected. They compare prover and verifier outputs against independently computed honest evaluations of `T`, `Y`, and the embedded index columns, so they pin the new transcript end to end.

`denominator_halves` becomes `denominator_eval`, and its test with it: the closed form is checked against an explicitly built `D[j] = c - iota(j)` multilinear for `m = 0..=6`.


crates/ip/src/logup_star/pushforward.rs
```diff
@@ -0,0 +1,178 @@
...
+	// Sumcheck binds variables highest-to-lowest; reverse to align with the low-to-high point z.
+	challenges.reverse();
+	let rho = challenges;
+
+	// Reconstruct the batched summand at the challenge point and check it against the reduced eval.
+	//
+	//     re-randomized leaf : eq(rho; z) * Y(rho)
+	//     product            : Y(rho) * T(rho)
+	let eq_eval = eq_ind::<C::Elem>(&rho, pushforward_eval_point);
+	let expected = eq_eval * y_eval.clone() + batch_coeff * y_eval.clone() * t_eval.clone();
```
jimpo: nit: factor out the y_eval

benji-potamus: Done — `y_eval * (eq_eval + batch_coeff * t_eval)`. It reads better as well as saving a clone: both claims read `Y`, so factoring it out is what the batched summand actually looks like. Noted that in the comment above it.

PR #2053, opened 2026-08-09
BINIUS-406: remove the stop-before-the-final-layer batch drivers

benji-potamus:
The follow-up [BINIUS-406](https://linear.app/jimpo/issue/BINIUS-406/logup-star-run-the-full-fracaddcheck-for-the-pushforward) calls for: *"we don't even need `batch_prove_until_final_layer` for either fracaddcheck or prodcheck — these can be removed entirely as a follow-up."*

Stacked on #2052, which removes their only caller and has now merged.

## What goes

`batch_prove_until_final_layer` existed so the logUp* table side could retain its final layer and splice it into the batched final-layer sumcheck. #2052 replaces that reduction, leaving both the fracaddcheck and prodcheck variants without a caller. Each `batch_prove` now runs its own layer loop for all `n_layers` rounds instead of `n_layers - 1` plus a hand-finished last one — the preconditions move up into it, and the two output structs go.

Two more things were only ever there for the same reduction:

- **`layer_prover`'s store column ids.** All three remaining callers already ignored them (`_cols`); it now returns `(remaining, layer_prover)`. `frac_add_mle::new_split_half` loses the ids with it.
- **`prodcheck::into_final_layer`**, which had no caller at all.

## What stays

`batch_prove_unequal_depths` still stops one layer short here. That behaviour has no caller either — it was added with the driver and has only ever been exercised by its own tests — but it is a different idiom, so removing it belongs with the change that makes the driver run every layer rather than with this one. BINIUS-407 does exactly that.

## Testing

Pure removal, no behavior change. The full `binius-ip` / `binius-ip-prover` / `binius-iop` / `binius-iop-prover` suites pass, including the fracaddcheck and prodcheck batch tests that exercise `batch_prove` directly (`test_batch_prove_*`, `test_unequal_depths_*`) and the logUp* round trips. Workspace clippy `-D warnings` is clean, which is what catches the now-dead imports and returns.

### DIFF
--- a/crates/ip-prover/src/fracaddcheck/mod.rs
+++ b/crates/ip-prover/src/fracaddcheck/mod.rs
@@ -24,7 +24,7 @@ use crate::{
 		batch::batch_prove_mle_with_coeff_and_write_evals,
 		common::MleCheckProver,
 		frac_add_mle::{self, FracAddFusedEvaluator},
-		mle_store::{ColId, MleStore},
+		mle_store::MleStore,
 		round_evaluator::{MleCheckRoundEvaluator, SharedMleCheckProver},
 	},
 };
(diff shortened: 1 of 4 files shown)


jimpo (approved):

benji-potamus:
Rebased onto the merged #2052 (`010a63bb` → `72160089`), so the diff is now just this PR's single commit — 4 files, 31+/196−. All checks green. Ready to enqueue whenever you are.

BINIUS-407 is up on top of it as #2055.

benji-potamus:
Rebased onto `40b5add7` (#2054) — `72160089` → `c0097993`, no conflicts, still the same single-commit diff. Re-verified green locally; CI re-running.

PR #2055, opened 2026-08-10
BINIUS-407: combine the looker-side and table-side fracaddchecks

benji-potamus:
Closes [BINIUS-407](https://linear.app/jimpo/issue/BINIUS-407/logup-star-combine-fracaddcheck-for-looker-side-and-table-side), using the unequal-depth batching from BINIUS-403 (#2042, #2043).

#2053 has merged, so this is rebased directly onto `main` and stands alone.

## The problem

logUp* ran *three* fractional-addition GKRs: one batch over the looker circuits (`n` layers), one over the table's (`m` layers), and — because a batch could only hold trees of the same depth — a third "top" circuit of `log_lookers` layers whose only job was to fractionally add the per-looker roots into a single claim. The two surviving roots then went on the transcript so the verifier could cross-multiply them and check the lookup identity itself.

## What changes

Unequal-depth batching removes the equal-depth constraint, so all of it collapses into **one** circuit of `k + max(n, m)` layers, `k = ceil(log2(n_instances))`:

- the lower `max(n, m)` layers run every instance — one per looker over `n` variables, the table's over `m` — as a single batch, with the shallower side padded by zero fractions. Padding costs `O(1)` per round, so the layer count reveals only the deeper side.
- the top `k` layers add those instances' root fractions together.

The table's fraction enters that sum **negated**, so the circuit's root is

```
sum_j num_j / den_j  -  num_r / den_r
```

whose numerator vanishes exactly when the lookup identity holds. So the numerator is never sent: the transcript carries **one scalar**, the root denominator, and the verifier supplies the zero itself. The identity is enforced by the shape of the claim rather than by a separate cross-multiplication — a prover whose lookups disagree cannot make the GKR meet a zero numerator — and `VerificationError::LookupSumMismatch` goes with the check it named. In characteristic 2 the negation is the identity, so the table's halves are used as they stand rather than spending a pass over `2^m` elements.

The `k` selector coordinates now fall out of the GKR's own rounds, so the verifier no longer samples them or interpolates the roots into a batch input claim. The prover runs two provers in sequence: the `k`-layer top circuit, whose output point seeds the batched unequal-depths driver for the rest.

The leaf claims come back on the *padded* witnesses. The prover divides each back onto its own circuit's leaves with `unpad_leaf_claim`; the verifier runs the same map forward through a new `fracaddcheck::pad_leaf_fraction` on the halves it can derive — the lookers' `gamma^j`-scaled equality indicators and the table's transparent `c - J` — then interpolates and checks them against the batch's leaf. That leaves the per-looker index evaluations and `Y` as the only leaf halves on the transcript.

A single-row looker no longer needs its own code path: its circuit has depth 0, which the batch carries as pure padding.

## Commits

1. **Run every layer in the unequal-depths fracadd driver.** `batch_prove_unequal_depths` stopped one layer short so a caller could batch the widest layer with other sumchecks. Nothing did — it was added with the driver and has had no caller outside its own tests. It now returns the plain `BatchProveOutput`, and `BatchProveUnequalDepthsOutput` goes. The precondition relaxes to admit depth-0 trees, which commit 2 needs.
2. **The combination itself**, as described above.

## Testing

- Round trips over the spread `(n, m) = (6,2) (5,3) (4,4) (3,5) (7,1)`, covering padding in both directions and the equal-depth case.
- `test_prove_verify_single_looker_row` now exercises the depth-0 circuit through the batch rather than a special case.
- New `test_unequal_depths_zero_depth_tree` (`&[0, 3]`) pins the driver's depth-0 handling directly.
- Full `binius-ip` / `binius-ip-prover` / `binius-iop` / `binius-iop-prover` suites pass; workspace clippy `-D warnings`, `fmt --check`, and rustdoc `-D warnings` clean.

### DIFF
--- a/crates/ip-prover/src/logup_star/prove.rs
+++ b/crates/ip-prover/src/logup_star/prove.rs
@@ -4,7 +4,7 @@
 
 use binius_compute::Allocator;
 use binius_field::{BinaryField, Divisible, Field, PackedField};
-use binius_ip::{MultilinearEvalClaim, logup_star::LogupOutput};
+use binius_ip::{MultilinearEvalClaim, fracaddcheck::unpad_leaf_claim, logup_star::LogupOutput};
 use binius_math::{FieldBuffer, FieldSlice, FieldVec, univariate::evaluate_univariate};
 use binius_utils::{checked_arithmetics::log2_ceil_usize, rayon::prelude::*};
 
@@ -26,7 +26,10 @@ use crate::{
 /// sharing the table. The lookers batch by a random linear combination: a challenge `gamma`
 /// scales looker `j`'s equality-indicator numerator by `gamma^j`, and the pushforward is the
 /// gamma-weighted sum of the per-looker pushforwards, still with only `2^m` entries. The
-/// looked-up vectors are never committed.
+/// looked-up vectors are never committed. Every fractional-addition circuit — the lookers' over
+/// `n` variables and the table's over `m` — is an instance of one GKR of
+/// `ceil(log2(#lookers + 1)) + max(n, m)` layers, with the shallower side padded by zero
+/// fractions.
 /// See [Soukhanov25] for the construction.
 ///
 /// [Soukhanov25]: <https://eprint.iacr.org/2025/946>
@@ -162,22 +165,23 @@ where
 {
 	let m = table.log_len();
 	let n = lookers[0].eval_point.len();
-	let log_lookers = log2_ceil_usize(lookers.len());
+	// One batch instance per looker, plus the table.
+	let k = log2_ceil_usize(lookers.len() + 1);
 
 	// Sample the logUp challenge c that randomizes the logarithmic-derivative denominators.
 	// This is the prover's first transcript action, mirroring the verifier.
 	// A committing caller must absorb the I, T, and Y commitments into the transcript before this.
 	let c = channel.sample();
 
-	// Build the fractional-addition circuits, one per looker plus the table side.
-	// Constructing a circuit computes every layer and returns its single root fraction.
+	// Build the fractional-addition circuits, one per looker plus the table side. Constructing a
+	// circuit computes every layer and returns its single root fraction.
 	//
 	//     looker j: gamma^j * eq_{r_j}(i) / (c - I_j(i))   over n variables
 	//     table:    Y(v)                  / (c - v)         over m variables
 	let circuits_guard = tracing::debug_span!("Build fracadd circuits").entered();
 	// The circuits are independent, so they build in parallel across lookers.
 	// The collect preserves looker order, so circuit j keeps numerator gamma^j.
-	let (looker_provers, looker_roots): (Vec<_>, Vec<_>) = (lookers, numerators)
+	let (mut provers, mut roots): (Vec<_>, Vec<_>) = (lookers, numerators)
 		.into_par_iter()
 		.map(|(looker, numerator)| {
 			let den = witness::looker_denominator::<A, F, P>(alloc, c, looker.index);
@@ -185,6 +189,8 @@ where
 			(prover, (root.0.get(0), root.1.get(0)))
 		})
 		.unzip();
+	// The table's fraction enters the sum of every instance negated, which is what makes that sum's
+	// numerator vanish. The negation rides on the denominator, which is built here anyway.
 	let table_den = witness::table_denominator::<A, F, P>(alloc, c, m);
 	// The pushforward is borrowed — a committing caller keeps it for the oracle opening — so the
 	// table circuit's leaf layer, which it folds in place, is a clone drawn from `alloc`.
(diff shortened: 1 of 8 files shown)


crates/ip/src/logup_star/verify.rs
```diff
@@ -109,100 +114,111 @@ where
...
 
-	// Read the root fractions of both fractional-addition GKR circuits.
-	//
-	//     looker side: num_L / den_L = sum_j gamma^j sum_{i in B_n} eq_{r_j}(i) / (c - I_j(i))
-	//     table  side: num_R / den_R = sum_{v in B_m} Y(v) / (c - v)
-	let [num_l, den_l, num_r, den_r] = channel
-		.recv_array()
+	// Read every circuit's root fraction: one per looker, then the table's.
+	let n_instances = lookers.len() + 1;
+	let roots = channel
```
jimpo: No, you don't need to read every root fraction. You'll run one fracaddcheck of depth `log_n_instances + max(n, m)` (log_n_instances is ceil log n_instances). The root must have a 0 numerator. So you only read one message, which is the denominator.

This reduces to claims on the looker columns and the negation of the table column.

One the prover side, do the sequence of two fracaddcheck provers, one for the single `log_n_instances` depth one, then the batched one, which is similar to how the code works now on the looker side.

jimpo (changes requested):

crates/ip/src/logup_star/verify.rs
benji-potamus: Done — one fracaddcheck of depth `k + max(n, m)`, `k = ceil(log2(n_instances))`.

The table circuit enters the top circuit's leaves negated, so the root is `sum_j num_j/den_j - num_r/den_r` and its numerator is zero. The transcript carries one scalar, the root denominator; the verifier supplies the zero itself and there is no separate identity check. `LookupSumMismatch` goes with it — a violated identity now surfaces as a GKR layer inconsistency.

On the prover, two provers in sequence as you described: `top_prover.prove(root_claim(F::ZERO, root_den), channel)` for the `k` layers, whose output point seeds `batch_prove_unequal_depths` for the remaining `max(n, m)`.

In characteristic 2 the negation is the identity, so the table's root and its leaf numerator claim are used as they stand — commented at both places rather than spending a pass over `2^m` elements to negate.

This also drops the verifier's explicit selector sampling and the root interpolation: those `k` coordinates now come out of the GKR's own rounds.

crates/ip/src/fracaddcheck.rs
```diff
@@ -83,6 +83,26 @@ where
...
+/// its denominator through the one-padding selector $\textsf{sel}(q, v) = 1 + (v - 1) q$. A
+/// verifier that rebuilds a padded batch's leaf claim from transparent parts applies this to each
+/// tree, so the two directions live together.
+///
+/// # Arguments
+///
+/// * `fraction` - The unpadded leaf's numerator and denominator.
+/// * `padding_point` - The padding coordinates, which are the lowest ones of the leaf point.
+pub fn pad_leaf_fraction<E: FieldOps>(fraction: (E, E), padding_point: &[E]) -> (E, E) {
+	let pad_eq = padding_point
```
jimpo: Use eq_ind_zero

crates/ip/src/logup_star/verify.rs
```diff
@@ -109,100 +116,88 @@ where
...
+		iter::zip(iter::zip(lookers, powers(gamma.clone())), &index_eval_claims)
+			.map(|((looker, power), index_eval)| {
+				let num = power * eq_ind::<C::Elem>(looker.eval_point, looker_content);
+				let den = c.clone() - index_eval.clone();
+				fracaddcheck::pad_leaf_fraction((num, den), &node_point[..looker_pad])
+			})
+			.unzip();
+
+	// The table's numerator is Y, just read, entering the sum negated so the root vanishes; in
+	// characteristic 2 that negation is the identity. Its denominator is the transparent c - J.
```
jimpo: Do the actual negation, don't depend on this being characteristic 2.

crates/ip/src/logup_star/verify.rs
```diff
@@ -109,100 +116,88 @@ where
...
+	let table_point = &node_point[table_pad..];
+
+	// Looker numerators are transparent: gamma^j scales the equality indicator at r_j. The
+	// denominators are c - I_j, from the index evaluations just read.
+	let (mut leaf_nums, mut leaf_dens): (Vec<_>, Vec<_>) =
+		iter::zip(iter::zip(lookers, powers(gamma.clone())), &index_eval_claims)
+			.map(|((looker, power), index_eval)| {
+				let num = power * eq_ind::<C::Elem>(looker.eval_point, looker_content);
+				let den = c.clone() - index_eval.clone();
+				fracaddcheck::pad_leaf_fraction((num, den), &node_point[..looker_pad])
```
jimpo: compute the sequence of prefix eq_ind_zero pads outside the loop do reduce duplicate computation

jimpo (changes requested):

crates/ip/src/logup_star/verify.rs
benji-potamus: Done, on both sides — relying on the characteristic was the wrong call.

Prover: the table circuit's numerator column is negated. The pushforward is borrowed, so that clone was already being made for the leaf layer; the negation rides along on it. Its root and leaf numerator then come out negated on their own, which is what the batched GKR needs — negating only at the top circuit's leaves would have left `roots` disagreeing with what the batch folds the table instance down to.

Verifier: `-pushforward_eval` when rebuilding the table's leaf fraction.

The transcript still carries `Y(z)` rather than `-Y(z)`: the prover negates the leaf claim back before sending, so both it and the pushforward reduction read the natural quantity and the negation stays inside the fracadd circuit.

crates/ip/src/fracaddcheck.rs
benji-potamus: Done. `pad_leaf_fraction` now takes the weight rather than the padding coordinates, so the caller computes it with `eq_ind_zero` — which also makes the hoisting in your other comment possible. That matches `unpad_claims(pad_eq_inv, ..)` on the prover side.

crates/ip/src/logup_star/verify.rs
benji-potamus: Hoisted — `pad_leaf_fraction` takes the weight now, so the loop just uses it.

One note on shape: there are only two distinct prefixes to compute here, not a sequence. Every looker has the same `n` (asserted above), so they all share `max(n, m) - n`, and the table has `max(n, m) - m`; one of the two is always the empty prefix, since only the shallower side is padded. So it is two `eq_ind_zero` calls above the loop rather than a prefix table:

```rust
let looker_pad_eq = eq_ind_zero::<C::Elem>(&node_point[..looker_pad]);
let table_pad_eq = eq_ind_zero::<C::Elem>(&node_point[..table_pad]);
```

Happy to build the running-prefix version instead if you'd rather have the general shape here — it would matter if lookers could differ in depth.

crates/ip-prover/src/logup_star/prove.rs
```diff
@@ -187,86 +191,95 @@ where
...
-	root_nums.resize(1 << log_lookers, F::ZERO);
-	root_dens.resize(1 << log_lookers, F::ONE);
-	let (top_prover, top_root) = FracAddCheckProver::new(
-		log_lookers,
-		alloc,
+	// table circuit's leaf layer, which it folds in place, is a clone drawn from `alloc`. The
+	// table's fraction enters the sum of every instance negated, which is what makes that sum's
+	// numerator vanish, so the clone is negated as it is taken.
+	let mut table_num = FieldBuffer::clone_from_slice(alloc, pushforward.to_ref());
+	table_num.as_mut().par_iter_mut().for_each(|packed| {
```
jimpo: Ah, I don't like this copy. Handle the negation like this instead: negate the table denominator instead of the numerator. On both the prover and verifier side. This will become part of the `table_denominator` computation itself and so is effectively free.

crates/ip-prover/src/logup_star/prove.rs
```diff
@@ -187,86 +191,95 @@ where
...
+
+	// Top circuit: interpolate every instance's root fraction into a multilinear pair over the k
+	// selector variables, padded with the zero fraction. Its own root is the fractional sum of all
+	// of them, and the table's numerator is already negated, so that sum is
+	//
+	//     sum_j num_j / den_j  -  num_r / den_r
+	//
+	// which is zero exactly when the logUp identity holds.
+	let (top_prover, (top_root_num, top_root_den)) = FracAddCheckProver::new(k, alloc, {
+		let mut root_nums = roots.iter().map(|&(num, _)| num).collect::<Vec<_>>();
```
jimpo: Use unzip instead of two collects

jimpo (changes requested):

crates/ip-prover/src/logup_star/prove.rs
benji-potamus: Much better — the copy is gone. `table_denominator` now builds `J - c` directly, so the negation costs nothing, and the pushforward clone goes back to a plain `clone_from_slice`.

Verifier side follows: `denominator_eval` returns `J - c`, and the table's leaf numerator is `Y` again rather than `-Y`. That also means the transcript needs no negate-and-undo — `table_leaf.num_eval` is `Y(z)` on its own.

Also updated the `c - J` wording in the two module docs, `witness`'s header, and the `IncorrectTableDenominator` message so they name what is actually built.

crates/ip-prover/src/logup_star/prove.rs
benji-potamus: Done: `let (mut root_nums, mut root_dens): (Vec<_>, Vec<_>) = roots.iter().copied().unzip();`

benji-potamus:
Rebased onto `fb85dfbc` (#2056) along with the review fixes — `ee3337ae` → `ffcc05dd`. Both commits re-verified green (tests across `binius-ip`/`ip-prover`/`iop`/`iop-prover`, workspace clippy `-D warnings`, rustdoc `-D warnings`, `fmt --check`).

benji-potamus:
Rebased onto `f185d2b4` (#2051) — `ffcc05dd` → `31fcdb39`, no conflicts.

Since #2051 added `proof_size_exactness_test.rs` and this PR shortens the logUp* transcript, I ran the spartan crates too rather than assume they were unaffected: `size_tracking_matches_real_proof_bytes` passes, so the size-tracking model follows the change. Both commits otherwise green — ip / ip-prover / iop / iop-prover / spartan-prover / spartan-verifier tests, workspace clippy `-D warnings`, rustdoc `-D warnings`, `fmt --check`.

jimpo (approved):

PR #2056, opened 2026-08-10
BINIUS-404: store the value vector without segment padding

benji-potamus:
Follows #2050, completing [BINIUS-404](https://linear.app/jimpo/issue/BINIUS-404/switch-valueindex-to-segment-aware-witnessindex).

**This is a prover/verifier change, not a storage cleanup.** I originally scoped it as the latter and was wrong twice; the tests corrected me both times. It touches the shift reduction's phase plumbing and breaks the `Verifier::verify` signature. Framing it accordingly so it gets reviewed as what it is.

## What changes

`ValueVec` stores exactly `n_const + n_inout + n_private` words — no padding anywhere. `ValueVecLayout` is down to five counts (`n_const`, `n_inout`, `n_witness`, `n_internal`, `n_scratch`), with `offset_inout()`, `offset_witness()` and `n_private()` as derived accessors:

```text
[ constants | inout ][ witness | internal ][ scratch ]
 \-- public values -/ \--- private values -/
```

The padding the protocol needs is now materialized only where the protocol needs it.

## Where the padding moved

- **`shift::prove` fills each segment once.** Both phases previously did `words.split_at(key_collection.public.n_words())` on the combined witness. That split is the reason this isn't a pure storage change: with a dense vector it reads private words as public. `prove` now takes the two declared segments, pads each to its committed width via `pad_to`, and hands both phases the result. `pad_to` **borrows** when the segment already fills its width — the usual case for the hidden segment, since private values normally outnumber public words — so the common path allocates only for the small public segment.
- **`SegmentWords`** carries the pair. Splitting the parameter pushed `prove_phase_2` past clippy's 7-argument limit, and a pair of slices threaded through three functions wanted a name regardless.
- **`Verifier::verify(public)` takes the dense `n_public_values`** rather than `1 << log_public_words`. `encode_public` pads internally, so the Fiat-Shamir transcript still matches what the prover packs. **This is a caller-visible break** — anything passing a pre-padded public vector must stop.
- **`ValueTable` (M4)** keeps its rows at the committed width with the private values at the front and the protocol's zero padding as the tail, so commitment sizes are byte-for-byte unchanged. Its `n_hidden_words()` now comes from `cs.n_hidden_words()` rather than the layout.

## Why commitment sizes don't change

Because `n_public_words` is a power of two:

```
log2_ceil(max(n_private, n_public_words)) == max(log2_ceil(n_private), log_public_words)
```

So the `max` in `n_hidden_words()` was never storage padding — it is the shift reduction's **word-axis width**, load-bearing because `r_y` has `log_witness_words` entries and `check_eval` slices `r_y[log_public_words..]`. Drop it and that slice goes out of bounds. Keeping it as a derived protocol quantity lets the storage go dense for free.

## Call sites the tests caught

Worth listing, because none were obvious from the type errors alone — each surfaced only at runtime:

- the ZK wrapper's symbolic verify built a padded dummy public vector and then failed its own length check;
- `tests/shift.rs` passed `combined_witness()` to `shift::prove`;
- `benches/shift_reduction.rs` did its own `split_at` in two places, plus two `prove` call sites.

The two M4 reference-fold tests gained `assert_eq!(..., "fixture must have no padding")` so those fixtures can't silently start depending on padding they don't exercise.

## Verification

With `RUSTFLAGS="-C target-cpu=native"`, on the current main (`fb45c50f`):

| check | result |
|---|---|
| `binius-core` | ✅ 98 passed |
| `binius-prover` | ✅ 37 + 5 + 13 passed (incl. all `prove_verify`, ZK and signature-of-knowledge) |
| `binius-circuits` | ✅ 332 passed |
| `binius-examples` | ✅ 22 + 3 passed |

### DIFF
--- a/crates/prover/benches/shift_reduction.rs
+++ b/crates/prover/benches/shift_reduction.rs
@@ -149,7 +149,8 @@ fn bench_prove_and_verify(c: &mut Criterion) {
 
 				prove::<F, P, _, _>(
 					&key_collection,
-					value_vec.combined_witness(),
+					value_vec.public(),
+					value_vec.non_public(),
 					OperatorClaims {
 						zero: prover_zero_data,
 						bitand: prover_bitand_data,
@@ -184,7 +185,8 @@ fn bench_prove_and_verify(c: &mut Criterion) {
 
 		prove::<F, P, _, _>(
 			&key_collection,
-			value_vec.combined_witness(),
+			value_vec.public(),
+			value_vec.non_public(),
 			OperatorClaims {
 				zero: prover_zero_data,
 				bitand: prover_bitand_data,
@@ -256,7 +258,10 @@ fn bench_shift_phases(c: &mut Criterion) {
 	let intmul_evals = [F::ZERO; 4];
 
 	let key_collection = build_key_collection(&cs);
-	let words = value_vec.combined_witness();
+	// The phase functions take each segment as the circuit declares it: `build_g_parts` zips the
+	// words with their key ranges and `fold_words` pads each fold to `log2_ceil(len)` variables.
+	let public_words = value_vec.public();
+	let hidden_words = value_vec.non_public();
 	let subspace = BinarySubspace::<AESTowerField8b>::with_dim(Word::LOG_BITS).isomorphic();
 
 	// Prepare the operator data. Lambda sampling is cheap and not part of any benched phase, so a
@@ -291,7 +296,6 @@ fn bench_shift_phases(c: &mut Criterion) {
 	// `build_g_parts` runs per key segment; the full g parts are the sum of the public and hidden
 	// segment parts.
 	let build_combined_g_parts = || {
-		let (public_words, hidden_words) = words.split_at(key_collection.public.n_words());
 		let mut g_parts = build_g_parts::<F, P, _>(
 			&GlobalAllocator,
 			public_words,
@@ -331,7 +335,6 @@ fn bench_shift_phases(c: &mut Criterion) {
 	let r_s = r_jr_s.split_off(Word::LOG_BITS);
 	let r_j = r_jr_s;
 	let r_j_tensor = eq_ind_partial_eval::<F>(&r_j);
-	let (public_words, hidden_words) = words.split_at(key_collection.public.n_words());
 	let public_folded = fold_words::<F, P, _>(&GlobalAllocator, public_words, r_j_tensor.as_ref());
 	let hidden_folded = fold_words::<F, P, _>(&GlobalAllocator, hidden_words, r_j_tensor.as_ref());
 	let (public_monster, hidden_monster) = build_monster_segments::<F, P, _>(
--- a/crates/prover/src/protocols/shift/prove.rs
+++ b/crates/prover/src/protocols/shift/prove.rs
@@ -1,7 +1,9 @@
 // Copyright 2025 Irreducible Inc.
 
+use std::borrow::Cow;
+
 use binius_compute::Allocator;
-use binius_core::word::Word;
+use binius_core::{ConstraintSystem, word::Word};
 use binius_field::{BinaryField, Field, PackedField, util::powers};
 use binius_ip::sumcheck::SumcheckOutput;
 use binius_ip_prover::channel::IPProverChannel;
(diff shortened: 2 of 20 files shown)


crates/prover/src/protocols/shift/prove.rs
```diff
@@ -155,6 +178,15 @@ where
 	Channel: IPProverChannel<F>,
 	A: Allocator,
 {
+	// Each segment is addressed at the width the protocol commits it at, which runs past the
+	// values the circuit declares. The words in between are the protocol's zero padding.
+	let public_words = pad_to(public_words, key_collection.public.n_words());
```
jimpo: Get rid of this padding, it shouldn't be necessary. The shift reduction prover should be able to handle non-power-of-two-sized segments. Make it so if not already.

jimpo (changes requested):

crates/prover/src/protocols/shift/prove.rs
benji-potamus: You were right, and it was already able to almost everywhere. I removed the padding and let the tests say what actually broke rather than reasoning about it.

**Nothing needed for non-power-of-two segments:**

- **Phase 1** already handles a short segment: `build_g_parts` *zips* the words with `segment.key_ranges`, so it stops at the last declared value. The words past it carry no keys and contribute nothing, so the zip ending early is exactly right.
- **Phase 2** already handles it too: `fold_words` zero-pads each fold up to `log2_ceil(len)` variables. And `log2_ceil(n_public_values) == log_public_words` for any `n_public_values >= 2`, so the fold lands on the same width the monster is built at without any help.

With the padding gone, 10 of the 13 `prove_verify` tests passed. The 3 that failed did so at `assert_eq!(public_monster.log_len(), public_folded.log_len())` with `left: 1, right: 0` — circuits whose public segment holds exactly **one** value, the all-one constant, against `MIN_WORDS_PER_SEGMENT = 2`.

So the one length the folds cannot infer is the minimum segment size, which `ConstraintSystem` applies to the public segment and the monster multilinear is therefore built at. That is now the only padding left:

```rust
let public_words = pad_to_min(public_words, ConstraintSystem::MIN_WORDS_PER_SEGMENT);
```

At most two words, borrowed whenever the segment already reaches it — so every real circuit takes the borrow. The hidden segment is no longer padded at all.

If you would rather that floor lived somewhere else, the alternative is to stop building the public monster at the minimum width — i.e. have `build_key_collection` size the public key segment by `n_public_values` instead of `n_public_words`. That is a bigger change and reaches the key collection's serialized form, so I left it; say the word if you would prefer it there.

Rebased onto `be66218c` and re-verified: prover 37 + 13 + 1, core 98, frontend 197 + 2, verifier 43 + 13, m4-prover 13, circuits 332, examples 22 + 3, workspace clippy `-D warnings`, rustdoc and fmt all clean.

crates/prover/benches/shift_reduction.rs
```diff
@@ -256,7 +258,15 @@ fn bench_shift_phases(c: &mut Criterion) {
 	let intmul_evals = [F::ZERO; 4];
 
 	let key_collection = build_key_collection(&cs);
-	let words = value_vec.combined_witness();
+	// The phase functions address each segment at the width the protocol commits it at, which runs
+	// past the values the circuit declares; the words in between are zero.
+	let pad_to = |words: &[Word], n_words: usize| {
```
jimpo: What's this for? Not necessary?

benji-potamus: Not necessary — removed. It was left over from when the prover still padded, and I didn't revisit the bench after taking that out.

The bench now hands `value_vec.public()` and `value_vec.non_public()` straight to `build_g_parts` and `fold_words`, same as the real prover does.

Confirmed by running it rather than just compiling it — `cargo bench --bench shift_reduction -- --test` passes every case, including `phase2_run_sumcheck`, which is where the segment-width assert lives:

```
Testing shift_reduction_phases/phase1_build_g_parts     Success
Testing shift_reduction_phases/phase1_build_h_parts     Success
Testing shift_reduction_phases/phase1_run_sumcheck      Success
Testing shift_reduction_phases/phase2_build_monster_segments  Success
Testing shift_reduction_phases/phase2_run_sumcheck      Success
```

(Clippy on `--benches` passes too, but that only type-checks, and this exact assert is what caught the padding removal in the prover — so it seemed worth exercising at runtime.)

jimpo (approved):

PR #2059, opened 2026-08-10
Count Merkle salts in the proof size, and clean up the scheme

tcoratger:
@jimpo I've mainly added a bunch of tests here because I've noticed that on the verifier side we didn't have that much unit tests.

Fixes a proof-size estimate that under-reported for hiding commitments, and cleans up the verifier-side Merkle scheme around it.

No change to what is verified or accepted — the transcript is byte-identical.

### The problem

The prover writes a salt before every branch it opens:

```
prove_opening:  [salt] [sibling_0] [sibling_1] ... [sibling_k]
```

The scheme's proof-size estimate counted only the digests. So for any non-zero salt length it under-reported by

```
n_queries * salt_len * BYTE_SIZE
```

The FRI estimate papered over this with a comment saying it *"assumes non-hiding proofs (salt_len = 0)"*. But the scheme is the one object that knows the salt length, and the fixed serialized width of the committed values hands it the byte size for free — so the bound that made this computable was already there, just unused.

### The idea

Count the salt where it is known.

The estimate splits into the digests the branches and layer cost, plus the salt every opened leaf reveals:

```
n_digests  = (log_len - layer_depth) * n_queries + 2^layer_depth
salt_bytes = n_queries * salt_len * BYTE_SIZE
```

That alone is not enough to make FRI exact. The terminal codeword is never queried, but it is decommitted in full, which reveals a salt for each of its leaves. So the scheme also gained a companion estimate for a full-vector opening, and FRI now adds it:

```
vector_proof_size(len) = len * salt_len * BYTE_SIZE
```

The optimal layer depth is unaffected: the salt term does not depend on it, so the argmin does not move.

### The change

```
proof_size:
- ((log_len - layer_depth) * n_queries + (1 << layer_depth)) * output_size()
+ n_digests * output_size() + n_queries * self.salt_len * T::BYTE_SIZE

fri::size_estimation:
+ merkle_sizes += vcs.vector_proof_size(1 << log_inv_rate);
- /// The estimate assumes non-hiding proofs (salt_len = 0).
```

`test_proof_size_hiding` pins it: a full FRI run at `salt_len = 3` now matches the predicted byte count exactly.

### Cleanup carried in the same pass

The file needed it, and the salt fix touches most of the same lines.

**A free function that should have been a method.** Layer folding took a compression function and a `Vec` through two generic parameters that were always the same two types, behind bounds — `Default + Send + Sync + Debug` — of which only `Clone` was used. It is now a private method. One caller was cloning an entire layer purely to satisfy the `Vec` signature; taking a slice removes that, and the scratch buffer is now `n/2` rather than `n`, allocated once and halved in place.

**Two allocations off the per-query path.** Branch siblings are read one at a time instead of collected into a `Vec`, and leaf hashing serializes straight into the hasher rather than staging the salt in a `Vec` first. Hashing dominates, so this is an allocation-count win, not a measurable speedup — worth it for the code.

**Preconditions where they belong.** Full-vector verification had three, of which one was documented. A zero batch size passed the multiple-of check on empty data and then panicked inside `chunks`; a non-power-of-two leaf count panicked inside the folding helper. Both are now asserted and documented on the trait, as is `layer_depth <= tree_depth`, whose subtraction could underflow.

**API tidying.**

- The compression getter had no callers anywhere in the workspace — removed.
- The derive-macro getters are written out, so the surviving one is `const` and carries a doc comment.
- `Clone` and `Debug` are hand-written: a derive would demand `T: Clone` / `T: Debug` on a type that holds no `T`.
- The hiding constructor takes a `NonZeroUsize`, so it can no longer name a security property while silently not providing it.
- Serialization failures get `#[from]`, matching every sibling variant.

**One thing deliberately not done.** Making `batch_size` a `NonZeroUsize` was tempting, but `leaf_size` is a plain `usize` throughout FRI and BaseFold, so the type would only force a `NonZeroUsize::new(..).expect(..)` at the single call site — the same assert, relocated somewhere it reads worse — while breaking the module's uniform documented-precondition convention.

### Tests

The scheme had **no tests in its own crate**; all coverage was prover-side round trips, and every one was a happy path. For a verifier that is backwards — nothing pinned that a bad proof is rejected.

It now has twelve tests and a proptest, built against a reference tree constructed straight from the hash and compression primitives. Checking the verifier against trees it built itself would pass even with a wrong folding rule.

- **acceptance:** every leaf at every layer depth, across batched and salted trees; single-leaf trees, where nothing folds.
- **rejection, one per tamper site:** branch digest (at each level in turn), leaf values, leaf salt, layer entry, vector value, vector salt — each pinned to the full error variant.
- **size:** predicted bytes equal written bytes for both opening kinds, salted and unsalted, with the bytes replayed through the verifier so a count cannot pass on a proof that would not.
- **proptest:** the chosen layer depth really minimizes the proof, compared by size since two depths can tie.

### Documentation

The type had no doc comment at all, and neither did its constructors, its fields, or the trait's digest type. The impl-level docs on two methods were *shadowing* the trait's own docs in rustdoc, and were worse than what they replaced.

The file is now documented to the repo's standard throughout: sectioned headers, one sentence per line, ASCII diagrams for the tree shape and the climb, and line-by-line narration in every body including the tests. The trait now carries the cost model behind the optimal layer depth, and the reason the committed values need a fixed serialized width — leaf encodings are injective, so two distinct leaves can never present the same bytes to the hash.

### Scope

- **changed:** the Merkle scheme, its trait, its error type, the FRI size estimate
- **new:** a test module in the scheme, one hiding FRI proof-size test
- **rippled:** hiding constructors take `NonZeroUsize` — 3 call sites plus tests
- **untouched:** the prover, the tree builder, the channel protocol

Also folded in: the FRI estimate was serializing a default field element into a heap `Vec` to learn its width, which the existing trait bound already provides as a constant.

### Verification

- `cargo clippy -p binius-iop -p binius-iop-prover --all-targets` — zero warnings
- `cargo test -p binius-iop -p binius-iop-prover` — 62 passed
- `cargo test` on the verifier, prover, and both Spartan crates — 82 passed

### DIFF
--- a/crates/iop-prover/src/fri/tests.rs
+++ b/crates/iop-prover/src/fri/tests.rs
@@ -1,7 +1,7 @@
 // Copyright 2024-2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::{iter, vec};
+use std::{iter, num::NonZeroUsize, vec};
 
 use binius_field::{
 	BinaryField, BinaryField128bGhash as B128, Field, PackedBinaryGhash1x128b, PackedField,
@@ -654,11 +654,14 @@ fn test_commit_prove_verify_lifted_multi_oracle() {
 
 /// Runs the FRI prover and returns the proof bytes along with the FRI params and Merkle scheme,
 /// so the caller can check the estimated proof size.
+///
+/// A non-zero `salt_len` runs the whole protocol with a hiding Merkle scheme.
 fn generate_fri_proof<F, P>(
 	log_dimension: usize,
 	log_inv_rate: usize,
 	log_batch_size: usize,
 	arities: &[usize],
+	salt_len: usize,
 ) -> (Vec<u8>, FRIParams<F>, binius_iop::merkle_tree::BinaryMerkleTreeScheme<F, StdHashSuite>)
 where
 	F: BinaryField,
@@ -680,11 +683,20 @@ where
 
 	let codeword = encode_interleaved(&params, 0, &ntt, msg.to_ref());
 
+	let salt_len = NonZeroUsize::new(salt_len);
 	let mut prover_transcript = ProverTranscript::new(StdChallenger::default());
-	let mut prover_channel =
-		ProverMerkleTranscriptChannel::<_, StdChallenger, _, StdHashSuite>::new(
+	let mut prover_channel = match salt_len {
+		Some(salt_len) => {
+			ProverMerkleTranscriptChannel::<_, StdChallenger, _, StdHashSuite>::hiding(
+				&mut prover_transcript,
+				&mut rng,
+				salt_len,
+			)
+		}
+		None => ProverMerkleTranscriptChannel::<_, StdChallenger, _, StdHashSuite>::new(
 			&mut prover_transcript,
-		);
+		),
+	};
 	let codeword_commitment =
 		prover_channel.send_merkle_commitment(codeword.to_ref(), 1 << log_batch_size);
 
@@ -701,7 +713,10 @@ where
 	round_prover.finish_proof(&mut prover_channel);
 	drop(prover_channel);
 
-	let scheme = binius_iop::merkle_tree::BinaryMerkleTreeScheme::new();
+	let scheme = match salt_len {
+		Some(salt_len) => binius_iop::merkle_tree::BinaryMerkleTreeScheme::hiding(salt_len),
+		None => binius_iop::merkle_tree::BinaryMerkleTreeScheme::new(),
+	};
 	let proof_bytes = prover_transcript.finalize();
 	(proof_bytes, params, scheme)
 }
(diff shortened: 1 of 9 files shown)


crates/iop-prover/src/fri/tests.rs
```diff
@@ -654,11 +654,14 @@ fn test_commit_prove_verify_lifted_multi_oracle() {
...
 /// Runs the FRI prover and returns the proof bytes along with the FRI params and Merkle scheme,
 /// so the caller can check the estimated proof size.
+///
+/// A non-zero `salt_len` runs the whole protocol with a hiding Merkle scheme.
 fn generate_fri_proof<F, P>(
 	log_dimension: usize,
 	log_inv_rate: usize,
 	log_batch_size: usize,
 	arities: &[usize],
+	salt_len: usize,
```
jimpo, line 664: We actually never use the salted Merkle tree. We could even remove salting altogether.

I haven't written it up, but the ZK construction relies on the fact that if there's one extra degree of freedom in the witness blinding, then that's sufficient to ensure zero-knowledge-ness of the BaseFold reduction.

crates/iop/src/fri/size_estimation.rs
```diff
@@ -23,13 +21,7 @@ where
...
 
 	// Serialized byte-size of a single field element.
-	let value_size = {
-		let mut buf = Vec::new();
-		F::default()
-			.serialize(&mut buf)
-			.expect("default element can be serialized to a resizable buffer");
-		buf.len()
-	};
+	let value_size = F::BYTE_SIZE;
```
jimpo, line 24: this files's changes are important. Maybe break this out into a PR separate from the hiding Merkle tree?

crates/iop-prover/src/fri/tests.rs
tcoratger: Just opened a PR: https://github.com/binius-zk/binius64/pull/2065

jimpo:
What do you want to do about this now that salts are gone?

tcoratger:
> What do you want to do about this now that salts are gone?

I'll handle via follow-ups progressively to add the tests mainly, let me close this big one for now to avoid confusion.

PR #2060, opened 2026-08-10
BINIUS-404: take only the inout words in verify

benji-potamus:
Follows #2056. Part of [BINIUS-404](https://linear.app/jimpo/issue/BINIUS-404/switch-valueindex-to-segment-aware-witnessindex).

## What changes

`verify` takes the inout words rather than the whole public segment:

```rust
pub fn verify<Channel>(&self, inout: &[Word], channel: &mut Channel) -> Result<(), Error>
```

It checks `inout.len() == cs.n_inout`, observes the encoded inout words in the channel, then concatenates `cs.constants` with them to get the public words the shift reduction reads — constants first, the order the value vector places them in.

A caller no longer restates the constants, which the constraint system already fixes. A caller that restated them *wrongly* was describing a different system than the one it was verifying against, and nothing checked that.

## Two things a reviewer should know

**The Fiat-Shamir transcript changed.** Both sides used to observe the whole padded public segment; now they observe only the inout words. So the prover moved in lockstep — it observes `encode_inout(witness.inout())` where it previously packed `witness.public()` to the padded public width.

To make the two impossible to drift apart, the encoding is a single `pub fn encode_inout` in `binius-verifier` that both sides call. It pads an odd word count with one zero word — enough to pair words into `B128` elements, with none of the power-of-two padding the old path carried.

**None of the caller breakage was compiler-visible.** `verify` still takes `&[Word]`, so every site passing `witness.public()` type-checked and would only have failed at runtime with a length mismatch. I found them by grep rather than by build errors:

- `tests/prove_verify.rs` (4 sites), `examples/src/lib.rs` (2, including `verify_sig`)
- `examples/benches/utils/runner.rs` (2)
- the end-to-end tutorial
- the **ZK prover's replay channel**, which drives the inner verifier and so must feed it exactly what it now expects
- the ZK verifier's symbolic-build dummy vector
- `bin/verifier.rs` — see below

## The one judgement call

The on-disk `ValuesData` still holds the **full** public segment, because the prover rebuilds its value vector from it via `value_vec_from_data`. So `bin/verifier.rs` slices the constants off at the call site rather than changing the file format. Happy to change the format instead if you'd rather the two sides agree on "public data means inout data".

## Fallout worth noting

`IOPProver::log_public_words` became dead code — clippy caught it. The prover no longer reasons about the padded public segment at all now, only the inout values. Removed.

Also adds `ValueVec::inout()` beside `public()`.

## Verification

With `RUSTFLAGS="-C target-cpu=native"`, on main at `fb85dfbc`:

| check | result |
|---|---|
| `binius-core` | ✅ 98 passed |
| `binius-prover` | ✅ 37 + 13 + 1 passed (incl. `prove_verify`, ZK, signature-of-knowledge) |
| `binius-circuits` | ✅ 332 passed |
| `binius-examples` | ✅ 22 + 3 passed |

The ZK and signature-of-knowledge tests are the ones that matter most here: they exercise the replay path, where a prover/verifier transcript mismatch shows up as a verification failure rather than a compile error.

### DIFF
--- a/crates/verifier/src/verify.rs
+++ b/crates/verifier/src/verify.rs
@@ -103,10 +103,10 @@ impl IOPVerifier {
 	/// `verify`, rather than duplicating it here.
 	pub fn oracle_specs(&self, is_zk: bool) -> Vec<OracleSpec> {
 		let mut channel = OracleSetupChannel::new(is_zk);
-		let public = vec![Word::ZERO; self.constraint_system.n_public_values()];
+		let inout = vec![Word::ZERO; self.constraint_system.n_inout];
 		// The result is discarded: the setup channel performs no real verification (all `recv_*`
 		// return zero, `assert_zero` is a no-op), so we only read back the recorded oracle specs.
-		let _ = self.verify(&public, &mut channel);
+		let _ = self.verify(&inout, &mut channel);
 		channel.into_oracle_specs()
 	}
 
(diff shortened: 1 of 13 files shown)


crates/verifier/src/verify.rs
```diff
@@ -537,14 +542,15 @@ where
...
-
-	let mut padded = public.to_vec();
-	padded.resize(n_public_words, Word::ZERO);
+/// Encodes the inout words as the field elements both sides observe in Fiat-Shamir.
+///
+/// Two words share one element, so an odd count is padded with a zero word. The prover and the
+/// verifier both go through here, which is what keeps their transcripts identical.
+pub fn encode_inout(inout: &[Word]) -> Vec<B128> {
+	let mut padded = inout.to_vec();
+	if !padded.len().is_multiple_of(2) {
```
jimpo: Don't actually make a padded copy, just inspect the remainder of as_chunks and conditionally push

jimpo (commented):
Change how ValuesData is handled -- serialize only the inout values. The prover and verifier should both have access to the constants through the constraint system.

crates/verifier/src/verify.rs
benji-potamus: Done — no copy now:

```rust
let (pairs, remainder) = inout.as_chunks::<2>();
let mut elems = pairs
    .iter()
    .map(|[w0, w1]| B128::new(((w1.as_u64() as u128) << 64) | w0.as_u64() as u128))
    .collect::<Vec<_>>();

// An odd word count leaves one word over, which takes the low half of a final element.
if let [w0] = remainder {
    elems.push(B128::new(w0.as_u64() as u128));
}
```

The odd word lands in the low half with a zero high half, which is the same element the padded version produced, so the transcript is unchanged.

Also fixed a stray line I left above it — the old `encode_public` doc comment's first line was still sitting on top of the new one.

benji-potamus:
Done — `ValuesData` now carries only the inout values.

- `cli.rs` serializes `witness.inout()` rather than `witness.public()`.
- `ConstraintSystem::value_vec_from_data(inout, private)` sources the constants from the system itself, mirroring what `verify` does. The CLI and both binaries reconstruct from inout + private.
- The `n_const` slicing in `bin/verifier.rs` is gone — that was the judgement call this overrules, and it was the wrong one. Both sides now get the constants from the constraint system, which is the only place they were ever authoritative.

Three tests broke **semantically** rather than at compile time, since only the meaning of the first argument changed. Two are worth naming because a less suspicious fix would have left them passing while checking nothing:

- `verify_rejects_a_value_vector_that_opens_a_constant_to_the_wrong_word` can no longer construct its own premise through `value_vec_from_data` — the constants come from the system now, so the disagreement it wants to inject is unrepresentable on that path. Rewritten to build the vector via `ValueVec::new_from_data` directly; the `verify` guard still matters for vectors the circuit fills.
- The tamper test picked its victim word from the ZERO constraints, and could pick a **constant** — which the rebuild would silently restore, leaving the test green while proving nothing. It now requires a non-constant segment.
- The roundtrip test serializes `values.inout()` and checks lengths against `n_inout` / `n_private`.

Workspace clippy `-D warnings` and rustdoc are clean; core 98, prover 37 + 13, verifier 43 + 13, examples 22 + 3 all pass. The full sweep across the remaining crates is still running and I will follow up if anything turns up.

Also, apologies for the delay on this one — I read your review's `state` but not its `body` when polling, so I only picked up the inline comment at first.

jimpo (approved):

PR #2061, opened 2026-08-10
BINIUS-409: open one committed oracle at several points in a single BaseFold batch

tcoratger:
@jimpo First step for being able to use M4 in Binius64, not sure if this is the approach you wanted to take, just a first item here, I've others in the pipeline on top.

Closes [BINIUS-409](https://linear.app/jimpo/issue/BINIUS-409).

Lets one committed oracle be opened at several points in the same batched BaseFold opening.
Single-claim batches produce byte-identical proofs — no protocol change for anything that exists today.

### The problem

The batched opening assumes exactly one relation per committed oracle.
The prover asserts it outright, with a `TODO` to remove the limitation.

A relation also carries the committed message it opens.
So a second claim on the same oracle means handing the channel a second copy of that message.
For a trace-sized oracle that is a full extra copy of the trace, per claim.

This blocks the chip architecture directly.
There, one trace commitment holds the main circuit's words and every chip's block.
The main reduction and each chip reduction each end with their own evaluation claim on that one commitment.

### A latent inconsistency in the masking

The two sides count the zero-knowledge masking values differently:

- the prover sends one `sigma` per ZK **relation**
- the verifier reads one `sigma` per ZK **oracle**

Those agree only while the one-relation-per-oracle assert holds, which is why nothing is broken today.
Lifting the assert without addressing this would desynchronise the transcript.

### The idea

Batch each oracle's claims into one before the opening runs, using a sampled `lambda`:

```
T_i = sum_j lambda_i^j * t_j     the combined transparent
S_i = sum_j lambda_i^j * s_j     the combined claim
```

The inner product is linear in the transparent.
So `<pi_i, T_i> = S_i` holds exactly when every `<pi_i, t_j> = s_j` does.

Everything downstream then sees one relation per committed oracle, exactly as today.
The `sigma` accounting stays per-oracle and becomes correct by construction rather than by assumption.

This is also cheaper than making the masking per-relation.
Batching first costs one sumcheck per oracle; masking per-relation would cost one per claim.

### The change

The prover-side item carries one message per oracle with a list of claims:

```diff
- (oracle, message, transparent, claim)
+ OracleOpening { oracle, message, claims: [(transparent, claim), ...] }
```

The verifier keeps its flat relation list and groups it by oracle inside the channel.
Groups follow first-arrival order, not ascending oracle index, so existing transcripts do not move.

`lambda` is drawn only when an oracle carries more than one claim, and always before `sigma`.
A `sigma` is an inner product against the **combined** transparent, so that order is load-bearing.

### Soundness

The batching is a random linear combination of inner-product claims on one message.
A prover holding a wrong claim survives with probability at most $(k - 1) / |F|$ for $k$ claims.

`lambda` is drawn after every claim is already bound to the transcript.
So no claim can be chosen as a function of it.

An oracle with a single claim draws no challenge at all.
So the existing protocol is left exactly as it is.

### Scope

- **changed:** the verifier channel in `binius-iop`; the channel trait, BaseFold channel and naive channel in `binius-iop-prover`
- **changed:** six call sites across `binius-prover`, `binius-m4-prover` and `binius-spartan-prover`, each a one-line switch to `OracleOpening::single`
- **removed:** the `TODO` and the one-relation-per-oracle assert it guarded
- **untouched:** the masking, sumcheck and FRI stages, which still see one relation per oracle

### Verification

- `clippy` on the five touched crates with `-D warnings` — clean
- `binius-iop-prover`, `binius-iop`, `binius-prover`, `binius-m4-prover`, `binius-spartan-prover` test suites pass

Seven new channel tests cover the new capability: two claims on one oracle, mixed claim counts with the verifier queueing relations interleaved across oracles, a mixed ZK/non-ZK batch, tampered claims in first, later and non-ZK positions, and a committed oracle left unopened.

**Regression check:** the `sha256` example's proof is byte-identical to `main` (same SHA-256 digest), confirming single-claim batches draw no extra challenge.

### DIFF
--- a/crates/iop-prover/src/basefold/channel.rs
+++ b/crates/iop-prover/src/basefold/channel.rs
@@ -2,10 +2,10 @@
 
 //! BaseFold ZK implementation of the IOP prover channel.
 
-use std::ops::Deref;
+use std::{iter, ops::Deref};
 
 use binius_compute::Allocator;
-use binius_field::{BinaryField, PackedField};
+use binius_field::{BinaryField, PackedField, util::powers};
 use binius_iop::{channel::OracleSpec, fri::FRIParams};
 use binius_ip_prover::{
 	channel::IPProverChannel,
@@ -16,7 +16,7 @@ use binius_ip_prover::{
 	},
 };
 use binius_math::{
-	FieldBuffer, FieldSlice, FieldSliceMut, FieldVec,
+	FieldBuffer, FieldSlice, FieldSliceMut,
 	inner_product::inner_product_par,
 	line::{extrapolate_line, extrapolate_line_packed},
 	multilinear::eq::{eq_ind_partial_eval_scalars, eq_ind_zero},
@@ -34,7 +34,7 @@ use rand::{Rng, SeedableRng, rngs::StdRng};
 
 use crate::{
 	basefold::prove_mlecheck_basefold,
-	channel::IOPProverChannel,
+	channel::{IOPProverChannel, OracleOpening, TransparentClaim},
 	fri::{self, FRIFoldProver, MaskedCodeword},
 	merkle_channel::MerkleIPProverChannel,
 };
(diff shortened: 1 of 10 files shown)


jimpo:
Will review

tcoratger:
> Will review

I've added another related PR here: https://github.com/binius-zk/binius64/pull/2063

Will stop in the process so that you have time to review and check the details to see if this is the good way to go before I continue with other PRs

jimpo:
Hmm, I had something different in mind.

I basically thing the `prove_oracle_relations`/`verify_oracle_relations` interface is almost fine, except that `prove_oracle_relations` couples in the responsibility of returning the owned oracle buffer to the channel. I would simply remove the message buffer from the `prove_oracle_relations` tuple and add a different, separate method for handing ownership of the buffers to the channel. Like this:

```rust
pub trait IOPProverChannel<P: PackedField, A: Allocator>: IPProverChannel<P::Scalar> {
	...

	fn prove_oracle_relation(
		&mut self,
		oracle: Self::Oracle,
		transparent: FieldVec<P, A>,
		claim: P::Scalar,
	);

	/// Gives ownership of the oracle buffer to the channel.
	///
	/// The [`Self::send_oracle`] method takes a borrowed reference to an oracle buffer and returns
	/// a handle to it. In order to prove the oracle relations without unnecessarily cloning the
	/// buffer, some channel implementations require ownership of the buffer.
	fn finalize_oracle(&mut self, oracle: Self::Oracle, buffer: FieldVec<P, A>);
}
```

It's more stateful, but I find it a bit more obvious what the intention of each call is. Notice that I also changed `prove_oracle_relations` to a `prove_oracle_relation` that does only one. Doing multiple at a time is a carryover to a past interface where `prove_oracle_relations` consumed self.

PR #2062, opened 2026-08-10
BINIUS-410: parallelize the shift evaluation, which dominates verification

tcoratger:
Closes [BINIUS-410](https://linear.app/jimpo/issue/BINIUS-410).

Accelerates the verifier phase that dominates XMSS verification.
No protocol change, no transcript change — proofs are byte-identical.

### The problem

One phase accounts for ~90% of XMSS verification: `shift::check_eval`, which re-derives the circuit's wiring from the constraint list.
Everything else — the AND check, the shift reduction, the PCS opening — is tens of microseconds.

Two parts of that phase leave work on the table:

- the equality-indicator tensors are built with the scalar routine, a plain sequential loop, even where the element type is a concrete field
- the constraint term loop is threaded, but one constraint names only a handful of terms, so rayon splits it below its own handoff cost

### The idea

The packed expansion is already threaded, and a base field element is its own single-element packing.
So the native path can simply call it instead of the scalar one — no new function.

The constraint term loop gets a minimum task size, so each worker receives real work.

### Benchmark

`hashsign` (XMSS), the dominant phase, median of 9 verifications:

| | main | this change |
|---|---|---|
| verify public-input phase | 6.72 ms | 6.45 ms |

### Scope

- the native monster-evaluation path takes the packed equality-indicator expansion
- the constraint term loop takes a minimum task size
- the generic path is untouched: a symbolic element cannot cross threads, so it keeps the scalar expansion

### What this deliberately leaves open

The term loop is the bulk of the phase and is already parallel, so speeding it up further is a data-layout question rather than a threading one.

### Verification

- `clippy -p binius-math -p binius-verifier -p binius-prover --all-targets` with `-D warnings` — clean
- `binius-math`, `binius-verifier` and `binius-prover` test suites pass, with and without `--features rayon`

**Regression check:** the `hashsign` and `sha256` proofs are byte-identical to `main` (same SHA-256 digest).

### DIFF
--- a/crates/verifier/src/protocols/shift/monster.rs
+++ b/crates/verifier/src/protocols/shift/monster.rs
@@ -8,8 +8,14 @@ use binius_field::{
 	BinaryField, FieldOps, WideMul,
 	util::{FieldFn, powers},
 };
-use binius_math::multilinear::eq::eq_ind_partial_eval_scalars;
-use binius_utils::{checked_arithmetics::log2_ceil_usize, rayon::prelude::*};
+use binius_math::multilinear::eq::{eq_ind_partial_eval, eq_ind_partial_eval_scalars};
+use binius_utils::{
+	checked_arithmetics::log2_ceil_usize,
+	rayon::{
+		prelude::*,
+		task_size::{IndexedParallelIteratorExt, WorkPerItem},
+	},
+};
 
 use super::SHIFT_COUNT;
 
@@ -196,7 +202,9 @@ where
 	fn call_native(&self, input: &[F]) -> F {
 		let (r_x_prime, lambda, shift_scalars, r_y_tensor) = self.split_input(input);
 
-		let r_x_prime_tensor = eq_ind_partial_eval_scalars(r_x_prime);
+		// The packed expansion threads the tensor's multiplications.
+		// It applies over the base field, which is its own single-element packing.
+		let r_x_prime_tensor = eq_ind_partial_eval::<F>(r_x_prime);
 		let operand_shift_scalars = operand_shift_scalar_table(shift_scalars.inner, *lambda, ARITY);
 
 		// One unreduced wide product per constraint. The constraints partition cleanly across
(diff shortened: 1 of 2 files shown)


jimpo:
Can we table this for a few days until M4 is integrated? That should solve the verification time problem completely.

crates/verifier/src/protocols/shift/monster.rs
```diff
@@ -216,18 +223,21 @@ where
 	fn call_native(&self, input: &[F]) -> F {
 		let (r_x_prime, lambda, shift_scalars, r_y_tensor) = self.split_input(input);
 
-		let r_x_prime_tensor = eq_ind_partial_eval_scalars(r_x_prime);
+		let r_x_prime_tensor = eq_ind_partial_eval_scalars_par(r_x_prime);
```
jimpo: This can use regular eq_ind_partial_eval. No scalars, no new method. That's the benefit of this `call_native` context.

jimpo (commented):
@tcoratger OK, let's go ahead with this. But in the implementation, just accelerate the `call_native` path using functions that are already available for concrete fields -- no need to introduce parallel versions of the "scalar" methods (over FieldOps).

crates/verifier/src/protocols/shift/monster.rs
tcoratger: Done — dropped the new `binius-math` entry points entirely.

`call_native` now calls `eq_ind_partial_eval::<F>` (and `scaled_eq_ind_partial_eval::<F>` for the word-index tensors in `verify.rs`). A base field element is its own single-element packing, so the packed expansion applies directly and it is already threaded — no new function, no `Send + Sync` variant over `FieldOps`. The generic path keeps the scalar expansion.

Also rebased onto `main` so the two-phase shift rewrite is picked up. Re-measured on `hashsign`: the phase goes 6.72 ms → 6.45 ms (median of 9), and the `hashsign`/`sha256` proofs are byte-identical to `main`.

tcoratger: Fixed

tcoratger:
@jimpo Fixed the comment here :)

PR #2063, opened 2026-08-10
BINIUS-412: reduce the monolithic and batched verifiers through one shared driver

tcoratger:
Closes [BINIUS-412](https://linear.app/jimpo/issue/BINIUS-412).

Both Binius64 verifiers now reduce constraints through one driver, parameterized by the instance count.
Pure refactor — monolithic proofs are byte-identical.

### The problem

Two verifiers reduce the same `ConstraintSystem`, and their reduction logic is duplicated.

`binius-m4-verifier`'s reduction is `binius-verifier`'s with `log_instances` added to every row count, plus one extra step that transports each operation's operand claims onto a shared instance point.

The two copies must stay in lockstep with each other *and* with their provers.
Every ordering invariant, skip guard and domain choice is written twice.
A chip architecture needs a third caller — one reduction per chip — which would make it three.

### The idea

One driver, parameterized by the instance count:

```text
k = 0   a batch of one instance -> the monolithic reduction
k > 0   a batch of 2^k          -> the batched reduction
```

The transport step is the only difference, and it is only needed when there is more than one instance point to unify. So it is guarded on exactly that:

```text
transport claims  <=>  log_instances > 0 and a multiplication is present
```

With one instance every operation's instance point is empty, so there is nothing to transport.
That guard is what keeps the monolithic transcript unchanged.

### The change

Both verifiers shrink to the same four steps:

```diff
- 200+ lines of reduction, per verifier
+ let trace_oracle = channel.recv_oracle(..)?;
+ let reduction = reduce_constraints(cs, log_instances, public, channel)?;
+ let eval_point = reduction.trace_point();
+ // ring-switch, then open
```

- **new:** `binius_verifier::reduction` — the shared driver, the per-operation claim types, and the instance re-randomization.
- **changed:** `IOPVerifier::verify` in both `binius-verifier` and `binius-m4-verifier`.
- The reduction output carries the trace opening point, so neither caller assembles `r_j || r_rho || r_y` by hand.

The two drivers lose 543 lines between them.

### The prover has to match

The batched prover's re-randomization guard is the mirror of the verifier's, so it gains the same instance condition.

Its no-transport branch previously assumed no multiplications existed and zeroed their claims.
That is wrong once a single instance can reach it: a present multiplication must still contribute, collapsed from its per-bit form.
A zero claim there would drop those constraints silently, so this is fixed alongside.

### Soundness

The reduction order is unchanged and still load-bearing.
IntMul and BinMul run before BitAnd, because BitAnd draws the univariate challenge that collapses their per-bit operand evaluations.
Committing those evaluations first is what stops a prover choosing them as a function of it.

Skipping the transport at one instance is an identity, not a shortcut.
Every claim already sits at the empty point, and $\widetilde{eq}$ of two empty points is one.

### Verification

- `clippy` on the four touched crates with `-D warnings` — clean
- nightly `fmt --all` — clean
- `binius-verifier`, `binius-prover`, `binius-m4-verifier`, `binius-m4-prover`, `binius-spartan-verifier` test suites pass

**Transcript neutrality.** Both provers are untouched on the monolithic path, so the existing prove-then-verify suites passing *is* the byte-equality test — the verifier consumes exactly the transcript it did before. Confirmed end to end: the `sha256` example's proof has the same SHA-256 digest as on `main`.

**New coverage.** `prove_integer_multiplication_single_instance` proves a batch of one instance with IMUL constraints. That is the one case where both sides skip the transport *and* the multiplication claims still have to reach the shift, so it pins the prover and verifier guards together.

Existing tests already cover the other three combinations: `k = 0` with multiplications (the whole `binius-prover` suite), `k > 0` with (`prove_integer_multiplication`), and `k > 0` without (blake3, keccak).

### DIFF
--- a/crates/m4-prover/src/prove.rs
+++ b/crates/m4-prover/src/prove.rs
@@ -16,7 +16,7 @@ use binius_ip_prover::sumcheck::{
 use binius_m4_verifier::{IOPVerifier, Verifier};
 use binius_math::{
 	BinarySubspace,
-	inner_product::inner_product,
+	inner_product::{inner_product, inner_product_scalars},
 	multilinear::eq::eq_ind_partial_eval_scalars,
 	ntt::{NeighborsLastMultiThread, domain_context::GenericPreExpanded},
 	univariate::lagrange_evals_scalars,
@@ -228,7 +228,8 @@ impl IOPProver {
 		//
 		// The re-randomization runs whenever IntMul or BinMul is present: BitAnd always enters,
 		// plus each present multiplication operation, all unified onto one shared `r_rho`.
-		let (r_rho, claims) = if mul.is_some() || bmul.is_some() {
+		// One instance has nothing to transport; the verifier's guard matches.
+		let (r_rho, claims) = if table.log_instances() > 0 && (mul.is_some() || bmul.is_some()) {
 			// Every present operation enters the re-randomization as operand columns with their
 			// oblong claims at their own instance point.
 			// BitAnd is already oblong.
@@ -268,9 +269,12 @@ impl IOPProver {
 			}
 			.prove::<P, _>(zero_data, &lagrange, z_challenge, channel, alloc)
 		} else {
-			// Neither IMUL nor BMUL constraints: the AND-check instance point is used directly.
-			// The IntMul and BinMul claims are zero claims at an empty point, contributing nothing
-			// to the shift.
+			// Nothing to transport: every claim already sits at the AND-check instance point.
+			// A present multiplication still contributes, collapsed from its per-bit form.
+			let lagrange = lagrange_evals_scalars::<B128, B128>(&shift_domain, &z_challenge);
+			let collapse = |evals: &[B128]| {
+				inner_product_scalars(evals.iter().copied(), lagrange.iter().copied())
+			};
 			(
 				r_rho_and.to_vec(),
 				OperatorClaims {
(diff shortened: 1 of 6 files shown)


crates/m4-prover/src/prove.rs
```diff
@@ -280,8 +284,34 @@ impl IOPProver {
 						r_zhat_prime: z_challenge,
 						r_x_prime: r_x_and.to_vec(),
 					},
-					intmul: OperatorData::zero_claim(z_challenge),
-					binmul: OperatorData::zero_claim(z_challenge),
+					intmul: mul.as_ref().map_or_else(
```
jimpo: Is this code just here for the case of a ValueTable with only 1 instance? We don't need special handling for log_instances = 0, the first branch with the re-randomization will be fine (and should be a 0-round sumcheck).

jimpo (commented):
Nice deduplication. One comment

crates/m4-prover/src/prove.rs
tcoratger: Yes exact this is just for the case with only 1 instance in order to unify things completely in the future, let me know if this is not what you want :)

jimpo: The shift reduction prover is special (and importantly optimized) in the single-instance case. My preference would be to drop this check for log_instances = 0 in the M4 case.

tcoratger: Ok no problem, fixed

PR #2067, opened 2026-08-10
BINIUS-418: batch the logUp* pushforward reduction over several tables

benji-potamus:
First of three PRs for [BINIUS-418](https://linear.app/jimpo/issue/BINIUS-418/logup-star-multi-table-batch-lookup) — multi-table batch lookup. This one is groundwork only: it generalizes the pushforward reduction to a batch of tables, but the caller still passes exactly one, so **the transcript is unchanged**.

## What changes

`prove_pushforward` / `verify_pushforward` take a slice of per-table claims instead of a single table's, and return one `(T_t, Y_t)` claim pair per table:

- **Prover** — one `SharedSumcheckProver` over `[Y_t, T_t]` per table, each wrapped in the existing `PaddedSumcheckDecorator` to raise it to the batch's round count, all driven by one `batch_prove`.
- **Verifier** — one `sumcheck::batch_verify(max_m, 2, …)` over `2 · n_tables` flattened claims, with each table's contribution carrying `eq_ind_zero` over its padding coordinates.

This is the same shape `iop/basefold/channel.rs` already uses to open oracles of unequal message length (`PaddedSumcheckDecorator` on one side, `eq_ind_zero(padding_coords)` on the other), so it is a pattern the codebase has already committed to rather than a new mechanism.

Two consequences worth naming, since they are easy to get backwards:

- The decorator pads at the **highest-indexed** variables, which are bound first. So table `t` ends up claimed at the **first `m_t`** coordinates of the shared reduced point — a prefix. That is the opposite end from the fractional-addition batch, which pads at the lowest variables and hands a shallow instance a **suffix** (`LogupOutput::index_eval_point`, #2064). Both are documented at their respective outputs.
- The deepest table is never padded, so it is in its inner phase every round and the batched round polynomial always has degree 2. That is what lets the verifier read a uniform degree, and it is why an all-padding round — which would emit fewer coefficients than the verifier reads — cannot occur.

## Why the transcript is unchanged

With one table the padding width is zero, so the decorator is a passthrough (`test_no_padding_passthrough` already covers this); the same two claims are batched in the same order; and `recv_many(2)` reads the same two scalars `recv_array::<2>` did. Corroborated by `size_tracking_matches_real_proof_bytes`, which pins a full Spartan proof's real byte count against its predicted size.

## What this PR does not do

It has no test that exercises more than one table, because `verify_pushforward` is private to `binius_ip::logup_star` and I would rather not widen the public API to reach it from a test. The multi-table round-trips land in the next PR in the stack, where the capability is reachable through `verify_reduction`. Until then the evidence for this PR is that it is transcript-preserving and every existing test passes unchanged.

The single-table callers each grow six lines that pull one claim back out of a returned `Vec`. That boilerplate disappears in the next PR, when `LogupOutput` itself becomes per-table.

## Verification

- `cargo test` green on `binius-ip`, `binius-ip-prover`, `binius-iop`, `binius-iop-prover`, `binius-prover`, `binius-verifier`, `binius-spartan-prover`, `binius-spartan-verifier` — no test changed.
- `cargo clippy --all --tests --benches --examples -- -D warnings` clean across the workspace.

## Stack

1. **this PR** — batch the pushforward reduction over several tables
2. the multi-table reduction itself: one logUp challenge `c_t` per table, lookers assigned to tables, fracadd instances become `lookers ++ tables`
3. one pushforward oracle per table in `iop`/`iop-prover`, plus the `intmul` call sites

### DIFF
--- a/crates/ip-prover/src/logup_star/pushforward.rs
+++ b/crates/ip-prover/src/logup_star/pushforward.rs
@@ -2,15 +2,15 @@
 
 //! The pushforward reduction that closes the table side.
 //!
-//! The table-side fractional-addition GKR runs all the way to its leaf, which claims the
-//! pushforward `Y` at a point `z`. Its denominator half is the public `D = J - c`, so the verifier
-//! checks that one itself. That leaves two claims that both read `Y`:
+//! Each table's fractional-addition GKR runs all the way to its leaf, which claims that table's
+//! pushforward `Y_t` at a point `z_t`. Its denominator half is the public `D = J - c_t`, so the
+//! verifier checks that one itself. That leaves two claims per table that both read `Y_t`:
 //!
-//! - the GKR leaf claim `Y(z)`,
-//! - the product claim `<T, Y> = e`.
+//! - the GKR leaf claim `Y_t(z_t)`,
+//! - the product claim `<T_t, Y_t> = e_t`.
 //!
-//! One batched `m`-variable sumcheck over the shared column pair `[Y, T]` reduces both to a single
-//! evaluation point, giving one evaluation of `Y` and one of `T`.
+//! One batched sumcheck reduces every one of them to a single evaluation point, giving one
+//! evaluation of `Y_t` and one of `T_t` per table.
 //!
 //! This is the prover mirror of the verifier's pushforward reduction in [`binius_ip::logup_star`].
 
@@ -21,6 +21,7 @@ use binius_math::FieldSlice;
 use crate::{
 	channel::IPProverChannel,
 	sumcheck::{
+		PaddedSumcheckDecorator,
 		batch::batch_prove,
 		bivariate_product_evaluator::BivariateProductEvaluator,
 		mle_store::MleStore,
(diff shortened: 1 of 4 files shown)


crates/ip/src/logup_star/pushforward.rs
```diff
@@ -2,94 +2,156 @@
...
 /// * `channel` - The verifier channel.
+///
+/// # Preconditions
+///
+/// - `tables` is non-empty.
 pub fn verify_pushforward<F, C>(
-	eval_claim: C::Elem,
-	pushforward_eval_claim: C::Elem,
-	pushforward_eval_point: &[C::Elem],
+	tables: &[TableClaim<'_, C::Elem>],
```
jimpo: Make this an `impl IntoIterator<Item=TableClaim<..>>`

crates/ip/src/logup_star/verify.rs
```diff
@@ -229,9 +229,25 @@ where
...
+	} = verify_pushforward::<F, C>(
+		&[TableClaim {
+			eval_claim: combined_eval_claim,
+			pushforward_eval_claim: pushforward_eval,
+			pushforward_eval_point: table_point,
+		}],
+		channel,
+	)?;
+	// The reduction runs over the one shared table, so it returns one claim of each kind.
+	let table_eval_claim = table_eval_claims
```
jimpo: Can these lines use destructuring assignment? Or `TryInto<[_; 1]>` at least?

crates/ip-prover/src/logup_star/pushforward.rs
```diff
@@ -29,107 +30,155 @@ use crate::{
...
+/// * For each table, `table.log_len() == pushforward.log_len() == pushforward_eval_point.len()`.
 #[tracing::instrument(skip_all, level = "debug", name = "logup* pushforward reduction")]
 pub fn prove_pushforward<'a, A, F, P>(
 	alloc: &'a A,
-	table: FieldSlice<'a, P>,
-	pushforward: FieldSlice<'a, P>,
-	eval_claim: F,
-	pushforward_eval_claim: F,
-	pushforward_eval_point: &[F],
+	tables: &[TableWitness<'a, P>],
```
jimpo: take by value. It's type should be `impl IntoIterator<Item=TableWitness<...>>`

jimpo (changes requested):

crates/ip/src/logup_star/pushforward.rs
benji-potamus: Done — now `tables: impl IntoIterator<Item = TableClaim<'a, C::Elem>>`, taken by value.

It is materialized once inside (`into_iter().collect()`) because the claims are walked three times: for the deepest table, for the batched sums, and for the reconstruction. That let the single-table caller drop its `&[…]` and pass the array directly, and in the multi-table PR above it lets `verify_reduction` pass its `map` chain straight through instead of collecting a `Vec` first.

crates/ip-prover/src/logup_star/pushforward.rs
benji-potamus: Done — `tables: impl IntoIterator<Item = TableWitness<'a, P>>`, by value, and the per-table loop now consumes it with `into_iter()` rather than borrowing.

Same as the verifier side: materialized once internally since it is walked twice (deepest table, then building the provers), and the callers now pass an array or a `map` chain directly.

crates/ip/src/logup_star/verify.rs
benji-potamus: Switched to `TryInto<[_; 1]>` with a destructuring `let`:

```rust
let [table_eval_claim]: [_; 1] = table_eval_claims
    .try_into()
    .unwrap_or_else(|_| unreachable!("the reduction runs over one table"));
```

`unwrap_or_else` rather than `expect` because `Result::expect` needs `E: Debug`, and the error here is `Vec<C::Elem>` — `C::Elem` carries no `Debug` bound on the verifier side. (`.ok().expect(…)` would sidestep that but trips `clippy::ok_expect`.) Applied the same shape to the mirrored spot in `ip-prover/prove.rs` and to the two single-table extractions in the `iop` layers in the PR above.

Worth noting these all disappear by the top of the stack: once `LogupOutput` carries per-table vectors (#2069) and the committed layer takes a table list (#2071), nothing pulls a single claim back out.

jimpo (approved):

PR #2069, opened 2026-08-10
BINIUS-418: let the logUp* lookers read several tables

benji-potamus:
Second of three PRs for [BINIUS-418](https://linear.app/jimpo/issue/BINIUS-418/logup-star-multi-table-batch-lookup). #2067 has merged, so this is now based on `main`. #2071 stacks on top.

One reduction now serves several tables, each with its own lookers, and no column need agree with any other on a length.

## The shape

Each table is passed as a `TableLookup` carrying its **own** batching challenge, its variable count, and its lookers:

```rust
pub struct TableLookup<'a, Elem> {
    pub gamma: Elem,
    pub n_vars: usize,
    pub lookers: Vec<LookerClaim<'a, Elem>>,
}
```

- **One logUp challenge `c` per table**, sampled inside the reduction; **one `gamma` per table**, sampled by the caller before that table's pushforward commitment.
- The fractional-addition batch runs `#lookers + #tables` instances, so `k = ceil(log2(#lookers + #tables))` and the depth is `max(max n, max m)`. Instances are laid out `lookers ++ tables`, lookers ordered table by table.
- Each table gets its own pushforward `Y`, the gamma-weighted scatter of just its own lookers, and its own product claim. All of them close in the one batched sumcheck from #2067.

Per-table `gamma` (rather than one global challenge indexed by a looker's position in the whole batch) makes a table's product claim exactly `evaluate_univariate(its claims, its gamma)`, and drops the global-position bookkeeping entirely. Tables now share nothing but the batching machinery.

## Why one logUp challenge per table

Table `t`'s contribution to the root fraction is

```
f_t(c_t) = sum_i gamma_t^i * sum_x eq_{r_i}(x)/(c_t - I_i(x))  -  sum_v Y_t(v)/(c_t - v)
```

— a rational function of `c_t` **alone**, vanishing as `c_t` grows. A sum of such functions in disjoint variables is identically zero only when every term is, so the single root-numerator check certifies each table separately. Under one shared `c` that argument collapses: two tables could miscount the same position in opposite directions and cancel inside the root numerator, and the batch would verify while being unsound. `test_verifier_rejects_lookup_against_the_wrong_table` pins the honest direction.

## Output shape

`LogupOutput` grows per-table vectors, and `index_eval_claims` is `Vec<Vec<F>>` — grouped per table, mirroring the input. Each side carries **one** point, and the two are read from **opposite ends**:

- **table side** — the pushforward batch pads at the high variables, so a table is claimed at the **first `m`** coordinates of `table_eval_point`;
- **index side** — the fracadd batch pads at the low variables, so a looker is claimed at the **last `n`** coordinates of `index_eval_point`.

That asymmetry is inherent to the two padding conventions rather than a choice; it is documented on `LogupOutput` and on both reductions.

## What is deliberately left alone

`iop` / `iop-prover` still commit exactly one pushforward, so they wrap a single `TableLookup`. Generalizing them is #2071. `intmul` is unchanged in substance and its transcript does not move — `size_tracking_matches_real_proof_bytes` still passes.

Also fixed in passing: `prove`'s doc comment had drifted above the `Looker` struct, leaving the function undocumented; and `logup_star/mod.rs` still described the pre-#2052 protocol. Both now match the code.

## Tests

The round-trip helper is parameterized by `[(table_n_vars, [looker_n_vars])]`, so one function covers the single-table, multi-looker, unequal-length and multi-table shapes; every prior case runs through it. It checks each table claim, each **pushforward** claim against the honestly rebuilt `Y` (weighted by that table's own gamma), and each index claim at its own suffix.

New coverage: `test_multi_table_round_trip` (five shapes, including one where the deepest instance is a *table* and one with equal-size tables, the no-padding path), `test_tables_are_independent`, `test_verifier_rejects_lookup_against_the_wrong_table`, and `test_table_without_lookers_panics` on both sides.

## Verification

- `cargo test` green on `binius-ip`, `binius-ip-prover`, `binius-iop`, `binius-iop-prover`, `binius-prover`, `binius-verifier`, `binius-spartan-prover`, `binius-spartan-verifier`, verified on this commit standing alone as well as at the top of the stack.
- `cargo clippy --all --tests --benches --examples -- -D warnings` clean.

### DIFF
--- a/crates/ip/src/logup_star/output.rs
+++ b/crates/ip/src/logup_star/output.rs
@@ -6,21 +6,38 @@
 ///
 /// Each claim must be verified separately by the caller.
 /// Verifying them is out of scope here.
+///
+/// The two sides each carry **one** point, shared by every table, and read it from opposite ends.
+/// That is not a choice: the pushforward reduction pads a table at its high variables, while the
+/// fractional-addition batch pads a looker at its low ones.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct LogupOutput<F> {
-	/// The `m`-coordinate point shared by the table and pushforward evaluation claims.
-	pub table_eval_point: Vec<F>,
-	/// The claimed evaluation of the table multilinear `T` at the table point.
-	pub table_eval_claim: F,
-	/// The claimed evaluation of the pushforward multilinear `Y` at the table point.
-	pub pushforward_eval_claim: F,
-	/// The point the index evaluation claims are drawn from, of `max_j n_j` coordinates.
+	/// The point the table and pushforward evaluation claims are drawn from, of `max m`
+	/// coordinates.
 	///
-	/// Looker `j`, whose column has `n_j` variables, is claimed at the **last `n_j`** coordinates.
-	/// Lookers of equal length therefore all share the whole point; a shorter looker's point is a
-	/// suffix of a longer one's, because the batch pads each instance at its low coordinates.
+	/// A table over `m` variables is claimed at the **first `m`** coordinates. Tables of equal
+	/// size therefore all share the whole point; a smaller table's point is a prefix of a larger
+	/// one's.
+	pub table_eval_point: Vec<F>,
+	/// The point the index evaluation claims are drawn from, of `max n` coordinates.
+	///
+	/// A looker whose column has `n` variables is claimed at the **last `n`** coordinates. Lookers
+	/// of equal length therefore all share the whole point; a shorter looker's point is a suffix
+	/// of a longer one's, because the batch pads each instance at its low coordinates.
 	pub index_eval_point: Vec<F>,
-	/// The claimed evaluations of the per-looker index multilinears `I_j`, each at its own suffix
-	/// of [`Self::index_eval_point`].
+	/// One entry per table, in the order the tables were given.
+	pub tables: Vec<LogupTableOutput<F>>,
+}
+
+/// The reduced claims belonging to one table.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LogupTableOutput<F> {
+	/// The claimed evaluation of the table multilinear `T` at its prefix of
+	/// [`LogupOutput::table_eval_point`].
+	pub eval_claim: F,
+	/// The claimed evaluation of the pushforward multilinear `Y` at the same prefix.
+	pub pushforward_claim: F,
+	/// The claimed evaluations of this table's lookers' index multilinears `I`, in its own looker
+	/// order, each at that looker's own suffix of [`LogupOutput::index_eval_point`].
 	pub index_eval_claims: Vec<F>,
 }
--- a/crates/ip/src/logup_star/verify.rs
+++ b/crates/ip/src/logup_star/verify.rs
@@ -12,10 +12,11 @@ use binius_math::{
 	},
 	univariate::evaluate_univariate,
 };
+use itertools::izip;
 
 use super::{
 	error::{Error, VerificationError},
-	output::LogupOutput,
+	output::{LogupOutput, LogupTableOutput},
 	pushforward::{Pushforward, TableClaim, denominator_eval, verify_pushforward},
 };
 use crate::{
@@ -23,7 +24,7 @@ use crate::{
 	fracaddcheck::{self, FracAddEvalClaim},
 };
 
-/// One looker's claim on its looked-up vector: `(I_j^* T)(eval_point) = eval_claim`.
+/// One looker's claim on its looked-up vector: `(I^* T)(eval_point) = eval_claim`.
 #[derive(Debug, Clone)]
 pub struct LookerClaim<'a, Elem> {
 	/// The `n`-coordinate evaluation point of this looker's claim.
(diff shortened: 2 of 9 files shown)


crates/ip/src/logup_star/verify.rs
```diff
@@ -23,76 +20,97 @@ use crate::{
 	fracaddcheck::{self, FracAddEvalClaim},
 };
 
-/// One looker's claim on its looked-up vector: `(I_j^* T)(eval_point) = eval_claim`.
+/// One looker's claim on its looked-up vector: `(I_j^* T_t)(eval_point) = eval_claim`.
 #[derive(Debug, Clone)]
 pub struct LookerClaim<'a, Elem> {
+	/// Which table this looker reads, as an index into the reduction's `table_n_vars`.
+	pub table: usize,
```
jimpo: No, I don't like referencing the tables by index here. The caller should pass the reduction verifier a struct per table. We can call it `struct TableLookup`. It has the `gamma` value, `n_vars` for the table, and a Vec of LookerClaims.

jimpo (changes requested):

crates/ip/src/logup_star/verify.rs
benji-potamus: Done — indexing is gone. `LookerClaim` no longer names a table; the reduction takes a `TableLookup` per table:

```rust
pub struct TableLookup<'a, Elem> {
    pub gamma: Elem,
    pub n_vars: usize,
    pub lookers: Vec<LookerClaim<'a, Elem>>,
}
```

Passed as `impl IntoIterator<Item = TableLookup<..>>`, matching the by-value shape you asked for on #2067.

Putting `gamma` in the struct is a genuine improvement I hadn't seen: it makes the challenge **per table** rather than one global challenge indexed by a looker's position across the whole batch. A table's product claim is then just `evaluate_univariate(its claims, its gamma)`, and all the "looker `j`'s weight is its global index, not its index within its table" bookkeeping disappears. Tables now share nothing but the batching machinery, which is a much easier invariant to state and to check. Soundness is the same argument applied per table.

Two knock-on choices, easy to change if you'd rather:

- `LogupOutput::index_eval_claims` became `Vec<Vec<F>>`, grouped per table to mirror the input. The alternative was to keep it flat in table-major order; grouping seemed right once the input is grouped.
- The `iop` layers can't take a `TableLookup` directly, since they sample `gamma` themselves (it must be drawn before that table's pushforward commitment). They take a `TableClaims` / `TableWitness` — the same struct minus `gamma` — rather than a bare tuple.

The prover mirrors it with `TableLookup { gamma, table, lookers }`, and `prove_reduction` now derives the per-table product claims itself, so the separate `combined_eval_claims` helper is gone.

crates/ip/src/logup_star/verify.rs
```diff
@@ -185,31 +237,48 @@ where
...
-		(pushforward_eval.clone(), denominator_eval::<F, C::Elem>(&c, table_point)),
-		pad_eqs[table_pad].clone(),
-	);
-	leaf_nums.push(table_num);
-	leaf_dens.push(table_den);
+	// A table's numerator is its Y_t, just read. Its denominator is the transparent J - c_t, the
+	// logUp denominator negated — every table's fraction enters the sum that way, which is what
+	// makes the root numerator vanish.
+	for ((c, point), pushforward_eval) in
+		iter::zip(iter::zip(&cs, &table_points), &pushforward_evals)
```
jimpo, line 280: nit: izip!

crates/ip/src/logup_star/verify.rs
```diff
@@ -220,39 +289,37 @@ where
...
-	// The product check binds <T, Y> to the gamma-combination of the looker claims.
-	let claims = lookers
-		.iter()
-		.map(|looker| looker.eval_claim.clone())
-		.collect::<Vec<_>>();
-	let combined_eval_claim = evaluate_univariate(&claims, gamma);
+	// Reduce every table's leaf claim on Y_t and its product claim <T_t, Y_t> = e_t to one shared
+	// evaluation point. A table's product check binds it to the gamma-combination of the claims of
+	// the lookers that read it.
+	let table_claims = iter::zip(iter::zip(&tables, pushforward_evals), &table_points).map(
```
jimpo: nit: izip!

crates/ip/src/logup_star/verify.rs
```diff
@@ -32,67 +32,98 @@ pub struct LookerClaim<'a, Elem> {
...
 
-/// Verify a logUp* indexed-lookup reduction over one or more lookers sharing a table.
+/// One table together with the lookers that read it.
+#[derive(Debug, Clone)]
+pub struct TableLookup<'a, Elem> {
+	/// The batching challenge for this table's lookers: looker `i` is weighted by `gamma^i`.
+	///
+	/// The caller samples it, and must do so before receiving this table's pushforward commitment:
+	/// the prover needs it to build the pushforward it commits.
+	pub gamma: Elem,
```
jimpo: I think it's sound to have a single gamma across all tables, since they are separated by the denominator challenges. So pull gamma out of TableLookup, back to a regular argument to verify_reduction.

crates/ip/src/logup_star/verify.rs
```diff
@@ -185,31 +237,48 @@ where
...
+	// The denominators are that table's c minus the index evaluation just read. The evaluations
+	// arrive table by table, so they are split back into per-table groups here.
+	let mut index_eval_claims = Vec::with_capacity(n_tables);
+	let mut remaining_index_evals = index_evals.as_slice();
+	let (mut leaf_nums, mut leaf_dens): (Vec<_>, Vec<_>) = iter::zip(&tables, &cs)
+		.flat_map(|(table, c)| {
+			let (table_evals, rest) = remaining_index_evals.split_at(table.lookers.len());
+			remaining_index_evals = rest;
+			index_eval_claims.push(table_evals.to_vec());
+			iter::zip(iter::zip(&table.lookers, powers(table.gamma.clone())), table_evals).map(
```
jimpo: izip

crates/ip/src/logup_star/output.rs
```diff
@@ -6,21 +6,33 @@
...
+	/// equal size therefore all share the whole point; a smaller table's point is a prefix of a
+	/// larger one's.
 	pub table_eval_point: Vec<F>,
-	/// The claimed evaluation of the table multilinear `T` at the table point.
-	pub table_eval_claim: F,
-	/// The claimed evaluation of the pushforward multilinear `Y` at the table point.
-	pub pushforward_eval_claim: F,
+	/// The claimed evaluations of the table multilinears `T_t`, each at its own prefix of
+	/// [`Self::table_eval_point`].
+	pub table_eval_claims: Vec<F>,
```
jimpo: Lets organize this into a struct `LogupTableOutput` with fields `eval_claim`, `pushforward_claim`, and `index_eval_claims`.

jimpo (changes requested):

crates/ip/src/logup_star/verify.rs
benji-potamus: Agreed and done — `gamma` is out of `TableLookup` and back to a regular argument of `verify_reduction` / `prove`. Looker `j` is weighted by `gamma^j` across the whole batch, so a table's product claim is the sum of its own lookers' claims at their **global** weights (rather than a per-table series restarting at `gamma^0`).

Your reasoning holds: the per-table denominator challenges already separate the tables, so the `gamma` combination only has to bind the lookers *within* each table, and distinct powers do that regardless of which table a looker sits in.

Worth noting this also simplified the layer above: with `gamma` gone from the struct, the `TableClaims` / `TableWitness` types I'd added to `iop` / `iop-prover` (which existed only because those layers sample `gamma` themselves) collapse back into the reduction's own `TableLookup`. One type per side instead of two.

One thing this caught: my round-trip test was rebuilding each table's honest pushforward with a *fresh* `powers(gamma)` per table, which passed under the old per-table scheme and failed under this one. The test was wrong, not the implementation — fixed to consume one running series across the batch.

crates/ip/src/logup_star/output.rs
benji-potamus: Done:

```rust
pub struct LogupOutput<F> {
    pub table_eval_point: Vec<F>,
    pub index_eval_point: Vec<F>,
    pub tables: Vec<LogupTableOutput<F>>,
}

pub struct LogupTableOutput<F> {
    pub eval_claim: F,
    pub pushforward_claim: F,
    pub index_eval_claims: Vec<F>,
}
```

The two shared points stay at the top level since they genuinely are shared — every table's claims are read from one point per side, from opposite ends. Everything per-table now lives in `LogupTableOutput`, which also removes the three parallel `Vec`s that had to be kept index-aligned by hand.

`iop` / `iop-prover` return the same `Vec<LogupTableOutput>` rather than re-wrapping, and `intmul` destructures the single entry.

crates/ip/src/logup_star/verify.rs
benji-potamus: Done — all three nested `iter::zip(iter::zip(..))` sites in `verify.rs` are now `izip!`, along with the same pattern in `ip-prover`. Added `itertools` to `binius-ip`'s dependencies; it was already a workspace dep and already used in `ip-prover`.

crates/ip/src/logup_star/verify.rs
```diff
@@ -102,46 +126,72 @@ pub struct LookerClaim<'a, Elem> {
...
+	let max_m = tables
+		.iter()
+		.map(|table| table.n_vars)
+		.max()
+		.expect("tables is non-empty");
+
+	// Looker `j` is weighted by gamma^j across the whole batch. A single challenge suffices even
+	// though the tables are independent: their fractions are separated by the per-table denominator
+	// challenges, so the combination only has to bind the lookers within each table.
+	let looker_powers = powers(gamma.clone()).take(n_lookers).collect::<Vec<_>>();
```
jimpo: We don't need this many. We only need the first `max_n` gamma powers. The powers can be reused across different tables because the denominator challenges differ.

jimpo (changes requested):

crates/ip/src/logup_star/verify.rs
benji-potamus: Done — the series now restarts per table, so only `max_table_lookers` powers are materialized (the largest table's looker count) instead of one per looker in the batch.

Your reasoning is the same separation argument as before, applied one level down: the combination only has to bind the lookers *inside* a table, and the per-table denominator challenges keep the tables from interfering, so reusing `gamma^0, gamma^1, …` across tables is fine.

Nice side effect: with each table's weights starting at `gamma^0` again, its product claim is back to a plain `evaluate_univariate(its claims, gamma)` on both sides, and the running-offset bookkeeping (`split_at` threading a cursor through the fold) is gone from the verifier, the prover, and the witness builder.

Same test caveat as last round, in the other direction — the round-trip's honest pushforward reconstruction had to go back to restarting `powers(gamma)` per table. It's the one place the weighting scheme is written down independently of the implementation, so it flips each time this does.

jimpo (approved):

PR #2070, opened 2026-08-10
BINIUS-413: run the shift reduction phase 1 as one 15-variate sumcheck

benji-potamus:
Closes [BINIUS-413](https://linear.app/jimpo/issue/BINIUS-413/run-the-shift-reduction-phase-1-sumcheck-as-one-15-variate-sumcheck), groundwork for [BINIUS-408](https://linear.app/jimpo/issue/BINIUS-408/double-shifted-value-indices).

Phase 1 ran as a summed batch of eight 12-variate sumchecks, one per shift variant: each round executed all eight provers, summed their round coefficients, and folded them with a shared challenge. Run it instead as a single **15-variate** sumcheck, with the shift variant as three more variables in the high index positions and `g`/`h` the multilinear interpolations over them.

The claimed sum is unchanged — summing `G · H` over the 15-cube is `Σ_v Σ_{j,s} g_v(j,s) h_v(j,s)`, exactly the total the eight sumchecks claimed. What changes is that the variant axis is folded by the sumcheck instead of summed eagerly, so the final evaluation is one product at a random `r_v`.

## What it buys

Beyond deleting the per-round `RoundCoeffs` batching (`run_phase_1_sumcheck` is now `bivariate_product_prover` + `prove_single`), the variant axis collapsing makes the downstream weights simpler:

- The verifier computes **one** `h` evaluation — the interpolation of the eight shift indicators over `r_v` — instead of an array of eight.
- The per-key shift scalar becomes a pure eq-tensor product, `h_eval · eq(r_v, variant) · eq(r_s, amount)`, on both sides.

That last point is the reason this is the right groundwork: under double shifts the variant axis becomes 6 bits and the amount axis 12, so a single flat sumcheck generalizes by widening axes. A batch of per-variant provers would have had to grow to 64, and the `SHIFT_VARIANT_COUNT * Word::BITS` scalar table would have gone quadratic.

## Notes for review

- **`build_h` lost its eight per-variant closures.** They collapsed into one `fill_h_row(variant, amount, row, l_tilde)` with `sll`/`srl`/`sar`/`rotr` helpers and a `halves` closure that covers all four `*32` forms — the same "one place that says what a variant does" shape `ShiftVariant::dispatch` already uses, instead of four near-duplicate half-word closures.
- **Renamed `build_g_parts` → `build_g`, `build_h_parts` → `build_h`, `build_g_parts_from_folded_words` → `build_g_from_folded_words`, `scatter_multilinear_parts` → `scatter_shift_rows`.** They each return one multilinear now, so "parts" was actively misleading. Tracing span names moved with them.
- **`h_op_consistency` was rewritten rather than dropped.** It now asserts that the built multilinear evaluated at `(r_j, r_s, r_v)` equals the interpolation of the eight succinct `evaluate_h_op` values over `r_v` — which is exactly the scalar phase 2 weights its keys by, so it pins the new contract.
- **`build_monster_segments` still computes one `h` evaluation per operation**, from that operation's own `r_zhat_prime`, even though the verifier uses a single one. That redundancy is pre-existing (the four are equal in practice); I kept the shape rather than fold it into this change.
- `evaluate_h_op` is untouched — it is still the right primitive; the interpolation happens at its call sites.

## Verification

`cargo check --workspace --all-targets` clean; clippy clean over prover/verifier/m4-prover/m4-verifier with `--all-targets -- -D warnings`; fmt applied. Tests: prover lib 37, prover `prove_verify` 13 (including the sha256 preimage and both zk round trips), m4-prover 43 + 5 integration, verifier 13, m4-verifier 3. Left the rest to CI.

### DIFF
--- a/crates/prover/src/protocols/shift/phase_1.rs
+++ b/crates/prover/src/protocols/shift/phase_1.rs
@@ -1,22 +1,22 @@
 // Copyright 2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::{array, iter, ops::Range};
+use std::{iter, ops::Range};
 
 use binius_compute::Allocator;
 use binius_core::word::Word;
 use binius_field::{BinaryField, Field, PackedField};
-use binius_ip::sumcheck::{SumcheckOutput, common::RoundCoeffs};
+use binius_ip::sumcheck::SumcheckOutput;
 use binius_ip_prover::{
 	channel::IPProverChannel,
-	sumcheck::{bivariate_product_prover, common::SumcheckProver},
+	sumcheck::{ProveSingleOutput, bivariate_product_prover, prove_single},
 };
 use binius_math::{BinarySubspace, FieldBuffer, FieldVec, inner_product::inner_product_buffers};
 use binius_utils::rayon::{
 	prelude::*,
 	task_size::{IndexedParallelIteratorExt, WorkPerItem},
 };
-use binius_verifier::protocols::shift::SHIFT_VARIANT_COUNT;
+use binius_verifier::protocols::shift::LOG_SHIFT_VARIANT_COUNT;
 use bytemuck::zeroed_vec;
 use itertools::izip;
 use tracing::instrument;
@@ -25,15 +25,18 @@ use super::{
 	SegmentWords,
 	claims::PreparedOperatorClaims,
 	key_collection::{DenseShiftEncoding, KeyCollection, KeySegment},
-	monster::build_h_parts,
+	monster::build_h,
 };
 
-// This is the number of variables in the g (and h) multilinears of phase 1.
-const LOG_LEN: usize = Word::LOG_BITS + Word::LOG_BITS;
+/// The number of variables in the g (and h) multilinear of phase 1.
+///
+/// The axes run, from the low index positions up: the bit position within a word, the shift
+/// amount, and the shift variant.
+pub const PHASE_1_LOG_LEN: usize = Word::LOG_BITS + Word::LOG_BITS + LOG_SHIFT_VARIANT_COUNT;
 
-/// Constructs the "g" multilinear parts for the BITAND, INTMUL and BMUL operations.
 /// Proves the first phase of the shift reduction.
-/// Computes the g and h multilinears and performs the sumcheck.
+///
+/// Builds the g and h multilinears and runs one sumcheck over their product.
 #[instrument(skip_all, name = "prover_phase_1")]
 pub fn prove_phase_1<F, P, Channel, A>(
 	key_collection: &KeyCollection,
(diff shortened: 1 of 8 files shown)


crates/prover/src/protocols/shift/phase_1.rs
```diff
@@ -49,119 +52,64 @@ where
...
 	alloc: &A,
 ) -> SumcheckOutput<F> {
-	// Build one shared bivariate-product prover per shift variant.
-	let mut provers = iter::zip(g_parts, h_parts)
-		.map(|(g_part, h_part)| {
-			let sum = inner_product_buffers(&g_part, &h_part);
-			bivariate_product_prover(alloc, [g_part, h_part], sum)
-		})
-		.collect::<Vec<_>>();
+	let sum = inner_product_buffers(&g, &h);
```
jimpo, line 93: As a follow-up, we should compute this in the caller (in the way the verifier does).

jimpo (approved):

crates/prover/src/protocols/shift/phase_1.rs
benji-potamus: Agreed, and it's more than a cost saving. Computing the sum from the buffers means the prover proves whatever `g · h` happens to sum to; a `g` that disagrees with the claims the verifier batched would only surface at the very end. Passing the claimed sum in — `prepared.zero.batched_eval + prepared.bitand.batched_eval + prepared.intmul.batched_eval + prepared.binmul.batched_eval`, which is exactly what the verifier feeds `verify_sumcheck` — makes the prover assert the claim it was given. The O(2^15) inner product going away is a bonus.

Filed as [BINIUS-419](https://linear.app/jimpo/issue/BINIUS-419/pass-the-phase-1-claimed-sum-in-from-the-caller).

benji-potamus: Correction to the link above: the follow-up is [BINIUS-423](https://linear.app/jimpo/issue/BINIUS-423/pass-the-phase-1-claimed-sum-in-from-the-caller), not 419 — I wrote the number before Linear assigned it.

benji-potamus: Correcting myself while implementing this: I implied above that passing the claim in makes a mismatch surface earlier. It doesn't. `binius_ip::sumcheck::verify` *recovers* each round polynomial's missing coefficient from the running sum rather than checking against it, so a wrong claim can't fail inside the sumcheck — the discrepancy rides through to the reduction's final evaluation check, the same place it surfaced before.

So the value is what you actually said: the prover proves the claim it was handed rather than one derived from its own buffers, and the O(2^15) inner product goes away. No change to where a bad witness is caught.

Found this because I first added a `debug_assert_eq!(sum, inner_product_buffers(&g, &h))` and `test_prove_verify_rejects_violated_zero_constraint` promptly failed — that test proves an intentionally violated witness, where the two *should* disagree. The assert was asserting the very property the reduction exists to establish, so it's gone.

PR #2071, opened 2026-08-10
BINIUS-418: commit one logUp* pushforward oracle per table

benji-potamus:
Last of three PRs for [BINIUS-418](https://linear.app/jimpo/issue/BINIUS-418/logup-star-multi-table-batch-lookup). #2067 and #2069 have merged, so this is now based on `main` and is the whole remaining diff.

#2069 made the bare reduction multi-table but left the committed layer pinned to one pushforward. This lifts that: `binius_iop::logup_star` and `binius_iop_prover::logup_star` take a list of tables and commit **one `Y` oracle per table**, in table order, before the reduction — so the per-table logUp challenges bind every commitment. Each relation `<Y, eq_r> = Y(r)` is opened at that table's own prefix of the shared reduced point.

One oracle per table is what the issue asked for ("for now, each pushforward can be a separate oracle"); a single concatenated oracle is left for later, noted in both docstrings.

Both layers take the reduction's own `TableLookup` rather than a bespoke type — with `gamma` now a plain argument, there was nothing left for a separate struct to carry. They return the same `Vec<LogupTableOutput>` too, rather than re-wrapping it.

`intmul` stays single-table — every limb column reads the one shared power table — so it passes a one-element list and destructures the single `LogupTableOutput`. Its transcript is unchanged, which `size_tracking_matches_real_proof_bytes` confirms.

## Tests

The committed round-trip helper is parameterized by `[(table_n_vars, [looker_n_vars])]`, the same shape as the bare one, so every prior case runs through it.

The one worth pointing at is **`test_basefold_round_trip`, now over two tables of different sizes** (`m = 2` and `m = 4`) with three lookers spread across them at three different lengths. That exercises the real FRI path with two *masked* ZK oracles of unequal message length, opened at different prefixes of the same reduced point — the end-to-end check that the padding conventions on both sides of the stack line up. It is the strongest evidence in the stack that the whole thing composes, since nothing about it is mocked.

Also added `test_multi_table_committed_round_trip` over four shapes on the naive channel.

## Verification

- `cargo test` green on `binius-ip`, `binius-ip-prover`, `binius-iop`, `binius-iop-prover`, `binius-prover`, `binius-verifier`, `binius-spartan-prover`, `binius-spartan-verifier`, re-run after the rebase onto merged `main`.
- `cargo clippy --all --tests --benches --examples -- -D warnings` clean.

### DIFF
--- a/crates/prover/src/protocols/intmul/prove.rs
+++ b/crates/prover/src/protocols/intmul/prove.rs
@@ -292,7 +292,16 @@ where
 			})
 			.collect::<Vec<_>>();
 		let log_cols = log2_ceil_usize(N_LIMB_COLUMNS);
-		let logup_proof = logup_star::prove(table, &lookers, self.channel, self.alloc);
+		// Every limb column reads the one shared power table, so the reduction runs over a
+		// single-table list.
+		let logup_proof = logup_star::prove(
+			[logup_star::TableLookup { table, lookers }],
+			self.channel,
+			self.alloc,
+		);
+		let [table_output] = logup_proof.tables.as_slice() else {
+			unreachable!("the reduction runs over the one power table")
+		};
 
 		// The index entries are the GF(2)-linear embeddings iota(e) = Σ_u basis(u) · bit_u(e),
 		// materialized by a table of all 2^LIMB_BITS embeddings.
@@ -311,7 +320,7 @@ where
 		// Collapse the per-column claims into a single claim on the eq(ρ)-folded column V by
 		// sampling ρ, so the final unification runs over the content variables only.
 		let rho = self.channel.sample_many(log_cols);
-		let mut padded_column_evals = logup_proof.index_eval_claims.clone();
+		let mut padded_column_evals = table_output.index_eval_claims.clone();
 		padded_column_evals.resize(1 << log_cols, F::ZERO);
 		let folded_index_claim =
 			evaluate(&FieldBuffer::<P>::from_values(&padded_column_evals), &rho);
--- a/crates/verifier/src/protocols/intmul/verify.rs
+++ b/crates/verifier/src/protocols/intmul/verify.rs
@@ -8,7 +8,7 @@ use binius_field::{BinaryField, BinaryField1b, ExtensionField, Field, field::Fie
 use binius_iop::{channel::IOPVerifierChannel, logup_star};
 use binius_ip::{
 	channel::IPVerifierChannel,
-	logup_star::LookerClaim,
+	logup_star::{LookerClaim, TableLookup},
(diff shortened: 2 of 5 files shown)


crates/verifier/src/protocols/intmul/verify.rs
```diff
@@ -283,16 +283,27 @@ where
 		})
 		.collect::<Vec<_>>();
 	let log_cols = log2_ceil_usize(N_LIMB_COLUMNS);
-	let logup_proof = logup_star::verify::<F, C>(LIMB_BITS, &looker_claims, channel)?;
+	// Every limb column reads the one shared power table, so the reduction runs over a single-table
+	// list and returns a single table claim.
+	let logup_proof = logup_star::verify::<F, C>(
+		&[TableLookup {
```
jimpo: Pass by value. It should take an impl IntoIterator

crates/prover/src/protocols/intmul/prove.rs
```diff
@@ -292,7 +292,16 @@ where
 			})
 			.collect::<Vec<_>>();
 		let log_cols = log2_ceil_usize(N_LIMB_COLUMNS);
-		let logup_proof = logup_star::prove(table, &lookers, self.channel, self.alloc);
+		// Every limb column reads the one shared power table, so the reduction runs over a
+		// single-table list.
+		let logup_proof = logup_star::prove(
+			&[logup_star::TableLookup { table, lookers }],
```
jimpo: Pass by value. it should take an impl IntoIterator

jimpo (changes requested):

crates/verifier/src/protocols/intmul/verify.rs
benji-potamus: Done — `binius_iop::logup_star::verify` now takes `impl IntoIterator<Item = TableLookup<'a, C::Elem>>` by value, and `intmul` passes the array directly.

It materializes once inside, since the tables are walked twice (receiving the commitments, then opening them). The variable counts are pulled out into a small `Vec` before the reduction consumes the tables, so no clone is needed for the opening pass.

crates/prover/src/protocols/intmul/prove.rs
benji-potamus: Done — same shape on the prover: `impl IntoIterator<Item = TableLookup<'a, P>>` by value, materialized once inside since the tables are walked three times (building the witnesses, running the reduction, opening the commitments).

Both layers now match the by-value convention from #2067, so every entry point in the stack — `verify_pushforward`, `prove_pushforward`, `verify_reduction`, `prove`, and both committed wrappers — takes its tables the same way.

jimpo (approved):

PR #2072, opened 2026-08-10
BINIUS-415: classify the composition of two shifts

benji-potamus:
Closes [BINIUS-415](https://linear.app/jimpo/issue/BINIUS-415/derive-the-shift-composition-table-by-brute-force-and-classify-every), Stage 1 of [BINIUS-408](https://linear.app/jimpo/issue/BINIUS-408/double-shifted-value-indices).

A second shift slot needs a merge rule, and the one the compiler has today cannot serve as one. `constraint_builder::Shift::compose` returns `None` both for pairs that *do* collapse to a single shift and for pairs whose composition is the **zero map**, so a caller cannot tell "keep both" from "drop the term". That is harmless while `None` only means "don't inline", but not once there is a slot to fill.

Give composition three outcomes:

```rust
pub enum Composition {
    Single(Shift),  // collapses to one shift, possibly the identity
    Zero,           // clears the word — the term contributes nothing
    Pair,           // genuinely needs both, in the order given
}
```

The answer is derived from what the shifts do to the bits, not from arithmetic over variants and amounts. A shift moves each output bit to at most one input bit (`Sar` included — the saturating positions all read the sign bit), so a shift *is* a `[u8; 64]` map and composition is indexing one map through the other. Identifying the result against the 377 canonical shifts is then a hash lookup.

That derivation catches what amount arithmetic misses:

| pair | today | derived |
|---|---|---|
| `Sar(5)` then `Sar(60)` | `None` (65 ≥ 64) | `Single(Sar(63))` |
| `Sll(40)` then `Sll(30)` | `None` | **`Zero`** |
| `Rotr(7)` then `Rotr(57)` | `Rotr(0)` | `Single(IDENTITY)` |
| `Srl(3)` then `Sll(3)` | `None` | `Pair` |

## Scope call I'd like checked

The issue asked for the `Pair` case to come back **canonicalized**, because the same linear map has up to 94 distinct `(inner, outer)` spellings — 108,571 irreducible spellings denote only 74,341 maps. **I did not implement that**, and I want that visible rather than buried.

Canonicalizing needs a map → canonical-pair table over all 74,341 maps: roughly 5 MB of lazily-initialized static in `binius-core`, paid by any process that composes a single shift. What it buys is ~1.46× fewer entries in the prover's shift alphabet — and only when a circuit actually derives the same map two different ways, which is rarer than the aggregate ratio suggests, on an alphabet that is a few dozen entries in practice. Trading a multi-megabyte static for that looked like the wrong side of "prefer simple logic", so I stopped at the classification, which is what the compiler needs to merge correctly.

If you want canonical pairs, it's an additive follow-up: the table is derived from the same `bit_map`, and `Composition::Pair` gains a payload. Happy to do it — I'd just rather it be a deliberate call than something I slipped in.

## Tests

Four, all exhaustive over the alphabet rather than sampled:

- `bit_map_matches_applying_the_shift` — pins the map to the real word operation by one-hot probing every input bit of every shift. This is the load-bearing one: everything else is derived from the map, so if it is wrong the rest is confidently wrong.
- `the_canonical_alphabet_has_no_two_spellings_of_one_shift` — all 377 shifts denote distinct maps, which is what lets a map identify a shift and therefore what makes `Single` well-defined.
- `compose_agrees_with_applying_both_shifts` — for all 377² ordered pairs, the classification is checked against actually applying both shifts to probe words.
- `compose_classifies_the_whole_alphabet_the_way_the_derivation_does` — the (23,046 / 10,512 / 108,571) split, as a regression. These numbers came out of an independent Python derivation during the BINIUS-408 design pass, and this Rust implementation reproduces them exactly.

`cargo check --workspace --all-targets`, clippy `-D warnings`, `cargo doc --no-deps --document-private-items` and fmt all clean; 103 core tests pass.

### DIFF
--- a/crates/core/src/constraint_system/shift.rs
+++ b/crates/core/src/constraint_system/shift.rs
@@ -168,15 +168,6 @@ impl ShiftVariant {
 		)
 	}
 
-	/// Whether this variant wraps the bits it moves out, rather than discarding them.
-	///
-	/// A cyclic variant loses nothing, so any two of its shifts compose however far they move;
-	/// every other variant drops what it carries past the end.
-	#[inline]
-	pub const fn is_cyclic(self) -> bool {
-		matches!(self, ShiftVariant::Rotr | ShiftVariant::Rotr32)
-	}
-
 	/// The exclusive upper bound on a valid shift amount for this variant.
 	///
 	/// - Half-word (`*32`) variants read only the lower 5 bits, so amounts run `0..32`.
@@ -342,10 +333,11 @@ impl Shift {
 	///
 	/// Panics if the amount is not below the variant's [`max_amount`](ShiftVariant::max_amount):
 	/// 32 for the half-word (`*32`) variants, 64 for the rest.
-	pub const fn new(variant: ShiftVariant, amount: usize) -> Self {
-		// A const context cannot format, so the amount and variant are left to the panic location
-		// rather than spelled into the message.
-		assert!(amount < variant.max_amount(), "shift amount out of range for this variant");
+	pub fn new(variant: ShiftVariant, amount: usize) -> Self {
+		assert!(
+			amount < variant.max_amount(),
+			"shift amount n={amount} out of range for {variant:?}"
+		);
 		Self {
 			variant,
 			// An amount below 64 always fits in the byte-sized field.
@@ -357,7 +349,7 @@ impl Shift {
 	///
 	/// # Panics
 	/// Panics if the shift amount is greater than or equal to 64.
-	pub const fn sll(amount: usize) -> Self {
+	pub fn sll(amount: usize) -> Self {
 		Self::new(ShiftVariant::Sll, amount)
 	}
 
@@ -365,7 +357,7 @@ impl Shift {
 	///
 	/// # Panics
 	/// Panics if the shift amount is greater than or equal to 64.
-	pub const fn srl(amount: usize) -> Self {
+	pub fn srl(amount: usize) -> Self {
 		Self::new(ShiftVariant::Slr, amount)
 	}
 
@@ -376,7 +368,7 @@ impl Shift {
 	///
 	/// # Panics
 	/// Panics if the shift amount is greater than or equal to 64.
-	pub const fn sar(amount: usize) -> Self {
+	pub fn sar(amount: usize) -> Self {
 		Self::new(ShiftVariant::Sar, amount)
 	}
 
@@ -386,7 +378,7 @@ impl Shift {
 	///
 	/// # Panics
 	/// Panics if the shift amount is greater than or equal to 64.
-	pub const fn rotr(amount: usize) -> Self {
+	pub fn rotr(amount: usize) -> Self {
 		Self::new(ShiftVariant::Rotr, amount)
 	}
 
@@ -396,7 +388,7 @@ impl Shift {
 	///
 	/// # Panics
 	/// Panics if the shift amount is greater than or equal to 32.
-	pub const fn sll32(amount: usize) -> Self {
+	pub fn sll32(amount: usize) -> Self {
 		Self::new(ShiftVariant::Sll32, amount)
 	}
 
@@ -406,7 +398,7 @@ impl Shift {
 	///
 	/// # Panics
 	/// Panics if the shift amount is greater than or equal to 32.
-	pub const fn srl32(amount: usize) -> Self {
+	pub fn srl32(amount: usize) -> Self {
 		Self::new(ShiftVariant::Srl32, amount)
 	}
 
@@ -417,7 +409,7 @@ impl Shift {
 	///
 	/// # Panics
 	/// Panics if the shift amount is greater than or equal to 32.
-	pub const fn sra32(amount: usize) -> Self {
+	pub fn sra32(amount: usize) -> Self {
 		Self::new(ShiftVariant::Sra32, amount)
 	}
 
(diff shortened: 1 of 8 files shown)


jimpo (changes requested):
I understand the motivation but not the bitmap implementation. We should be able to do this with simple casework on the pairs of variants. The only ones that commute (I believe) are when the two variants are the same or SRL followed by SRA where the SRL has a >0 amount.

benji-potamus:
Taking the casework direction — the bitmap belongs in a test, not in the shipped path. But before I write it: I ran your enumeration against the exhaustive derivation, and it's incomplete. Of the 23,046 pairs that collapse to a single shift, your cases cover 17,402 and miss **5,644**. Worth knowing, because it decides how big the casework actually is.

Your cases, confirmed: identity on either side (753), same variant (14,696), and `Slr(a>0)` then `Sar` (1,953).

The missing 5,644 fall into four families, all structured:

**1. Half-word meets full-width, once the full-width shift has cleared the halfway point** — 3,937 pairs. Shifting left by ≥32 leaves the low half all zeros, so a following half-wise left shift acts like a full one:

```
Sll(32)   then Sll32(1)  -> Sll(33)          Sll32(1)  then Sll(32) -> Sll(33)
Slr(32)   then Srl32(1)  -> Slr(33)          Srl32(1)  then Slr(32) -> Slr(33)
Sar(32)   then Sra32(1)  -> Sar(33)          Sra32(1)  then Sar(32) -> Sar(33)
Slr(33)   then Sra32(1)  -> Slr(34)          Srl32(1)  then Sar(32) -> Slr(33)
```

**2. Your `Slr → Sar` rule, in the half-word width** — 465 pairs. `Srl32(1)` then `Sra32(1)` is `Srl32(2)`, for the same reason: the logical shift zeroes each half's sign bit, so the arithmetic one acts logically. This is really your case 4 stated per width family rather than for `Slr`/`Sar` specifically.

**3. A saturated word absorbs any rotation** — 94 pairs. `Sar(63)` makes every bit the sign bit, so the word is all-zeros or all-ones and any rotation is a no-op: `Sar(63) then Rotr(k) -> Sar(63)`, likewise `Rotr32`. Same for `Sra32(31) then Rotr32(k)`.

**4. Keeping only the bit an arithmetic shift preserves** — 188 pairs. `Slr(63)` keeps just bit 63, which `Sar` never disturbs, so `Sar(a) then Slr(63) -> Slr(63)` for every `a`. Same for `Sra32(a) then Srl32(31)`.

So the closed form is about five rule families / ~15 match arms rather than two — still very writable, just bigger than the shape suggests.

Two notes on how complete it needs to be. Missing a collapse is **conservative, not incorrect**: reporting `Pair` where `Single` would do costs a shift slot, never a wrong answer. So we can deliberately trim families 3 and 4 (282 pairs, all degenerate — a saturated source or a single surviving bit) if the arms aren't worth their weight; I'd keep 1 and 2 since `Sll(32) then Sll32(k)` is the kind of thing an inlined `*32` gadget will actually produce.

I'll implement all five families, with the exhaustive derivation kept as a `#[cfg(test)]` oracle checking the closed form on all 377² pairs — so the casework is verified rather than believed, and nothing bitmap-shaped ships. Say the word if you'd rather I drop families 3 and 4.

benji-potamus:
Rewritten as casework — pushed `294805b`. Nothing bitmap-shaped ships; the whole thing is now three small functions:

- **`chained(inner, outer) -> Option<(ShiftVariant, usize)>`** — do the two move bits the same way, and how far in total. Same variant always chains; then your `Slr → Sar` rule (in both widths); then the eight half-word/full-width pairs.
- **`degenerate(inner, outer) -> Option<Shift>`** — the two flattening cases: a saturated word absorbs a rotation, and `Slr(63)` keeps the one bit an arithmetic shift never disturbs.
- **`chained_composition(variant, distance)`** — one place for the overflow rules: logical shifts clear the word past the width, arithmetic ones saturate at the sign, rotations wrap.

That last split is what made the casework smaller than I expected. `Single` and `Zero` are not separate analyses — the same chaining families produce a direction and a distance, and the direction's own overflow rule decides which you get. So the 10,512 zero pairs need no rules of their own.

Guard values worth a second look, since they're the parts I'd most expect to be wrong:

- `Slr → Sra32` needs `inner.amount >= 33`, not `>= 32`. At 32 the original high half lands in the low half with its top bit intact, so the half-wise arithmetic shift still has a sign to replicate. Every other cross-width pair needs `>= 32`.
- `Sra32 → Sar` and `Sar → Sra32` need no upper bound, because saturation absorbs any excess; the left/logical-right pairs do.

Verification is the exhaustive part now: `compose_matches_the_bit_level_model` builds the independent bit-level model in `#[cfg(test)]` and checks the closed form against it on all 377² = 142,129 pairs. It passes, so the rules above are right and complete — including the four families your enumeration didn't cover. `bit_map_matches_applying_the_shift` pins the model itself to the word operations by one-hot probing, so the oracle isn't self-certifying. The (23,046 / 10,512 / 108,571) count regression stays.

Two gate notes: `degenerate` had to become `const fn` (`clippy::missing_const_for_fn`), and the doc comment can't link `chained`/`degenerate` by intra-doc link since they're private and `compose` is public — `rustdocflags` has `-Dwarnings`, so that's a hard error rather than a lint.

The canonicalization question from the original PR description is unchanged and still open: `Pair` comes back as spelled, not canonicalized.

jimpo (approved):

PR #2076, opened 2026-08-10
BINIUS-416: unify the frontend shift algebra with the core Shift type

benji-potamus:
Closes [BINIUS-416](https://linear.app/jimpo/issue/BINIUS-416/unify-the-frontend-shiftshiftkind-with-the-core-shift-type), the last of Stage 1 for [BINIUS-408](https://linear.app/jimpo/issue/BINIUS-408/double-shifted-value-indices).

**Stacked on #2072** — it consumes `Shift::compose`. Base is the 415 branch; I'll retarget to main once that lands.

The compiler carried its own `Shift` and `ShiftKind` alongside core's, differing only in spelling the identity separately and in discriminant order. Both are gone: the frontend uses `binius_core::constraint_system::Shift` throughout, and `ShiftedWire::to_shifted_value_index` drops from a nine-arm translation to a field copy. `ShiftSummary` re-keys on `ShiftVariant`, and `patch.rs` takes its merge rule from core's `Composition`. Net −89 lines.

## The one behavior change, measured

Collapsing `Shift::None` into `Sll(0)` is not cosmetic. `expr::sll(w, 0)` builds an unnormalized zero-amount shift, and `ShiftSummary` used to treat that as pinning a path to the `Sll` variant — so `sll(x, 0)` feeding `srl(y, 0)` forced a commit, despite both being the identity. Unified, `is_identity()` covers both spellings and the pair composes.

`test_zero_shift_composition` asserted the old behavior, with the comment "Zero shifts still carry their type; mixed types do not compose → commit". It now asserts the corrected one. That test is the *only* thing in the tree that noticed:

| circuit | ZERO | AND | vs. before |
|---|---|---|---|
| sha256 | 2,005 | 6,575 | unchanged |
| keccak | 0 | 4,783 | unchanged |
| blake3 | 1,997 | 2,760 | unchanged |

So the wart was real but unreachable from the gadgets we actually build — no snapshot movement, and the correctness argument stands on its own rather than on a constraint-count win.

## Notes for review

- **`Composition::Zero` drops the term rather than panicking** ([jimpo's review](https://github.com/binius-zk/binius64/pull/2076#discussion_r3752726045)). An operand is an XOR, so a term the two shifts clear contributes nothing. `Pair` keeps the panic and is now the only arm meaning `commit_set` and `patch` disagree about what is inlinable.
- **`patch.rs` adopting core's stronger `compose` is invisible**, which is worth stating because it looks like it shouldn't be. `commit_set` decides what may be inlined and `patch` only executes it, so `commit_set` stays the binding constraint and the extra collapses core knows about are never reached.
- **The `ShiftSummary` proptests are now one-directional.** They asserted the summary answers *exactly* as the collapsed chain does; against core's rule that is false, because the summary tracks a total distance rather than the composition — `sar(40)` then `sar(40)` saturates to `sar(63)`, but the summary sees 80 and refuses. The honest invariant is soundness: whatever the summary permits really does compose. That is also precisely what `patch::process_term`'s remaining panic rests on, so the test now pins the property the code depends on rather than a stronger one that happened to hold.
- **Core's `Shift` and `ShiftedValueIndex` constructors became `const fn`**, so the frontend's `expr::*` and `gate/shift.rs` stay `const`. The cost is that `Shift::new`'s panic message can no longer interpolate the amount and variant — a const context cannot format — so it is a static string and the values are left to the panic location.
- `ShiftVariant` gains `is_cyclic()`; `width()` was already there as `max_amount()`.

## Verification

`cargo check --workspace --all-targets`, clippy `-D warnings` over core + frontend, `cargo doc --no-deps --document-private-items`, fmt — all clean. Tests: core 104, frontend 197 + 2 integration + doctest. Constraint counts measured above.

### DIFF
--- a/crates/frontend/src/compiler/gate_fusion/patch.rs
+++ b/crates/frontend/src/compiler/gate_fusion/patch.rs
@@ -1,12 +1,14 @@
 // Copyright 2026 The Binius Developers
 // Copyright 2025 Irreducible Inc.
 
+use binius_core::constraint_system::{Composition, Shift};
+
 use super::legraph::LeGraph;
 use crate::compiler::{
 	Wire,
 	constraint_builder::{
-		ConstraintBuilder, Shift, ShiftedWire, WireAndConstraint, WireBmulConstraint,
-		WireImulConstraint, WireLinearConstraint, WireOperand,
+		ConstraintBuilder, ShiftedWire, WireAndConstraint, WireBmulConstraint, WireImulConstraint,
+		WireLinearConstraint, WireOperand,
 	},
 	gate_fusion::legraph::ConstraintRef,
 };
@@ -239,12 +241,16 @@ fn process_term(
 			// Compose shifts: we're applying 'shift' to 'inner_term'
 			// So we need Shift::compose(inner_term.shift, shift)
 			match Shift::compose(inner_term.shift, shift) {
-				Some(composed_shift) => {
+				Composition::Single(composed_shift) => {
 					// Recursively process this term with the composed shift
 					process_term(cb, leg, new_operand, subsumes, inner_term.wire, composed_shift);
 				}
-				None => {
-					// Incompatible shifts - this shouldn't happen if commit set is correct
+				// The two shifts clear the word, so the term is identically zero. An operand is
+				// an XOR, so a zero term contributes nothing and the inlining simply drops it.
+				Composition::Zero => {}
+				// Needing both slots means the commit set inlined a definition it should have
+				// committed: the two passes disagree about what is inlinable.
+				Composition::Pair => {
 					panic!(
 						"Incompatible shifts during inlining: {:?} followed by {:?} for wire {:?}",
 						inner_term.shift, shift, inner_term.wire
@@ -336,11 +342,11 @@ mod tests {
 				vec![
 					ShiftedWire {
 						wire: w(0),
-						shift: Shift::Rotr(0),
+						shift: Shift::IDENTITY,
 					},
 					ShiftedWire {
 						wire: w(1),
-						shift: Shift::Rotr(0),
+						shift: Shift::IDENTITY,
 					},
 				],
 			)],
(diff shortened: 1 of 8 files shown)


crates/frontend/src/compiler/gate_fusion/patch.rs
```diff
@@ -239,12 +241,14 @@ fn process_term(
...
+				Composition::Single(composed_shift) => {
 					// Recursively process this term with the composed shift
 					process_term(cb, leg, new_operand, subsumes, inner_term.wire, composed_shift);
 				}
-				None => {
-					// Incompatible shifts - this shouldn't happen if commit set is correct
+				// The commit set only inlines a definition whose shifts compose into one, so a
+				// term that needs both slots — or that the two shifts clear outright — means the
+				// two passes disagree about what is inlinable.
+				Composition::Zero | Composition::Pair => {
```
jimpo: Seems OK to inline in the Zero case -- inlining a zero just removes the term, right?

benji-potamus: Yes — an operand is an XOR, so a term that is identically zero contributes nothing and dropping it is exactly right. Done in `0d5c744`: `Zero` now falls through without pushing the term, in both `process_term` and `expand_term_recursive`. `Pair` keeps the panic, which is now the only arm that means the two passes disagree.

I'd convinced myself the arm was unreachable and left it panicking on that basis, which was the weaker call. The argument does hold today — `ShiftSummary::composable` requires `total + amount < width` for non-cyclic variants, and rejects outright once a path carries two variants, so a chain it permits can never compose to zero. But that is a reachability argument resting on two passes agreeing, and it stops holding the moment `commit_set` gets more permissive — which is exactly what the gate-fusion half of BINIUS-408 will do. Dropping the term is correct either way and needs no such argument.

No behavior change today (the arm is unreachable, so nothing hits it), and an empty operand is already well-defined as the constant zero — `WireOperand::is_empty` says as much, and `ZeroConstraint` with an empty operand is the trivially satisfied `0 = 0`.

jimpo (approved):

PR #2078, opened 2026-08-10
BINIUS-426: build the M4 fixtures with inout wires

benji-potamus:
Closes [BINIUS-426](https://linear.app/jimpo/issue/BINIUS-426/m4-build-the-test-and-bench-circuits-with-inout-wires-instead-of-force). Follow-up to [BINIUS-424](https://linear.app/jimpo/issue/BINIUS-424) (#2077), now rebased on [#2079](https://github.com/binius-zk/binius64/pull/2079) which added `CircuitBuilder::mark_inout`.

Every circuit fixture in `binius-m4-prover` allocated its inputs as witness wires and kept its outputs alive with `force_commit`. That was forced on them: the batch witness table rejected inout wires, so a fixture had no way to declare public inputs and outputs. #2077 lifted the restriction.

## The change

Build the fixtures the way a real circuit is built: the inputs are inout wires and each gadget output is promoted to one with `mark_inout`, which is what keeps the gadget alive under dead-code elimination. **`force_commit` now has no use anywhere in the crate** — all 51 calls are gone.

## It costs nothing

An earlier revision of this PR declared separate inout outputs and `assert_eq`'d the gadget against them. That was measurably worse, and prompted #2079. Promoting instead lands exactly on the `force_commit` baseline:

| circuit | committed words | AND | ZERO |
|---|---|---|---|
| `keccak_f1600` | 625 → **625** | 600 → **600** | 0 → 0 |
| `blake3_compress_2x` | 596 → **596** | 336 → **336** | 232 → **232** |
| `sha256_compress_2x` | 934 → **934** | 728 → **728** | 182 → **182** |

Every count is unchanged from the `force_commit` form, while 50 / 36 / 32 words respectively are now public. The assert form would have cost Keccak 25 words and 25 AND constraints — 4% of the circuit — because its lanes leave the chi step as AND outputs that must be committed regardless.

So the bench circuits are byte-for-byte the same shape they were; **bench numbers remain comparable across this commit**.

## What this buys

The operand columns now resolve inout indices through their own `ValueSegment::InOut` arm, which no fixture reached before — #2077 added that arm covered only by its own round trip.

## Note on the reference implementations

The intermediate assert-based revision filled the public outputs from reference implementations (`ref_compress`, the `keccak` crate), which incidentally checked each gadget against its reference. Promotion derives the outputs from the gate instead, so those fills — and the `keccak` dev-dependency — are gone. No coverage is lost: `binius-circuits` already tests these gadgets against references itself (`compress_matches_reference` for BLAKE3, and the Keccak gadget against the `sha3` crate). Verifying hash gadgets is that crate's job, not the M4 prover's.

## Tests

No new tests; this rebuilds existing fixtures. All 44 unit tests, the 6 hash-primitive integration tests, clippy and the full `prek` hook set pass, and the benches build.

### DIFF
--- a/crates/m4-prover/benches/assemble_operation_witness.rs
+++ b/crates/m4-prover/benches/assemble_operation_witness.rs
@@ -24,19 +24,19 @@ const LOG_INSTANCES: usize = 13;
 /// The number of 64-bit lanes in a Keccak-f1600 state.
 const STATE_LANES: usize = 25;
 
-/// Builds a circuit that applies one Keccak-f1600 permutation to a witness-input state and
-/// force-commits the permuted output words. Returns the circuit and the 25 input state wires.
+/// Builds a circuit that applies one Keccak-f1600 permutation to a public input state and promotes
+/// the permuted lanes to public outputs. Returns the circuit and the 25 input state wires.
 fn build_keccak_circuit() -> (Circuit, [Wire; STATE_LANES]) {
 	let builder = CircuitBuilder::new();
-	let input: [Wire; STATE_LANES] = array::from_fn(|_| builder.add_witness());
+	let input: [Wire; STATE_LANES] = array::from_fn(|_| builder.add_inout());
 
 	// Permute a copy of the input wires in place; `state` then holds the output wires.
 	let mut state = input;
 	keccak_f1600(&builder, &mut state);
 
-	// Pin the outputs so dead-code elimination keeps the whole permutation.
+	// Promoting the permuted state keeps the whole permutation alive under dead-code elimination.
 	for wire in state {
-		builder.force_commit(wire);
+		builder.mark_inout(wire);
 	}
 
 	(builder.build(), input)
--- a/crates/m4-prover/benches/blake3_compressions.rs
+++ b/crates/m4-prover/benches/blake3_compressions.rs
@@ -33,12 +33,12 @@ const DEFAULT_LOG_INV_RATE: usize = 1;
 /// Compressions computed per instance: the two-lane core runs two at once.
 const COMPRESSIONS_PER_INSTANCE: u64 = 2;
 
-/// The witness input wires of one two-lane BLAKE3 compression instance.
+/// The public wires of one two-lane BLAKE3 compression instance.
 ///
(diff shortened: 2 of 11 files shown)


benji-potamus:
⚠️ **Hold off reviewing this as final** — a measured circuit-size regression turned up, and there may be a better mechanism. Full analysis on [BINIUS-426](https://linear.app/jimpo/issue/BINIUS-426/m4-build-the-test-and-bench-circuits-with-inout-wires-instead-of-force); summary here.

I measured each `prove_hash_primitives` circuit both ways. "Committed" is `n_inout + n_private`, which is exactly what the M4 table stores per instance:

| circuit | committed | AND | ZERO |
|---|---|---|---|
| `blake3_compress_2x` | 596 → **596** | 336 → 344 | 232 → 224 |
| `sha256_compress_2x` | 934 → **934** | 728 → 736 | 182 → 174 |
| `keccak_f1600` | 625 → **650** (+4.0%) | 600 → **625** (+4.2%) | 0 → 0 |
| `imul` | 4 → **6** (+50%) | 0 → 2 | 0 → 0 |

The assert costs a committed word exactly when the gadget output is a wire that must be committed regardless — a non-linear gate output. Keccak's lanes come out of the chi AND and IMUL's `hi`/`lo` are IMUL outputs, so the wire stays committed *and* an inout word is added beside it. BLAKE3 and SHA-256 end on XOR, so gate fusion inlines the output into the assert and nothing grows.

The better mechanism is a `CircuitBuilder::mark_inout(wire)` that moves an existing gate output from the internal segment to the inout segment, instead of allocating a fresh inout wire and asserting equality. That looks straightforward: wire kind is just an entry in `graph.wires`, and the routing loop in `compiler/mod.rs:318-345` is the only thing that reads it for placement. Liveness comes free, since `dedup_gates` already treats an inout output as observable (`cse.rs:84`) — the same test `force_commit` satisfies. It would put Keccak back at 625 words / 600 AND constraints, identical to the `force_commit` baseline, while still making the output public.

Awaiting a call on whether to hold this PR for `mark_inout` or land it and convert afterwards. Not merging either way.

jimpo (approved):

PR #2083, opened 2026-08-10
BINIUS-427 (2/2): add the WordIP channel traits

benji-potamus:
Second of two PRs for [BINIUS-427](https://linear.app/jimpo/issue/BINIUS-427/add-wordip-channel-traits-and-generalize-the-iop-verifier-stack-over), split per review. #2088 landed the BaseFold `Elem` generalization, so this is now rebased onto `main` and stands alone.

#2088 let the BaseFold and FRI verifiers run over a symbolic element type. This one does the same for the words a protocol carries.

## The trait pair

`WordIPVerifierChannel` (binius-ip) and `WordIPProverChannel` (binius-ip-prover), each with a `Word` associated type: `Word` for a channel over concrete values, a wire type for one that builds a circuit. Both crates gain a `binius-core` dependency, which the architecture allows.

Per review, lifting and index arithmetic are bounds on the associated type rather than channel methods:

```rust
type Word: Clone + From<Word> + Shr<u32, Output = Self::Word>;
```

so protocol code writes `word.into()` and `index >> n` whatever the channel is, and a symbolic channel folds those over its builder the way `CircuitElem` does for arithmetic. That removed 20 method impls across ten channels.

The remaining three operations do need the channel, since they either touch the Fiat-Shamir state or emit a sub-circuit: `observe_words`, `subset_sum` and `select`.

## Why the words have to be symbolic

Query indices come from `sample_bits`, so they cannot stay `usize`. They do not stay inside the Merkle channel either — they escape into `fri` in five places:

* `*index >>= arity` and `terminate_codeword[index]` in `fri/verify.rs`
* `index >> log_lift` and `index >> coset_log_size` before `recv_openings`, in `fri/batch.rs`
* `coset[index & coset_mask]` in `FRIOracle::reduce_queries`
* `chunk_index` feeding `DomainContext::twiddle` in `fold_coset`

The `Shr` bound covers three, `select` the fourth, `subset_sum` the fifth.

`subset_sum` is what a twiddle at a symbolic block index needs. `DomainContext::twiddle` is F2-linear in that argument and vanishes at zero, so the twiddle of a block is the sum of the twiddles of its individual bits. The basis is read through the trait as `twiddle(layer, 1 << bit)` rather than reaching into `GenericOnTheFly`'s subspace representation.

The same word problem appears one layer up in `shift::check_eval`, which is generic over `IPVerifierChannel` rather than the Merkle channel — hence the traits sit below `MerkleIP*` rather than on it.

## Two consequences worth reviewer attention

**The transcript changes for an odd `n_inout`.** `observe_words` replaces `encode_inout`, which is deleted. `encode_inout` packed two words into a `B128` serialized with `put_u128_le`, so for an even count the byte stream was already the words at eight little-endian bytes each and is unchanged. An odd count left a half-element with eight bytes of zero padding; observing words directly drops it. Prover and verifier move together, and the example snapshots hold circuit statistics rather than proof bytes, so nothing needs reblessing.

**Three items became dead code.** `ProxTestOracle::log_len`, `FRIOracle::log_len` and `BrakedownOracle::depth` served only the index-range preconditions, which cannot be checked against a symbolic word. Those asserts are gone, so the items are too (CI runs `-D warnings`), and the guarantee moves onto `sample_bits`, which masks its result to the requested width. `BrakedownOracle::new` loses its `depth` parameter as a result.

## Scope boundary

`IOPVerifier::verify` is pinned to `WordIPVerifierChannel<B128, Word = Word>`, and `IOPProver::prove` to the prover equivalent, so `observe_words` type-checks against the still-native `&[Word]` parameter. Relaxing that to a free associated type is [BINIUS-433](https://linear.app/jimpo/issue/BINIUS-433/take-the-inner-statement-symbolically-in-iopverifierverify), and is what actually lets a circuit-building channel run the verifier end to end.

Every channel over concrete values takes `type Word = Word` and the obvious implementation, so behaviour is unchanged: `subset_sum` is an XOR-sum over natively known bits emitting no constraints, `select` is an array index.

## Testing

Rebased onto `main` at `c7e9fc29`; no conflicts.

* `cargo check --workspace --all-targets`: clean, including every example.
* `cargo test --workspace --exclude binius-examples --exclude binius-arith-bench`: **1400 passed, 0 failed, 19 ignored** across 50 binaries.
* `cargo clippy --all-features --tests -- -D warnings`: clean on all eight crates touched.

`binius-examples` is excluded for disk, not correctness — linking its 328 example binaries needs ~6.3G my volume cannot spare; they all compile under `cargo check --all-targets`. `binius-arith-bench` does not build on aarch64 at stock `HEAD` and depends on nothing here.

### DIFF
--- a/crates/iop/src/fri/batch.rs
+++ b/crates/iop/src/fri/batch.rs
@@ -2,7 +2,8 @@
 
 use std::iter;
 
-use binius_field::{BinaryField, FieldOps};
+use binius_field::{BinaryField, FieldOps, util::expand_subset_sums};
+use binius_ip::channel::WordIPVerifierChannel;
 use binius_math::{line::extrapolate_line, multilinear, ntt::DomainContext};
 
 use crate::merkle_channel::MerkleIPVerifierChannel;
@@ -22,26 +23,22 @@ pub trait ProxTestOracle<F: BinaryField, E> {
 	/// The Merkle commitment handle for the committed oracle.
 	type Commitment;
 
-	/// The base-2 logarithm of the length of the virtual oracle.
-	///
-	/// The virtual oracle is defined by the committed oracle and the folding challenges. Indices
-	/// passed to [`Self::open_queries`] must lie in the range `0..2^self.log_len()`.
-	fn log_len(&self) -> usize;
-
 	/// Opens queried locations on the virtual oracle.
 	///
 	/// This has a batch interface for verifying multiple queries because opening multiple Merkle
 	/// tree locations at once amortizes the proof size.
 	///
 	/// ## Preconditions
-	/// The `indices` must lie in the range `0..2^self.log_len()`.
+	/// The `indices` must address the virtual oracle, which spans the committed tree depth
+	/// duplicated over the lift factor. The channel guarantees the bound by masking what
+	/// [`WordIPVerifierChannel::sample_bits`] returns.
 	///
 	/// ## Returns
 	/// The values of the virtual oracle at the queried indices. The virtual oracle is defined by
 	/// the committed oracle and the folding challenges.
 	fn open_queries<Channel>(
 		&self,
-		indices: &[usize],
+		indices: &[Channel::Word],
 		channel: &mut Channel,
 	) -> Result<Vec<E>, Error>
 	where
@@ -54,8 +51,6 @@ pub trait ProxTestOracle<F: BinaryField, E> {
 pub struct BrakedownOracle<E, C> {
 	challenges: Vec<E>,
 	commitment: C,
-	/// The depth of the committed Merkle tree.
-	depth: usize,
 	/// log2 the lift factor (oracle padding). The committed codeword is virtually duplicated
 	/// `2^log_lift` times to reach the common first-round length; a query at global index `k`
 	/// reads the committed codeword at `k >> log_lift`. Zero when no lifting is needed.
@@ -65,14 +60,13 @@ pub struct BrakedownOracle<E, C> {
 impl<E, C> BrakedownOracle<E, C> {
 	/// Constructs a new oracle from the committed interleaved codeword and the folding challenges.
 	///
-	/// `depth` is the depth of the committed Merkle tree. `log_lift` is the oracle-padding lift
-	/// factor (the committed codeword is virtually duplicated `2^log_lift` times to reach the
-	/// common first-round length); pass `0` when no lifting is needed.
-	pub const fn new(challenges: Vec<E>, commitment: C, depth: usize, log_lift: usize) -> Self {
+	/// `log_lift` is the oracle-padding lift factor (the committed codeword is virtually
+	/// duplicated `2^log_lift` times to reach the common first-round length); pass `0` when no
+	/// lifting is needed.
+	pub const fn new(challenges: Vec<E>, commitment: C, log_lift: usize) -> Self {
 		Self {
 			challenges,
 			commitment,
-			depth,
 			log_lift,
 		}
 	}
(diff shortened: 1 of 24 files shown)


jimpo (changes requested):
Split this into two PRs:

1. Remove the Elem = F bound for the BaseFold channels, allowing them to be symbolic over Elem.
2. The rest of the change with the WordIP channels

Also, instead of the `constant_word` and `shift_word` methods on `WordIP` channel, the associated Word type should have `From<Word>` and `Shr<usize>` type bounds. We'll implement this in the symbolic channel similarly to the Spartan builder channel.

crates/iop/src/fri/batch.rs
```diff
@@ -299,29 +274,58 @@ where
...
+		let shift = log_size - 1;
+		let chunk_basis = twiddle_basis[shift..]
+			.iter()
+			.copied()
+			.map(E::from)
+			.collect::<Vec<_>>();
+		let chunk_twiddle = channel.subset_sum(&chunk_basis, chunk_index);
+
+		for index_offset in 0..1 << shift {
+			let offset_twiddle = (0..shift)
```
jimpo: Just compute all (0..1 << shift) of them outside the loop using the same algorithm as expand_subset_xors.

jimpo (changes requested):

crates/iop/src/fri/batch.rs
benji-potamus: Done — the offset twiddles are now a table built once per round by doubling, so the twiddle work drops from one addition per set bit to one per entry:

```rust
let mut offset_twiddles = vec![F::ZERO; 1 << shift];
for (bit, &basis_bit) in twiddle_basis[..shift].iter().enumerate() {
    let (lo_half, hi_half) = offset_twiddles[..1 << (bit + 1)].split_at_mut(1 << bit);
    for (lo, hi) in iter::zip(lo_half, hi_half) {
        *hi = *lo + basis_bit;
    }
}
```

Written inline rather than calling `expand_subset_sums_array`, since `shift` is a runtime value and that helper is const-generic over `[P; N]`. Happy to add a slice-based sibling in `binius_field::util` if you'd rather have it shared — it just widens this PR past channels.

The `twiddle_basis` construction stays inside the per-challenge loop: `layer` decreases each round so the basis genuinely differs. Only the offset expansion was loop-invariant.

crates/iop/src/fri/batch.rs
```diff
@@ -299,29 +274,65 @@ where
...
 	E: FieldOps<Scalar = F> + From<F>,
 	DC: DomainContext<Field = F>,
+	Channel: WordIPVerifierChannel<F, Elem = E>,
 {
 	let mut log_size = challenges.len();
 	for challenge in challenges {
-		for index_offset in 0..1 << (log_size - 1) {
+		let layer = log_len - 1;
+		// The twiddle basis of this layer: one entry per bit of a block index.
+		let twiddle_basis = (0..layer)
```
jimpo: Can't you use `domain_context.subspace(layer)` instead?

crates/iop/src/fri/batch.rs
```diff
@@ -299,29 +274,65 @@ where
...
+			.iter()
+			.copied()
+			.map(E::from)
+			.collect::<Vec<_>>();
+		let chunk_twiddle = channel.subset_sum(&chunk_basis, chunk_index);
+
+		// Every offset's twiddle, indexed by the offset: entry `mask` is the sum of the basis
+		// elements its set bits select. Building the table by doubling, as `expand_subset_xors`
+		// does, costs one addition per entry rather than one per set bit.
+		let mut offset_twiddles = vec![F::ZERO; 1 << shift];
```
jimpo: Make this a function in `binius_field::util::expand_subset_sums` (like expand_subset_sums_array, but dynamically sized, returning a Vec)

jimpo (changes requested):

crates/iop/src/fri/batch.rs
benji-potamus: Nearly — it's `subspace(layer + 1).basis()[1..]`, not `subspace(layer)`. The leading element is the normalizing one that `twiddle` skips (`BinarySubspace::new_unchecked(&v[1..])` in `GenericOnTheFly::twiddle`).

I checked rather than reasoned it out, across all four `DomainContext` impls for layers 1..8:

```
GenericOnTheFly:       twiddle basis == subspace(layer + 1).basis()[1..]  ✓   subspace(layer)  ✗
GenericPreExpanded:    ✓ / ✗
GaoMateerOnTheFly:     ✓ / ✗
GaoMateerPreExpanded:  ✓ / ✗
```

Taken with that correction — it's now one `subspace` call per round plus a slice, where before it was one `twiddle` call per bit, each of which rebuilds a `BinarySubspace` internally and sums it.

Worth noting the relationship isn't stated in the `DomainContext` trait docs, only in the two methods' individual definitions in terms of Ŵ. It holds uniformly across the impls, so relying on it seems fine, but a line on the trait saying so would make it contractual rather than incidental.

crates/iop/src/fri/batch.rs
benji-potamus: Added as `binius_field::util::expand_subset_sums` — same doubling as `expand_subset_sums_array` but over a slice, returning a `Vec`, with a precondition that `elems.len() < usize::BITS`. `fold_coset` now just calls it, so the whole offset expansion is one line.

jimpo (approved):

PR #2085, opened 2026-08-10
BINIUS-435: constrain assert gates with ZERO constraints instead of AND

benji-potamus:
Closes [BINIUS-435](https://linear.app/jimpo/issue/BINIUS-435/frontend-constraint-assert-gates-with-a-zero-constraint-instead-of-and).

The frontend's `ConstraintBuilder` could only reach a ZERO constraint through a *linear* constraint, which names a destination wire. An assertion has no destination, so every assert gate paid an AND constraint against the all-ones (or MSB-one) constant instead.

Two commits:

**1. `assert_eq` / `assert_zero` — `ConstraintBuilder::zero`**

Adds a `zero()` method with its own bucket, and teaches gate fusion to treat a ZERO constraint as an inlining *root*: a linear definition feeding an assertion now inlines into it exactly as it inlines into an AND operand (`ConstraintRef::Zero`). Without that, converting the assertions would have stranded their operands' linear definitions and forced them to commit.

**2. `assert_true` / `assert_false` / `assert_non_zero`'s MSB check**

`sar(x, 63)` broadcasts the most-significant bit across the word — all-1 when set, 0 when clear — so the MSB test reads as a ZERO constraint. `assert_false` now names no constant at all; `assert_true` swaps `MSB_ONE` for the all-ones word the gate graph already seeds.

`assert_eq_cond` stays an AND: it is genuinely conditional (`(x ⊕ y) ∧ (cond ~>> 63) = 0`).

### Soundness note

`assert_non_zero`'s second constraint carried a comment explaining it was emitted as an AND *specifically* so gate fusion could not inline it and substitute the constant back into the carry constraint. That property is preserved: a ZERO constraint is an assertion and defines no wire, so fusion never reads it as a definition. `test_assert_non_zero_forge_zero_rejected` still passes.

### Constraint counts

Rebased onto `main` at #2084 (select → BMUL), which lowered AND counts across the board.

| example | AND before → after | padded AND | ZERO before → after |
|---|---|---|---|
| ethsign | 133,816 → 114,068 (−14.8%) | **2^18 → 2^17** | 2,173 → 21,921 |
| ec_msm | 132,481 → 112,695 (−14.9%) | **2^18 → 2^17** | 2,140 → 21,926 |
| bitcoin_p2pkh | 98,901 → 84,233 (−14.8%) | 2^17 → 2^17 | 1,914 → 16,582 |
| zklogin | 203,408 → 187,912 (−7.6%) | 2^18 → 2^18 | 22,862 → 38,358 |
| hashsign | 142,844 → 136,383 (−4.5%) | 2^18 → 2^18 | 110,514 → 116,975 |
| bitcoin_headers | 15,234 → 15,104 (−0.9%) | 2^14 → 2^14 | 4,614 → 4,744 |
| sha256 | 6,575 → 6,503 (−1.1%) | 2^13 → 2^13 | 2,005 → 2,077 |

No example gains a committed word; the Value Vector section of every snapshot is unchanged.

### Performance

`build_operation_columns` sizes the AND operand columns at `n_constraints.next_power_of_two()`, so the BitAnd sumcheck costs the *padded* size. **ethsign and ec_msm now cross a boundary (2^18 → 2^17), halving their AND reduction.** Measured on ethsign (quiet machine, interleaved A/B, 8 pairs, release + `target-cpu=native`):

| span | baseline | branch | delta |
|---|---|---|---|
| Prove (total) | 136.0 ms | 134.0 ms | −1.5% |
| **BitAnd check** | 13.6 ms | 8.2 ms | **−39.9%** |
| Shift Reduction | 17.9 ms | 19.0 ms | +6.1% |

Examples that don't cross a boundary stay a wash on total prove time — the AND-count reduction is headroom there. Full numbers and caveats in the comments below.

### DIFF
--- a/crates/examples/snapshots/bitcoin_headers.snap
+++ b/crates/examples/snapshots/bitcoin_headers.snap
@@ -5,16 +5,16 @@ Gates & Instructions
 └─ Number of evaluation instructions: 47,238
 
 Constraints
-├─ ZERO constraints: 4,734 used (57.8% of 2^13)
-│  [▓▓▓▓▓░░░░░] spare: 3,458
-├─ AND constraints: 15,114 used (92.2% of 2^14)
-│  [▓▓▓▓▓▓▓▓▓░] spare: 1,270
+├─ ZERO constraints: 4,744 used (57.9% of 2^13)
+│  [▓▓▓▓▓░░░░░] spare: 3,448
+├─ AND constraints: 15,104 used (92.2% of 2^14)
+│  [▓▓▓▓▓▓▓▓▓░] spare: 1,280
 ├─ IMUL constraints: 0 used (0.0% of 0)
 │  [░░░░░░░░░░] spare: 0
 ├─ BMUL constraints: 110 used (85.9% of 2^7)
 │  [▓▓▓▓▓▓▓▓░░] spare: 18
-└─ Distinct value indices: 47,303
-   ├─ Distinct shifted value indices: 27,142
+└─ Distinct value indices: 47,323
+   ├─ Distinct shifted value indices: 27,162
    └─ Distinct unshifted value indices: 20,161
 
 Value Vector
--- a/crates/examples/snapshots/bitcoin_headers.snap
+++ b/crates/examples/snapshots/bitcoin_headers.snap
@@ -5,10 +5,10 @@ Gates & Instructions
 └─ Number of evaluation instructions: 47,238
 
 Constraints
-├─ ZERO constraints: 4,614 used (56.3% of 2^13)
-│  [▓▓▓▓▓░░░░░] spare: 3,578
-├─ AND constraints: 15,234 used (93.0% of 2^14)
-│  [▓▓▓▓▓▓▓▓▓░] spare: 1,150
+├─ ZERO constraints: 4,734 used (57.8% of 2^13)
+│  [▓▓▓▓▓░░░░░] spare: 3,458
+├─ AND constraints: 15,114 used (92.2% of 2^14)
+│  [▓▓▓▓▓▓▓▓▓░] spare: 1,270
 ├─ IMUL constraints: 0 used (0.0% of 0)
 │  [░░░░░░░░░░] spare: 0
 ├─ BMUL constraints: 110 used (85.9% of 2^7)
--- a/crates/examples/snapshots/bitcoin_p2pkh.snap
+++ b/crates/examples/snapshots/bitcoin_p2pkh.snap
@@ -5,16 +5,16 @@ Gates & Instructions
 └─ Number of evaluation instructions: 176,783
 
 Constraints
-├─ ZERO constraints: 9,769 used (59.6% of 2^14)
-│  [▓▓▓▓▓░░░░░] spare: 6,615
-├─ AND constraints: 91,046 used (69.5% of 2^17)
-│  [▓▓▓▓▓▓░░░░] spare: 40,026
+├─ ZERO constraints: 16,582 used (50.6% of 2^15)
+│  [▓▓▓▓▓░░░░░] spare: 16,186
+├─ AND constraints: 84,233 used (64.3% of 2^17)
+│  [▓▓▓▓▓▓░░░░] spare: 46,839
 ├─ IMUL constraints: 15,186 used (92.7% of 2^14)
 │  [▓▓▓▓▓▓▓▓▓░] spare: 1,198
 ├─ BMUL constraints: 12,666 used (77.3% of 2^14)
 │  [▓▓▓▓▓▓▓░░░] spare: 3,718
-└─ Distinct value indices: 267,507
-   ├─ Distinct shifted value indices: 131,632
+└─ Distinct value indices: 275,067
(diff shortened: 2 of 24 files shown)


benji-potamus:
## Performance measurement

**Summary: the AND-count reduction does not move prove time today. It buys headroom.**

### Why, from the code

`build_operation_columns` (`crates/prover/src/prove.rs:577`) sizes the AND operand columns at `n_constraints.next_power_of_two()`. The BitAnd sumcheck therefore costs the *padded* size, not the true count. No example crosses a power-of-two boundary:

| example | padded AND before → after | padded ZERO before → after |
|---|---|---|
| ethsign | 2^18 → 2^18 | 2^12 → 2^15 |
| ec_msm | 2^18 → 2^18 | 2^12 → 2^15 |
| zklogin | 2^18 → 2^18 | 2^15 → 2^16 |
| bitcoin_p2pkh | 2^17 → 2^17 | 2^11 → 2^15 |
| sha256 | 2^13 → 2^13 | 2^11 → 2^12 |

So the win is 12% more AND capacity before the next doubling — and a doubling is a 2× cost step — rather than time on today's circuits.

### Measured

zklogin, 5 prove runs per side, release + `RUSTFLAGS="-C target-cpu=native"`, aarch64 (4 cores), quiet machine. Baseline is this branch's merge-base.

| span | baseline | branch | delta (min) |
|---|---|---|---|
| Prove (total) | 96.6 / 96.9 / 96.9 / 97.4 / 103 ms | 96.7 / 96.7 / 97.1 / 97.2 ms | wash |
| BitAnd check | min 16.8 ms | min 16.5 ms | −1.8% |
| Shift Reduction | min 24.7 ms | min 25.2 ms | **+2.0%** |

The small BitAnd gain is the operand-evaluation half of the phase, which *is* proportional to the true count. It is roughly cancelled by the shift reduction getting wider: ZERO rows grow, and the MSB asserts in commit 2 add `sar(x, 63)` terms (ethsign: +10,211 distinct shifted value indices).

### Honesty caveats

- **ethsign has no valid comparison.** Its branch batch ran on a quiet machine (139–142 ms) but its baseline batch hit contention (571–789 ms). Interleaved A/B retries were also noise-dominated — the shared box was at load average 25 on 4 cores with other agents' `rustc` running, giving 277–534 ms swings against a ~0.5 ms effect. I did not manage to get a clean ethsign pair.
- The zklogin figures above are the one pair where both sides ran quiet (spreads under 1% on the branch side, under 6% on baseline).

### Splitting, if you want it

Commit 2 (the MSB asserts) is the part that adds the `sar` terms and widens the shift reduction, and it contributes little AND-side on the AND-heavy circuits — zklogin goes 219,690 → 219,628, i.e. 62 constraints. Commit 1 delivers most of the AND reduction with essentially no new shifted terms. The two are separate commits so commit 2 can be dropped if the shift-reduction widening isn't worth it.

benji-potamus:
## Corrected performance measurement (after rebase onto #2084)

**This supersedes my earlier comment.** That comment said "no example crosses a power-of-two boundary, so measured prove time is a wash." That was true on the old base. It is no longer true: #2084 (select gates → BMUL) lowered AND counts enough that this change now tips **ethsign and ec_msm from 2^18 down to 2^17**, halving their BitAnd reduction.

### Measured — ethsign

Interleaved A/B (8 pairs, alternating baseline and branch binaries so machine drift cancels), quiet machine (load ~2 on 4 cores), release + `RUSTFLAGS="-C target-cpu=native"`, aarch64. Baseline is `main` at 06abf12.

| span | baseline min / med | branch min / med | delta (med) |
|---|---|---|---|
| Prove (total) | 135.0 / 136.0 ms | 133.0 / 134.0 ms | −1.5% |
| **BitAnd check** | 13.4 / 13.6 ms | 8.1 / 8.2 ms | **−39.9%** |
| Shift Reduction | 17.8 / 17.9 ms | 18.9 / 19.0 ms | +6.1% |

The BitAnd phase drops ~40%, which is the 2^18 → 2^17 halving showing up directly. Total prove moves only −1.5% because BitAnd is roughly 10% of ethsign's prove time — IntMul dominates it at ~67%.

The shift reduction gets ~6% slower: ZERO rows grow, and the MSB asserts in commit 2 add `sar(x, 63)` terms (+10,211 distinct shifted value indices on ethsign). That cost is real and partly offsets the BitAnd win on the total.

### Where it doesn't help

Circuits whose padded AND size is unchanged see a wash on total prove time. Measured earlier on zklogin (5 runs/side, quiet): total prove unchanged, BitAnd −1.8%, shift reduction +2.0%. For those the AND reduction is headroom — capacity before the next doubling, which is a 2× step — rather than time today.

### Caveats

- ec_msm also crosses 2^18 → 2^17 by the same margin, but I did not time it; the mechanism is identical to ethsign.
- Whether the +6% shift-reduction cost is worth paying on circuits that *don't* cross a boundary is a judgement call. Commit 2 is the part responsible for it (it adds the `sar` terms) and contributes little AND-side on AND-heavy circuits — zklogin moves 62 constraints. The two commits are separate so commit 2 can be dropped independently.

jimpo (approved):

PR #2088, opened 2026-08-11
BINIUS-427 (1/2): make the BaseFold channels generic over Channel::Elem

benji-potamus:
First of two PRs for [BINIUS-427](https://linear.app/jimpo/issue/BINIUS-427/add-wordip-channel-traits-and-generalize-the-iop-verifier-stack-over), split per review on #2083. This one only removes the `Elem = F` bound so the BaseFold and FRI verifiers can run over a symbolic element type. The `WordIP` channel traits follow in #2083, which is now stacked on this branch.

The IP layer is already generic over `Channel::Elem` — that is how `IronSpartanBuilderChannel` works — but everything above it pins `Elem = F`, so a channel that symbolically executes the verifier to build a circuit cannot be plugged in.

## What changes

`MerkleIPVerifierChannel::recv_openings` and `recv_committed_vector` return `Vec<Self::Elem>` rather than `Vec<F>`. The `Elem = F` bounds then come off `basefold::{channel, compiler, opening}` and `fri::{verify, batch, fold}`.

The FRI oracle types (`BrakedownOracle`, `BatchBrakedownOracle`, `FRIOracle`, `FRIQueryVerifier`) are parameterized by the element type instead of the base field, since the folding challenges they hold were sampled from the channel. Twiddles stay in `F` and convert at use, so the bound is `E: FieldOps<Scalar = F> + From<F>`.

Two math calls swap to their generic forms: `extrapolate_line_packed` → `extrapolate_line`, and `eq_ind_partial_eval` → `eq_ind_partial_eval_scalars`. Everything else on the path already had a `FieldOps` form.

Query indices stay `usize` and `sample_bits` stays on `MerkleIPVerifierChannel` — making those symbolic is #2083's job, along with the `binius-core` dependency that comes with it. Nothing here touches `Cargo.toml`.

## The FRI error surface narrows

A symbolic element has no equality, so the checks that compared an opened value against a claim become channel assertions: `fri::batch::reduce_queries` and `verify_terminal_queries` call `channel.assert_zero` where they returned `ClaimMismatch` / `IncorrectFold` / `IncorrectDegree`. A channel over concrete values still rejects, but reports `IPChannel(InvalidAssert)` instead of naming the query round and index. Nothing produces those three variants any more, so they are removed.

This is the one behavioural change in the PR, and it is forced by the generalization rather than incidental to it. Say the word if the query-round/index detail is worth keeping — it could be preserved by threading a position through the assertion.

## Testing

7 files, all in `binius-iop`.

* `cargo check --workspace --all-targets`: clean, including every example.
* `cargo test --workspace --exclude binius-examples --exclude binius-arith-bench`: **1396 passed, 0 failed, 19 ignored** across 50 binaries. The `iop-prover` FRI and BaseFold round-trips are the checks that the path over concrete values is unchanged.

`binius-examples` is excluded for disk, not correctness — linking its 328 example binaries needs ~6.3G my volume cannot spare; they all compile under `cargo check --all-targets`. `binius-arith-bench` does not build on aarch64 at stock `HEAD` and depends on nothing here.

### DIFF
--- a/crates/iop/src/fri/batch.rs
+++ b/crates/iop/src/fri/batch.rs
@@ -2,7 +2,7 @@
 
 use std::iter;
 
-use binius_field::BinaryField;
+use binius_field::{BinaryField, FieldOps};
 use binius_math::{line::extrapolate_line, multilinear, ntt::DomainContext};
 
 use crate::merkle_channel::MerkleIPVerifierChannel;
@@ -15,7 +15,10 @@ use crate::merkle_channel::MerkleIPVerifierChannel;
 /// implementation therefore holds a handle to the committed oracle along with the folding
 /// challenges, and receives openings over a Merkle channel in order to evaluate the virtual oracle
 /// at queried locations.
-pub trait ProxTestOracle<F: BinaryField> {
+///
+/// The oracle is parameterized by the element type `E` of the channel it is opened over, since the
+/// folding challenges it holds were sampled from that channel.
+pub trait ProxTestOracle<F: BinaryField, E> {
 	/// The Merkle commitment handle for the committed oracle.
 	type Commitment;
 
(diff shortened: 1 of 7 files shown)


crates/iop/src/fri/batch.rs
```diff
@@ -276,16 +282,16 @@ where
...
-			.zip(claims)
-			.map(|((coset, (&coset_index, &index)), &claim)| {
-				// Verify the claimed base-codeword value against the opened coset.
-				if coset[index & coset_mask] != claim {
-					return Err(Error::ClaimMismatch { index });
-				}
-				Ok(self.fold_coset(coset_index, coset.to_vec()))
-			})
-			.collect()
+		let mut folds = Vec::with_capacity(indices.len());
```
jimpo: What was wrong with collect? It's in the style guide to prefer functional style

benji-potamus: Nothing — I converted it without checking, and you're right that it didn't need to change. `assert_zero` takes `&mut channel`, but the closure can hold that mutable borrow perfectly well since `values` and `coset_indices` are owned by then, and `?` converts through the `From<binius_ip::channel::Error>` impl. Restored to `.map(...).collect()`.

Two other places where I'd made the same swap, both fixed in the same push:

* `verify_terminal_queries`' claim check — `iter::zip(claims, indices).try_for_each(...)`.
* the repetition-codeword check, which had been `.iter().any(...)` before I turned it into a loop — now `.iter().try_for_each(...)`.

jimpo (approved):

PR #2093, opened 2026-08-11
BINIUS-431: binius-recursion, an end-to-end recursion skeleton

benji-potamus:
**Draft — a skeleton, and not sound.** The gadgets a real recursive verifier needs are unwritten; see below. Nothing outside the new crate depends on it, and the one change elsewhere is a relaxed trait bound.

Part of [BINIUS-425](https://linear.app/jimpo/issue/BINIUS-425/create-a-merkleipverifierchannel-for-constructing-a-recursive-circuit) / [BINIUS-431](https://linear.app/jimpo/issue/BINIUS-431/binius-recursion-circuit-building-merkle-ip-verifier-channel), on the channel traits from #2083.

## What runs

A CRC-64 circuit is proved, its verifier is run symbolically to produce a recursive circuit, that circuit's witness is filled by replaying the same verifier over the real transcript, and the witness validates:

```
recursive circuit: 165653 gates, 0 AND, 130245 BMUL, 3915 ZERO, 3336 recorded inputs
```

Zero AND constraints because no hashing happens yet — this is the arithmetic half alone, dominated by evaluating the monster multilinear symbolically. The SHA-256 gadgets will add millions of AND and swamp it.

## How the witness gets filled

The builder channel records every wire the circuit cannot derive, in the order it reached them. `WitnessFillerChannel` wraps a real transcript channel, runs the same verifier, and writes each value into the next of those wires. One cursor keeps the runs aligned, which works because no shape in the protocol depends on a received value — every length comes from the FRI parameters and the oracle specs. `Shared` holds the builder and that record together so an operation can allocate a wire wherever it happens.

## Changes since the first review

* **`select` uses `binius_circuits::multiplexer::multi_wire_multiplex`**, per review. `Word::bit_mask` became `bit_at_msb` — one gate, since `select` reads its condition's MSB.
* **The inner statement arrives as builder inout wires**, per review.
* **The `IOPVerifierChannel` impl is gone.** The channel is a `MerkleIPVerifierChannel` and `BaseFoldVerifierCompiler::create_channel` wraps it, exactly as it wraps the transcript channel, so the oracle layer comes for free rather than being stubbed.
* **`invert_or_zero` and `square_transpose` are frontend hints** rather than recorded inputs — the evaluator derives them, so the replay need not know they exist. That is also the shape the real gadgets want: keep the hint, add the constraints that pin it.

## The one change outside the crate

`IOPVerifier::verify` no longer pins `Word = Word`, so a channel carrying words as wires can drive it. The statement is lifted through `From<Word>` and so arrives concrete, which leaves the recursive circuit tied to one instance — taking it symbolically is [BINIUS-433](https://linear.app/jimpo/issue/BINIUS-433/take-the-inner-statement-symbolically-in-iopverifierverify). That is a one-line change in place of the full refactor.

## What is not constrained

- **The Fiat-Shamir state.** `sample` and `sample_bits` return free wires, so nothing ties a challenge to the transcript that produced it.
- **The Merkle openings.** `recv_openings` and `recv_committed_vector` return free wires rather than values checked against a root.
- **The two field hints**, which carry the right value but are not pinned to their inputs.

So a circuit built here accepts proofs it should reject. Each gadget that lands removes entries from the recorded input list; with all of them in place the only input left is the proof itself. That is stated in the crate docs, not just at the call sites, because the circuit builds and reports normal statistics either way.

## Testing

`cargo test --workspace --exclude binius-examples --exclude binius-arith-bench`: **1409 passed, 0 failed, 19 ignored** across 54 binaries. `cargo clippy --all-features --tests -- -D warnings` clean on the crate.

`binius-examples` is excluded for disk, not correctness — the volume here cannot hold this workspace's artifacts twice. All targets compile under `cargo check --workspace --all-targets`.

### DIFF
--- a/crates/recursion/src/channel.rs
+++ b/crates/recursion/src/channel.rs
@@ -9,7 +9,7 @@ use binius_core::word::Word;
 use binius_field::{BinaryField128bGhash as B128, Field, FieldOps, util::FieldFn};
 use binius_frontend::{Circuit, Wire};
 use binius_iop::merkle_channel::{self, MerkleIPVerifierChannel};
-use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel};
+use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel, select_word, subset_sum_word};
 
 use crate::{
 	shared::Shared,
@@ -38,9 +38,7 @@ pub struct Recorded {
 	/// The compiled circuit.
 	pub circuit: Circuit,
 	/// The wires the witness must supply, in the order the verifier reached them.
-	pub inputs: Vec<Wire>,
-	/// The wires holding the inner statement.
-	pub inout: Vec<Wire>,
+	pub inputs: Vec<crate::shared::Input>,
 }
 
 /// A channel that records a verifier run as a Binius64 circuit.
@@ -57,23 +55,21 @@ pub struct Recorded {
 /// See the crate docs for what this does and does not constrain.
 pub struct Binius64BuilderChannel {
 	shared: Rc<Shared>,
-	inout: Vec<Wire>,
+	/// How many assertions have been recorded, so each gets a distinct name.
+	n_assertions: usize,
 }
 
 impl Binius64BuilderChannel {
-	/// Creates a channel over a fresh builder, with an inner statement of `n_inout` words.
-	pub fn new(n_inout: usize) -> Self {
-		let shared = Rc::new(Shared::new());
-		let inout = (0..n_inout).map(|_| shared.builder().add_inout()).collect();
-		Self { shared, inout }
-	}
-
-	/// The inner statement, as this circuit's public input.
-	pub fn statement(&self) -> Vec<SymbolicWord> {
-		self.inout
-			.iter()
-			.map(|&wire| SymbolicWord::wire(&self.shared, wire))
-			.collect()
+	/// Creates a channel over a fresh builder.
+	///
+	/// The inner statement is not among the circuit's inputs yet: `IOPVerifier::verify` lifts it
+	/// through `From<Word>`, so it reaches the channel already concrete and is baked in. Taking it
+	/// as inout wires is BINIUS-433.
+	pub fn new() -> Self {
+		Self {
+			shared: Rc::new(Shared::new()),
+			n_assertions: 0,
+		}
 	}
 
 	/// Consumes the channel and compiles what it recorded.
@@ -81,22 +77,26 @@ impl Binius64BuilderChannel {
 	/// Every [`SymbolicElem`] and [`SymbolicWord`] derived from this channel must be dropped first:
 	/// they hold weak handles to the builder, and using one afterwards panics.
 	pub fn build(self) -> Recorded {
-		let Self { shared, inout } = self;
-		let shared = Rc::try_unwrap(shared).unwrap_or_else(|_| {
+		let shared = Rc::try_unwrap(self.shared).unwrap_or_else(|_| {
 			panic!("SymbolicElem and SymbolicWord values hold only weak handles")
 		});
 		Recorded {
 			inputs: shared.inputs(),
 			circuit: shared.builder().build(),
-			inout,
 		}
 	}
 
 	/// Allocates the `(lo, hi)` pair one field element occupies, as circuit inputs.
-	fn input_elem(&mut self) -> SymbolicElem {
-		let lo = self.shared.input_wire();
-		let hi = self.shared.input_wire();
+	fn input_elem(&mut self, kind: &'static str) -> SymbolicElem {
+		let lo = self.shared.input_wire(kind);
+		let hi = self.shared.input_wire(kind);
 		SymbolicElem::wires(&self.shared, lo, hi)
+	}
+}
+
+impl Default for Binius64BuilderChannel {
+	fn default() -> Self {
+		Self::new()
 	}
 }
 
@@ -104,18 +104,18 @@ impl IPVerifierChannel<B128> for Binius64BuilderChannel {
 	type Elem = SymbolicElem;
 
 	fn recv_one(&mut self) -> Result<SymbolicElem, binius_ip::channel::Error> {
-		Ok(self.input_elem())
+		Ok(self.input_elem("recv_one"))
 	}
 
 	fn sample(&mut self) -> SymbolicElem {
 		// UNCONSTRAINED: a challenge is the Fiat-Shamir state's output, which needs the
 		// challenger's hashing in-circuit. Until then it is an input the replay supplies.
-		self.input_elem()
+		self.input_elem("sample")
 	}
 
 	fn observe_one(&mut self, _val: B128) -> SymbolicElem {
 		// UNCONSTRAINED: nothing absorbs the value.
-		self.input_elem()
+		self.input_elem("observe_one")
 	}
 
 	fn assert_zero(&mut self, val: SymbolicElem) -> Result<(), binius_ip::channel::Error> {
@@ -129,8 +129,16 @@ impl IPVerifierChannel<B128> for Binius64BuilderChannel {
 				}
 			}
 			SymbolicElem::Wires { lo, hi, .. } => {
-				self.shared.builder().assert_zero("channel", lo);
-				self.shared.builder().assert_zero("channel", hi);
+				// Number the assertions so an unsatisfied circuit names which check failed, and
+				// which half of the element it failed in.
+				let n = self.n_assertions;
+				self.n_assertions += 1;
+				self.shared
+					.builder()
+					.assert_zero(format!("assert_zero[{n}].lo"), lo);
+				self.shared
+					.builder()
+					.assert_zero(format!("assert_zero[{n}].hi"), hi);
 				Ok(())
 			}
 		}
@@ -159,6 +167,13 @@ impl WordIPVerifierChannel<B128> for Binius64BuilderChannel {
 	fn subset_sum(&mut self, elems: &[SymbolicElem], word: &SymbolicWord) -> SymbolicElem {
 		assert!(elems.len() <= Word::BITS); // precondition
 
+		// A word fixed while the circuit is built decides the selection there too, so it costs no
+		// gates. `fold_coset` reaches this on every round of the terminal fold, where the coset
+		// index is a constant.
+		if let SymbolicWord::Constant(word) = word {
+			return subset_sum_word(elems, *word);
+		}
+
 		// Each element is kept or dropped by its own bit, and the survivors are summed. The sum is
 		// XOR, which the constraint system absorbs, so the cost is the keep-or-drop.
 		let builder = self.shared.builder();
@@ -180,6 +195,11 @@ impl WordIPVerifierChannel<B128> for Binius64BuilderChannel {
 	fn select(&mut self, elems: &[SymbolicElem], word: &SymbolicWord) -> SymbolicElem {
 		assert!(!elems.is_empty() && elems.len().is_power_of_two()); // precondition
 
+		// As in `subset_sum`, a constant index picks its element while the circuit is built.
+		if let SymbolicWord::Constant(word) = word {
+			return select_word(elems, *word);
+		}
+
 		// One multiplexer over the `(lo, hi)` pairs, which is the same select-gate tree per wire
 		// position with the index bits read inside.
 		let builder = self.shared.builder();
@@ -199,7 +219,7 @@ impl WordIPVerifierChannel<B128> for Binius64BuilderChannel {
 	fn sample_bits(&mut self, _bits: usize) -> SymbolicWord {
 		// UNCONSTRAINED, twice over: the index should come from the Fiat-Shamir state, and it
 		// should be masked to `bits` bits, which the FRI code relies on rather than asserting.
-		SymbolicWord::wire(&self.shared, self.shared.input_wire())
+		SymbolicWord::wire(&self.shared, self.shared.input_wire("sample_bits"))
 	}
 }
 
@@ -211,7 +231,7 @@ impl MerkleIPVerifierChannel<B128> for Binius64BuilderChannel {
 		leaf_size: usize,
 		depth: usize,
 	) -> Result<Commitment, merkle_channel::Error> {
-		let root = std::array::from_fn(|_| self.shared.input_wire());
+		let root = std::array::from_fn(|_| self.shared.input_wire("merkle_root"));
 		Ok(Commitment {
 			root,
 			leaf_size,
@@ -227,7 +247,7 @@ impl MerkleIPVerifierChannel<B128> for Binius64BuilderChannel {
 		// UNCONSTRAINED: nothing binds the opened values to the root. Hashing each leaf and
 		// climbing the path is the gadget this skeleton leaves out.
 		Ok((0..indices.len() * commitment.leaf_size)
-			.map(|_| self.input_elem())
+			.map(|_| self.input_elem("opening"))
 			.collect())
 	}
 
@@ -237,7 +257,7 @@ impl MerkleIPVerifierChannel<B128> for Binius64BuilderChannel {
 	) -> Result<Vec<SymbolicElem>, merkle_channel::Error> {
 		// UNCONSTRAINED: nothing rebuilds the tree over the vector.
 		Ok((0..commitment.leaf_size << commitment.depth)
-			.map(|_| self.input_elem())
+			.map(|_| self.input_elem("committed_vector"))
 			.collect())
 	}
 }
--- a/crates/recursion/src/channel.rs
+++ b/crates/recursion/src/channel.rs
@@ -1,0 +1,267 @@
+// Copyright 2026 The Binius Developers
+
+//! The channel that turns a verifier run into a circuit.
+
+use std::rc::Rc;
+
+use binius_field::{BinaryField128bGhash as B128, Field, FieldOps, util::FieldFn};
+use binius_frontend::{Circuit, CircuitBuilder, Wire};
+use binius_iop::{
+	channel::{IOPVerifierChannel, OracleLinearRelation, OracleSpec},
+	merkle_channel::{self, MerkleIPVerifierChannel},
+};
+use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel};
+
+use crate::{Elem, Word};
+
+/// Number of 64-bit words a SHA-256 digest occupies.
+const DIGEST_WORDS: usize = 4;
+
+/// A Merkle commitment received by the builder channel.
+///
+/// The root is wires rather than bytes, since the prover supplies it in the proof. The shape is
+/// fixed by the protocol, so it stays concrete.
+#[derive(Clone)]
+pub struct Commitment {
+	/// The commitment root, as `DIGEST_WORDS` wires.
+	pub root: [Wire; DIGEST_WORDS],
+	/// Field elements in each leaf.
+	pub leaf_size: usize,
+	/// Base-2 logarithm of the number of leaves.
+	pub depth: usize,
+}
+
+/// A channel that records a verifier run as a Binius64 circuit.
+///
+/// Drive a verifier written against the channel traits with this in place of a transcript channel
+/// and the result is a circuit rather than a verdict. See the crate docs for what is and is not
+/// constrained yet.
+///
+/// The builder is shared through an [`Rc`] because [`IOPVerifierChannel`] requires
+/// `Elem: 'static`, so an element cannot borrow the builder it was built on. The frontend's
+/// `CircuitBuilder` takes `&self` throughout, so no interior-mutability wrapper is needed on top.
+pub struct Binius64BuilderChannel {
+	builder: Rc<CircuitBuilder>,
+	/// Wires holding the proof, in the order the verifier reads them. Filling a witness means
+	/// writing the proof into these and letting the evaluator derive everything else.
+	transcript: Vec<Wire>,
+	oracle_specs: Vec<OracleSpec>,
+	next_oracle_index: usize,
+}
+
+impl Binius64BuilderChannel {
+	/// Creates a channel over a fresh builder, expecting the given oracles.
+	pub fn new(oracle_specs: Vec<OracleSpec>) -> Self {
+		Self {
+			builder: Rc::new(CircuitBuilder::new()),
+			transcript: Vec::new(),
+			oracle_specs,
+			next_oracle_index: 0,
+		}
+	}
+
+	/// The wires the proof is written into, in read order.
+	pub fn transcript(&self) -> &[Wire] {
+		&self.transcript
+	}
+
+	/// Consumes the channel and compiles the recorded circuit.
+	///
+	/// Every [`Elem`] and [`Word`] derived from this channel must be dropped first: they hold weak
+	/// handles to the builder, and using one afterwards panics.
+	pub fn build(self) -> Circuit {
+		Rc::try_unwrap(self.builder)
+			.unwrap_or_else(|_| panic!("Elem and Word values hold only weak handles"))
+			.build()
+	}
+
+	/// Allocates a witness wire and records it as the next word of the proof.
+	fn recv_word(&mut self) -> Wire {
+		let wire = self.builder.add_witness();
+		self.transcript.push(wire);
+		wire
+	}
+
+	/// Allocates the `(lo, hi)` pair one received field element occupies.
+	fn recv_elem(&mut self) -> Elem {
+		let lo = self.recv_word();
+		let hi = self.recv_word();
+		Elem::wires(&self.builder, lo, hi)
+	}
+}
+
+impl IPVerifierChannel<B128> for Binius64BuilderChannel {
+	type Elem = Elem;
+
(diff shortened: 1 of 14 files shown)


crates/recursion/src/elem.rs
```diff
@@ -0,0 +1,230 @@
...
+/// absorbs into its operands for free.
+///
+/// A `Constant` folds while the circuit is built. That matters more than it looks: the verifier's
+/// arithmetic is full of build-time constants — subspace bases, Lagrange weights, eq-indicator
+/// evaluations at fixed points — and folding them costs no constraints at all.
+#[derive(Clone)]
+pub enum Elem {
+	Constant(B128),
+	Wires {
+		builder: Weak<CircuitBuilder>,
```
benji-potamus: **Decision 1 of 4: every element carries a `Weak<CircuitBuilder>`.**

This is the thing I'd least like to be wrong about, since it's in the hot data type.

It's forced by two constraints rather than chosen. `IOPVerifierChannel` requires `Elem: 'static` (`iop/src/channel/mod.rs:96`), so `Elem<'a>` borrowing the builder is not allowed. And `WordIPVerifierChannel` asks for `From<Word>` and `Shr` as plain operators on the associated type, so those have to work with no channel in hand — same for `Add`/`Mul` on `Elem`, which `FieldOps` requires as operators.

Where it differs from Spartan's `CircuitElem`: no `RefCell`. The frontend's `CircuitBuilder` takes `&self` throughout and does its own interior mutability, so `Rc<CircuitBuilder>` is enough.

The cost is 8 bytes and an `upgrade()` per operation on every element, and the panic path in `combine` when an element outlives the channel. If you'd rather pay something else — e.g. a thread-local or a builder passed explicitly with `FieldOps` implemented on a wrapper — now is the time.

crates/recursion/src/channel.rs
```diff
@@ -0,0 +1,267 @@
...
+/// constrained yet.
+///
+/// The builder is shared through an [`Rc`] because [`IOPVerifierChannel`] requires
+/// `Elem: 'static`, so an element cannot borrow the builder it was built on. The frontend's
+/// `CircuitBuilder` takes `&self` throughout, so no interior-mutability wrapper is needed on top.
+pub struct Binius64BuilderChannel {
+	builder: Rc<CircuitBuilder>,
+	/// Wires holding the proof, in the order the verifier reads them. Filling a witness means
+	/// writing the proof into these and letting the evaluator derive everything else.
+	transcript: Vec<Wire>,
```
benji-potamus: **Decision 2 of 4: the witness story is this one field.**

Every `recv_*` allocates witness wires and pushes them here, so `transcript` ends up holding the proof in the order the verifier read it. The claim that follows is that these are the circuit's *only* inputs — challenges, folded values and Merkle digests are all gate outputs — so populating a witness is:

```
write the proof bytes into transcript[..]  →  evaluator derives everything else
```

and there is no second pass replaying the verifier against a filler channel, which is what the Spartan wrapper needs.

Two things that could break the claim, worth checking now rather than at BINIUS-432:

1. The inner statement is not in here. It arrives via `observe_words`, and once BINIUS-433 makes it symbolic it becomes a second input — presumably `inout` wires rather than witness. Should `transcript` be two vectors from the start?
2. It assumes control flow never depends on a received value. That holds today (every shape comes from `FRIParams` and the oracle specs), but nothing enforces it.

crates/recursion/src/channel.rs
```diff
@@ -0,0 +1,267 @@
...
+		assert!(elems.len() <= binius_core::word::Word::BITS); // precondition
+
+		// Each element is kept or dropped by its bit, then the survivors are summed. `select`
+		// against zero is the keep-or-drop, and the sum is XOR, which costs nothing.
+		(0..elems.len())
+			.map(|bit| select_elem(&self.builder, &word.bit_mask(bit), &elems[bit], &Elem::zero()))
+			.sum()
+	}
+
+	fn select(&mut self, elems: &[Elem], word: &Word) -> Elem {
```
benji-potamus: **Decision 3 of 4: `select` is built from the `select` gate, so every mux lands in the BMUL column.**

This is the one I'd most like a second opinion on, because I picked it without thinking and the alternative is a one-line change.

Since BINIUS-437, `select` is 1 BMUL (`gate/select.rs`). But `select(mask, t, f)` and `f ^ (mask & (t ^ f))` compute the same thing, and `band` is 1 AND (`gate/band.rs`) with the XORs free. So this code chooses which constraint column every mux in the FRI verifier lands in.

Scale, at rate 1 with 232 queries:

| call site | entries | gates per query |
| -- | -- | -- |
| `FRIOracle::reduce_queries` | 2^arity ≈ 8–16 | 14–30 |
| `verify_terminal_queries` | 2^(n_final+rate) ≈ 2048 | 4094 |

The terminal mux alone is ~950k gates — currently 950k BMUL, or 950k AND if built from `band`.

It isn't obvious which is better. Constraint counts pad to a power of two per column independently (`stat` reports e.g. `AND constraints: 368,687 used (70.3% of 2^19)`). A recursive verifier's AND column will be dominated by SHA-256 in the millions and its BMUL column by sumcheck multiplications in the thousands, so there may be a million already-paid-for AND slots to put muxes in — or piling them into BMUL may balance the two columns better. Opposite answers, and which wins depends on where the totals land, which BINIUS-430/432 measure.

Happy to leave it and revisit with numbers, unless you already expect one column to be scarce.

crates/recursion/src/word.rs
```diff
@@ -0,0 +1,82 @@
...
+			Self::Constant(word) => builder.add_constant_64(word.as_u64()),
+			Self::Wire { wire, .. } => *wire,
+		}
+	}
+
+	/// Returns the bit at position `i` as an all-ones or all-zeros mask word.
+	///
+	/// Masks rather than a single bit because that is what `select` consumes: the frontend's
+	/// `select` gate reads the most significant bit of its condition.
+	pub fn bit_mask(&self, _i: usize) -> Wire {
```
benji-potamus: **Decision 4 of 4: `bit_mask` returns a broadcast mask, not a bit.**

The frontend's `select` reads the MSB of its condition, and a `band`-based mux needs an all-ones/all-zeros word. Either way the consumer wants the bit broadcast across the word, so this returns a mask and the gadget owes the broadcast rather than just a shift.

Naming it `bit_mask` rather than `bit` is meant to make that contract visible at the call sites in `subset_sum` and `select`. If you'd rather it return a plain 0/1 word and have the muxes broadcast, say so before BINIUS-430 builds a Merkle path on it — that gadget wants the same operation for its pair ordering, so the two should agree.

crates/recursion/src/channel.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+impl IPVerifierChannel<B128> for Binius64BuilderChannel {
+	type Elem = Elem;
+
+	fn recv_one(&mut self) -> Result<Elem, binius_ip::channel::Error> {
+		Ok(self.recv_elem())
+	}
+
+	fn sample(&mut self) -> Elem {
```
benji-potamus: Not a decision, a warning about how this fails.

`sample`, `sample_bits` (line 177) and `recv_openings` (line 200) all return plausible values while constraining nothing. A circuit built from this channel compiles, reports sensible statistics, and accepts any proof — there is no `todo!()` to trip over. That is why the crate docs say it outright rather than leaving it to these comments.

If you'd rather the skeleton refuse to run at all until the gadgets land, I can make these `todo!()` too. The reason they aren't is that returning values lets a verifier run end to end against the channel, which is how I'd like to find the *next* structural problem before writing several thousand lines of SHA-256 gadget.

benji-potamus (commented):
Five inline comments: four decisions I'd like challenged, and one warning about the failure mode.

Summary, so you can triage before opening files:

1. **`elem.rs:31`** — every `Elem` carries a `Weak<CircuitBuilder>`. Forced by `Elem: 'static` on `IOPVerifierChannel` plus the operator-shaped bounds on `Word`, but it's in the hot type, so worth confirming.
2. **`channel.rs:47`** — `transcript: Vec<Wire>` is the whole witness story. Claims the proof is the circuit's only input, so no replay pass is needed. Two things that could break that claim.
3. **`channel.rs:160`** — `select` is built from the `select` gate, putting ~950k gates per proof in the BMUL column rather than AND. One-line change either way; I picked it without thinking.
4. **`word.rs:52`** — `bit_mask` returns a broadcast mask rather than a bit, which sets the contract the Merkle gadget will inherit.
5. **`channel.rs:100`** — the unconstrained stubs return plausible values rather than panicking, so the circuit builds and accepts anything. Deliberate, but say if you'd rather it refuse to run.

crates/recursion/src/elem.rs
jimpo: This is fine. The Spartan builder does the same pattern.

crates/recursion/src/channel.rs
jimpo: 1. inout for the circuit should come in as wires for the builder.
2. Yes, that is an essential assumption and we'll keep things that way.

crates/recursion/src/channel.rs
jimpo: Forget the padding, that's been mitigated significantly. We want to reduce # of constraints. The select gate and the multiplexer circuit (`binius_circuits::multi_wire_multiplex`) is the way to go.

jimpo: Use the multiplexer circuit from binius_circuits

crates/recursion/src/word.rs
```diff
@@ -0,0 +1,78 @@
...
+/// This is the `Word` associated type of
+/// [`WordIPVerifierChannel`](binius_ip::channel::WordIPVerifierChannel). The trait requires
+/// `From<Word> + Shr<u32>` on it rather than offering channel methods, so the operations have to be
+/// available without a channel in hand — hence the [`Weak`] handle back to the builder, which is
+/// the same shape `CircuitElem` uses in the Spartan wrapper.
+///
+/// A `Constant` folds while the circuit is built and costs nothing. The FRI query indices, which
+/// arrive from `sample_bits`, are `Wire`s.
+#[derive(Clone)]
+pub enum Word {
```
jimpo: rename to SymbolicWord to avoid conflict with binius_core::word::Word. You can `use binius_core::word::Word;`

crates/recursion/src/word.rs
```diff
@@ -0,0 +1,78 @@
...
+			Self::Constant(word) => builder.add_constant_64(word.as_u64()),
+			Self::Wire { wire, .. } => *wire,
+		}
+	}
+
+	/// Returns bit `i` moved into the most significant position, where a `select` gate reads it.
+	///
+	/// The bits below it are whatever the shift carried up and are ignored, so this is one gate
+	/// rather than a broadcast. `single_wire_multiplex` selects the same way.
+	pub fn bit_at_msb(&self, builder: &CircuitBuilder, i: usize) -> Wire {
```
jimpo: This is weird. Just impl `Shl<32>` for SymbolicWord.

crates/recursion/src/elem.rs
```diff
@@ -0,0 +1,285 @@
...
+///
+/// The wire pair is the `(lo, hi)` split the frontend's `bmul` gate takes, so one field
+/// multiplication is one BMUL constraint and addition is two XORs, which the constraint system
+/// absorbs into its operands for free.
+///
+/// A `Constant` folds while the circuit is built. That matters more than it looks: the verifier's
+/// arithmetic is full of build-time constants — subspace bases, Lagrange weights, eq-indicator
+/// evaluations at fixed points — and folding them costs no constraints at all.
+#[derive(Clone)]
+pub enum Elem {
```
jimpo: Rename to SymbolicElem to be coherent with SymbolicWord. You can even put both of them in a module called `symbolic`

jimpo (changes requested):
Move this out of draft, it's ready for final review after these inline comments are addressed.

jimpo:
@tcoratger Please take a look at this. Once it's in, there will be several follow-ons, then we'll have minimal viable recursion!

crates/recursion/src/word.rs
benji-potamus: Renamed to `SymbolicWord`, and `channel.rs` now does a plain `use binius_core::word::Word;`. The clash was worse than it looked — `subset_sum`'s precondition reads `elems.len() <= Word::BITS` against core's `Word`, so the file genuinely needs both names in scope.

`SymbolicElem` likewise, and both moved into a `symbolic` module per your other comment.

crates/recursion/src/word.rs
benji-potamus: Agreed, it was weird. `bit_at_msb` is gone; there is now `impl Shl<u32> for SymbolicWord` alongside the existing `Shr`, and both route through one `shift` helper that folds a constant or emits the gate.

The call site in `subset_sum` reads:

```rust
let sel = (word.clone() << (Word::BITS - 1 - bit) as u32).to_wire(builder);
```

which says what it does — move the bit where `select` reads it — instead of hiding it behind a method name.

crates/recursion/src/elem.rs
benji-potamus: Done — `SymbolicElem`, and both types moved into `src/symbolic/`, with the module doc covering what they share: each is a constant that folds while the circuit is built, or wires plus a `Weak` handle to the builder, and the handle is what lets the arithmetic be plain operators rather than channel methods.

crates/recursion/src/channel.rs
benji-potamus: This one was already done in `cd0da11d`, before this round of comments — flagging it so it doesn't read as unaddressed. `select` builds on `binius_circuits::multiplexer::multi_wire_multiplex` over the `(lo, hi)` pairs:

```rust
let groups = pairs.iter().map(|pair| pair.as_slice()).collect::<Vec<_>>();
let sel = word.to_wire(builder);
let selected = multi_wire_multiplex(builder, &groups, sel);
```

crates/recursion/src/channel.rs
benji-potamus: Flagging that **the answer you gave here no longer describes the code**, since GitHub has marked the thread outdated and that would otherwise bury it.

Your point 1 is done — the statement comes in as builder inout wires. Point 2 holds and the code depends on it.

What changed is the claim I made in the comment above: that the proof is the circuit's only input, so no replay pass is needed. That is no longer true. `transcript: Vec<Wire>` is gone; the record moved to `Shared`, and there is now a `WitnessFillerChannel` that replays the verifier over the real transcript.

The reason is the unconstrained stubs. A circuit can only derive a challenge if the challenger runs inside it, and can only derive an opened value if the Merkle check does. Neither is built, so those values have to be supplied, and the only thing that knows them is a real verifier run. The replay is a consequence of the gadgets being absent, not a design change — as each lands, entries leave the recorded list, and with all of them in place the proof is once again the only input and the replay can go.

Concretely, the crc64 test currently records 3,336 input wires, which is a direct measure of what is still missing.

crates/recursion/src/channel.rs
```diff
@@ -0,0 +1,243 @@
...
+impl Binius64BuilderChannel {
+	/// Creates a channel over a fresh builder, with an inner statement of `n_inout` words.
+	pub fn new(n_inout: usize) -> Self {
+		let shared = Rc::new(Shared::new());
+		let inout = (0..n_inout).map(|_| shared.builder().add_inout()).collect();
+		Self { shared, inout }
+	}
+
+	/// The inner statement, as this circuit's public input.
+	pub fn statement(&self) -> Vec<SymbolicWord> {
```
tcoratger: `statement()` isn't called anywhere, and neither are the inout wires.

`IOPVerifier::verify` lifts the statement with `Channel::Word::from`, so it arrives as `Word::Constant` and these wires feed no gate. The test fills them, which makes it look wired up.

Either drop `statement()` and `n_inout` until BINIUS-433, or pass the statement through the channel so the wires are the only path. The half-way state is the confusing one.

crates/recursion/src/channel.rs
```diff
@@ -0,0 +1,243 @@
...
+
+impl WordIPVerifierChannel<B128> for Binius64BuilderChannel {
+	type Word = SymbolicWord;
+
+	fn observe_words(&mut self, _words: &[SymbolicWord]) {
+		// UNCONSTRAINED: the statement must reach the Fiat-Shamir state for the circuit to be
+		// bound to what it claims to verify.
+	}
+
+	fn subset_sum(&mut self, elems: &[SymbolicElem], word: &SymbolicWord) -> SymbolicElem {
```
tcoratger, line 167: This and `select` below both throw away the `Constant` case.

`binius_ip::channel` has `subset_sum_word` and `select_word` for exactly that, so a word known at build time costs no gates.

It happens on a real path: `fri/verify.rs:247` builds a constant coset index and `fold_coset` calls `subset_sum` with it once per fold round.

The `Shl` doesn't save you here — the frontend doesn't const-fold through `Select`, so a constant condition still emits the gate. Match on the word first and delegate.

crates/recursion/src/channel.rs
```diff
@@ -0,0 +1,243 @@
...
+			// A build-time constant is decided here; a non-zero one is unsatisfiable.
+			SymbolicElem::Constant(c) => {
+				if c == B128::ZERO {
+					Ok(())
+				} else {
+					Err(binius_ip::channel::Error::InvalidAssert)
+				}
+			}
+			SymbolicElem::Wires { lo, hi, .. } => {
+				self.shared.builder().assert_zero("channel", lo);
```
tcoratger: Give these distinct names. Every assertion in the circuit is called "channel", so an unsatisfied circuit can't tell you which check failed, or which half of the element.

crates/recursion/src/symbolic/elem.rs
```diff
@@ -0,0 +1,285 @@
...
+	}
+}
+
+impl Mul<&Self> for SymbolicElem {
+	type Output = Self;
+
+	fn mul(self, rhs: &Self) -> Self {
+		// A product with a zero factor is zero without a constraint, which is worth catching: the
+		// verifier multiplies by eq-indicator terms that are often zero at build time.
+		if matches!(&self, Self::Constant(c) if *c == B128::ZERO)
```
tcoratger, line 126: Catch `ONE` here too — `x * 1` currently emits a BMUL. Same reasoning as zero: eq-indicator products and Lagrange weights hit it.

crates/recursion/src/filler.rs
```diff
@@ -0,0 +1,164 @@
...
+			filler,
+			wires: wires.into_iter(),
+		}
+	}
+
+	/// Checks that the replay consumed exactly the wires the build recorded.
+	///
+	/// A mismatch means the two runs diverged, which would leave the witness silently wrong rather
+	/// than merely incomplete.
+	pub fn finish(self) {
```
tcoratger, line 56: This only checks the total.

If two operations disagree in opposite directions the counts still add up, every later value goes into the wrong wire, and neither this nor `fill_word` says anything. The whole witness story rests on the two runs staying in step, so it's worth making a divergence loud.

Record which operation allocated each wire and assert it matches on fill.

crates/recursion/src/filler.rs
```diff
@@ -0,0 +1,164 @@
...
+	fn recv_merkle_commitment(
+		&mut self,
+		leaf_size: usize,
+		depth: usize,
+	) -> Result<Self::Commitment, merkle_channel::Error> {
+		let commitment = self.inner.recv_merkle_commitment(leaf_size, depth)?;
+		// The builder reads a root as four wires. The concrete channel keeps it as a digest rather
+		// than words, so the replay has nothing to copy and leaves them zero — one more thing the
+		// Merkle gadget will settle.
+		for _ in 0..super::channel::DIGEST_WORDS {
```
tcoratger: Filling the root with zeros means the end-to-end test passes with four wires set to the wrong value. Harmless while nothing constrains the root, but it becomes a confusing failure the moment the Merkle gadget lands.

Serialize the digest into words here instead.

crates/recursion/src/hints.rs
```diff
@@ -0,0 +1,79 @@
...
+
+	fn execute(&self, dimensions: &[usize], inputs: &[Word], outputs: &mut [Word]) {
+		let degree = dimensions[0];
+		let mut elems = (0..degree)
+			.map(|i| elem_of(&inputs[2 * i..]))
+			.collect::<Vec<_>>();
+
+		// The transpose is over the subfield the degree names. Ring switching only ever asks for
+		// `B1`, which is the case the degree of 128 selects.
+		match degree {
```
tcoratger: Dispatching on the degree throws away `FSub`, which the caller in `elem.rs` already knows.

Correct for B128 today, since only B1 has degree 128. But a call with any other subfield panics inside a verifier. Pass a discriminant you control, or do the transpose generically.

tcoratger (commented):
Skeleton looks reasonable. I've also done #2103, #2104, #2106, #2107 related so we should check what to do with them (I think they are complementary as noted just after). 

Main thing: the inout wires aren't actually used, so review point 1 isn't done yet. Rest are small.

Also, the gadgets this leaves out are up already — challenger in #2103, Merkle in #2104, and a constrained element + channel in #2106 — so the unconstrained parts here have replacements ready rather than needing to be written again.

crates/recursion/src/channel.rs
benji-potamus: Good catch, and it's measurable. Both now match on `SymbolicWord::Constant` first and delegate to `subset_sum_word` / `select_word`.

On the crc64 end-to-end test the recursive circuit goes 130,245 → 129,959 BMUL and 165,653 → 165,050 gates — and that's a four-word inner circuit, so the fold rounds you point at are nearly all of it.

I haven't verified your claim about `Select` not const-folding, so I'm not repeating it: matching on the word first is right regardless, since it skips the work rather than relying on a later pass to undo it.

crates/recursion/src/filler.rs
benji-potamus: Agreed, and this is the one I'm most glad you caught — the alignment is the whole witness story, and a check that two divergences can cancel out of is barely a check.

Each recorded wire now carries the operation that allocated it:

```rust
pub struct Input {
    pub wire: Wire,
    pub kind: &'static str,
}
```

and every fill asserts the two agree, so a divergence names the wire it happened at instead of showing up as a wrong value thousands of wires later:

```
the replay diverged from the build: it filled a sample where the build recorded a opening
```

`finish()` still checks the total, since that catches the case where the counts differ.

crates/recursion/src/channel.rs
benji-potamus: You're right that the half-way state is the confusing one, and I've dropped `statement()`, `n_inout` and `Recorded::inout` rather than half-wire them. The test no longer fills them either, so nothing looks connected that isn't.

Worth naming for @jimpo: this partly defers his earlier request on the other thread that the inout come in as builder wires. It can't be honoured while `IOPVerifier::verify` lifts the statement through `From<Word>` — it arrives concrete whatever the channel does with it. BINIUS-433 changes that signature, and the wires come back then, doing something.

crates/recursion/src/filler.rs
benji-potamus: Fixed — the digest is serialized into the four wires now.

This forced a change I think is an improvement anyway: the filler was generic over `Inner: MerkleIPVerifierChannel`, whose `Commitment` is opaque, so it had no way to read a root. It is now specialised to `VerifierMerkleTranscriptChannel`, which is the only thing it ever wraps, and `TranscriptMerkleCommitment` exposes the root.

Agreed on the reasoning too: zeros are harmless while nothing constrains the root and would have turned into a puzzling failure the moment BINIUS-430 started reading it.

crates/recursion/src/channel.rs
benji-potamus: Done — assertions are numbered in reach order and each half is named, so a failure reports `assert_zero[17].hi` rather than one name repeated across every check in the run. The counter lives on the channel, which is also what makes it the reach order.

crates/recursion/src/symbolic/elem.rs
benji-potamus: Caught, on both sides — `Mul` now folds a constant `ONE` in either operand as well as `ZERO`.

Measured on the end-to-end test, `ONE` on its own is worth little here: the recursive circuit went from 130,245 to 129,959 BMUL, and most of that is the constant-word change on the other thread. Right to add anyway — it costs nothing and multiplying by one shouldn't reach the constraint system at all.

crates/recursion/src/hints.rs
benji-potamus: Agreed the degree was doing work it can't do: `ExtensionField<B1>` and `ExtensionField<B8>` over `B128` have different degrees today, but nothing says a future pair won't collide, and the hint would have transposed the wrong one silently.

Rather than pass a discriminant I dropped the pretence of genericity, since a hint can't be generic anyway — it registers under one constant `NAME` and receives words, not typed elements, so it has no way to recover `FSub` whatever it's handed. So it's `SquareTransposeB1Hint`, `B1` only, and the check moved up to `SymbolicElem::square_transpose`, where `FSub` is still in scope: degree 1 returns early, `B1`'s degree calls the hint, anything else asserts. Ring switching only asks for `B1`, so nothing is lost, and the assert is what fails loudly if that changes.

Doing it generically is the real answer and belongs with the field gadgets — a bit-matrix transpose over the wire pairs, which is cheap. This is a stand-in until then.

jimpo (approved):

PR #2095, opened 2026-08-11
Prove the phase-1 shift sumcheck over sparse shift rows

benji-potamus:
Closes [BINIUS-460](https://linear.app/jimpo/issue/BINIUS-460/shift-sparse-sumcheck-over-the-shift-variant-and-amount-variables).

Prover-side only; the verifier is untouched and the transcript is byte-identical.

> **Revised twice.** It first hand-rolled the sparse rounds, then routed `g` through `SparseDenseProductSumcheckProver` (#2105), and now has the `Phase1SumcheckProver` @jimpo asked for. **The two performance comments above measure designs that no longer exist — disregard them**; current numbers are in the newest comment.

## What

The phase-1 sumcheck binds 15 variables: the high 9 index the `(shift variant, shift amount)` row, the low 6 the bit position within it. A constraint system names only a few dozen of the 512 rows — 16 for SHA256, 40 for Keccak — so `g` is zero almost everywhere over the row axis. But *within* a named row it is dense: every bit position carries a value. Its sparsity is in long strides, not scattered points.

`Phase1SumcheckProver` changes strategy halfway to match that:

- **The 9 row rounds** read the stored rows alone, so their cost follows the shifts the system names rather than the 512 the space holds. `h` stays dense across them and folds as any multilinear does, which is what keeps this simple — a round reads `h` at whichever row a live `g` row pairs with, so nothing has to work out in advance which rows of `h` a later round will reach.
- **The 6 bit rounds** run on the shared `bivariate_product_prover`. By then both multilinears are one dense row and there is no sparsity left to exploit.

It is an adaptation of `SparseDenseProductSumcheckProver` and inherits its central simplification: both sampled evaluations and the fold are **linear in `g`**, so a stored row contributes on its own. Rows are never paired across the split or merged, and the list stays a flat run of the same length for the whole protocol.

That linearity pays off twice more. The two key segments' rows simply **concatenate** — rows at a repeated index add up wherever `g` is read — so the merge step is gone, and with it the `2^15` buffer each segment used to be scattered into and the loop that summed the two.

## Correctness

Both provers sum the same product over the same hypercube, so field arithmetic being exact, every round message must agree exactly. Three tests, covering different things — each checked by mutation rather than assumed:

- `sparse_rounds_match_the_dense_prover` (and an empty-`g` variant) asserts a **byte-identical transcript**, the same challenges and the same reduced evaluation against a dense `bivariate_product_prover`. Verified non-vacuous by dropping the `index & half` guard on `y_1`, which made it fail.
- `from_segments_concatenates_the_two_encodings` pins the concatenation and that a repeated index adds up where `g` is read.
- `tests/shift.rs::test_shift_prove_and_verify` covers the whole reduction end to end against the real verifier, which builds `h` independently.

`cargo test --workspace` (56 suites), `cargo clippy --all --tests --benches --examples -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items --no-deps` and `cargo +nightly-2026-07-01 fmt --check` are clean.

`m4-prover` and the `shift_reduction` bench are ported alongside, since the signatures change.

### DIFF
--- a/crates/prover/src/protocols/shift/phase_1.rs
+++ b/crates/prover/src/protocols/shift/phase_1.rs
@@ -1,17 +1,26 @@
 // Copyright 2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::{iter, ops::Range};
+use std::{
+	iter,
+	ops::{Deref, Range},
+};
 
 use binius_compute::Allocator;
 use binius_core::word::Word;
-use binius_field::{BinaryField, Field, PackedField};
-use binius_ip::sumcheck::SumcheckOutput;
+use binius_field::{BinaryField, Field, PackedField, WideMul};
+use binius_ip::sumcheck::{RoundCoeffs, SumcheckOutput};
 use binius_ip_prover::{
 	channel::IPProverChannel,
-	sumcheck::{ProveSingleOutput, bivariate_product_prover, prove_single},
+	sumcheck::{
+		ProveSingleOutput, bivariate_product_evaluator::BivariateProductEvaluator,
+		bivariate_product_prover, common::SumcheckProver, prove_single, round_evals::RoundEvals,
+		round_evaluator::SharedSumcheckProver,
+	},
 };
-use binius_math::{BinarySubspace, FieldBuffer, FieldVec};
+use binius_math::{
+	BinarySubspace, FieldBuffer, FieldVec, multilinear::fold::fold_highest_var_inplace,
+};
 use binius_utils::rayon::{
 	prelude::*,
 	task_size::{IndexedParallelIteratorExt, WorkPerItem},
@@ -34,6 +43,13 @@ use super::{
 /// amount, and the shift variant.
 pub const PHASE_1_LOG_LEN: usize = Word::LOG_BITS + Word::LOG_BITS + LOG_SHIFT_VARIANT_COUNT;
 
+/// The number of variables of the phase-1 multilinears that index the shift.
+///
+/// Both multilinears hold one row of `Word::BITS` scalars per `(shift variant, shift amount)`
+/// pair, the bit position running within a row. These are the variables above that row: the shift
+/// amount, then the shift variant.
+pub const LOG_SHIFT_ROWS: usize = PHASE_1_LOG_LEN - Word::LOG_BITS;
+
 /// Proves the first phase of the shift reduction.
 ///
 /// Builds the g and h multilinears and runs one sumcheck over their product.
@@ -52,13 +68,14 @@ where
 	Channel: IPProverChannel<F>,
 	A: Allocator,
 {
-	// Build the g multilinear for the public and hidden segments separately, then sum them. The
+	// Accumulate the g rows of the public and hidden segments separately, then collect both. The
 	// public words are the prefix of `words`, and each segment's key ranges are segment-relative.
-	let mut g = build_g::<_, P, _>(alloc, words.public, &key_collection.public, prepared);
-	let hidden_g = build_g::<_, P, _>(alloc, words.hidden, &key_collection.hidden, prepared);
-	for (slot, add) in iter::zip(g.as_mut(), hidden_g.as_ref()) {
-		*slot += *add;
-	}
+	let public = build_g::<_, P>(words.public, &key_collection.public, prepared);
+	let hidden = build_g::<_, P>(words.hidden, &key_collection.hidden, prepared);
+	let g = SparseShiftRows::from_segments([
+		(&public, &key_collection.public.dense_shift_enc),
+		(&hidden, &key_collection.hidden.dense_shift_enc),
+	]);
 
 	// BitAnd, IntMul and BinMul share the same `r_zhat_prime`.
 	let h = build_h(alloc, domain_subspace, prepared.bitand.r_zhat_prime);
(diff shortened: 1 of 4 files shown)


benji-potamus:
## Performance

Machine: this sandbox, `RUSTFLAGS="-C target-cpu=native"`, release. Baseline is `main` (`ada5b905`) built in a **separate scratch jj workspace with its own target dir**, and both sides were run back to back so they see the same machine conditions.

### Criterion — `shift_reduction_phases` (SHA256, 2^14-byte message)

| bench | baseline | branch | delta |
|---|---|---|---|
| **`phase1_run_sumcheck`** | **699.8 µs** | **344.0 µs** | **−356 µs, −51%** |
| `phase1_build_g_parts` | 21.65 ms | 22.28 ms | +3% |
| `phase1_build_h_parts` | 70.5 µs | 80.4 µs | untouched |
| `phase2_build_monster_segments` | 7.61 ms | 8.71 ms | untouched |
| `phase2_run_sumcheck` | 5.57 ms | 4.73 ms | untouched |

The untouched phases wander ±15% between the two runs — that is the noise floor here, and the sumcheck's −51% is well clear of it. `phase1_build_g_parts` no longer scatters, but it is dominated by the key scan (~500 K keys), so the change is invisible in it.

### Tracing spans, `binius-examples`

**keccak `--message-len 136`** (1 179 AND constraints) — 5 runs per side, best-of-5, since the runs are short and load-sensitive:

| span | baseline | branch |
|---|---|---|
| `scatter_shift_rows` (×2) | 376 µs + 373 µs | **gone** |
| phase-1 `run_sumcheck` | 502 µs | 299 µs |
| `build_h` | 782 µs | 845 µs (untouched) |
| `prover_phase_1` | 2.81 ms | **1.83 ms** (−35%) |
| `Prove` total | 6.76 ms | **5.90 ms** (−13%) |

The ~0.98 ms drop in `prover_phase_1` accounts for itself: ~0.75 ms of removed scatter plus ~0.2 ms of faster sumcheck.

**sha256 `--message-len 16384`** (93 863 AND constraints) — 3 runs per side:

| span | baseline | branch |
|---|---|---|
| `scatter_shift_rows` (×2) | 79 µs + 88 µs | **gone** |
| phase-1 `run_sumcheck` | 534–674 µs | 231–308 µs |
| `Prove` total | 155–208 ms | 155–269 ms — **no signal** |

That is the honest result on a large circuit: ~0.5 ms of addressable work out of a ~155 ms prove is ~0.3%, and end-to-end prove time is far noisier than that. The win is real but it is a **fixed** cost being removed, so it only shows up when the circuit is small enough for a fixed `2^15`-element cost to matter.

### Two corrections to numbers I quoted on the Linear issue

- The keccak baseline I posted there (~28 ms prove, `prover_phase_1` 41–51%, `build_h` 1–13 ms) was measured while the machine was loaded from a build. On a quiet machine the same circuit proves in ~6–9 ms with `build_h` at ~0.8 ms. The *proportions* held up — `prover_phase_1` is still ~40% of a small prove — but the absolute figures were inflated.
- Likewise, the first criterion baseline I took is not comparable to the after-numbers: phases I never touched had moved 2–3× between the two runs. Everything above comes from the paired scratch-workspace run instead.

### Still on the table, not in this PR

`build_h` allocates `zeroed_vec::<F>(1 << PHASE_1_LOG_LEN)` and then copies it into a pool buffer via `FieldBuffer::from_values_in` (`crates/prover/src/protocols/shift/monster.rs:118-127`) — two 512 KB allocations and a copy for 512 KB of data. It is now the largest fixed cost left in phase 1 on small circuits (~0.85 ms of the remaining 1.83 ms), and it is where the run-to-run variance concentrates. Independent of this change; worth its own issue.

benji-potamus:
## Performance — supersedes the earlier comment

**Ignore the perf comment above.** It measured the hand-rolled design this PR replaced, and its baseline was taken on a loaded machine. Everything here is fresh, paired, and on a quiet box: untouched phases land within ±4% between arms, versus ±15% in that earlier session.

Machine: this sandbox, `RUSTFLAGS="-C target-cpu=native"`, release. Each arm built in its own jj workspace with its own target dir, all run back to back with nothing else on the CPU.

Three arms, because "is this faster than `main`" turned out to be the less interesting question:

- **dense** — `main` (`d1bc6ffc`), the current bivariate-product prover over a materialized `g`.
- **rows** — the hand-rolled version this PR replaced (`4191742c`), sparse over `g`'s *rows*.
- **entries** — this PR (`d924e76e`), `g` as `(index, value)` entries through `SparseDenseProductSumcheckProver`.

### Phase-1 sumcheck (tracing `run_sumcheck` span, best-of-N)

| circuit | shifts → entries | dense | rows | **entries (this PR)** |
|---|---|---|---|---|
| sha256 `--message-len 16384` | 16 → 1024 | 474 µs | **235 µs** | 389 µs |
| keccak `--message-len 136` | 40 → 2560 | 463 µs | **286 µs** | 914 µs |

Criterion `shift_reduction_phases` (sha256 only, dense vs entries): `phase1_run_sumcheck` 532.4 µs → 439.0 µs, −17.5%. Consistent with the trace.

### What the numbers say

**The dense cost is constant** — a fixed `2^15` hypercube regardless of circuit. Confirmed: 474 µs on sha256 against 463 µs on keccak, for circuits two orders of magnitude apart in size.

**The entry cost is linear in entry count, with no per-round decay.** Entries are `n_shifts × 64` and every one of the 15 rounds does two random-access scalar `dense.get()` lookups per entry. 2.5× the entries buys 2.35× the time — almost exactly linear. Crossover against dense is around 1300 entries, i.e. **about 20 distinct shifts**: sha256 sits under it and wins, Keccak sits over it and **loses ~2×**.

**The rows arm beat both, on both circuits** — 1.65× faster than this PR on sha256, 3.2× on Keccak. It walked packed rows with `P::wide_mul` (SIMD, one deferred reduction per round) and its row count shrank as rows merged, where the shared prover does scalar random access at a count that never decays.

### Net effect on the phase

| `prover_phase_1`, best-of-N | dense | rows | entries (this PR) |
|---|---|---|---|
| sha256 | 17.3 ms | 15.5 ms | 16.4 ms |
| keccak | 2.54 ms | 1.59 ms | 2.20 ms |

So this PR does still improve phase 1 against `main` on both circuits — but on Keccak that is **entirely** because dropping `scatter_shift_rows` (2 × ~380 µs) more than pays for a sumcheck that got ~450 µs worse. The win is the scatter removal, not the sumcheck. End-to-end prove time shows no signal on sha256 (~145 ms, all arms) and is too noisy on Keccak to call.

### The trade-off, stated plainly

This PR buys a real simplification — +353/−151 instead of +593/−126, no bespoke sumcheck machinery in the shift prover, and reuse of a component that has its own proptests — and pays 1.65–3.2× on the phase-1 sumcheck against the version it replaced. On the largest circuits that is ~0.15 ms of a ~145 ms prove and plainly worth it; on shift-heavy small circuits it is the dominant term in phase 1.

I'm not reverting on my own judgement — @jimpo asked for this refactor and the simplification is genuine. But the cost is larger than I expected when I made the change, so it should be a decision rather than a side effect. Three ways forward, in the order I'd rank them:

1. **Keep it as is.** The absolute numbers are small and the code is much simpler.
2. **Keep it and make the shared prover faster** — packed access instead of scalar `get`, and a deferred reduction as the dense evaluators use. That helps every future caller, not just this one; it belongs on #2105's follow-up, not here.
3. **Go back to the rows version.** Fastest, but reintroduces ~200 lines of bespoke sumcheck the shared prover already generalizes.

Happy to do any of them.

jimpo (commented):
OK, actually SparseDenseProductSumcheckProver isn't exactly the right choice here, because the g polynomial is dense in the J variables. (Meaning there's sparsity of long strides). Create a custom Phase1SumcheckProver in the phase 1 module, that is an adaptation of the SparseDense one. It should switch strategies after the first 9 rounds, and then switch to a dense-dense product sumcheck (using bivariateproductcheckprover) for the J variable rounds.

benji-potamus:
Done — `Phase1SumcheckProver` in the phase-1 module, switching to `bivariate_product_prover` after the 9 row rounds, and simple in the SparseDense sense: linear per row, no pairing, no merging. Pushed as `0e57a8f8`.

Dropping the merge simplified more than the prover itself. Since a row contributes on its own, the two key segments' rows just **concatenate** — a shift both name contributes one row each at the same index and they add up wherever `g` is read. `merge_segments` and the sorted-index invariant are gone.

One thing to note: `RoundState` is private to `binius-ip-prover`, so the row stage keeps its own claim/coeffs pair rather than reusing it. Easy to switch if you'd rather export it.

## Performance — supersedes both earlier comments

Paired against `main` (`3a3382d7`) built in its own workspace, run back to back on a quiet machine.

### Phase-1 sumcheck

| | main (dense) | **this PR** |
|---|---|---|
| sha256, criterion `phase1_run_sumcheck` | 593.0 µs | **373.8 µs** (−37%) |
| sha256, trace best-of-3 | 488 µs | **270 µs** (−45%) |
| keccak, trace best-of-5 | 479 µs | **477 µs** (a wash) |

The two circuits differ because cost now tracks the **row count**, and the row list no longer shrinks: sha256 carries 8 + 16 = 24 rows, Keccak 1 + 40 = 41, each visited in all 9 row rounds. Against a dense cost that is constant at ~480 µs, 24 rows wins comfortably and 41 rows breaks even.

### What the no-merge simplification costs

Worth putting the number on it, since it is your call and not mine. Measured against the dense baseline in each measurement session, so the two are comparable as ratios:

| phase-1 sumcheck ÷ dense | merging rows (earlier) | **this PR (no merge)** | entries (#2105 prover) |
|---|---|---|---|
| sha256, 24 rows | 0.50 | **0.55** | 0.82 |
| keccak, 41 rows | 0.62 | **1.00** | 1.97 |

On sha256 the simplification is nearly free. On Keccak it gives up the win: merging collapsed the list as rows collided — and with 41 rows in a space that halves to 256, 128, … they collide a lot — so it stayed ahead of dense, where this version levels out.

Both are far better than routing through the entry-based prover, which was the thing actually worth fixing.

So: **simple and free on sha256, simple and break-even on Keccak.** If you want the Keccak win back it costs the merge; if you'd rather keep the code this shape, this is what it costs. Happy either way — say which.

Absolute numbers are small throughout (~0.2–0.5 ms of a 145 ms sha256 prove, ~5 ms Keccak), and end-to-end prove time shows no signal on either.

crates/prover/src/protocols/shift/phase_1.rs
```diff
@@ -52,25 +68,325 @@ where
...
+	/// Spreads the rows over the full shift space, as a dense multilinear of [`PHASE_1_LOG_LEN`]
+	/// variables. Rows at a repeated index add up; every row the system does not name stays zero.
+	///
+	/// The prover never needs this — [`Phase1SumcheckProver`] reads the rows where they are. It is
+	/// the dense `g` that the tests and the `shift_reduction` benchmark measure against.
+	///
+	/// ## Preconditions
+	///
+	/// * `self.log_rows() == LOG_SHIFT_ROWS`, so the rows still span the whole shift space
+	pub fn scatter<A: Allocator>(&self, alloc: &A) -> FieldVec<P, A> {
```
jimpo: If it's only used in tests, move it to the tests module.

crates/prover/src/protocols/shift/phase_1.rs
```diff
@@ -91,23 +407,21 @@ where
 /// `SumcheckOutput` containing the challenge vector and the final evaluation `gamma`.
 #[instrument(skip_all, name = "run_sumcheck")]
 pub fn run_phase_1_sumcheck<
-	F: Field,
+	F: BinaryField,
```
jimpo: Why can't this be just Field?

jimpo (changes requested):

crates/prover/src/protocols/shift/phase_1.rs
benji-potamus: Moved — it's now a `#[cfg(test)] impl SparseShiftRows` block inside the tests module.

The reason it was `pub` was that `m4-prover`'s phase-1 test used it cross-crate, where `#[cfg(test)]` can't reach. That test only wanted `sum g·h`, so it now sums `<row, h at shift_index * Word::BITS ..>` straight from its own rows and encodings instead. Nothing outside the tests depends on it any more.

crates/prover/src/protocols/shift/phase_1.rs
benji-potamus: It can — `PackedField` already has `WideMul` as a supertrait, so `BinaryField` was buying nothing here. I'd carried it over from the old signature without checking.

Relaxed to `Field` on `run_phase_1_sumcheck`, both `Phase1SumcheckProver` impl blocks, `shift_round_coeffs` and `build_g`. `prove_phase_1` keeps `BinaryField`, since `build_h` needs it for the Lagrange evaluation over the binary subspace.

jimpo (approved):

PR #2096, opened 2026-08-11
BINIUS-440: widen ShiftedValueIndex to a sequence of two shifts

tcoratger:
@jimpo I've stacked couple of PRs here to showcase the work from zero to end, we can go step by step and rebase along the way, no problem :) This one is the first step

Closes [BINIUS-440](https://linear.app/jimpo/issue/BINIUS-440).

Widens a constraint term from one shift to a sequence of two, applied inner first.
Pure representation change — no protocol change, no behavior change, transcripts byte-identical.

### The problem

A term names one value and one shift.
That cannot express maps a single shift has no spelling for.

Clearing the low nibble and leaving everything else where it started is the canonical example:

```
srl(4) then sll(4)      // clears bits 0..3, other bits unmoved
```

No single shift both drops bits and returns the rest in place.
Today the frontend has to commit an intermediate value for such a path.
Per the BINIUS-408 design pass, essentially every committed linear definition in the SHA family exists for exactly this reason — sha256 commits 2,005 of them.

### The idea

Give a term two shift slots instead of one.

```rust
pub struct ShiftedValueIndex {
    pub value_index: ValueIndex,
    pub shift_seq: [Shift; 2],
}
```

`shift_seq[0]` applies first, `shift_seq[1]` to its result.

The struct **stays at 8 bytes**: `ValueIndex` is a `u32`, and two `Shift { variant: u8, amount: u8 }` fill the 4-byte alignment exactly.
That matters — constraint systems carry millions of these on the prover hot path — so `shifted_value_index_fits_in_a_word` still holds.

### Canonical form

A lone shift goes in the **inner** slot, which splits every term into three classes:

| class | slots | spelling |
|---|---|---|
| unshifted | `s_1 = s_2 = 0` | `Shift::IDENTITY` twice |
| singly shifted | `s_2 = 0 != s_1` | the lone shift sits inner |
| doubly shifted | `s_2 != 0` | both slots carry work |

`validate` gains the two rules that make the pair meaningful:

- **`shift_seq[0].is_identity()` implies `shift_seq[1].is_identity()`.**
  Were the outer slot allowed to hold the lone shift, one map would have two spellings, and two terms denoting the same shifted word would not compare equal.
- **A genuine pair must not collapse** — `Shift::compose` reports `Composition::Pair`.
  `Single` means the frontend failed to merge two shifts that compose into one.
  `Zero` means it emitted a term whose every bit is cleared, which should have been deleted rather than encoded.

It deliberately does **not** canonicalize at the map level.
Per BINIUS-415 the `[Shift; 2]` spelling is not canonical — 108,571 irreducible spellings denote 74,341 distinct maps — and buying that back needs a table far larger than the few dozen shifts a real system uses.
Nothing in the reduction depends on it for correctness.

### The change

```
- ShiftedValueIndex { value_index, shift: Shift::srl(3) }
+ ShiftedValueIndex::single(value_index, Shift::srl(3))     // [srl(3), IDENTITY]
+ ShiftedValueIndex::new(value_index, [Shift::srl(3), Shift::sll(3)])   // a genuine pair
```

- `eval` applies both shifts in order.
- Accessors `inner()`, `outer()`, `is_unshifted()`, `is_doubly_shifted()` keep call sites reading like prose.
- The 9 existing constructors (`plain`, `sll`, `srl`, `sar`, `rotr`, and the `*32` family) delegate to `single`, so they keep working unchanged.
- Serialization writes both slots; `ConstraintSystem::SERIALIZATION_VERSION` 9 → 10 with a regenerated reference binary.

### Why nothing can mis-verify yet

The reduction still folds one shift per key.
So the outer slot is the identity at every construction site, and three places `debug_assert` it:

- `prover/protocols/shift/key_collection.rs` — where keys are built
- `verifier/protocols/shift/monster.rs` — both the generic and native evaluation paths
- `m4-prover/witness/operand_columns.rs` — where operand stripes are materialized

A term carrying both slots would be reduced against the wrong shifted word.
Widening the reduction to the pair is what lifts these.

### Scope

- **changed:** 16 files — `core` (the type, `validate`, 2 error variants), `frontend` (the lowering and its tests, `stat.rs`), `prover` + `verifier` + `m4-prover` (read `inner()`, assert on `outer()`)
- **new:** `crates/core/test_data/constraint_system_v10.bin`, replacing the v9 reference
- **left untouched:** the reduction itself, and the frontend's own single-shift `WireExprTerm` — the frontend keeps emitting lone shifts

### Verification

- `clippy -p binius-core -p binius-frontend -p binius-prover -p binius-verifier -p binius-m4-prover --all-targets` — clean
- `binius-core` 113 tests, `binius-frontend` 199, shift reduction 9+9, `binius-m4-prover` 44 — all pass
- `prove_verify` 14/14 and the shift round trip pass **unchanged**, which is the evidence the transcript did not move
- `sha256`/`keccak`/`blake3` `check-snapshot` all match — no count moves while the outer slot is the identity

### Tests added

- serialization round-trips both slots on a genuine pair
- `eval` applies the two shifts inner-first — order matters, and the reverse order clears a different nibble
- the three term classes read off the sequence
- `validate` rejects a lone shift in the outer slot, a pair collapsing to `Single`, and a pair collapsing to `Zero`
- `validate` accepts a genuine pair, and checks the amount bound on *both* slots


crates/verifier/src/protocols/shift/monster.rs
```diff
@@ -193,9 +200,10 @@ where
 			let mut constraint_eval = E::zero();
 			for (operand_id, operand) in constraint.as_ref().iter().enumerate() {
 				for svi in operand {
-					let variant = svi.shift.variant as usize;
+					debug_assert!(!svi.is_doubly_shifted(), "{DOUBLE_SHIFT_UNSUPPORTED}");
```
jimpo:
```suggestion
					assert!(!svi.is_doubly_shifted(), "{DOUBLE_SHIFT_UNSUPPORTED}");
```

crates/verifier/src/protocols/shift/monster.rs
```diff
@@ -233,9 +241,10 @@ where
 				let mut constraint_eval = F::ZERO;
 				for (operand_id, operand) in constraint.as_ref().iter().enumerate() {
 					for svi in operand {
-						let variant = svi.shift.variant as usize;
-						let index =
-							(variant * Word::BITS + svi.shift.amount as usize) * ARITY + operand_id;
+						debug_assert!(!svi.is_doubly_shifted(), "{DOUBLE_SHIFT_UNSUPPORTED}");
```
jimpo:
```suggestion
						assert!(!svi.is_doubly_shifted(), "{DOUBLE_SHIFT_UNSUPPORTED}");
```

jimpo (approved):

jimpo:
Thanks! I guess we're going for it :-)

PR #2097, opened 2026-08-11
BINIUS-442: re-key the prover and verifier shift tables on the shift pair

tcoratger:
Closes [BINIUS-442](https://linear.app/jimpo/issue/BINIUS-442).

Moves both sides off "one shift per key" and onto "one shift *sequence* per key", while the outer slot is still always the identity.
Pure refactor — no protocol change, transcripts byte-identical.

> Rebased onto main after the shift-reduction surgery (#2095, #2113, #2114, #2115). See **What the rebase changed** below for what that absorbed and what is left.
>
> Now also on top of #2121, which is why the earlier CI runs on this branch were red. Compiled with `-Ctarget-cpu=native`, an AVX-512 runner picks a packed width of 4, and the one-gate circuit in `cli_artifacts` had a lift block narrower than that — the pre-existing panic #2121 fixes. It reproduced identically on this branch's base, so it was never this diff.

### The problem

Both sides sized their shift tables by the whole shift space, using the fact that it is small.

- **Prover:** building the dense re-encoding collected a segment's shifts into a `[None; 512]` on the stack, and inverted it through a second 512-entry table.
- **Verifier:** `shift_scalars` was a `[E; 512]`, fanned out to `512 * arity` by the operand batching coefficients.

Keyed on a *sequence* of two shifts, that space is $512^2 = 262{,}144$.
The stack arrays cannot survive that, and the verifier's fan-out would reach roughly 1.5M multiplications at BMUL's arity of six.

### The idea

**The space grew; the occupancy did not.**
A segment still names at most one sequence per shifted value index — a few dozen in practice.
So stop indexing by a table over the space, and index by search over what is present.

And on the verifier side, **never build the square at all**, because the per-key weight factorizes:

```
key weight = eq(r_v1, v1) * eq(r_s1, s1)  *  eq(r_v2, v2) * eq(r_s2, s2)  *  lambda^(operand+1)
             \_____ inner table ______/     \_____ outer table ______/
```

Two 512-entry tables, not one of 262,144.
The cost is **one extra multiply per term**.

### Core

Where a shift sits in the enumeration of `(variant, amount)` spellings is a property of the shift, not of either side of the protocol. It had grown four copies — a free function in the prover, another in the verifier, and two test-local closures — so it moves onto the type as `Shift::index()`:

```
variant indexes runs of Word::BITS, amount indexes within a run

    [ Sll 0 .. Sll 63 | Slr 0 .. Slr 63 | ... | Rotr32 0 .. Rotr32 63 ]
      0          63     64         127          448             511
```

The prover's row axis and the verifier's weight tables now read the same definition, which is what makes "the two sides agree on the layout" a fact rather than a convention. It panics on an amount at or above the word width, which would index into the next variant's run and share an entry with another shift — reachable only by building a `Shift` field-wise, since the constructor already rejects it.

### Prover

```
- shifts: Vec<(ShiftVariant, u8)>          // built via [None; 512], inverted via [u16; 512]
+ shifts: Vec<[Shift; 2]>                  // BTreeSet collects + orders; binary search inverts
```

- The sorted set dedupes and orders in one pass, over the sequences present rather than the space they come from.
- `inverse()` — a 512-entry lookup table — becomes `dense_idx()`, a binary search over the sequences the segment uses.
- `Key::dense_shift_idx` is a `u16`, so the build now checks that bound explicitly instead of relying on an implicit "cannot exceed 512".
- Serialization writes both slots per entry and rejects a list that is not strictly ascending, which is what makes a dense index meaningful.
- `shift_indices()` places a sequence on the reduction's row axis by its **inner** shift. The axis is one slot wide, so the outer slot must be the identity there; a `debug_assert` says so, naming the shared `DOUBLE_SHIFT_UNSUPPORTED` reason.
- Phase 2's scalars gain the outer slot's factor. `OuterSlotWeights` holds one tensor per outer axis, so the weights stay at `2 * SHIFT_COUNT` entries rather than `SHIFT_COUNT^2`.

### Verifier

`shift_scalars` becomes a two-field struct, one table per slot:

```
- shift_scalars: &[E; SHIFT_COUNT]
+ ShiftScalars { inner: &[E; SHIFT_COUNT], outer: &[E; SHIFT_COUNT] }
```

- `encode_operation_input` lays the inner table before the outer one; the split is the inverse.
- `operand_shift_scalar_table` keeps the lambda fan-out on the **inner** table only.
- A term's outer weight multiplies in where the term is read.

New `SHIFT_COUNT` names $512$ — the spellings one shift slot can take — replacing the open-coded `SHIFT_VARIANT_COUNT * Word::BITS`.

### What the rebase changed

Main got there first on two of the pieces this PR originally carried, so they are gone from the diff:

- **The sparse re-encoding already exists.** #2095 landed `DenseShiftEncoding` keyed on a single shift, and `SparseShiftRows` keyed on the full-space shift index. What is left here is widening that key to a pair and removing the two stack arrays the square would break.
- **`scatter_shift_rows` is gone.** #2095 replaced it with `SparseShiftRows`, so `crates/m4-prover/src/shift.rs` needs no change at all now and `phase_1.rs` only moves its test onto `shift_indices()`.

One deliberate difference from the pre-rebase version: the h evaluation stays **out** of the inner table. #2115 moved it into `check_eval`, and this keeps it there rather than folding it back in.

### Why today's tests still pass unchanged

The outer slot is the identity on every key, so the outer table is the equality indicator at the all-zero point:

```
index 0 = (Sll, 0) -> one
every other index  -> zero
```

A well-formed term's weight is therefore exactly what a single-shift reduction gives it.
That is the point of landing this before any protocol change depends on it.

### Scope

- **added:** `Shift::index` in core `shift.rs`, replacing four copies of the layout formula
- **changed:** prover `key_collection.rs` (the encoding), prover `monster.rs` (per-slot weights), prover `mod.rs` (the shared assertion reason), verifier `monster.rs` + `verify.rs` (the two tables), verifier `mod.rs` (`SHIFT_COUNT`), and two tests in prover `phase_1.rs`
- **left untouched:** the reduction's phase structure — that is BINIUS-445

### Verification

Run on the rebased commit in a fresh worktree:

- `clippy -p binius-prover -p binius-verifier -p binius-m4-prover --all-targets` — clean
- shift reduction: 23 prover + 6 verifier tests pass, plus the `Shift::index` tests in core
- `test_shift_prove_and_verify` and the 60 `binius-m4-prover` tests pass **unchanged**, which is the evidence the transcript did not move

### Tests added

- the encoding distinguishes two sequences that share an inner shift but differ outside — keyed on the inner shift alone they would collide into one row
- indexing is the inverse of decoding, on input given unsorted and with a repeat, which also pins the sort and the dedupe
- serialization rejects an unordered list, a repeated sequence, and an out-of-range amount in **either** slot
- the outer weight reaches the evaluation *as a factor*: scaling the one entry every term reads scales the whole evaluation, zeroing it kills the evaluation, and filling the entries no term names changes nothing
- `Shift::index` places a shift by variant then amount, and every spelling in the alphabet gets a slot of its own
- `Shift::index` panics on an amount at the word width, the case that would alias two shifts onto one entry

### DIFF
--- a/crates/core/src/constraint_system/shift.rs
+++ b/crates/core/src/constraint_system/shift.rs
@@ -450,6 +450,29 @@ impl Shift {
 		!self.is_identity() || matches!(self.variant, ShiftVariant::Sll)
 	}
 
+	/// Where this shift sits in the enumeration of every `(variant, amount)` spelling.
+	///
+	/// The variant indexes runs of `Word::BITS`, and the amount indexes within a run:
+	///
+	/// ```text
+	///     [ Sll 0 .. Sll 63 | Slr 0 .. Slr 63 | ... | Rotr32 0 .. Rotr32 63 ]
+	///       0          63     64         127          448             511
+	/// ```
+	///
+	/// A reduction keying one table entry per spelling addresses it by this index.
+	/// So does the prover's multilinear over the same axis pair, which is what lets the two agree.
+	///
+	/// # Panics
+	///
+	/// Panics if the amount is not below `Word::BITS`.
+	/// Above it a shift would index into the next variant's run, sharing an entry with another
+	/// shift.
+	#[inline]
+	pub const fn index(self) -> usize {
+		assert!((self.amount as usize) < Word::BITS, "shift amount is not below the word width");
+		self.variant as usize * Word::BITS + self.amount as usize
+	}
+
 	/// Applies this shift to a word and returns the result.
 	///
 	/// # Performance
@@ -1282,6 +1305,42 @@ mod tests {
 			deserialize_amount(ShiftVariant::Sll, 63).unwrap(),
 			ShiftedValueIndex::sll(ValueIndex::constant(0), 63)
 		);
+	}
+
+	#[test]
+	fn index_places_a_shift_by_variant_then_amount() {
+		// Runs of `Word::BITS` amounts, one run per variant. A table indexed the other way round
+		// would weight a shifted word by another shift's scalar.
+		//
+		//     [ Sll 0 .. Sll 63 | Slr 0 .. Slr 63 | ... ]
+		//       0          63     64         127
+		assert_eq!(Shift::IDENTITY.index(), 0);
+		assert_eq!(Shift::sll(5).index(), 5);
+		assert_eq!(Shift::srl(0).index(), Word::BITS);
+		assert_eq!(Shift::srl(3).index(), Word::BITS + 3);
+		assert_eq!(Shift::rotr32(31).index(), ShiftVariant::Rotr32 as usize * Word::BITS + 31);
+
+		// Every spelling lands in its own slot, inside the enumeration.
+		let mut seen = vec![false; ShiftVariant::ALL.len() * Word::BITS];
+		for variant in ShiftVariant::ALL {
+			for amount in 0..variant.max_amount() {
+				let index = Shift::new(variant, amount).index();
+				assert!(!seen[index], "{variant:?} {amount} shares an index");
+				seen[index] = true;
+			}
+		}
+	}
+
+	// An amount at the word width would index into the next variant's run, aliasing another shift.
+	// The fields are public, so a hand-built shift can carry one even though `new` rejects it.
+	#[test]
+	#[should_panic(expected = "shift amount is not below the word width")]
+	fn index_rejects_an_amount_at_the_word_width() {
+		let shift = Shift {
+			variant: ShiftVariant::Sll,
+			amount: Word::BITS as u8,
+		};
+		let _ = shift.index();
 	}
 
 	#[test]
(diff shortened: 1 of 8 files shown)


jimpo:
Needs rebase onto main

tcoratger:
> Needs rebase onto main

@jimpo Done :)

tcoratger:
@jimpo Let me know when ready to merge, I've seen a lot of conflicts coming i can rebase when needed :)

jimpo:
@tcoratger Sorry about that, did major surgery on shift reduction to prepare for double shifts. Please fix this up, it'll be the next thing to go in for shift reduction!

jimpo (approved):

PR #2098, opened 2026-08-11
BINIUS-443: evaluate the inner-phase shift-indicator weight at the intermediate point

tcoratger:
Closes [BINIUS-443](https://linear.app/jimpo/issue/BINIUS-443).

> **Stacked on #2097** ([BINIUS-442](https://linear.app/jimpo/issue/BINIUS-442)) → #2096 ([BINIUS-440](https://linear.app/jimpo/issue/BINIUS-440)). Review in that order; this PR's diff is only its own increment.

Adds the second of the two weight families the three-phase reduction needs.
Almost no new evaluation code — the existing evaluator already had the right shape.

### The problem

Today's reduction has one weight family: the shift indicator contracted against the *oblong* factor.

The three-phase reduction needs two, one per shift slot:

```
outer   H(K, S_2) = sum_i delta_D(r_ihat, ihat) * shift-ind_op2(i, K, S_2)
inner   h(J, S_1) = shift-ind_op1(r_k, J, S_1)
```

The outer family is exactly today's, unchanged.
The inner one is the **bare** indicator at the intermediate point $r_k$ the outer phase produced — no oblong factor, because that phase already summed the output bit index out.

### The idea

Both are the *same contraction* under different weights.

A multilinear extension in the output-bit argument **is** the equality-weighted sum over that argument:

```
shift-ind~_op1(r_k, j, s_1) = sum_i eq~(r_k, i) * shift-ind_op1(i, j, s_1)
```

The existing evaluator is already generic in that weight vector — it takes an arbitrary 64-entry slice, not specifically the Lagrange basis.
So the second family is the same kernel called with the equality indicator of $r_k$ in place of the Lagrange weights.

### The change

```
+ pub fn evaluate_inner_h_op(r_k, r_j, r_s) -> [E; SHIFT_VARIANT_COUNT] {
+     let eq_r_k = eq_ind_partial_eval_scalars(r_k);
+     evaluate_h_op(&eq_r_k, r_j, r_s)     // the shared kernel, different weights
+ }
```

That is the whole of it.
The shared kernel's doc now names it as a kernel serving two families rather than as one function with one purpose.

### Scope

- **new:** one function and its three tests, in the verifier's shift module
- **changed:** the shared evaluator's header doc only — no behavior change
- **left untouched:** every existing caller, which still gets the outer family

### Verification

- `clippy -p binius-verifier --all-targets` — clean
- verifier shift module: 14 tests pass (11 existing + 3 new)

### Tests added

- **at a hypercube vertex the weight is the integer indicator.** An extension agrees with the function it extends at a vertex, so driving all three points to vertices must reproduce the integer conditions — checked against a direct reference for all eight variants, over 1,024 random $(k, j, s)$ triples.
- **an identity shift weights by plain equality.** A degenerate slot contributes `eq(r_k, r_j)` and nothing else, so composing it with the other slot's weight leaves that weight standing alone. This is the compatibility property the staged rollout leans on. All eight variants are the identity at amount zero, so all eight must agree — including the half-word forms.
- **multilinear in each of the three points.** The reduction binds two of them by sumcheck and reads the third from the outer phase, so soundness needs linearity in all three.


jimpo:
This seems to be based on the old spec. See the new writeup here: https://github.com/binius-zk/binius.xyz/pull/124

tcoratger:
@jimpo Instead just opened https://github.com/binius-zk/binius64/pull/2126

PR #2099, opened 2026-08-11
BINIUS-444: build the outer-phase G tables from intermediate words

tcoratger:
Closes [BINIUS-444](https://linear.app/jimpo/issue/BINIUS-444).

> **Stacked on #2098** ([BINIUS-443](https://linear.app/jimpo/issue/BINIUS-443)) → #2097 → #2096. Review in that order; this PR's diff is only its own increment.

The one piece of the double-shift prover that does not already exist in some form.

### The idea

The outer phase's multilinear is defined with an inner sum over the source bit index:

```
G[k][s_2] = sum_y sum_{op1, s_1} Z~(r'_x, y, s_1, s_2) * [ sum_j shift-ind_op1(k, j, s_1) * w[y]_j ]
                                                          \_________ this inner sum _________/
```

That inner sum has a closed form:

```
sum_j shift-ind_op1(k, j, s_1) * w[y]_j  =  op1(w[y], s_1)_k
```

It is **just bit $k$ of the intermediate word**.

Why: every shift variant has the *at-most-one-source-bit* property.
An output position reads exactly one source position — arithmetic right shift merely has *many* output positions read the sign bit, which is many-to-one, not one-to-many.
So the sum has at most one nonzero term, and the indicator is a permutation with holes.

**The intermediate word therefore costs one machine shift, not 64 field products.**

### The change

The new builder differs from the existing one in exactly two places:

| | row named by | bits scattered |
|---|---|---|
| inner slot | the inner shift | `w[y]` |
| outer slot | the outer shift | `op1(w[y], s_1)`, the intermediate word |

Both are fixed by one slot parameter, so this is **one function** rather than two copies of a ninety-line loop.
Everything else is shared: the per-key accumulation, the packed-mask bit scatter, the rayon fold and reduce, and the final scatter over the shift axes.

```
+ pub fn build_outer_g(..)  -> build_slot_g(.., ShiftSlot::Outer)
  pub fn build_g(..)        -> build_slot_g(.., ShiftSlot::Inner)
```

A shift that clears the word costs nothing: the scatter loop is driven by the remaining nonzero bytes.

### A subtlety keying on sequences introduced

Two sequences can agree on the slot a phase binds:

```
[srl(3), sll(3)]  \
                   >  agree on the outer shift -> must accumulate into ONE row
[srl(5), sll(3)]  /
```

The dense index is per *sequence*, so an accumulator keyed by it would give these two separate rows — and the scatter would then copy both to the same offset, silently dropping one.

So the accumulator holds one row per distinct shift **that slot** takes, and a per-slot projection maps each sequence onto its row.
This was latent before: with the outer slot always the identity, distinct sequences implied distinct inner shifts, so the old keying happened to be safe.

### Scope

- **new:** the outer builder, the slot parameter, the per-slot row projection, and a predicate for whether any sequence needs both slots
- **changed:** the existing builder becomes a thin wrapper; the scatter keys off slot rows
- **left untouched:** the reduction still runs two phases — the outer builder has no caller yet, which is BINIUS-445. The two-phase entry point now asserts no sequence carries an outer shift.

### Verification

- `clippy -p binius-prover --all-targets` — clean
- prover shift module: 18 tests pass (13 existing + 5 new)
- `prove_verify` 14/14 pass **unchanged**

### Tests added

The load-bearing one is a **naive indicator reference**: it forms the intermediate word's bit $k$ as the *field sum* over all 64 source positions, spelled out, rather than with the machine shift.

Two source bits reaching one output position would cancel in that sum and survive the machine shift — so agreeing with it is what pins the at-most-one-source-bit property the whole phase rests on.

- the built form equals that reference on a fixture of genuine pairs, including two that share an outer shift and one whose inner shift clears the word
- and again over **every ordered variant combination** (all 64), so no variant's source map can be wrong unnoticed
- a degenerate inner slot makes the outer builder equal the inner builder on mirrored sequences — the compatibility property the staged rollout leans on
- the per-slot projection merges sequences sharing the bound slot, and each sequence maps to the row of its own shift
- the "needs a third phase" predicate: false for lone shifts, true for one genuine pair, false for an empty segment

### DIFF
--- a/crates/prover/src/protocols/shift/key_collection.rs
+++ b/crates/prover/src/protocols/shift/key_collection.rs
@@ -121,6 +121,111 @@ impl DenseShiftEncoding {
 		// `new` bounds the length, so every index it yields fits.
 		index as u16
 	}
+
+	/// Whether any sequence carries work in its outer slot.
+	///
+	/// False means every sequence is a lone shift.
+	/// The outer phase would then bind axes whose weight is already the identity indicator, so a
+	/// reduction can skip that phase entirely.
+	pub fn has_outer_shift(&self) -> bool {
+		self.shifts
+			.iter()
+			.any(|shift_seq| !shift_seq[ShiftSlot::Outer as usize].is_identity())
+	}
+
+	/// Groups the segment's sequences by the shift one slot of them takes.
+	///
+	/// A phase binds one slot's axes, so its accumulator holds one row per distinct shift *that
+	/// slot* takes — not one per sequence.
+	/// Several sequences can share a slot shift, and their contributions belong in one row:
+	///
+	/// ```text
+	/// [srl(3), sll(3)]  \
+	///                    >  agree on the outer shift -> one row
+	/// [srl(5), sll(3)]  /
+	/// ```
+	///
+	/// This is the map that sends each sequence to its row.
+	pub fn slot_rows(&self, slot: ShiftSlot) -> SlotRows {
+		let shift_of = |shift_seq: &[Shift; 2]| shift_seq[slot as usize];
+
+		// The distinct shifts the slot takes, ascending, one row each.
+		let shifts = self
+			.shifts
+			.iter()
+			.map(shift_of)
+			.collect::<BTreeSet<_>>()
+			.into_iter()
+			.collect::<Vec<_>>();
+
+		// Each sequence's row, found by search over that ordered list.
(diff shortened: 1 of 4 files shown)


tcoratger:
Obsolete as per [binius.xyz#124](https://github.com/binius-zk/binius.xyz/pull/124)

PR #2100, opened 2026-08-11
BINIUS-441: materialize m4 operand columns through both shifts

tcoratger:
Closes [BINIUS-441](https://linear.app/jimpo/issue/BINIUS-441).

Rebased onto `main` and retargeted there. It was stacked under the three-phase reduction only to keep that stack linear; it never depended on any of it, and it does not depend on the reworked protocol either.

Its subject is the **value a shifted index denotes**, which [binius.xyz#124](https://github.com/binius-zk/binius.xyz/pull/124) carries over unchanged:

```
val(y, op_1, s_1, op_2, s_2) = op_2( op_1( w[y], s_1 ), s_2 )
```

### The problem

The single-instance prover gets the second shift for free. It reads operands through the value vector, which applies a term's whole shift sequence.

m4 does not. It applies the shift straight into the output stripe, so its term representation carried exactly one:

```
Shifted { row, variant, amount }        // one shift, streamed into the stripe
```

Anything doubly shifted was rejected by a `debug_assert`.

### The change

A doubly shifted case that chains the pair, and a dispatch on which slots do work:

```rust
match (inner.is_identity(), outer.is_identity()) {
    (true,  true)  => TermWords::Plain(row),
    (false, true)  => single(inner),
    (true,  false) => single(outer),
    (false, false) => TermWords::DoublyShifted { row, inner, outer },
}
```

Chaining needs **no scratch stripe**. The intermediate word lives in a register per instance:

```
out[i] = outer(inner(row[i]))
```

So the stripe is still walked once and nothing is allocated. The ticket suggested a scratch stripe plus a second pass; this is simpler and strictly cheaper.

The trade is that both shifts are resolved per word rather than hoisted out of the loop — two predictable branches an instance, paid only by terms that genuinely need both slots. Composing the pair into a single source-bit map would hoist them back out if a profile ever asks.

### The fast paths are kept

| term | path | cost |
|---|---|---|
| unshifted | copy the row | one pass |
| singly shifted | stream through the resolved shift | one pass, dispatch hoisted |
| doubly shifted | chain per word | one pass, dispatch per word |
| constant | shifted once here, splatted | free per instance |

A degenerate slot drops out, so **a lone shift costs exactly what it did before sequences existed** — no no-op second pass over the stripe. Constants are shifted once and shared across every instance, now through both slots.

The canonical form places a lone shift inner, so the outer-only arm is unreachable for a canonical term. It costs nothing and keeps the match correct for a term that arrives spelled the other way.

### Tests

The existing proptest compared every batch row against the single-instance reference for one drawn shift. It now draws a **sequence**:

- the outer slot is the identity in some draws and a working shift in others, so both the singly and the doubly shifted branches are reached;
- the pair is deliberately **not** filtered for irreducibility, since both routes apply whatever sequence they are given and a collapsible pair is a valid input.

The single-instance path is the reference throughout, which is what makes this a check that m4 now agrees with the semantics rather than a restatement of the new code.

### Scope

- **changed:** one file — the term enum grows a case, the two stripe writers gain an arm, and the term resolver dispatches on which slots do work
- **removed:** the `debug_assert` rejecting doubly shifted terms
- **left untouched:** the column layout, the parallel stripe partitioning, and every other operand path

### Verification

- `clippy -p binius-m4-prover --all-targets` — clean
- `binius-m4-prover` 38 tests pass, including the 17 in `witness::operand_columns`

### DIFF
--- a/crates/m4-prover/src/witness/operand_columns.rs
+++ b/crates/m4-prover/src/witness/operand_columns.rs
@@ -272,6 +272,13 @@ struct ValueWords<'a> {
 }
 
 /// The words one operand term contributes, one per instance of the batch.
+///
+/// A term carries a sequence of two shifts, and the cases below split on how much of it does work.
+///
+/// - A constant is shifted once here and shared, whatever its sequence.
+/// - A hidden row keeps the single-shift streaming path when one slot is degenerate.
+///
+/// So a lone shift costs one pass over the stripe, as it does without a sequence at all.
 enum TermWords<'a> {
 	/// One already-shifted constant, repeated across every instance.
 	Splat(Word),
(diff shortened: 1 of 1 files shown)


tcoratger:
@jimpo I think that this one still is valid :)

jimpo (approved):

PR #2101, opened 2026-08-11
BINIUS-445: wire the three-phase shift reduction end to end

tcoratger:
Closes [BINIUS-445](https://linear.app/jimpo/issue/BINIUS-445).

> **Stacked on #2100** ([BINIUS-441](https://linear.app/jimpo/issue/BINIUS-441)) → #2099 → #2098 → #2097 → #2096. Review in that order; this PR's diff is only its own increment.

**This is the protocol change.** Prover and verifier move together.
Where double-shifted value indices first actually verify.

### The shape

Three sumchecks in sequence, peeling a term's shift sequence from the output end inward:

| phase | binds | against | weight |
|---|---|---|---|
| **outer** | `(k, s_2, v_2)` | the intermediate words | the oblong factor |
| **inner** | `(j, s_1, v_1)` | the witness words | the bare indicator at $r_k$ |
| **words** | `r_y` | the monster multilinear | both slots' weights, multiplied |

Each slot phase is one 15-variate bivariate product, so one runner serves both.

### Why the order is forced

This is the load-bearing detail, and it is now in a doc comment.

The oblong factor $\delta_D(r_{\hat i}, \hat i)$ attaches to the **output** bit index.
That index couples to the rest of the sum *only* through the outer indicator.

- Sum it out → a function of `(k, s_2)` alone → **one 15-variate sumcheck**.
- Bind the inner shift first → an output-side factor still depending on `s_2` → the sumcheck runs over `(j, s_1, s_2)`: **18 variates**.

### How the phases compose

The outer sumcheck closes on a product. The inner phase's claim *is* that product, because the outer phase's own evaluation is folded into the inner phase's per-key weight:

```
key scale = H_eval * eq(r_s2, s_2) * eq(r_v2, v_2)
```

So `sum(h_inner * g_scaled) = H_eval * G_eval = gamma_outer`, and no division is needed anywhere.

The words phase then weights a key by the product of both slots' tables, plus each phase's `h` evaluation — the two-table structure #2097 put in place, now with real challenges on both sides.

### The output interface does not change

One evaluation claim on the witness at $(r_j, r_y)$, whatever ran.
Ring-switching and everything downstream is untouched — and there is a test that pins exactly that.

### The degeneration stopgap — included

The ticket listed this as pending a call. It is implemented, and it is worth a look.

When no term needs both slots, the outer phase is **skipped entirely**.
Both sides derive that predicate from the constraint system, so it selects the same structure on each with **nothing sent between them**.

- **Upside:** a circuit of lone shifts produces the transcript it produced before this change, *byte for byte* — not one with an extra phase whose weight is the identity indicator on every key. Every existing circuit pays nothing.
- **Downside to weigh:** the three-phase path is then exercised only by the new double-shift tests, never by the existing suite.

### Scope

- **renamed:** the phase modules, to names that survive a third phase (`phase_1` → `slot_phase`, `phase_2` → `words_phase`), with the benchmark following
- **new:** the outer and inner phase provers, the shared slot-phase output, the per-slot weights and per-operation scalars, the verifier's optional outer point, and the shared phase predicate
- **changed:** the monster builder now takes both slots' weights instead of deriving one; the verifier's evaluator branches on whether the outer phase ran
- **lifted:** the verifier's assertion that no term is doubly shifted — both evaluation paths now weight a term by both slots

### Verification

- `clippy -p binius-prover -p binius-verifier -p binius-m4-prover --all-targets` — clean
- prover shift 18, verifier shift 14, `binius-core` 113, `binius-m4-prover` 44 — pass
- shift integration **5/5 in both debug and release**; `prove_verify` 14/14 **unchanged**

> Run in **both** profiles deliberately: a stale `debug_assert` on the verifier's evaluation path only fires with debug assertions on, and a release-only run would have missed it. That assertion is lifted here.

### Tests added

- **genuine shift pairs prove and verify**, on a hand-built system whose AND and ZERO constraints carry irreducible pairs, with the outer phase confirmed to have run
- **every irreducible variant combination alone.** Run together, one pair's contribution could mask another's; one constraint system per pair keeps a mis-handled variant from hiding.
- **an unsatisfied claim is rejected**
- **lone shifts skip the outer phase**, and both sides agree they do

### DIFF
--- a/crates/m4-prover/src/shift.rs
+++ b/crates/m4-prover/src/shift.rs
@@ -14,9 +14,9 @@ use binius_prover::{
 	fold_word::fold_words,
 	protocols::shift::{
 		KeyCollection, KeySegment, OperatorClaims, PreparedOperatorClaims,
-		monster::{build_h, build_monster_segments},
-		phase_1::{PHASE_1_LOG_LEN, build_g, run_phase_1_sumcheck},
-		phase_2::run_sumcheck,
+		monster::{SlotWeights, build_h, build_monster_segments, outer_h_evals},
+		slot_phase::{SLOT_PHASE_LOG_LEN, SlotPhaseOutput, build_g, run_slot_sumcheck},
+		words_phase::run_sumcheck,
 	},
 };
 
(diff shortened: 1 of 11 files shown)


tcoratger:
Obsolete as per https://github.com/binius-zk/binius.xyz/pull/124

PR #2102, opened 2026-08-11
BINIUS-446: the batched three-phase shift reduction

tcoratger:
Closes [BINIUS-446](https://linear.app/jimpo/issue/BINIUS-446).

> **Top of the stack.** #2101 ([BINIUS-445](https://linear.app/jimpo/issue/BINIUS-445)) → #2100 → #2099 → #2098 → #2097 → #2096. Review bottom-up; this PR's diff is only its own increment.

Gives the m4 prover the outer phase, and the builder for its intermediate rows.

### The idea — cleaner than it sounds

In the single-instance prover the intermediate word is formed with **one machine shift** over packed bits.

Here the instance axis has already been folded, so a bit position is no longer a bit — it is a field element, and a machine shift cannot touch it.

**It does not need to.** Every shift variant reads *at most one source bit per output position*, and folding over instances is linear in those bits. So the shift and the fold commute:

```
fold(op(w, s))[k] = fold(w)[source_bits[k]]
```

Forming the intermediate row is therefore a **gather** over the 64 folded elements — exactly the permutation-with-holes the unfolded path applies to bits.

`sar` is no exception: it makes *many* output positions read the sign bit, which is many-to-one, not one-to-many.

### Benchmark

The alternative is contracting against the shift indicator: `sum_j shift-ind(k, j, s) * row[j]` over all 64 source positions.

| route | 8,192 intermediate rows |
|---|---|
| indicator contraction | 73.5 ms |
| gather | 0.65 ms |
| **speedup** | **113×** |

Both routes are asserted equal before timing. Reported by an `#[ignore]`d test, so it runs on demand rather than in CI.

### The change

```
+ Shift::source_bits() -> [Option<u8>; 64]        // in binius-core
+ shift_folded_word(row, shift)                    // the gather
+ build_outer_g_from_folded_words(..)              // alongside the existing inner builder
```

and `m4_prover::shift::prove` gains the same two-branch phase structure as the single-instance prover, reusing its subroutines throughout:

| phase | hidden segment built by |
|---|---|
| outer | the folded row gathered through its inner shift |
| inner | the folded row itself |
| words | the row contracted along its bit axis |

The public segment goes through the single-instance builder in each case, and the two are summed.
`build_h`, `build_inner_h`, `build_monster_segments` and the words sumcheck are reused unchanged.

### On exposing the source map

The gather needs it, so `binius-core` now exposes it.

The composition tests already carried a **private copy of the same map** as their independent oracle. They now share the public one — still pinned against the word operations by its own test, so the chain that validates the composition rules is unchanged, just no longer duplicated.

### Scope

- **new:** the source map on the shift type, the folded-row gather, the outer builder, and the third phase in the batched prover
- **changed:** the existing batched builder is now a thin wrapper on a slot-parameterized one; the composition tests drop their duplicate map
- **left untouched:** every m4 caller — the reduction's output interface does not depend on how many phases ran

### Verification

- `clippy --workspace --all-targets` — clean
- `binius-m4-prover` 47, `binius-core` 113, shift integration 5/5 — pass
- `prove_verify` 14/14 pass **unchanged**
- run in both debug and release

### Tests added

- **the gather equals folding the shifted words**, over every canonical shift, so a variant whose source map is wrong cannot hide
- **the batched round trip on genuine shift pairs**, verified against the single-instance verifier, with the outer phase confirmed to have run
- **the outer-phase identity**: summing the `G·H` inner products reconstructs the lambda-batched operand claim — which is what pins the gather against the operand columns the AND check actually produced
- the source map is checked against the word operations at every canonical shift, and by doctest at the two interesting positions

> One fixture subtlety worth knowing if you write more of these: the batched AND check *derives* its third column as the conjunction of the other two, but the shift reduction proves each **declared** column against its claim. A fixture that does not satisfy the AND relation must therefore read its third claim off the declared operand rather than derive it — which is what the new helper does.

### DIFF
--- a/crates/core/src/constraint_system/shift.rs
+++ b/crates/core/src/constraint_system/shift.rs
@@ -8,6 +8,12 @@ use bytes::{Buf, BufMut};
 use super::{ValueIndex, ValueVec};
 use crate::word::Word;
 
+/// The width the half-word (`*32`) shift variants act over.
+///
+/// Those variants shift the upper and lower halves of a word independently.
+/// So every rule stated over a word's positions is applied once per half of this width.
+const HALF_WORD_BITS: usize = 32;
+
 /// A different variants of shifting a value.
 ///
 /// Note that there is no shift left arithmetic because it is redundant.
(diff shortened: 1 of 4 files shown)


tcoratger:
Obsolete as per https://github.com/binius-zk/binius.xyz/pull/124

PR #2103, opened 2026-08-11
BINIUS-429: reproduce the Fiat-Shamir challenger in-circuit

tcoratger:
Closes [BINIUS-429](https://linear.app/jimpo/issue/BINIUS-429).

Reproduces the transcript's SHA-256 Fiat-Shamir challenger as a circuit gadget, byte for byte.
Rebased onto the merged recursion skeleton (#2093), so this is now purely additive: one new module, plus three lines of manifest and five of crate docs.

### Where this slots in

#2093 landed the pipeline but left the Fiat-Shamir state unconstrained, which its own crate docs call out. Three sites in the merged channel are waiting on this gadget:

```
channel.rs:110  sample        UNCONSTRAINED: a challenge is the Fiat-Shamir state's output,
                              which needs the challenger's hashing in-circuit
channel.rs:219  sample_bits   UNCONSTRAINED, twice over: the index should come from the
                              Fiat-Shamir state, and it should be masked to `bits` bits
channel.rs:163  observe_words UNCONSTRAINED: the statement must reach the Fiat-Shamir state
```

This PR adds the gadget and its differential test against the native challenger. It deliberately does **not** rewire those three call sites — that is a separate, reviewable change, and doing it here would mix a new primitive with a change to the channel's soundness story. So the skeleton stays exactly as sound as it is today, and the crate docs now say where the missing piece lives.

### The problem

A recursive verifier is a circuit whose statement is "this proof verifies".
It cannot be handed its challenges by an outer verifier — that would just move the trust.
So it has to derive them itself, which means recomputing the challenger's byte stream in-circuit.

### Why this is affordable

Only bytes read as transcript *messages* reach the challenger.
Decommitment reads bypass it and are bound instead by the Merkle roots, which is where the bulk of a proof lives.
For the 240 KiB proof a Linear^{19}$-trace circuit produces, the observed portion is a few KiB — on the order of 150 blocks, against the thousands the Merkle paths cost.

### The idea

The native challenger's control flow never reads the data it hashes.
All of this follows from the *sequence* of operations alone:

- buffer occupancy, and where block boundaries fall
- the running count of sampled bytes consumed
- the padding bytes and the length field

So each of those lives in an ordinary Rust field, settled while the circuit is built, and only byte *values* reach a wire.

Two things fall out:

- The circuit is a straight line, with no multiplexer, comparison or conditional anywhere.
- The bookkeeping is free, since counts and padding become constants and constants cost nothing.

### What is reproduced exactly

```
construction       hasher fed SHA-256(""), same digest seeds the sample buffer
buffer refill      finalize and reset, digest fed forward and served as 32 sampled bytes
sample -> observe  count of sampled bytes appended, little-endian in 8 bytes
observe -> sample  buffer retired, so the next byte read forces a refill
```

An *epoch* is the span between two resets.
Its input always opens with the previous digest's 32 bytes, so the first 32 bytes served are constants and compress nothing.

### Wire conventions

```
byte wire:  one stream byte in bits [0, 8), and bits [8, 64) must be zero
word wire:  eight stream bytes little-endian, stream byte k in bits [8k, 8k + 8)
```

Little-endian is the transcript's own convention, not a choice made here.
A 128-bit challenge is 16 little-endian bytes and a bit sample reads four, so a word wire lines up with a transcript word without shuffling.

### Cost

| item | AND constraints |
|---|---|
| one compression core, covering two chained blocks | 720 |
| one observed 64-byte block, amortized | 400 |
| one observed word, for the byte reversal | 5 |
| one 128-bit challenge, refills excluded | 12 |
| one bit sample, refills excluded | 4 |
| a refill's digest masking | 8 |

No multiplication constraints are emitted, at any size — this gadget is pure bit manipulation, so it never competes for the columns the surrounding verifier spends on field arithmetic and select gates.

Consecutive blocks of an epoch chain, so they compress in pairs through one two-lane core.
A one-lane compression costs about 690, so pairing nearly halves the dominant term.

At ~150 observed blocks that is roughly 60k AND constraints, a few percent of what the Merkle paths cost.

These are standalone measurements. Once the gadget is driven from `sample` and `sample_bits`, the per-call figures shift a little as gate fusion absorbs shifts across the boundary.

### Bounded bit samples

Sampling a bit width masks the result with a real gate, not by convention.
So the value provably lies below `2^bits`, which is what lets a query index be used without a separate range check — the second half of what `sample_bits` is marked unconstrained for today.

### Scope

- **new:** `crates/recursion/src/challenger.rs`
- **changed:** `Cargo.toml` gains three dev-dependencies; `lib.rs` gains the module and a docs paragraph
- **left out:** wiring the gadget into `sample` / `sample_bits` / `observe_words`, and the Merkle gadgets (#2104)

### Verification

- `cargo clippy -p binius-recursion --all-targets --all-features -- -D warnings` — clean
- `cargo test -p binius-recursion` and `--all-features` — 7 new tests pass, and the skeleton's 3 still pass

The acceptance test is differential against the native challenger, over a recorded script covering partial blocks, transitions in both directions, sampling that spans a buffer refill, zero-length observes, and bit widths including non-multiples of 8 and the clamped case.
Every reference value is pinned to a public wire by an equality assertion, so a mismatch fails witness population rather than a Rust comparison — the check lives in the constraints, not beside them.
A proptest covers random operation sequences, and the hardcoded seed is pinned against the real hash of the empty string.

### DIFF
--- a/crates/recursion/Cargo.toml
+++ b/crates/recursion/Cargo.toml
@@ -22,3 +22,6 @@ digest.workspace = true
 [dev-dependencies]
 binius-prover = { path = "../prover" }
 binius-verifier = { path = "../verifier" }
+proptest.workspace = true
+rand.workspace = true
+sha2.workspace = true
--- a/crates/recursion/src/challenger.rs
+++ b/crates/recursion/src/challenger.rs
@@ -1,0 +1,774 @@
+// Copyright 2026 The Binius Developers
+
+//! The proof system's Fiat-Shamir challenger, re-expressed as a circuit.
+//!
+//! A recursive verifier derives its own challenges instead of being handed them.
+//! So it must recompute the native challenger's byte stream exactly, over wires.
+//!
+//! # Build-time shape, run-time bytes
+//!
+//! The native challenger's control flow never reads the data it hashes.
+//! All of this follows from the *sequence* of operations alone:
+//!
+//! - buffer occupancy, and where block boundaries fall
+//! - the running count of sampled bytes consumed
+//! - the padding bytes and the length field
+//!
+//! Each of those therefore lives in an ordinary Rust field, settled while the circuit is built.
+//! Only byte *values* ever reach a wire.
+//!
+//! Two things fall out of that split:
+//!
+//! - The circuit is a straight line, with no multiplexer, comparison or conditional anywhere.
+//! - The bookkeeping is free, since counts and padding become constants and constants cost nothing.
+//!
+//! # The state machine being mirrored
+//!
+//! The native challenger is one SHA-256 hasher plus two halves that take turns:
+//!
+//! ```text
+//!     observe:  bytes are appended to the hasher's input
+//!     sample:   bytes are read out of a 32-byte buffer holding the latest digest
+//! ```
+//!
+//! Four events matter, and each is reproduced here exactly.
+//!
+//! 1. **Construction** — the hasher is fed the digest of the empty string, which also seeds the
+//!    sample buffer.
+//! 2. **Buffer refill** — the hasher is finalized *and reset*, its digest fed back into the fresh
+//!    hasher and also served as the next 32 sampled bytes.
+//! 3. **Sample to observe** — the count of sampled bytes consumed so far is appended, little-endian
+//!    in eight bytes.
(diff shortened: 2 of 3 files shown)


jimpo (approved):
Please rebase onto main and resolve conflicts

PR #2106, opened 2026-08-11
BINIUS-431: constrain the two field gadgets

tcoratger:
Closes [BINIUS-431](https://linear.app/jimpo/issue/BINIUS-431).

Constrains the two field gadgets the recursion skeleton left as bare hints. Rebased onto `main`, which now carries the skeleton (#2093), the challenger (#2103) and the Merkle gadgets (#2104).

### Scope changed after the rebase

Most of what this PR originally added is already upstream — the review on #2093 took the constant-word fast path, the numbered assertions, the typed input records and the dead statement wires. What was left is the third bullet of the crate's own "not sound" list:

> **Two field gadgets.** `invert_or_zero` and `square_transpose` are hints, so they carry the right value but nothing pins them to their inputs.

That bullet is now gone. The other two — the Fiat-Shamir state and the Merkle openings — still stand; wiring in the gadgets from #2103 and #2104 is the next step, and it is a decision about the witness design rather than a rebase.

### Inversion

The hint stays, since a circuit cannot divide. What is new is what pins it. Write `p` for the product of the value and the claimed inverse, then assert **both**:

```
x * p == x        at x != 0 this divides down to x * inv == 1
inv * p == inv    at x == 0 this reads inv * 0 == inv, forcing inv == 0
```

Neither suffices alone, and the cheaper form is unsound:

- Keeping only the first admits **any** claimed inverse at `x = 0`, since both sides vanish.
- Folding the two into one assertion admits `inv = x`, since in characteristic 2 that also vanishes on both sides.

Cost: 3 BMUL, plus 4 ZERO for the two asserted wire pairs.

### The subfield transpose

Was a hint; now real. The basis is the successive powers of the generator, so reading bit `j` of a wire pair as subfield coefficient `j` makes the whole operation a 128x128 bit-matrix transpose.

Split into four 64x64 blocks. Transposing swaps the off-diagonal blocks, so that swap is pure rewiring — only which wire is read:

```
M^T = [[A^T, C^T], [B^T, D^T]]
```

Each block runs the six-stage shift-and-mask network, where the mask is the only non-linear step:

| | cost |
|---|---|
| hinted, before | 0 constraints, and unsound |
| bit transpose, now | **768 AND**, 0 BMUL |

For contrast, the Spartan wrapper's equivalent hints a `d x d` coefficient matrix and spends 97,920 multiplications — the word representation is where this gets cheap.

A runtime assertion pins the monomial-basis assumption rather than leaving it implicit.

### What this removes

`SquareTransposeB1Hint` is deleted. The inverse hint stays, so the replay is unaffected either way: a hint is derived from wires the circuit already holds and never becomes a circuit input, so the recorded input list does not change and the end-to-end test still passes untouched.

### Scope

- **changed:** `symbolic/elem.rs` — the two gadgets plus two tests; `hints.rs` — transpose hint deleted; `lib.rs` — one bullet removed from the status list
- **left out:** driving the challenger and Merkle gadgets from the channel

### Verification

- `cargo test -p binius-recursion` — 28 + 2 + 1 pass, which is the 26 already there plus the 2 added

Two tests, both driving the production code rather than a re-derivation:

- the inverse constraints are driven with a caller-chosen inverse, checking the honest witness passes at a random value and at zero, and that both forgeries above are rejected
- the transpose is checked against the field's own `square_transpose` on 128 random elements, with the 768 AND count and the absence of any field multiplication pinned

### DIFF
--- a/crates/recursion/src/symbolic/elem.rs
+++ b/crates/recursion/src/symbolic/elem.rs
@@ -3,6 +3,7 @@
 //! The GHASH-field element a circuit-building channel carries.
 
 use std::{
+	array,
 	iter::{Product, Sum},
 	ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 	rc::{Rc, Weak},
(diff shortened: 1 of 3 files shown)


tcoratger:
@jimpo Just rebased this one as well with the latest changes

crates/recursion/src/symbolic/elem.rs
jimpo, line 219: follow-up: also implement `unsafe invert` with a custom impl that doesn't do the zero checking. This will get used after BINIUS-468.

jimpo (approved):

PR #2107, opened 2026-08-11
BINIUS-432: verify a BaseFold opening end to end inside a circuit

tcoratger:
Closes [BINIUS-432](https://linear.app/jimpo/issue/BINIUS-432).

The acceptance test for [BINIUS-425](https://linear.app/jimpo/issue/BINIUS-425): a polynomial-commitment opening proved natively, then verified **inside a Binius64 circuit**.
This is the first real measurement of the fold-and-Merkle share of a recursive verifier.

### What changed on the rebase

The stack this branch sat on landed in pieces while it waited, so the old "review only the second commit" note is gone — there is no longer a stack base.

| piece | landed as | on main |
|---|---|---|
| Fiat-Shamir challenger | #2103 | byte-identical |
| Merkle gadgets | #2104 | byte-identical |
| in-circuit field element | #2106 | reworked, `Elem` → `SymbolicElem` |
| **the channel that assembles them** | — | **never landed** |

That last row is why the rebase was not mechanical. `Binius64BuilderChannel` on main is still the skeleton from #2093: it records the verifier's operations but leaves the hashing out, so a replay supplies what the circuit cannot derive. The gadgets exist; nothing drives them. With no constrained channel there was nothing for this test to run against.

So the channel is carried here, as `merkle_channel`, rebuilt on main's `SymbolicElem` and `SymbolicWord` rather than on the branch's own element type:

```
skeleton (untouched):  proof -> replay -> wires the circuit could not derive
merkle_channel:        proof -> wires, and every other value is a gate output
```

Deriving those values is what costs the SHA-256, and what makes an unsatisfied circuit mean a rejected proof. The skeleton and its replay path are left exactly as they are — both channels compile, and `recursion_end_to_end.rs` still passes.

This subsumes [BINIUS-469](https://linear.app/jimpo/issue/BINIUS-469), [BINIUS-470](https://linear.app/jimpo/issue/BINIUS-470), [BINIUS-471](https://linear.app/jimpo/issue/BINIUS-471) and [BINIUS-472](https://linear.app/jimpo/issue/BINIUS-472) for the new channel, though not for the skeleton, which keeps its six `UNCONSTRAINED` markers. **If you would rather those four land incrementally against the skeleton instead, say so and I will close this and reopen the test on top of them.**

Two small additions outside the channel:

- `Shared` gains a constructor anchoring on a builder the caller already holds, which is how the statement reaches the circuit on inout wires.
- `SymbolicWord` gains a `value` accessor, so a lookup over a protocol-fixed index settles in Rust rather than spending select gates.

Three things in the test moved with main: `optimal_for_batch` takes the Merkle scheme as a type parameter, the NTT domain is Gao-Mateer rather than a naive subspace (#2094), and the channel traits grew `pack_words`.

### The two halves

The native round trip is reused at both ends:

```
prover:   native and unchanged, writing one real transcript
verifier: the same code, over a channel that emits gates instead of checking values
```

A satisfied constraint system is then the statement "this transcript opens this claim".
An outer proof can check that statement, which is what makes this one step of recursion.

### What is an input

Two things enter the circuit, and nothing else:

```
the proof byte stream   one witness wire per eight-byte proof word
the inner statement     the claimed evaluation and the point, on inout wires
```

Challenges, folded values and recomputed digests are all gate outputs.

The statement deliberately does **not** enter as a build-time constant.
A constant would fold one opening's numbers into the gates and quietly destroy the next property.

### Build once, verify many

Every shape the circuit depends on is settled before any proof exists — the code rate, the fold arities, the oracle layout, the tree depths, the query count.
So one circuit serves every opening of that shape, and the expensive compilation is paid once per protocol rather than once per proof.

A test asserts that directly: build once, then verify two different openings (different witness, different point, different transcript).

### Negative tests

Five corruption tests each flip one byte and require the system to become unsatisfiable, and each states which mechanism catches it:

| corrupted | caught by |
|---|---|
| opened value | the Merkle path to the layer |
| Merkle sibling | the Merkle path to the layer |
| layer digest | the layer fold to the root |
| terminal codeword entry | the rebuilt tree, or a fold equality |
| round polynomial | see below |

The round polynomial is the interesting one, and the reason the distinction is worth keeping.
It is an **observed** message, so corrupting it moves the Fiat-Shamir state, and therefore the sampled query indices:

```
before:  round value  -> challenge -> ... -> query indices -> the committed positions
after:   round value' -> a different challenge from that round on, indices included
```

The opened values and siblings on the tape are unchanged, but the indices addressing them are not, so every opening now climbs to the wrong entry of the decommitted layer.

A tampered statement wire is covered too.
Each test asserts on every field of the resulting error, with no discarded fields.

### The measurement

At `n_vars = 8`, `log_inv_rate = 1` and 32 queries, over a 13,536-byte proof: **229,675 AND**, 15,463 BMUL, 71,362 ZERO, 524,288 committed words.

| phase | AND | BMUL | AND % |
|---|---|---|---|
| Merkle openings | 212,315 | 9,216 | 92.4% |
| Fiat-Shamir challenger | 9,643 | 0 | 4.2% |
| terminal codeword | 7,773 | 0 | 3.4% |
| MLE-check arithmetic | 0 | 16 | 0.0% |
| fold and its equalities | 0 | 6,231 | 0.0% |

So the verifier is SHA-256 and almost nothing else:

- the three hashing phases are the whole AND column
- Merkle openings alone are 92% of it
- the field arithmetic the fold and the sum-check actually reduce spends **no AND at all**, only 6,247 BMUL

The AND column is what binds.

The AND column is unchanged from the pre-rebase measurement at every shape in the sweep below. Only BMUL moved, from 14,887 to 15,463, which is #2106 charging for the two field gadgets it constrained.

### The cost surface

One sweep, moving a single lever at a time:

| shape | AND | vs base |
|---|---|---|
| `n_vars` 8, rate 1, 32 queries | 229,675 | -- |
| `n_vars` 10, rate 1, 32 queries | 346,681 | +51% |
| `n_vars` 8, rate 2, 32 queries | 285,585 | +24% |
| `n_vars` 8, rate 1, 16 queries | 146,249 | -36% |
| `n_vars` 8, rate 1, 64 queries | 280,193 | +22% |

Two readings worth keeping, both of which contradict the estimates on the issue:

1. **Queries are sublinear** — 16 to 32 costs +57% but 32 to 64 only +22%, because the shared decommitment layer deepens as the query count grows and every climb gets shorter.
2. **Committed size is not flat** — four times the committed data costs 51% more, not "a few percent". Growth is logarithmic, but with a steep constant: every extra Merkle level is paid once per query across every tree in the fold schedule, and here the terminal codeword grew with it from 7,773 to 26,781 AND.

So doubling the query count is cheaper than it looks, and the layer depth is the reason.

### Where this sits

Against the 2.5M-5M AND estimate for a full recursive verifier, one opening of this shape is around a tenth of the lower bound.

A production shape costs more: more queries, deeper trees, several oracles, and the constraint-system checks above the opening.
The levers are the same three, in this order:

1. fewer SHA-256 compressions per Merkle level
2. fewer Merkle levels
3. the layer-depth trade-off that sets how many levels each query climbs

That last one is [BINIUS-428](https://linear.app/jimpo/issue/BINIUS-428)'s territory, and the sweep above is the surface it would be tuned against.

### Scope

3,053 insertions, 16 deletions, all inside `crates/recursion`.

| file | lines | what |
|---|---|---|
| `src/merkle_channel.rs` | +693 | the constrained channel |
| `src/merkle_channel/tests.rs` | +1018 | its unit tests, pinning it against the native channel |
| `tests/basefold_opening.rs` | +1301 | the end-to-end test and the measurement |
| `src/lib.rs` | +19/-15 | module registration, and crate docs that claimed nothing was constrained |
| `src/shared.rs` | +8/-1 | `Shared::with_builder` |
| `src/symbolic/word.rs` | +10 | `SymbolicWord::value` |
| `Cargo.toml` | +4 | `thiserror`, and three dev-dependencies for the test |

Untouched: the skeleton channel, the filler, and everything outside the crate.

Worth reading in that order — the channel is the design, the two test files are volume. Two unit tests carry most of the soundness weight:

- `a_full_script_matches_the_native_channel` runs a scripted verifier over both the in-circuit channel and the native one, and pins every sampled challenge, query index and received value between them. If the in-circuit Fiat-Shamir state ever diverged, this is what catches it.
- `only_the_root_and_prover_messages_are_observed` pins the message-versus-decommitment split, which is the load-bearing property of the module: getting it wrong does not make the circuit unsatisfiable, it makes the circuit disagree with every real transcript.

### Verification

- `cargo check -p binius-recursion --all-targets` — clean
- `cargo test -p binius-recursion` — 44 lib + 10 basefold + 2 channel-builds + 1 end-to-end passed

`cargo check --workspace --all-targets` fails on this machine, but it fails identically on a clean `upstream/main` checkout, so it is not from this branch.

The two measurement tests print their tables; run with `--nocapture` to see them.

### DIFF
--- a/crates/recursion/Cargo.toml
+++ b/crates/recursion/Cargo.toml
@@ -18,6 +18,7 @@ binius-ip = { path = "../ip" }
 binius-transcript = { path = "../transcript" }
 binius-utils = { path = "../utils" }
 digest.workspace = true
+thiserror.workspace = true
 
 [dev-dependencies]
 binius-compute = { path = "../compute" }
--- a/crates/recursion/Cargo.toml
+++ b/crates/recursion/Cargo.toml
@@ -20,7 +20,10 @@ binius-utils = { path = "../utils" }
 digest.workspace = true
 
 [dev-dependencies]
+binius-compute = { path = "../compute" }
 binius-iop-prover = { path = "../iop-prover" }
+binius-ip-prover = { path = "../ip-prover" }
+binius-math = { path = "../math", features = ["test-utils"] }
 binius-prover = { path = "../prover" }
 binius-verifier = { path = "../verifier" }
 proptest.workspace = true
--- a/crates/recursion/src/lib.rs
+++ b/crates/recursion/src/lib.rs
@@ -48,6 +48,7 @@ mod channel;
 mod filler;
 mod hints;
 pub mod merkle;
+pub mod merkle_channel;
 mod shared;
 pub mod symbolic;
 
--- a/crates/recursion/src/lib.rs
+++ b/crates/recursion/src/lib.rs
@@ -18,30 +18,33 @@
 //! operations in the same order, and one cursor pairs what the second saw with the wires the first
 //! allocated.
 //!
-//! # Status: a skeleton, and not sound
+//! # Two channels
 //!
-//! The gadgets a real recursive verifier needs are not written. In their place, values that ought
-//! to be derived are left as circuit inputs for the replay to supply:
+//! [`Binius64BuilderChannel`] is a skeleton, and not sound.
+//! Values that ought to be derived are left as circuit inputs for a replay to supply:
 //!
 //! - **The Fiat-Shamir state.** `sample` and `sample_bits` return free wires rather than the
 //!   challenger's output, so nothing ties a challenge to the transcript that produced it.
 //! - **The Merkle openings.** `recv_openings` and `recv_committed_vector` return free wires rather
 //!   than values checked against a commitment root.
 //!
-//! What *is* constrained is the verifier's arithmetic: the sumcheck folding, the eq-indicator and
-//! Lagrange evaluations, the monster multilinear, every `assert_zero` along the way, and both field
-//! gadgets that used to be bare hints.
+//! What it does constrain is the verifier's arithmetic alone:
 //!
-//! So a circuit built here accepts proofs it should reject. It is useful for measuring that
-//! arithmetic and for keeping the pipeline honest while the gadgets are written, not for proving
-//! anything. Each gadget that lands removes entries from the recorded input list; with all of them
-//! in place the only input left is the proof itself.
+//! - the sumcheck folding, and the eq-indicator and Lagrange evaluations
+//! - the monster multilinear, and every `assert_zero` along the way
 //!
-//! The gadget for the first bullet now exists in [`challenger`], reproducing the native
-//! challenger's byte stream over wires.
-//! The gadgets for the second exist in [`merkle`], matching the native binary Merkle scheme.
-//! Driving them from `sample`, `sample_bits`, `recv_openings` and `recv_committed_vector` is what
-//! removes the two remaining bullets.
+//! So a circuit built there accepts proofs it should reject.
+//! It is for measuring that arithmetic, not for proving anything.
+//!
+//! [`merkle_channel`] closes both holes.
+//! It drives the [`challenger`] and [`merkle`] gadgets from those same four methods:
+//!
+//! ```text
+//!   skeleton:        proof -> replay -> wires the circuit could not derive
+//!   merkle_channel:  proof -> wires, and every other value is a gate output
+//! ```
+//!
+//! An unsatisfied circuit there means a rejected proof, which the skeleton cannot say.
 
 pub mod challenger;
 mod channel;
--- a/crates/recursion/src/merkle_channel.rs
+++ b/crates/recursion/src/merkle_channel.rs
@@ -1,0 +1,693 @@
+// Copyright 2026 The Binius Developers
+
+//! A verifier channel that builds a circuit instead of checking a proof.
+//!
+//! A verifier in this codebase never touches a transcript directly.
+//! It talks to a channel, and the protocol code across the interactive-proof, IOP and verifier
+//! crates is generic over which channel it is handed.
+//!
+//! Running that code against this channel verifies nothing.
+//! It *records* the verification as a Binius64 circuit, satisfiable exactly when the proof it was
+//! handed would have been accepted.
+//!
+//! Three gadgets are assembled here: field arithmetic over wires, the Fiat-Shamir state over wires,
+//! and the Merkle commitment checks.
+//!
+//! # Against the skeleton channel
+//!
+//! [`Binius64BuilderChannel`](crate::Binius64BuilderChannel) records the same operations.
+//! It leaves the hashing out, so a replay supplies what it could not derive:
+//!
+//! ```text
+//!     skeleton:  proof -> replay -> wires the circuit could not derive
+//!     this one:  proof -> wires, and every other value is a gate output
+//! ```
+//!
+//! Deriving those values is what costs the SHA-256.
+//! It is also what makes an unsatisfied circuit mean a rejected proof.
+//!
+//! # The proof stream
+//!
+//! A verifier transcript holds one byte tape, and two readers consume it strictly in order.
+//! They differ in one respect only:
+//!
+//! ```text
+//!     message reader:       reads the tape and feeds what it read to the Fiat-Shamir state
+//!     decommitment reader:  reads the tape and feeds the Fiat-Shamir state nothing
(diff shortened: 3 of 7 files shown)


jimpo:
needs rebase/conflict resolution. Seems like many pieces have been merged -- should we just close this? Is anything else left unconstrained?

tcoratger:
I think this still is valid, a short summary.

**Should we close it?** The pieces that merged are the *gadgets*. The thing that calls them didn't.

Here is main today:

```
sample()         -> a free wire     nothing hashes the transcript
sample_bits()    -> a free wire     nothing derives the query index
recv_openings()  -> free wires      nothing checks a Merkle path
```

The challenger from #2103 and the Merkle gadgets from #2104 are both there, sitting uncalled. `Binius64BuilderChannel` hands out free wires at six sites and lets the replay fill them, so a circuit built from main accepts a proof with whatever challenges you like.

The channel that *drives* those gadgets and reads the proof tape was in this branch's base commit, but it was never its own PR — #2103, #2104 and #2106 took the challenger, the Merkle gadgets and the field element, and the channel fell through the gap. With it missing there was nothing for this test to run against, so the rebase could not be a text merge. It is carried here now as `merkle_channel`, rebuilt on `SymbolicElem` rather than the branch's own element type.

**Anything else unconstrained?** Yes — those six sites, all on the skeleton, all ticketed as BINIUS-469/470/471/472. The new channel has none of them: every challenge, folded value and recomputed digest is a gate output, and the only inputs are the proof bytes and the statement.

That is what makes the negative tests mean anything. Flip one bit of the first round polynomial:

```
before:  round value  -> challenge -> ... -> query indices -> committed positions
after:   round value' -> a different challenge from that round on, indices included
```

The opened values on the tape are untouched, but the indices addressing them are not, so every opening climbs to the wrong entry of the decommitted layer and the system is unsatisfiable. On the skeleton that test cannot exist — the indices are free wires, so nothing moves when the transcript changes.

**The number you wanted:** 229,675 AND at `n_vars` 8, rate 1, 32 queries. Merkle openings are 92% of it, the challenger 4%, the terminal codeword 3%. The field arithmetic that the fold and the sum-check actually reduce spends **no AND at all**. Identical to the pre-rebase measurement at every shape in the sweep; only BMUL moved, by the 576 that #2106 charges for constraining the two field gadgets.

**One call for you.** This lands 469–472 in one go through a second channel type, instead of incrementally against the skeleton the way you split them. If you would rather have the four separate PRs, say so and I will close this and reopen the test on top of them.

tcoratger:
(And also there are a lot of added lines of code here for end to end test, which I think is useful).

jimpo:
I think this is too big. I broke BINIUS-425 down into a few subissues that I think we should do and merge independently. Could you pick that up please?

tcoratger:
> I think this is too big. I broke BINIUS-425 down into a few subissues that I think we should do and merge independently. Could you pick that up please?

Ok no problem, just opened the first one here: https://github.com/binius-zk/binius64/pull/2128

tcoratger:
Closing this as outdated. It is replaced by #2147, which does the same thing in a tenth of the code.

You were right that this was too big, and the reason has since gone away. This branch had to carry its own tape-reading channel — 693 lines of it, plus 1,018 lines of channel tests — because the channel on `main` constrained nothing at the time.

The four subissues you split out have all landed:

| | |
|---|---|
| #2128 | BINIUS-471, bind the Merkle openings to their roots |
| #2130 | BINIUS-472, take the inner statement as a public input |
| #2138 | BINIUS-469, drive the Fiat-Shamir challenger |
| #2143 | BINIUS-470, derive and mask the FRI query index |

So the channel on `main` *is* the verifier now, and the acceptance test just drives it:

| | this PR | #2147 |
|---|---|---|
| new channel | 693 lines | none |
| channel tests | 1,018 lines | none |
| test + source | 1,342 lines | 709 |

The measurements survive too. Both readings this PR reported — queries are sublinear, committed size is not flat — reproduce on the merged channel, and the AND counts land within a couple of percent of what is here.

PR #2109, opened 2026-08-11
Add chip composition for M4 circuits

jimpo:
## Overview

An M4 constraint system is a composition of chips. Each chip is an embedded constraint system validating a relation on its local inout values, and runs once per call that reaches it; a distinguished main chip runs once and nothing calls it. A chip call names a callee and passes operands over the caller&#39;s value vector, one per inout value of the callee, and is served by the callee instance at the call&#39;s position.

Four commits: the feature, a pure item-relocation pass (verified: added and removed line multisets are identical), then the two review rounds.

## What each crate gains

**binius-core** — the constraint-system form: `ConstraintSystemM4`, built from `EmbeddedConstraintSystem` and `ChipCall`.
- `validate` checks call operands the way constraint operands are checked (the per-term body of `validate_operand` moves to `operand_fault`, which both wrap under their own naming), range-checks callee IDs and call arity, and requires the chips to be in topological order.
- `verify` is the reference checker for the composition, in the manner of `ConstraintSystem::verify`: the main chip&#39;s constraints, every instance&#39;s (padding instances included, since every instance is committed), and every call positionally against the instance serving it. Instances are read through the `ChipInstances` accessor, one at a time.

**binius-frontend** — the circuit form: a `chip` module with `EmbeddedCircuit` and `CircuitM4`, and a builder API.
- `CircuitBuilder::add_chip` registers a system and returns an opaque `ChipRef`, flattening nested systems into one topologically ordered chip list.
- `call_chip` takes a `ChipRef` and the wires of one invocation.
- `build_m4` compiles, resolves each call&#39;s wires to value indices, and counts each chip&#39;s active instances off the call graph. `CircuitM4::to_constraint_system` lowers to the form the protocol consumes.
- `Circuit` now carries its inout wires in segment order, so a chip&#39;s interface is its circuit&#39;s inout segment.

**binius-m4-prover** — `WitnessM4`. `generate` populates the main circuit and one `ValueTable` per chip, one instance per invocation in caller order, the count padded to a power of two by repeating the last invocation. `verify` hands the tables to the core checker, which reads their instances one at a time.

## The model

A chip call is a constraint on the words its wires hold. The compiler passes have no notion of a call, so `compile` folds call wires into the pinned set they do understand and counts them as constraint-read when the value vector is laid out. The expected usage computes a chip&#39;s outputs from its inputs with a hint and passes both: the hint hands the witness its values, and the call is the constraint that makes them correct. A chip does not say which of its inout words are inputs and which are outputs; a call constrains them all alike.

The chip-accelerated CRC-64 test (`crates/m4-prover/tests/crc64_chip.rs`) drives the whole path: a chip mixes one message word into the register, main calls it once per word on hinted states, and the witness verifies whole. A hint returning a wrong register state fails verification with a `CallMismatch` naming the call, instance and word.

## Ordering

The chips are in topological order: a chip calls only chips with a higher ID, and main, which is not one of them, calls any. Call positions are defined by enumerating the chips in ID order, and both witness generation and `verify` walk them that way, so every caller of a chip is reached before the chip itself.

Both `CircuitM4::validate` and `ConstraintSystemM4::validate` require it, as one comparison per call. Acyclicity follows, and is not separately checked: an acyclic graph is not enough, because a backward call between two chips that never form a cycle is still served by an instance the walk has already passed.

## Validation

`CircuitM4::validate` is the single check on the circuit form. It rejects a call naming a chip the system does not have, a call passing more operands than the callee has inout values, an operand reading an uncommitted word of its caller, a call to an earlier chip, a declared active-instance count that disagrees with the calls reaching the chip, a chip nothing calls, and an instance count that outgrows a `usize`.

A system that passes it lowers to one that passes `ConstraintSystemM4::validate`, which requires the same ordering and additionally validates each chip&#39;s compiled constraint system — the one thing the compiler rather than that check keeps well-formed. So nothing downstream has to run the lowered check to know its preconditions hold, and that check remains the one for a hand-built `ConstraintSystemM4` that never went through the frontend.

## Instance access

`ConstraintSystemM4::verify` reads chip instances through `ChipInstances`, which serves one instance at a time by `(chip_id, row)`. A witness that stores its instances packed — as the prover&#39;s `ValueTable` does, one column per instance — is therefore never expanded into value vectors whole. For a chip of 451 words at 2^20 instances, expanding it was several gigabytes on top of the table itself, for a check whose working set is one value vector. The cost is that an instance serving a call is rebuilt when the call is checked, having already been built for its own constraints: roughly 2x the materializations for O(1) live instead of O(total).

`[Vec<ValueVec>]` implements the trait, so a hand-built witness still passes a slice.

## Known deferrals and open questions

- **Fusion rewrite site**: a chip call is not yet a rewrite site of gate fusion, so a linear argument such as `a ^ b` is committed together with its defining ZERO constraint rather than fusing into the call operand.
- **Protocol gap**: nothing proves a `WitnessM4` yet; positional matching is the reference stand-in for the eventual permutation/lookup argument. Core M4 types have no serialization yet.
- **`build_m4` does not validate its result**: a registered-but-never-called chip surfaces as a `generate` panic instead of `CircuitM4Error::NeverCalled`. Every test calls `validate()` by hand right after `build_m4()`; the call may belong inside.
- **Chip circuits with `add_witness` wires** compile but fail late in generation (only `circuit.inout()` is assigned). A build-time check that a chip declares its whole interface as inout could catch this earlier.
- **Cross-builder `ChipRef`**: a ref from one builder used on another either panics on an index or silently names the wrong chip. Fixing needs builder provenance inside `ChipRef`.
- **Naming**: `m4::ConstraintSystemM4` repeats the module in the identifier, unlike its siblings `EmbeddedConstraintSystem`/`ChipCall`; kept for symmetry with `CircuitM4`/`WitnessM4`.

### DIFF
--- a/crates/core/src/constraint_system/m4.rs
+++ b/crates/core/src/constraint_system/m4.rs
@@ -104,10 +104,11 @@ pub struct ChipCall {
 /// Validity invariants:
 /// - all embedded constraint systems in `chips` must have a chip_id value that indexes into `chips`
 ///
-/// [`Self::validate`] accepts any acyclic call graph, but call positions are defined by
-/// enumerating the chips in ID order, and witness generation walks them the same way — so only a
-/// system whose calls run to higher IDs is provable. The frontend's
-/// `CircuitM4::validate` is what enforces that ordering.
+/// The chips are in topological order: a chip calls only chips with a higher ID, and main, which
+/// is not one of them, calls any. Call positions are defined by enumerating the chips in ID order
+/// and witness generation walks them the same way, so every caller of a chip is reached before the
+/// chip itself. [`Self::validate`] requires it, which makes the call graph acyclic as a
+/// consequence.
 pub struct ConstraintSystemM4 {
 	/// The entry point of the system. It calls chips, but no chip ID names it, so nothing calls
 	/// it.
@@ -121,32 +122,25 @@ pub struct ConstraintSystemM4 {
 }
 
 impl ConstraintSystemM4 {
-	/// Checks every chip and every chip call, and rejects call-graph cycles.
+	/// Checks every chip and every chip call, and requires the chips to be in topological order.
 	pub fn validate(&self) -> Result<(), ConstraintSystemError> {
-		let n_chips = self.chips.len();
-
 		self.main.validate()?;
 		self.validate_calls(None, &self.main.chip_calls)?;
 
-		// The chips each chip calls, indexed by the caller's chip ID.
-		let mut callees = vec![Vec::new(); n_chips];
 		for (chip_index, (chip, _)) in self.chips.iter().enumerate() {
 			chip.validate()?;
 			self.validate_calls(Some(chip_index), &chip.chip_calls)?;
-			callees[chip_index].extend(chip.chip_calls.iter().map(|call| call.chip_id));
-		}
-
-		if !is_acyclic(&callees) {
-			return Err(ConstraintSystemError::CyclicChipCalls);
 		}
 
 		Ok(())
 	}
 
-	/// Checks that one caller's calls name existing chips and pass no more operands than the
-	/// callee has inout values.
+	/// Checks that one caller's calls name existing chips, run only to later chips, and pass no
+	/// more operands than the callee has inout values.
 	///
 	/// A call may pass fewer: the inout values past its operands are constrained to zero.
+	///
+	/// Main is not one of the numbered chips and runs before all of them, so it may call any.
 	fn validate_calls(
 		&self,
 		chip_index: Option<usize>,
@@ -159,6 +153,14 @@ impl ConstraintSystemM4 {
 					chip_index,
 					chip_id: call.chip_id,
 					n_chips,
+				});
+			}
+			if let Some(caller) = chip_index
+				&& call.chip_id <= caller
+			{
+				return Err(ConstraintSystemError::CallOutOfOrder {
+					chip_index: caller,
+					callee: call.chip_id,
 				});
 			}
 			let n_inout = self.chips[call.chip_id].0.cs.n_inout;
@@ -318,38 +320,8 @@ impl ConstraintSystemM4 {
 	}
 }
 
-/// Checks that the call graph given by the callee list of each chip is acyclic.
-///
-/// This is Kahn's algorithm: chips with no remaining callers are removed one at a time, and any
-/// chip still left when none remain lies on a cycle. A chip that calls itself is a cycle.
-fn is_acyclic(callees: &[Vec<usize>]) -> bool {
-	let mut n_callers = vec![0usize; callees.len()];
-	for &callee in callees.iter().flatten() {
-		n_callers[callee] += 1;
-	}
-
-	let mut uncalled = (0..callees.len())
-		.filter(|&chip| n_callers[chip] == 0)
-		.collect::<Vec<_>>();
-
-	let mut n_removed = 0;
-	while let Some(chip) = uncalled.pop() {
-		n_removed += 1;
-		for &callee in &callees[chip] {
-			n_callers[callee] -= 1;
-			if n_callers[callee] == 0 {
-				uncalled.push(callee);
-			}
-		}
-	}
-
-	n_removed == callees.len()
-}
-
 #[cfg(test)]
 mod tests {
-	use proptest::prelude::*;
-
 	use super::{
 		super::{ShiftedValueIndex, ValueIndex, ValueSegment, ValueVecLayout, ZeroConstraint},
 		*,
@@ -432,34 +404,52 @@ mod tests {
 	}
 
 	#[test]
-	fn validate_accepts_an_acyclic_call_graph() {
-		// Main calls 0, which calls 1 and 2; 1 calls 2, and 2 is a leaf.
+	fn validate_accepts_calls_running_only_to_later_chips() {
+		// Main calls 0, which calls 1 and 2; 1 calls 2, and 2 is a leaf. Main is not one of the
+		// numbered chips, so its call to 0 is not a backward one.
 		let cs = system(&[0], vec![chip(&[1, 2]), chip(&[2]), chip(&[])]);
 		cs.validate().unwrap();
 	}
 
 	#[test]
 	fn validate_rejects_a_chip_that_calls_itself() {
+		// A chip is not later than itself, so self-recursion is out of order rather than a
+		// separate case.
 		let cs = system(&[0], vec![chip(&[0])]);
-		assert!(matches!(cs.validate(), Err(ConstraintSystemError::CyclicChipCalls)));
+		assert!(matches!(
+			cs.validate(),
+			Err(ConstraintSystemError::CallOutOfOrder {
+				chip_index: 0,
+				callee: 0,
+			})
+		));
 	}
 
 	#[test]
-	fn validate_rejects_a_cycle_reachable_from_an_uncalled_chip() {
-		// Chip 0 is called by nobody, so removing it leaves the cycle 1 -> 2 -> 1.
+	fn validate_rejects_a_chip_that_calls_an_earlier_chip() {
+		// Chip 2 calls back to chip 1, which is populated before it. The graph as a whole need not
+		// be cyclic for this to break the one pass over the chips.
 		let cs = system(&[0], vec![chip(&[1]), chip(&[2]), chip(&[1])]);
-		assert!(matches!(cs.validate(), Err(ConstraintSystemError::CyclicChipCalls)));
+		assert!(matches!(
+			cs.validate(),
+			Err(ConstraintSystemError::CallOutOfOrder {
+				chip_index: 2,
+				callee: 1,
+			})
+		));
 	}
 
 	#[test]
 	fn validate_rejects_a_call_to_a_chip_that_does_not_exist() {
-		let cs = system(&[0], vec![chip(&[0, 7])]);
+		// Chip 0's first call is to a later chip, so the out-of-range second one is the only
+		// fault and the range check is what has to catch it.
+		let cs = system(&[0], vec![chip(&[1, 7]), chip(&[])]);
 		assert!(matches!(
 			cs.validate(),
 			Err(ConstraintSystemError::OutOfRangeChipId {
 				chip_index: Some(0),
 				chip_id: 7,
-				n_chips: 1,
+				n_chips: 2,
 			})
 		));
 	}
@@ -587,51 +577,5 @@ mod tests {
 			),
 			"{err}"
 		);
-	}
-
-	/// Whether the call graph is acyclic, by a three-colour depth-first search.
-	///
-	/// A chip on the stack that is reached again closes a cycle. This is the independent reference
-	/// [`is_acyclic`]'s Kahn's algorithm is checked against.
-	fn acyclic_by_depth_first_search(callees: &[Vec<usize>]) -> bool {
-		#[derive(Clone, Copy, PartialEq)]
-		enum Colour {
-			Unvisited,
-			OnStack,
-			Done,
-		}
-
-		fn visit(chip: usize, callees: &[Vec<usize>], colour: &mut [Colour]) -> bool {
-			match colour[chip] {
-				Colour::OnStack => return false,
-				Colour::Done => return true,
-				Colour::Unvisited => {}
-			}
-			colour[chip] = Colour::OnStack;
-			for &callee in &callees[chip] {
-				if !visit(callee, callees, colour) {
-					return false;
-				}
-			}
-			colour[chip] = Colour::Done;
-			true
-		}
-
-		let mut colour = vec![Colour::Unvisited; callees.len()];
-		(0..callees.len()).all(|chip| visit(chip, callees, &mut colour))
-	}
-
-	proptest! {
-		// Kahn's algorithm and a depth-first search decide acyclicity by unrelated means, so
-		// agreeing on random graphs covers what the cases above are each written for one of:
-		// multi-edges, diamonds, self-loops and components nothing reaches.
-		#[test]
-		fn is_acyclic_agrees_with_a_depth_first_search(
-			graph in (1usize..7).prop_flat_map(|n_chips| {
-				prop::collection::vec(prop::collection::vec(0..n_chips, 0..4), n_chips)
-			})
-		) {
-			prop_assert_eq!(is_acyclic(&graph), acyclic_by_depth_first_search(&graph));
-		}
 	}
 }
--- a/crates/core/src/constraint_system/m4.rs
+++ b/crates/core/src/constraint_system/m4.rs
@@ -1,12 +1,46 @@
 // Copyright 2026 The Binius Developers
 
-use std::iter;
+use std::{borrow::Cow, iter};
 
 use super::{ValueVec, constraint::Operand};
 use crate::{
 	ConstraintSystem, Word,
 	error::{ConstraintSystemError, VerificationM4Error},
 };
+
+/// The chip instances of an M4 witness, addressed by chip ID and row.
+///
+/// [`ConstraintSystemM4::verify`] reads one instance at a time and never holds two, so this is all
+/// it needs of a witness. A witness that stores its instances packed — as the prover's tables do,
+/// one column per instance — can therefore serve them one at a time rather than expanding the whole
+/// witness into value vectors first, which for a system of any size is most of its memory.
+///
+/// Instances are served, not lent, because a packed witness has to build one to hand it over. One
+/// that already holds value vectors lends them and pays nothing.
+pub trait ChipInstances {
+	/// The number of chips the witness covers, which must be the number the system has.
+	fn n_chips(&self) -> usize;
+
+	/// The number of instances of the given chip, which must be at least its active-instance count.
+	fn n_instances(&self, chip_id: usize) -> usize;
+
+	/// The value vector of one instance of one chip.
+	fn instance(&self, chip_id: usize, row: usize) -> Cow<'_, ValueVec>;
+}
+
+impl ChipInstances for [Vec<ValueVec>] {
+	fn n_chips(&self) -> usize {
+		self.len()
+	}
+
+	fn n_instances(&self, chip_id: usize) -> usize {
+		self[chip_id].len()
+	}
+
+	fn instance(&self, chip_id: usize, row: usize) -> Cow<'_, ValueVec> {
+		Cow::Borrowed(&self[chip_id][row])
+	}
+}
 
 /// A constraint system that represents a single chip in a [`ConstraintSystemM4`].
 ///
@@ -157,38 +191,42 @@ impl ConstraintSystemM4 {
 	/// This is the reference the proving protocol's argument is checked against, in the manner of
 	/// [`ConstraintSystem::verify`].
 	///
-	/// Malformed systems are [`Self::validate`]'s to reject; verifying one may panic or
-	/// misreport.
+	/// Instances are read through [`ChipInstances`] one at a time, so a witness that stores them
+	/// packed is never expanded whole. The price is that an instance serving a call is built again
+	/// when the call is checked, having already been built for its own constraints.
+	///
+	/// Malformed systems are [`Self::validate`]'s to reject; verifying one may panic, misreport, or
+	/// pass. In particular a call passing more operands than the callee has inout values — which
+	/// [`Self::validate`] rejects — has the operands past them ignored here.
 	///
 	/// # Errors
 	///
 	/// Reports the first failure found, in the order listed above.
-	pub fn verify(
+	pub fn verify<I: ChipInstances + ?Sized>(
 		&self,
 		main: &ValueVec,
-		chip_instances: &[Vec<ValueVec>],
+		chip_instances: &I,
 	) -> Result<(), VerificationM4Error> {
-		if chip_instances.len() != self.chips.len() {
+		if chip_instances.n_chips() != self.chips.len() {
 			return Err(VerificationM4Error::WrongChipCount {
-				n_witness_chips: chip_instances.len(),
+				n_witness_chips: chip_instances.n_chips(),
 				n_chips: self.chips.len(),
 			});
 		}
 
 		self.main.cs.verify(main)?;
-		for (chip_id, ((chip, n_active), instances)) in
-			iter::zip(&self.chips, chip_instances).enumerate()
-		{
-			if instances.len() < *n_active {
+		for (chip_id, (chip, n_active)) in self.chips.iter().enumerate() {
+			let n_instances = chip_instances.n_instances(chip_id);
+			if n_instances < *n_active {
 				return Err(VerificationM4Error::MissingInstances {
 					chip_id,
-					n_instances: instances.len(),
+					n_instances,
 					n_active: *n_active,
 				});
 			}
-			for (instance, values) in instances.iter().enumerate() {
+			for instance in 0..n_instances {
 				chip.cs
-					.verify(values)
+					.verify(&chip_instances.instance(chip_id, instance))
 					.map_err(|source| VerificationM4Error::ChipInstance {
 						chip_id,
 						instance,
@@ -203,10 +241,11 @@ impl ConstraintSystemM4 {
 		self.check_caller(None, &self.main.chip_calls, main, &mut cursor, chip_instances)?;
 		for (caller_chip, (chip, n_active)) in self.chips.iter().enumerate() {
 			for caller_instance in 0..*n_active {
+				let values = chip_instances.instance(caller_chip, caller_instance);
 				self.check_caller(
 					Some((caller_chip, caller_instance)),
 					&chip.chip_calls,
-					&chip_instances[caller_chip][caller_instance],
+					&values,
 					&mut cursor,
 					chip_instances,
 				)?;
@@ -234,18 +273,14 @@ impl ConstraintSystemM4 {
 	/// `values` is that caller's value vector, which the call operands are evaluated on. A call
 	/// past its chip's active instances is counted but compared to nothing; the caller reports the
 	/// overshoot when the cursors are read back.
-	fn check_caller(
+	fn check_caller<I: ChipInstances + ?Sized>(
 		&self,
 		caller: Option<(usize, usize)>,
 		calls: &[ChipCall],
 		values: &ValueVec,
 		cursor: &mut [usize],
-		chip_instances: &[Vec<ValueVec>],
+		chip_instances: &I,
 	) -> Result<(), VerificationM4Error> {
-		let (caller_chip, caller_instance) = match caller {
-			Some((chip, instance)) => (Some(chip), instance),
-			None => (None, 0),
-		};
 		for (call_index, call) in calls.iter().enumerate() {
 			let chip_id = call.chip_id;
 			let (chip, n_active) = &self.chips[chip_id];
@@ -255,8 +290,11 @@ impl ConstraintSystemM4 {
 				continue;
 			}
 
+			// Only the callee's own inout values are compared. An operand past them is one
+			// `validate` rejects, and nothing here would have a word to compare it to.
 			let n_inout = chip.cs.n_inout;
-			let served = chip_instances[chip_id][row].inout();
+			let instance = chip_instances.instance(chip_id, row);
+			let served = instance.inout();
 			for word in 0..n_inout {
 				let passed = call
 					.inout
@@ -267,8 +305,7 @@ impl ConstraintSystemM4 {
 					return Err(VerificationM4Error::CallMismatch {
 						chip_id,
 						row,
-						caller_chip,
-						caller_instance,
+						caller,
 						call_index,
 						word,
 						passed: passed.as_u64(),
@@ -311,23 +348,27 @@ fn is_acyclic(callees: &[Vec<usize>]) -> bool {
 
 #[cfg(test)]
 mod tests {
+	use proptest::prelude::*;
+
 	use super::{
-		super::{ShiftedValueIndex, ValueIndex, ValueSegment, ValueVecLayout},
+		super::{ShiftedValueIndex, ValueIndex, ValueSegment, ValueVecLayout, ZeroConstraint},
 		*,
 	};
-	use crate::error::OperandFault;
+	use crate::error::{OperandFault, VerificationError};
+
+	/// The value-vector layout every chip in these tests is shaped by.
+	const LAYOUT: ValueVecLayout = ValueVecLayout {
+		n_const: 0,
+		n_inout: 8,
+		n_witness: 8,
+		n_internal: 0,
+		n_scratch: 0,
+	};
 
 	/// A chip that calls each of `callees` once, over a value vector of 8 inout and 8 private
 	/// words.
 	fn chip(callees: &[usize]) -> EmbeddedConstraintSystem {
-		let cs = ValueVecLayout {
-			n_const: 0,
-			n_inout: 8,
-			n_witness: 8,
-			n_internal: 0,
-			n_scratch: 0,
-		}
-		.constraint_system_shape(vec![]);
+		let cs = LAYOUT.constraint_system_shape(vec![]);
 		let chip_calls = callees
 			.iter()
 			.map(|&chip_id| ChipCall {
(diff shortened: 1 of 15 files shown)


jimpo:
@tcoratger This is a big, important one to review. I'm gonna do a final review tomorrow.

crates/frontend/src/chip.rs
```diff
@@ -0,0 +1,236 @@
...
+	/// Checks that this system can be populated in one pass over the chips, in ID order.
+	///
+	/// Specifically checks that:
+	///
+	/// - every chip call names a chip of this system;
+	/// - the chips are in topological order, each calling only chips with a higher ID, so every
+	///   caller of a chip is populated before the chip itself;
+	/// - each chip's declared active-instance count is the number of invocations that reach it, and
+	///   no chip is left uncalled.
+	pub fn validate(&self) -> Result<(), CircuitM4Error> {
```
tcoratger, line 82: This doesn't check call arity, and I think it should.

I built a system by hand with a `ChipCall` passing 4 operands to a 3-inout chip:

```
CircuitM4::validate()       -> Ok
ConstraintSystemM4::validate() -> "the main chip's call #0 passes 4 operands to chip 0, which has 3 inout values"
WitnessM4::generate()       -> Ok
WitnessM4::verify()         -> Ok
```

So the fourth operand is dropped and nothing anywhere complains. `check_caller` loops `0..n_inout` and `generate` does the same, so operands past `n_inout` are unreachable by construction. The only thing that catches it is `ConstraintSystemM4::validate`, which nothing on the frontend or prover path calls.

Easiest fix is to add the arity loop here against `self.chips[call.chip_id].0.circuit.inout().len()` — three lines. Or call `self.to_constraint_system().validate()` at the end and wrap the error, if the clone is acceptable.

crates/m4-prover/src/m4_witness.rs
```diff
@@ -0,0 +1,477 @@
...
+	///
+	/// `fill_main` assigns the witness inputs of the main circuit; every other value in the system
+	/// is derived from them. Each chip's table holds one instance per invocation that reaches it,
+	/// ordered by caller: main's calls first, then the calls of each chip in ID order, and within a
+	/// chip by instance. The instance count is rounded up to a power of two by repeating the last
+	/// invocation, which satisfies the chip because the invocation it copies does.
+	///
+	/// # Panics
+	///
+	/// Panics if the system does not pass [`CircuitM4::validate`].
```
tcoratger: This isn't accurate — a system can pass `CircuitM4::validate` and still panic here.

Repro: hand-build a call whose third operand is `ShiftedValueIndex::plain(ValueIndex::scratch(0))`.

```
CircuitM4::validate()          -> Ok
ConstraintSystemM4::validate() -> "chip call #0 has a malformed operand #2: it refers to a scratch value"
WitnessM4::generate()          -> panicked at crates/core/src/constraint_system/shift.rs:820:
                                  index out of bounds: the len is 4 but the index is 4
```

Not reachable through `call_chip` (it pins its wires out of scratch), but `EmbeddedCircuit` / `ChipCall` are pub with pub fields and the tests below build them by hand, so it's a path people will use.

Either the precondition should name `ConstraintSystemM4::validate` instead, or `CircuitM4::validate` should cover the operand checks too (see my other comment).

crates/m4-prover/src/m4_witness.rs
```diff
@@ -0,0 +1,477 @@
...
+
+			// Push the calls of the chips populated so far. Calls only run to higher IDs, so
+			// zipping against the tables built up to here yields exactly this chip's callers.
+			for ((caller, caller_active), table) in iter::zip(&circuit.chips, &tables) {
+				if !caller.chip_calls.iter().any(|call| call.chip_id == chip_id) {
+					continue;
+				}
+				let constants = &caller.circuit.constraint_system().constants;
+				for instance in 0..*caller_active {
+					let values = table.instance_value_vec(instance, constants);
```
tcoratger: The loop nest is callee-major, so a caller that calls k different chips has each of its instances rebuilt k times.

I instrumented this with a 3-instance caller calling two leaves:

```
MATERIALIZE caller=0 instance=0 for callee=1
MATERIALIZE caller=0 instance=1 for callee=1
MATERIALIZE caller=0 instance=2 for callee=1
MATERIALIZE caller=0 instance=0 for callee=2   <- same three rows again
MATERIALIZE caller=0 instance=1 for callee=2
MATERIALIZE caller=0 instance=2 for callee=2
```

`instance_value_vec` isn't free: it allocates a zeroed `ValueVec` of `combined_len + n_scratch` and gathers the hidden column word by word. For the CRC mix chip in this PR that's 259 + 192 = 451 words (3.6 KB) allocated and 257 words copied, per instance, per callee. And this whole phase is serial, between two `populate_parallel` calls.

Flipping the nest fixes it and keeps the ordering `verify` expects, since pushes still happen main-first / ID order / instance-major / call-minor:

```rust
let mut pending: Vec<Vec<Vec<Word>>> = vec![Vec::new(); circuit.chips.len()];
for call in &circuit.main.chip_calls {
    pending[call.chip_id].push(eval_call(&main_values, call));
}
for (chip_id, (chip, n_active)) in circuit.chips.iter().enumerate() {
    let call_data = mem::take(&mut pending[chip_id]);
    // ... build `table` exactly as today ...
    let constants = &chip.circuit.constraint_system().constants;
    for instance in 0..*n_active {
        let values = table.instance_value_vec(instance, constants);  // once
        for call in &chip.chip_calls {
            pending[call.chip_id].push(eval_call(&values, call));
        }
    }
    tables.push(table);
}
```

Also drops the `any(|call| ...)` rescan per (caller, callee) pair.

crates/m4-prover/src/m4_witness.rs
```diff
@@ -0,0 +1,477 @@
...
+		if self.tables.len() != cs.chips.len() {
+			return Err(VerificationM4Error::WrongChipCount {
+				n_witness_chips: self.tables.len(),
+				n_chips: cs.chips.len(),
+			});
+		}
+		let chip_instances = iter::zip(&cs.chips, &self.tables)
+			.map(|((chip, _), table)| {
+				(0..table.n_instances())
+					.map(|instance| table.instance_value_vec(instance, &chip.cs.constants))
```
tcoratger: This expands the entire witness into `Vec<Vec<ValueVec>>` and holds all of it live at once. Each `ValueVec` is its own heap allocation, carries a copy of the constants, and carries a zeroed scratch tail that `verify` never reads.

For the CRC mix chip (259 committed + 192 scratch words) at 2^20 instances: the `ValueTable` is already ~2.1 GB, and this line puts ~3.8 GB more on top of it via a million allocations. Roughly triples peak memory for a check whose working set is one value vector.

The forcing function is the core signature `verify(&self, main: &ValueVec, chip_instances: &[Vec<ValueVec>])`. `verify` only ever needs random access by `(chip_id, row)`, so an accessor would do:

```rust
pub trait ChipInstances {
    fn n_instances(&self, chip_id: usize) -> usize;
    fn instance(&self, chip_id: usize, row: usize) -> Cow<'_, ValueVec>;
}
```

and `WitnessM4` implements it straight over `ValueTable`. If that's too much for this PR, at least restructure so only one chip's instances are live at a time.

crates/frontend/src/chip.rs
```diff
@@ -0,0 +1,236 @@
...
+				});
+			}
+			if *n_active == 0 {
+				return Err(CircuitM4Error::NeverCalled { chip_index });
+			}
+
+			// Only the active instances of this chip have their calls enforced, so only they
+			// demand an instance of the callee.
+			for call in &chip.chip_calls {
+				n_calls[call.chip_id] += n_active;
```
tcoratger: These counts multiply down the graph — a chain where each chip calls the next twice gives 2^depth, so ~64 chips overflow `usize` and wrap silently in release. `recompute_n_active` (same line lower down) has it too, and it runs inside `build_m4` before anything can validate.

Long before overflow, `generate` does `log2_ceil_usize(call_data.len())` and allocates `1 << log_instances` columns with nothing in between. A `checked_add` here into a `TooManyInstances { chip_index }` and a ceiling on `log_instances` would turn an OOM into a message.

tcoratger (commented):
Read through the whole thing. The model makes sense and the code is clean.

I checked the one thing I was most worried about: that `WitnessM4::generate` and `ConstraintSystemM4::verify` walk the calls in the same order. They do — main first, then callers in ID order, instance-major, call-minor, on both sides. And the `cursor == n_active` check at the end is what makes it a bijection and not just an injection. That part is right.

What I found is two contract problems and two things that will hurt once tables get big. Details inline. Nothing blocking correctness of what's here today, since no protocol consumes it yet.

On tests: `verify` has no unit tests in `m4.rs` at all — the test module there only covers `validate`. `MissingInstances`, `WrongInvocationCount` and `ChipInstance` are constructed and never asserted anywhere. `WrongInvocationCount` in particular is the check that makes the call/instance map a bijection, so it'd be good to have one test that breaks it on purpose.

Two behaviours also have no test:
- a caller with 2+ instances making **two** calls to the same callee. Every fixture makes at most one call per callee, so the instance-major/call-minor interleaving isn't pinned by anything. If the two files ever drift apart, no test fails.
- a call with fewer operands than the callee's `n_inout` (the documented zero-fill). `call_chip` asserts exact arity so only a hand-built call reaches it.

And `is_acyclic` would be a nice proptest — random `Vec<Vec<usize>>` with in-range indices, compare against a DFS three-colour check. Cheap, and covers multi-edges / diamonds / disconnected components that the five hand-written cases don't.

crates/core/src/constraint_system/m4.rs
```diff
@@ -0,0 +1,433 @@
...
+			let (chip, n_active) = &self.chips[chip_id];
+			let row = cursor[chip_id];
+			cursor[chip_id] += 1;
+			if row >= *n_active {
+				continue;
+			}
+
+			let n_inout = chip.cs.n_inout;
+			let served = chip_instances[chip_id][row].inout();
+			for word in 0..n_inout {
```
tcoratger, line 300: Minor, but: operands past `n_inout` are silently ignored here rather than reported. The doc above says malformed systems are `validate`'s to reject and this may "panic or misreport" — but it does neither, it quietly accepts. Combined with `CircuitM4::validate` not checking arity, that's how the dropped-operand case I mentioned gets all the way through `verify` clean.

crates/core/src/constraint_system/system.rs
```diff
@@ -325,25 +325,80 @@ impl ConstraintSystem {
 		constraint_index: usize,
 		operand_name: &'static str,
 	) -> Result<(), ConstraintSystemError> {
-		for term in operand {
+		match self.operand_fault(operand) {
```
tcoratger, line 328: Splitting out `operand_fault` is the right call, but this adapter is now ~55 lines mapping six `OperandFault` variants onto six near-identical `ConstraintSystemError` ones. Every fault exists twice in the public API and adding a seventh means touching three places with nothing to catch a miss.

The shape you already wrote one screen down for `ChipCallOperand` works here too:

```rust
#[error("{constraint_kind} #{constraint_index} operand {operand_name} is malformed: {source}")]
ConstraintOperand {
    constraint_kind: ConstraintKind,
    constraint_index: usize,
    operand_name: &'static str,
    #[source] source: OperandFault,
},
```

`validate_operand` becomes a three-line `map_err` and six public variants go away. It does change the error surface so it may belong in a follow-up — but the asymmetry between the two operand checks is a bit odd as it stands.

crates/core/src/error.rs
```diff
@@ -160,3 +224,66 @@ pub enum VerificationError {
...
+	#[error(
+		"{n_invocations} invocations reach chip #{chip_id}, which has {n_active} active instances"
+	)]
+	WrongInvocationCount {
+		chip_id: usize,
+		n_invocations: usize,
+		n_active: usize,
+	},
+	#[error(
+		"call #{call_index} of {} reaches chip #{chip_id} as invocation #{row}, passing {passed:016x} \
```
tcoratger, line 242: `{passed:016x}` renders as `passing 000000000000dead`. `{passed:#018x}` gives the `0x` prefix, which matches how words print elsewhere in the crate.

crates/core/src/constraint_system/m4.rs
```diff
@@ -0,0 +1,433 @@
...
+		}
+		Ok(())
+	}
+}
+
+/// Checks that the call graph given by the callee list of each chip is acyclic.
+///
+/// This is Kahn's algorithm: chips with no remaining callers are removed one at a time, and any
+/// chip still left when none remain lies on a cycle. A chip that calls itself is a cycle.
+fn is_acyclic(callees: &[Vec<usize>]) -> bool {
```
tcoratger: Maybe in the future we will need to put that somewhere more appropriate than isolated here but we can handle that later

crates/core/src/constraint_system/system.rs
```diff
@@ -325,25 +325,80 @@ impl ConstraintSystem {
...
+			}
+			Some(OperandFault::CollapsibleShiftSequence { composition }) => {
+				Err(ConstraintSystemError::CollapsibleShiftSequence {
+					constraint_kind,
+					constraint_index,
+					operand_name,
+					composition,
+				})
+			}
+			Some(OperandFault::ScratchValueIndex) => {
```
tcoratger: Maybe we need to add some fields to `OperandFault::ScratchValueIndex` to enrich the enum variant with adequate data no?

crates/core/src/error.rs
```diff
@@ -62,6 +62,70 @@ pub enum ConstraintSystemError {
...
+
+/// Names the chip of an M4 constraint system that a diagnostic is about.
+///
+/// The main chip is not one of the numbered chips, so it has no index.
+fn chip_name(chip_index: Option<usize>) -> String {
+	match chip_index {
+		Some(chip_index) => format!("chip #{chip_index}"),
+		None => "the main chip".to_string(),
+	}
+}
```
tcoratger: This is a bit strange to have this isolated method here just for the format of the error, probably there is a better way

crates/core/src/error.rs
```diff
@@ -160,3 +224,66 @@ pub enum VerificationError {
...
+
+/// Names the chip instance an M4 diagnostic blames a call on.
+///
+/// The main chip runs once, so only a numbered chip's instance is worth naming.
+fn caller_name(chip_index: Option<usize>, instance: usize) -> String {
+	match chip_index {
+		Some(chip_index) => format!("chip #{chip_index} instance #{instance}"),
+		None => "the main chip".to_string(),
+	}
+}
```
tcoratger: Same here a bit strange to have such a small helper for the format of the error, used only one time

crates/frontend/src/chip.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// Names the circuit of an M4 system that a diagnostic is about.
+///
+/// The main circuit is not one of the numbered chips, so it has no index.
+fn circuit_name(chip_index: Option<usize>) -> String {
+	match chip_index {
+		Some(chip_index) => format!("chip #{chip_index}"),
+		None => "the main circuit".to_string(),
+	}
+}
```
tcoratger: Same here I think we have the exact same in another file

crates/frontend/src/chip.rs
benji-potamus: Fixed — and thanks for the concrete repro, that made it obvious the check was missing rather than merely duplicated.

`CircuitM4::validate` now does the arity check, and the operand-fault check alongside it (your other comment), in a new `validate_calls` helper mirroring the core one. Rather than lower and clone, `ConstraintSystem::operand_fault` is now `pub` so the frontend can ask the same question of a borrowed system. Two new variants, `CircuitM4Error::WrongCallArity` and `CircuitM4Error::CallOperand`.

That makes the frontend check a superset of what the lowered one would catch, up to the two things it doesn't need to: acyclicity, which the stronger topological-order requirement already rules out, and each chip's own compiled constraint system, which the compiler is what keeps well-formed. So the layering note in the PR description is now wrong and I've rewritten it — validation is no longer split across the two, and nothing downstream has to run the lowered check to know its preconditions hold.

Test: `validate_rejects_a_call_passing_more_operands_than_the_callee_takes`.

crates/m4-prover/src/m4_witness.rs
benji-potamus: Took the second option: `CircuitM4::validate` now covers the operand checks, so the precondition as written is true rather than needing to name a method on the lowered system. `validate_calls` runs `ConstraintSystem::operand_fault` over every call operand against the caller's own system, reporting `CircuitM4Error::CallOperand`.

Your scratch-operand repro is now a test — `validate_rejects_a_call_operand_naming_a_scratch_value`. You're right that this is a path people will use: `call_chip` pins its wires, but `EmbeddedCircuit` and `ChipCall` are `pub` with `pub` fields and the tests build them by hand throughout.

The `# Panics` line now also says what the precondition covers, since "does not pass `validate`" was doing a lot of unstated work: both the ordering `generate` walks the chips in and the well-formedness of the operands it evaluates.

crates/m4-prover/src/m4_witness.rs
benji-potamus: Done, essentially your version. The nest is now caller-major: main's calls are evaluated into a `pending: Vec<Vec<Vec<Word>>>` up front, and each chip takes its own list with `mem::take`, builds its table, then materializes each active instance **once** and pushes every call it makes onto the callees' lists. The `any(|call| ...)` rescan is gone with it — replaced by an `is_empty()` guard so a leaf chip doesn't take the constants reference for nothing.

The ordering argument is the same one that made the old nest correct, just read the other way: pushes still happen main-first, then callers in ID order, instance-major and call-minor, which is what `verify` walks.

That order was the thing I wanted a test to hold, since with one call per callee it's unobservable — `generate_interleaves_a_callers_instances_before_its_calls` has a 2-instance caller calling one leaf twice with distinguishable operands, and asserts the four rows land as `a0, b0, a1, b1` rather than `a0, a1, b0, b1`.

crates/m4-prover/src/m4_witness.rs
benji-potamus: Took the trait. You're right that the core signature is the forcing function, and it's much cheaper to change now than once something proves a `WitnessM4`.

`binius_core::m4::ChipInstances`, close to your sketch plus an `n_chips` so `verify` can still report `WrongChipCount`:

```rust
pub trait ChipInstances {
    fn n_chips(&self) -> usize;
    fn n_instances(&self, chip_id: usize) -> usize;
    fn instance(&self, chip_id: usize, row: usize) -> Cow<'_, ValueVec>;
}
```

`verify` is now generic over `I: ChipInstances + ?Sized`, with an impl for `[Vec<ValueVec>]` so hand-built witnesses (and the tests) pass a slice unchanged. On the prover side a private `TableInstances` pairs the tables with the system, since the constants are the one part of an instance a table doesn't store; `WitnessM4::verify` is now a single delegating call and the `WrongChipCount` check moves into core, where it guards the indexing.

The cost is that an instance serving a call gets rebuilt when the call is checked, having already been built for its own constraints — roughly 2x the materializations for O(1) live instead of O(total). Both the trait doc and `verify`'s doc say so explicitly.

crates/frontend/src/chip.rs
benji-potamus: Good catch — this one I'd missed entirely.

`recompute_n_active` now uses `saturating_add`, and `validate` does the accumulation with `checked_add` into a new `CircuitM4Error::TooManyInstances { chip_index }`. The pairing matters: saturating first means an overflowed count stays *too large* for the invocations that reach the chip rather than wrapping to a plausible one, so `validate` sees the disagreement instead of two wrapped counts agreeing with each other. Blame lands on the callee whose total left the range, which is where the doubling actually happened.

Test: `validate_rejects_an_instance_count_that_outgrows_a_usize` builds your chain — 70 chips, each calling the next twice — and asserts it's chip 64 that's reported.

I left the `log_instances` ceiling out. Overflow is a correctness bug (a wrapped count is silently wrong); a 2^40-instance system is a resource limit, and any ceiling I picked would be arbitrary — there's no bound the types imply, since instances are separate value vectors rather than one index space. Happy to add one if you have a number in mind.

crates/core/src/constraint_system/m4.rs
benji-potamus: Right on both counts — the behaviour wasn't in the documented set, and the reason it mattered was the missing frontend check.

The hole is closed at the source: `CircuitM4::validate` now rejects the over-long call, so nothing reaches here with operands past `n_inout`. For this loop I've kept the `0..n_inout` bound and said why in a comment, and widened `verify`'s doc from "may panic or misreport" to "may panic, misreport, or pass", naming this case specifically as the one it silently ignores. Comparing operands the callee has no inout value for isn't something `verify` could do anyway — there's no word on the other side.

crates/core/src/constraint_system/system.rs
benji-potamus: Agreed, and done in this PR rather than a follow-up — the asymmetry was mine to create, so it's mine to not ship.

`ConstraintOperand` with your exact message shape, and `validate_operand` is now:

```rust
match self.operand_fault(operand) {
    None => Ok(()),
    Some(source) => Err(ConstraintSystemError::ConstraintOperand {
        constraint_kind,
        constraint_index,
        operand_name,
        source,
    }),
}
```

Six public variants gone, ~55 lines down to 9, and adding a seventh fault is now one place. The error surface change is contained: every construction and match was inside `system.rs` and its tests, so nothing outside the crate moved.

This also answers your `ScratchValueIndex` comment below — the arm that motivated enriching the variant no longer exists, and the context it was missing (kind, index, operand name) now comes from the wrapper for every fault rather than being duplicated into each one.

crates/core/src/constraint_system/system.rs
benji-potamus: This one's resolved by the collapse in the thread above rather than by adding fields: the whole `Some(OperandFault::ScratchValueIndex) => Err(ConstraintSystemError::ScratchValueIndex { .. })` arm is gone.

`OperandFault` deliberately says only *how* a term is malformed, never *where* — that's what lets one fault be reported under two namings, by a constraint and by a chip call. The data you'd want here (constraint kind, index, operand name) is the "where", and it now comes from the `ConstraintOperand` wrapper uniformly for all six faults. `ScratchValueIndex` itself genuinely has nothing more to say: the index is a scratch index, and which one it is doesn't help — unlike `OutOfRangeValueIndex`, where the segment and the bound do.

crates/core/src/error.rs
benji-potamus: Fair — and it turned out to be the same smell as your comment on `caller_name` and on the frontend's copy, so I've fixed all three together.

`chip_name` is now `pub struct ChipName(pub Option<usize>)` with a `Display` impl. That drops the `String` allocation on the formatting path, and — the actual win — the frontend can use it instead of carrying `circuit_name`, which was the same function with "the main circuit" for "the main chip". Since the frontend's field is literally `CircuitM4::chips`, "chip #N" was the right noun on both sides anyway.

If you'd still rather not have a type here at all, the alternative is inlining a `match` into each `#[error]` attribute, which thiserror allows but reads worse at three call sites across two crates. Happy to go that way if you prefer it.

crates/core/src/error.rs
benji-potamus: Same treatment, with the single-use point taken: `CallerName` is a `Display` newtype like `ChipName`, but **private** to `error.rs` since nothing outside it names a caller.

It also let me simplify the caller side. `CallMismatch` carried `caller_chip: Option<usize>` and `caller_instance: usize` as separate fields, which meant `check_caller` destructured its own `Option<(usize, usize)>` parameter into two just to store it back. The variant now holds `caller: Option<(usize, usize)>` — the shape it's given — and the destructuring is gone. So the helper isn't purely a formatting concern any more: it's the Display for a field that's now honestly one field.

crates/frontend/src/chip.rs
benji-potamus: You're right, and it was — `circuit_name` here was `chip_name` in `binius-core`'s `error.rs`, differing only in saying "the main circuit" where the other said "the main chip".

Deleted. Both crates now use `binius_core::error::ChipName`. Unifying the wording is a fix rather than a loss: `CircuitM4`'s field is literally `chips`, so "chip #N" was already the right noun on this side, and "the main chip" matches `CircuitM4::main`'s role in the model.

crates/core/src/constraint_system/m4.rs
benji-potamus: Agreed — leaving it here for now. It has exactly one caller, and moving it before there's a second (a graph utility module, or wherever the eventual permutation argument wants its own traversal) would be guessing at where it belongs.

It did pick up a proptest against an independent three-colour DFS, per your review comment, which is the thing that would make moving it safe later.

crates/core/src/error.rs
benji-potamus: I checked this one and I think the premise is inverted, so I've left it as `{passed:016x}` — pushing back rather than declining, so overrule me if you disagree.

`016x` is what the crate does: 13 occurrences, all of them here in `error.rs`, including every arm of `ConstraintViolation` and the `ConstantMismatch` variant directly above this one:

```rust
#[error("{val:016x} != 0")]
#[error("({a:016x} & {b:016x}) ^ {c:016x} = {residue:016x} != 0")]
#[error("value {value_index} is {actual:016x}, but the system declares the constant {expected:016x}")]
```

`#018x` appears three times in the whole tree, and the two that matter are `Word`'s own `Debug` impl (`Word({:#018x})`), where the prefix is disambiguating the debug representation rather than formatting a word inline in prose.

So switching this one variant would make it the odd one out among its immediate siblings. I'd rather change all fourteen or none — and if you do want the `0x` prefix everywhere, that's a clean standalone commit that shouldn't ride along with a feature PR. Say the word and I'll open one.

benji-potamus:
Thanks @tcoratger — this was a genuinely useful review. Pushed 51ce2bd addressing all of it; replies inline on each thread, summary here.

**Taken as proposed** (4 of 5 inline findings):

- **Call arity unchecked on the frontend path.** `CircuitM4::validate` now checks arity *and* call-operand faults, so your dropped-fourth-operand case is rejected before generation. `ConstraintSystem::operand_fault` went `pub` to make the second check possible without a clone.
- **`generate`'s panic contract inaccurate.** Fixed by making the contract true rather than by weakening it — your scratch-operand repro is now a test.
- **Callee-major loop nest.** Flipped, essentially your code. Each instance is built once regardless of how many chips it calls; the `any(|call| ...)` rescan is gone.
- **`verify` expanding the whole witness.** Took the trait. `ChipInstances` with your three methods plus `n_chips`; `verify` is generic over it, with an impl for `[Vec<ValueVec>]` so hand-built witnesses are unaffected, and the prover serves instances off its tables one at a time. Peak memory for the check goes from O(total instances) to O(1), paying ~2x materializations.
- **Instance-count overflow.** `recompute_n_active` saturates, `validate` reports `TooManyInstances`. I left the `log_instances` ceiling out and said why on the thread — happy to add one if you have a number.

**Pushed back on one:** the `{passed:#018x}` suggestion. `016x` is what this file does in all 13 other places, including the variant directly above; details on the thread.

**Tests** — all four gaps you named, plus the proptest:

| Test | Gap |
|---|---|
| `verify_rejects_a_chip_with_fewer_instances_than_active_ones` | `MissingInstances` never asserted |
| `verify_rejects_more_invocations_than_a_chip_has_active_instances` | `WrongInvocationCount` — the bijection, not just the injection |
| `verify_reports_the_instance_whose_own_constraints_fail` | `ChipInstance` never asserted |
| `verify_rejects_a_witness_covering_the_wrong_number_of_chips` | `WrongChipCount`, which moved into core |
| `verify_constrains_the_inout_values_a_call_passes_no_operand_for` | the documented zero-fill, both directions |
| `generate_interleaves_a_callers_instances_before_its_calls` | 2-instance caller, two calls to one callee — pins instance-major/call-minor on both sides |
| `is_acyclic_agrees_with_a_depth_first_search` | proptest vs. an independent three-colour DFS |
| `validate_rejects_a_call_passing_more_operands_than_the_callee_takes` | frontend arity |
| `validate_rejects_a_call_operand_naming_a_scratch_value` | frontend operand faults |
| `validate_rejects_an_instance_count_that_outgrows_a_usize` | your 2^depth chain, at 70 chips |

You were right that the interleaving was pinned by nothing: every fixture made at most one call per callee, so `generate` and `verify` could have drifted apart silently. The new test uses a caller whose two calls forward distinguishable operands, so `a0, b0, a1, b1` and `a0, a1, b0, b1` are now distinguishable outcomes.

I've also rewritten the PR description. Two of its "known deferrals" are no longer true — validation is no longer layered across the frontend and core, and the `verify` memory note is obsolete — and it would have been misleading to leave them standing.

The three deferrals that remain are unchanged and still real: `build_m4` doesn't call `validate` itself, a chip circuit with `add_witness` wires fails late rather than at build time, and `ChipRef` has no builder provenance.

@jimpo this is ready for your pass whenever. Everything in this round is either a contract fix or a memory/CPU fix on paths nothing proves yet; the model and the call/instance correspondence you'd be reviewing are unchanged.

crates/core/src/constraint_system/m4.rs
```diff
@@ -0,0 +1,637 @@
...
+
+		// The chips each chip calls, indexed by the caller's chip ID.
+		let mut callees = vec![Vec::new(); n_chips];
+		for (chip_index, (chip, _)) in self.chips.iter().enumerate() {
+			chip.validate()?;
+			self.validate_calls(Some(chip_index), &chip.chip_calls)?;
+			callees[chip_index].extend(chip.chip_calls.iter().map(|call| call.chip_id));
+		}
+
+		if !is_acyclic(&callees) {
```
jimpo: Let's be stricter and require that the chips are listed in topological order -- that is, they only reference later chips. This is what is already required in the CircuitM4, so it's easy to construct, and easier to check.

benji-potamus: Done in c847b61 — and it's a real fix, not just a tightening. The acyclicity check was the wrong invariant here.

`verify` walks the chips in ID order, and `check_caller` advances the cursor of each callee as it goes. So a *backward* call — chip 2 calling chip 1 — is served by a row of an instance the walk has already passed, even though the graph is perfectly acyclic. Kahn's algorithm accepted that, so core was accepting systems its own `verify` cannot check. It was only safe because `CircuitM4::validate` happened to be stricter, and the doc on `ConstraintSystemM4` admitted as much: "only a system whose calls run to higher IDs is provable. The frontend's `CircuitM4::validate` is what enforces that ordering." Core now enforces the invariant it actually depends on.

The check is one comparison per call, in `validate_calls` next to the range and arity checks:

```rust
if let Some(caller) = chip_index
    && call.chip_id <= caller
{
    return Err(ConstraintSystemError::CallOutOfOrder { chip_index: caller, callee: call.chip_id });
}
```

Main is `None` there — it isn't one of the numbered chips and runs before all of them, so it may call any.

What went with it:

- `is_acyclic` and its Kahn's-algorithm implementation are deleted, along with the `callees: Vec<Vec<usize>>` that `validate` was building only to feed it. That also moots @tcoratger's two threads on it — where it should live, and his suggestion of a proptest against a three-colour DFS. The ordering check doesn't need one.
- `ConstraintSystemError::CyclicChipCalls` → `CallOutOfOrder { chip_index, callee }`, same name, shape and message as the frontend's variant.
- A chip calling itself is now reported as the out-of-order call it is, rather than as a separate cycle case — one fewer concept.
- The `ConstraintSystemM4` doc no longer needs to explain that validation is weaker than what's provable, and `CircuitM4::validate`'s doc no longer claims it's the stricter of the two.

Tests: `validate_rejects_a_chip_that_calls_itself` and a new `validate_rejects_a_chip_that_calls_an_earlier_chip` (chip 2 → chip 1, an acyclic graph the old check passed); `validate_accepts_an_acyclic_call_graph` becomes `validate_accepts_calls_running_only_to_later_chips`. I also had to fix a fixture — `validate_rejects_a_call_to_a_chip_that_does_not_exist` used a chip calling itself *and* an out-of-range ID, so the ordering check now fires first; its first call is a forward one now, leaving the range check as the only thing that can catch it.

CI is running. It's a separate commit on top of 51ce2bd4 rather than an amend, so the diff against what you commented on is just this change.

PR #2111, opened 2026-08-11
BINIUS-433: take the inner statement symbolically in IOPVerifier::verify

benji-potamus:
Next in the recursion stack, on top of #2093. Closes [BINIUS-433](https://linear.app/jimpo/issue/BINIUS-433/take-the-inner-statement-symbolically-in-iopverifierverify).

`IOPVerifier::verify` took `inout: &[Word]`, so a channel that builds a circuit received the statement already concrete and baked it in as constants. The circuit then verified one instance of the inner system — the one it happened to be built with. It now takes `&[Channel::Word]`, and a builder channel hands back wires that are its circuit's own public input.

Two places read the statement.

**Fiat-Shamir** — `observe_words` already took the channel's word type, so the `From<Word>` lift at the call site just goes away.

**The public-words MLE** — `shift::check_eval` evaluated the public segment through `compute_public_value`, which carries field elements and so cannot carry symbolic words; the old code smuggled the words past it by capturing them in `PublicWordsEvalFn`. It now runs the algorithm symbolically, over `subset_sum`. `compute_public_value` is for values too slow to rebuild — the monster multilinear, proportional to the inner circuit's size. The public segment is small, so it is not one of them, and `PublicWordsEvalFn` is gone.

## The Spartan wrapper carries its statement symbolically too

That wrapper built its circuit from `dummy_inout_words`, which worked only while nothing in the circuit read a statement word's bits. `subset_sum` does, and a concrete word decides *which* elements the sum runs over, so folding it would write the dummy into the circuit's shape.

So the wrapper channels carry a `CircuitWord`: either a `Constant`, for a word the protocol fixes, or the word's bits. `subset_sum` is `Σ bits[i]·elems[i]` and `select` a mux over them, built the same way whatever the statement is. The statement enters as inout wires whose bits are hinted — a hint over a public input is not a value a prover chooses, since the outer verifier recomputes it — and `dummy_inout_words` is gone.

The cost is exactly the statement. On the sha256 preimage test:

| | before | after |
|---|---|---|
| inout wires | 324 | **332** |
| derived wires | 129 | 129 |
| public (padded) | 1024 | 1024 |
| mul constraints | 49,366 | 49,366 |

Every operation over the bits is public-derivable, so `ConstraintBuilder::mul` emits no constraint, and the intermediates feed only other derived wires, so `WitnessLayout::sparse` gives them no slot. Outer proof bytes change, since the outer constraint system does.

`ZKWrappedVerifierChannel::observe_words` is the one place a value still leaves the circuit: it reads the words back off the wires to feed the inner Fiat-Shamir state, which lives outside the circuit and which the inner prover mirrors. A read, not a decision — nothing the circuit contains depends on it.

## Also in here

- `reduce_constraints` and `check_eval` take the **inout words** rather than the whole public segment, reading the constants off the constraint system. The M4 verifier was passing `&cs.constants` for a batch that hides its inout words; it passes nothing now. A caller can no longer restate the constants wrongly — which the existing comment in `verify` already said was the point.
- `evaluate_words_mle` is the channel-generic one; the native form the prover needs (`WordIPProverChannel` has no `subset_sum`, so a prover only ever has native words) is `evaluate_words_mle_native`, the suffix `subset_sum_word` and `select_word` already use.

## Cost, recursion side

On the crc64 end-to-end test the recursive circuit goes from 129,959 to 130,599 BMUL and 3,912 to 4,259 ZERO: the statement's four words at one `subset_sum` each (64 selects over the `(lo, hi)` pair), plus the eq-indicator tensor over the word index.

```
recursive circuit: 166312 gates, 0 AND, 130599 BMUL, 4259 ZERO, 3336 recorded inputs
```

## Testing

`a_statement_the_proof_is_not_about_fails` is what says the recursion wires are wired in rather than merely present: filling them with a corrupted statement leaves the shift reduction's consistency check unsatisfied (`.assert_zero[1].lo`) instead of passing regardless. `CircuitWord`'s bit arithmetic is tested against the concrete helpers — `subset_sum`, `select`, `>>`, and the word/element embedding.

`cargo test --workspace --exclude binius-arith-bench`: **1454 passed, 0 failed, 21 ignored** across 58 binaries. Clippy clean on every crate touched; rustdoc clean; `fmt --check` clean.

Pre-existing breakages on this machine, excluded and confirmed against pristine `main`: `binius-arith-bench` does not compile here (aarch64 `poly64x2_t: Underlier`), and `cargo clippy --all-features` fails in `binius-ip-prover` (`Data cannot be shared between threads safely`) and `binius-examples` (a `needless_borrow` from a newer clippy).

### DIFF
--- a/crates/spartan-verifier/src/wrapper/builder_channel.rs
+++ b/crates/spartan-verifier/src/wrapper/builder_channel.rs
@@ -11,10 +11,10 @@ use std::{
 use binius_core::word::Word;
 use binius_field::{Field, util::FieldFn};
 use binius_iop::channel::{IOPVerifierChannel, OracleLinearRelation, OracleSpec};
-use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel, select_word, subset_sum_word};
+use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel, select_word};
 use binius_spartan_frontend::circuit_builder::{CircuitBuilder, ConstraintBuilder};
 
-use super::circuit_elem::CircuitElem;
+use super::circuit_elem::{CircuitElem, hinted_subset_sum};
 
 /// A channel that symbolically executes a verifier, building up an IronSpartan constraint system.
 ///
@@ -132,10 +132,14 @@ impl<F: Field> WordIPVerifierChannel<F> for IronSpartanBuilderChannel<F> {
 	fn observe_words(&mut self, _words: &[Word]) {}
 
 	fn subset_sum(&mut self, elems: &[Self::Elem], word: &Word) -> Self::Elem {
-		// The word is concrete, so which elements the sum runs over is settled while building.
-		subset_sum_word(elems, *word)
+		// Hinted rather than folded: the statement is a dummy while building. See
+		// `hinted_subset_sum`.
+		hinted_subset_sum(&self.builder, elems, *word)
 	}
 
+	// Folded rather than hinted, unlike `subset_sum`: the only words reaching this are a code
+	// proximity test's query indices, and that test is not run symbolically here — which is also
+	// why `sample_bits` below can answer with zero.
 	fn select(&mut self, elems: &[Self::Elem], word: &Word) -> Self::Elem {
 		select_word(elems, *word)
 	}
--- a/crates/spartan-verifier/src/wrapper/builder_channel.rs
+++ b/crates/spartan-verifier/src/wrapper/builder_channel.rs
@@ -9,12 +9,12 @@ use std::{
 };
 
 use binius_core::word::Word;
-use binius_field::{Field, util::FieldFn};
+use binius_field::{BinaryField1b as B1, ExtensionField, Field, util::FieldFn};
 use binius_iop::channel::{IOPVerifierChannel, OracleLinearRelation, OracleSpec};
-use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel, select_word};
+use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel};
 use binius_spartan_frontend::circuit_builder::{CircuitBuilder, ConstraintBuilder};
 
-use super::circuit_elem::{CircuitElem, hinted_subset_sum};
+use super::{circuit_elem::CircuitElem, circuit_word::CircuitWord};
 
 /// A channel that symbolically executes a verifier, building up an IronSpartan constraint system.
 ///
@@ -46,6 +46,23 @@ impl<F: Field> IronSpartanBuilderChannel<F> {
 	fn alloc_inout_elem(&self) -> CircuitElem<F, ConstraintBuilder<F>> {
 		let wire = self.builder.borrow_mut().alloc_inout();
 		CircuitElem::wire(&self.builder, wire)
+	}
+
+	/// Allocates the inner statement as this circuit's public input.
+	///
+	/// Pass the result to the verifier this channel drives. The words are wires, so what is
+	/// recorded is a circuit about the inner system rather than about one statement — which it
+	/// could not be anyway, since the build precedes every instance.
+	pub fn statement(&mut self, n_inout: usize) -> Vec<CircuitWord<F, ConstraintBuilder<F>>>
+	where
+		F: ExtensionField<B1>,
+	{
+		(0..n_inout)
+			.map(|_| {
+				let wire = self.builder.borrow_mut().alloc_inout();
+				CircuitWord::from_wire(&self.builder, wire)
+			})
+			.collect()
 	}
 
 	fn alloc_precommit_elem(&self) -> CircuitElem<F, ConstraintBuilder<F>> {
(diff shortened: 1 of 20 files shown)


crates/spartan-verifier/src/wrapper/builder_channel.rs
```diff
@@ -132,10 +132,14 @@ impl<F: Field> WordIPVerifierChannel<F> for IronSpartanBuilderChannel<F> {
 	fn observe_words(&mut self, _words: &[Word]) {}
 
 	fn subset_sum(&mut self, elems: &[Self::Elem], word: &Word) -> Self::Elem {
-		// The word is concrete, so which elements the sum runs over is settled while building.
-		subset_sum_word(elems, *word)
+		// Hinted rather than folded: the statement is a dummy while building. See
+		// `hinted_subset_sum`.
+		hinted_subset_sum(&self.builder, elems, *word)
```
jimpo: This looks wrong. Seems like `IronSpartanBuilderChannel` needs a symbolic Word.

crates/verifier/src/protocols/shift/verify.rs
jimpo: Do we still need this? If not, or it's only used in tests or something, get rid of it. And rename `channel_words_mle` to `evaluate_words_mle`.

jimpo (changes requested):

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Still needed — `prover/src/protocols/shift/phase_2.rs:300` computes the same public evaluation while proving, and it can't share the channel path: `WordIPProverChannel` deliberately has no `subset_sum` ("the arithmetic over a word's bits is the verifier's alone"), so a prover only ever has native words.

Renamed as you asked: the channel one is `evaluate_words_mle`, and the native one is `evaluate_words_mle_native` — the suffix `subset_sum_word` and `select_word` already use for the concrete case. Pushed in 92c3f10e.

crates/spartan-verifier/src/wrapper/builder_channel.rs
benji-potamus: You're right that it's the wrong shape, and for a sharper reason than I gave in the PR body: `subset_sum` is a general primitive with no public-derivability contract in its docs, and `CircuitBuilder::hint` allocates a *private* wire when its inputs aren't all public. So a future caller passing elements derived from private wires would get an unconstrained free variable rather than a sum. That is a footgun `compute_public_value` avoids only because its contract is written down.

A symbolic `Word` removes it: the statement enters the wrapper circuit as wires, `subset_sum` is a real sub-circuit over their bits, and the shape stops depending on the value, so no hint is needed anywhere.

Scoping it, since it's bigger than it looks:

* A `CircuitWord<F, B>` next to `CircuitElem`: either `Constant(Word)` or 64 bit-elements. `From<Word>` folds, `Shr<u32>` shifts the bit vector, `subset_sum` is `Σ bits[i]·elems[i]`, `select` is a mux. Over public/derived wires every one of those multiplications is free, and the bit decomposition is a hint over a public input, so this costs the wrapper nothing in constraints either — it just stops lying about where the statement is.
* All three channels need the type and a way to get the statement words, allocated in the same order (the `inout_alloc` pattern they already share): the builder allocates inout wires, the instance and witness generators fill them.
* `dummy_inout_words` goes away. The drivers are the easy part — `prover/src/zk_config.rs:139` and `verifier/src/zk_config.rs:185` already hold the real statement; only `setup` passes dummies today.

That is ~300 lines across `spartan-verifier`, `spartan-prover` and both `zk_config`s, and it changes the outer constraint system (n_inout more public wires), so ZK proof bytes change.

Which do you want? I'd rather file it as its own issue and keep this PR to the Binius64 verifier signature — but if you want it here, say so and I'll do it. Either way it removes the wrapper's last use of the dummy-statement escape hatch, which the BINIUS-433 description already called out as the thing a real recursive verifier can't rely on.

benji-potamus: Done in 2e7a5f65 — the hint is gone and the wrapper carries a `CircuitWord`: either a `Constant`, for a word the protocol fixes, or the word's bits. `subset_sum` is `Σ bits[i]·elems[i]` and `select` a mux over them, built the same way whatever the statement is. `dummy_inout_words` is gone with it; the statement enters as inout wires whose bits are hinted, which is safe in the way the old hint was not — a hint over a *public* input is not a value a prover chooses.

The cost is exactly the statement and nothing else. On the sha256 preimage test the wrapper's inout wires go 324 → 332, the inner circuit's 8:

| | before | after |
|---|---|---|
| inout wires | 324 | **332** |
| derived wires | 129 | 129 |
| public (padded) | 1024 | 1024 |
| mul constraints | 49,366 | 49,366 |

Every operation over the bits is public-derivable, so `ConstraintBuilder::mul` emits no constraint, and the intermediates feed only other derived wires, so `WitnessLayout::sparse` doesn't even give them a slot. Outer proof bytes change, since the outer constraint system does.

New tests cover the bit arithmetic against the concrete helpers — `subset_sum`, `select`, `>>` and the word/element embedding. Full suite 1454 passed, 0 failed.

One thing to flag, since it is the one place a value still leaves the circuit: `ZKWrappedVerifierChannel::observe_words` reads the words' values back off the wires to feed the *inner* Fiat-Shamir state, which lives outside the circuit and which the inner prover mirrors. That is a read, not a decision — nothing about what the circuit contains depends on it, which is what keeps the build instance-independent.

jimpo:
This is bad. I'm now thinking we should just do ring-switching on the public segment. That way the public inputs are just field elements.

jimpo:
replaced by #2116

PR #2113, opened 2026-08-11
Bind the shift indicator bit index with a sumcheck

benji-potamus:
Closes [BINIUS-464](https://linear.app/jimpo/issue/BINIUS-464/shift-sumcheck-to-bind-the-i-variables).

The shift verifier computed the h evaluation itself, summing the shift indicators over the output bit index `i` through the partial-evaluation recurrences. A sumcheck over that index replaces the summation: the prover claims the h evaluation and proves it in `Word::LOG_BITS` rounds, leaving the verifier one evaluation of the Lagrange weights and one of the eight shift indicators, both at a single point. The recurrences move to the prover crate along with the summation they serve.

### Commits

1. **Evaluate the shift indicators at a point** — adds a closed-form evaluation of all eight indicator multilinears at an arbitrary point, alongside the existing recurrences. It is built from the two carry indicators of `i + s`: `sll` is `srl` with the two bit indices exchanged, `rotr` adds the wrapped carry, `sra` adds the overflow indicator at the sign position, and the `*32` variants are the same four rules over five variables times the equality of the halves the indices fall in. Purely additive — nothing else changes.
2. **Compute the shift h evaluation once, before phase 2** — the h evaluation was computed inside `build_monster_segments`, once per operation, from four `r_zhat_prime` values that are always the same one (phase 1 already builds its h multilinear from the BitAnd claim's challenge alone). It moves up to `prove`, and `Phase1Output` replaces the raw `SumcheckOutput` at the phase boundary so the challenge axes are split once.
3. **Bind the shift indicator bit index with a sumcheck** — the protocol change. The prover sends the h evaluation after phase 2 and proves it with a bivariate-product sumcheck over the Lagrange weights and the shift indicators; the verifier verifies the sumcheck and checks its reduced claim against its own evaluation of both. `partial_eval_sigmas` / `partial_eval_sigmas_transpose` / `partial_eval_phi` move to `binius-prover`, where `build_shift_ind` assembles the eight indicators from them into the multilinear the sumcheck runs over.

Two tests tie the prover's and verifier's definitions together: `build_matches_the_verifier_point_evaluation` checks the assembled multilinear against `evaluate_shift_inds` at every bit-index vertex, and `claimed_sum_is_the_weighted_indicator_sum` checks the claimed sum. The recurrences keep their `*_matches_reference` tests against a brute-force indicator.

### Verifier cost

The verifier drops the eight partial-evaluation recurrences over the 64-point bit-index cube (and the `r_j` section of the monster evaluation's input) in exchange for six sumcheck rounds, one extra transcript element, and two 64-element evaluations. The `h_eval` factor is pulled out of the monster `FieldFn` and multiplied in by `check_eval`, because it is a prover message and `compute_public_value` may only take public inputs; the reduced-claim check is a plain function, not a `compute_public_value` call.

### Testing

`binius-verifier` and `binius-prover` shift tests, the `shift` prove/verify integration test, `binius-m4-prover`, `binius-recursion`, rustdoc, and clippy over the touched crates. The full workspace run is left to CI.

### DIFF
--- a/crates/prover/src/protocols/shift/shift_ind.rs
+++ b/crates/prover/src/protocols/shift/shift_ind.rs
@@ -1,0 +1,393 @@
+// Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
+
+//! The sumcheck binding the bit index the h polynomials sum over, and the shift indicator
+//! partial evaluations it runs over.
+
+use binius_compute::Allocator;
+use binius_core::word::Word;
+use binius_field::{BinaryField, FieldOps, PackedField};
+use binius_ip_prover::{
+	channel::IPProverChannel,
+	sumcheck::{bivariate_product_prover, prove_single},
+};
+use binius_math::{
+	BinarySubspace, FieldBuffer, FieldVec, inner_product::inner_product,
+	multilinear::eq::eq_ind_partial_eval, univariate::lagrange_evals,
+};
+
+/// The number of bit variables a half-word (`*32`) shift variant acts over.
+const HALF_WORD_LOG_BITS: usize = Word::LOG_BITS - 1;
+
+/// The shift reduction's last sumcheck, over the bit index the h polynomials sum over.
+///
+/// The scalar phase 2 weights every shift key by is that sum, at the phase-1 challenges:
+///
+/// $$
+/// h(r_j, r_s, r_v) = \sum_i \widetilde{L}(i) \cdot
+///     \sum_{\text{op}} \widetilde{eq}(r_v, \text{op}) \cdot \text{ind}_{\text{op}}(i, r_j, r_s)
+/// $$
+///
+/// Proving it as a sumcheck over the two multilinears in `i` leaves the verifier one evaluation
+/// of each at a random point, which it computes directly with
+/// [`evaluate_shift_inds`](binius_verifier::protocols::shift::evaluate_shift_inds) — no
+/// summation over the bit index, and no partial evaluation.
+pub struct ShiftIndSumcheck<P: PackedField, A: Allocator> {
+	/// The Lagrange evaluations at the univariate challenge, over the bit index.
+	l_tilde: FieldVec<P, A>,
+	/// The shift indicators interpolated over the shift variant, over the bit index.
+	shift_ind: FieldVec<P, A>,
+	/// The claimed sum of their product: the h evaluation.
+	h_eval: P::Scalar,
+}
+
+impl<F: BinaryField, P: PackedField<Scalar = F>, A: Allocator> ShiftIndSumcheck<P, A> {
+	/// Builds the two multilinears the sumcheck runs over, and their claimed sum.
+	///
+	/// # Arguments
+	///
+	/// - `domain_subspace`: the univariate evaluation domain.
+	/// - `r_zhat_prime`: the univariate challenge, shared by every operation.
+	/// - `r_j`, `r_s`, `r_v`: phase 1's challenges — the input bit position, the shift amount and
+	///   the shift variant.
+	pub fn new(
+		alloc: &A,
+		domain_subspace: &BinarySubspace<F>,
+		r_zhat_prime: F,
+		r_j: &[F],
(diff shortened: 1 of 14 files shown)


crates/prover/src/protocols/shift/shift_ind.rs
```diff
@@ -0,0 +1,93 @@
...
+	) -> Self {
+		let l_tilde = lagrange_evals(domain_subspace, r_zhat_prime);
+		let l_tilde = l_tilde.as_ref();
+
+		// The eight shift indicators are interpolated over the variant axis, which phase 1 folded
+		// into its own sumcheck, leaving one value per bit index.
+		let r_v_tensor = eq_ind_partial_eval::<F>(r_v);
+		let shift_ind = (0..Word::BITS)
+			.map(|index| {
+				let shift_inds = evaluate_shift_inds(&bit_index_vertex(index), r_j, r_s);
```
jimpo: Restore the code for partially evaluating the shift indicators, but move it to the prover crate (it used to be in the verifier crate).

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -301,54 +327,69 @@ where
...
+	// axis. Both are functions of public-channel-derived elements, so they are computed the same
+	// way as `monster_eval` below.
+	let h_reduced_eval = {
+		let inputs: Vec<C::Elem> = iter::once(r_zhat_prime)
+			.chain(r_i.iter().cloned())
+			.chain(r_j.iter().cloned())
+			.chain(r_s.iter().cloned())
+			.chain(r_v.iter().cloned())
+			.collect();
+		channel.compute_public_value(&inputs, ShiftIndEvalFn { subspace })
```
jimpo: This doesn't need to be compute_public_value. Just a regular function. compute_public_value is only for the MonsterEvalFn (at least for binius64 -- spartan uses it too in one place).

jimpo (changes requested):

crates/prover/src/protocols/shift/shift_ind.rs
benji-potamus: Done. `partial_eval_sigmas`, `partial_eval_sigmas_transpose` and `partial_eval_phi` are back verbatim, now private to `crates/prover/src/protocols/shift/shift_ind.rs`, and `build_shift_ind` assembles the eight indicators from them (plus the two 5-variable copies and the `same_half` tensor for the `*32` variants) instead of evaluating the closed form at each vertex. Their four `*_matches_reference` tests moved with them.

I restructured the stack so this reads as a move rather than a delete-and-restore: commit 1 is now purely additive on the verifier side — it adds `evaluate_shift_inds` next to the untouched recurrences — and this commit takes them across to the prover. Two tests tie the two definitions together: `build_matches_the_verifier_point_evaluation` checks the prover's multilinear against `evaluate_shift_inds` at every bit-index vertex, and `claimed_sum_is_the_weighted_indicator_sum` checks the claimed sum.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done — `ShiftIndEvalFn` is gone. `check_eval` now computes the two evaluations inline:

```rust
let l_tilde = lagrange_evals_scalars(subspace, r_zhat_prime);
let l_tilde_eval = inner_product_scalars(l_tilde, eq_ind_partial_eval_scalars(r_i));
let shift_ind_eval = inner_product_scalars(
    evaluate_shift_inds(r_i, r_j, r_s),
    eq_ind_partial_eval_scalars(r_v),
);
channel.assert_zero(h_sumcheck_eval.clone() - l_tilde_eval * shift_ind_eval)?;
```

`MonsterEvalFn` is left as the only `compute_public_value` caller here. One knock-on: `r_zhat_prime` is now only read by reference, so the parameter became `&C::Elem` (clippy's `needless_pass_by_value` otherwise), which touched the three call sites in `reduction.rs` and the two shift tests.

jimpo (approved):

PR #2114, opened 2026-08-12
BINIUS-466: ring-switch the public segment's evaluation

benji-potamus:
Closes [BINIUS-466](https://linear.app/jimpo/issue/BINIUS-466/ring-switching-for-the-public-segment), the first step of [BINIUS-465](https://linear.app/jimpo/issue/BINIUS-465/remove-wordipchannel-requirement-for-binius64-iopverifier).

> **Revised.** The first version summed the ring-switched claim over the hypercube and hid that behind `compute_public_value`. @jimpo asked for the sumcheck instead, which removes both that sum and the escape hatch. The description below is the current design.

The shift closes over the public segment as a **bit matrix**: its multilinear is claimed at `r_j` over the bit within a word and the low coordinates of `r_y` over the word index. `check_eval` evaluated that itself, from the verifier's public words — which costs the verifier the bits of every one of them.

Now the prover states that evaluation, and it is reduced in two steps.

## The reduction

```
prover                                    verifier
  public_eval ------------------------->  reads the claim
  s_hat_v (128 elems) ----------------->  ring_switch::verify: partial eval at r_j || r_y[0]
  <--------------------------- r''        row-batching query
                                          claim: sum_x P(x) A(x)
  log_public_elems sumcheck rounds <-->   verify_sumcheck(.., degree 2, ..)
                                          claim: P(r) * A(r), both computed by the verifier
```

1. The **ring-switch** moves the claim onto the segment's *packed* form, two words to a field element, leaving `sum_x P(x) A(x)` against the ring-switching indicator.
2. The **sumcheck** over the packed segment's own variables leaves one evaluation of each factor.

Nothing here is committed, so the verifier finishes on its own, with field arithmetic and **no word bits**:

```rust
let packed_eval = evaluate_inplace_scalars(packed_public, &r_public);
let rs_eq_eval = eval_rs_eq(&eval_point[log_packing..], &r_public, &eq_r_double_prime);
channel.assert_zero(packed_eval * rs_eq_eval - eval)?;
```

Entirely separate from the trace's ring-switch: its own partial evaluations, its own row-batching query, its own sumcheck, drawn between the shift and the trace's reduction.

## What changed

- **`shift::check_eval`** takes `public_eval` in place of `public: &[Word]`. It is the segment over the shift's *whole* word-index space, so the eq-zero padding above the segment's own length moves out to the caller with the rest of the segment's layout — `check_eval` no longer knows `log_public_words` at all, and `PublicWordsEvalFn` is gone.
- **`reduction::verify_public_eval`** reads the claim, ring-switches it, verifies the sumcheck, and closes against the two evaluations it computes itself.
- **`pack_words`** brings back [`encode_inout`](https://github.com/binius-zk/binius64/pull/2083/files#diff-29768aff84a68cb55d0379fea96a6827e0483e2530f5078ffb68efca40e04819L374), over the public segment rather than just the inout words: word `2i` low, word `2i + 1` high, matching how the trace is packed, so a bit of the packed segment sits where the shift's bit-index challenges address it.
- **`ring_switch::prove_public_eval`** on the prover side, which reuses `ring_switch::prove` and then runs the sumcheck on the shared `bivariate_product_prover`. The final multilinear evaluations need no message, since the verifier computes both.

### Whole field elements

The claim spans whole field elements, so a public segment shorter than one still spans one and reads the words past its end as zero. This is not hypothetical: an M4 system's public half is the shared constants alone, so `log_public_elems` is 0 there and this runs a zero-round sumcheck. That path is covered by the M4 round trips.

### No new channel method

The issue anticipated a `WordIPVerifierChannel` method for packing words into `Elem`s. It is not needed yet: the statement still arrives as concrete `Word`s, so packing natively and lifting through the `From<B128>` bound `reduce_constraints` already carries produces the same constants a wrapper channel would build, with no new trait surface. That method becomes necessary when the statement goes symbolic ([BINIUS-433](https://linear.app/jimpo/issue/BINIUS-433)). Happy to add it now if you'd rather have the seam in place.

`observe` is untouched — the inout words still enter Fiat-Shamir as words.

## Verification

Tests for the four affected crates, `cargo clippy --all --tests --benches --examples -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` are clean.

The end-to-end round trips cover the reduction: `prove_verify` (15 cases, including the ZK and the public-wider-than-hidden ones) and the M4 protocol round trips exercise both sides of the new messages.

The check is **non-vacuous**, verified by mutation rather than assumed. Two independent mutations each fail `test_prove_verify_sha256_preimage`:

- swapping the word order inside `pack_words` — so the packed values are really bound;
- dropping the sumcheck challenge reversal — so the sumcheck's variable order is really exercised, not incidentally satisfied.

The two shift-protocol tests drive the shift alone, with no reduction around it, so they evaluate the public segment directly where the full reduction reads it from the prover.

### DIFF
--- a/crates/verifier/src/reduction.rs
+++ b/crates/verifier/src/reduction.rs
@@ -17,28 +17,31 @@
 //! A batch of one instance still runs that sumcheck, over no rounds.
 //! It is the batched protocol either way, because its prover folds a batched witness.
 
-use std::array;
+use std::{array, iter};
 
 use binius_core::{
 	constraint_system::{ConstraintSystem, InoutSegment},
 	word::Word,
 };
-use binius_field::{AESTowerField8b as B8, FieldOps};
+use binius_field::{AESTowerField8b as B8, ExtensionField, FieldOps};
 use binius_iop::channel::IOPVerifierChannel;
 use binius_ip::{
 	channel::IPVerifierChannel,
-	sumcheck::{BatchSumcheckOutput, batch_verify},
+	sumcheck::{BatchSumcheckOutput, SumcheckOutput, batch_verify, verify as verify_sumcheck},
 };
 use binius_math::{
 	BinarySubspace,
 	inner_product::inner_product_scalars,
-	multilinear::eq::eq_ind,
+	multilinear::{
+		eq::{eq_ind, eq_ind_zero},
+		evaluate::evaluate_inplace_scalars,
+	},
 	univariate::{evaluate_univariate, lagrange_evals_scalars},
 };
 
 use crate::{
 	Error,
-	config::B128,
+	config::{B1, B128, LOG_WORDS_PER_ELEM},
 	protocols::{
 		binmul::{BinMulOutput, verify as verify_binmul_reduction},
 		bitand::AndCheckOutput,
@@ -46,6 +49,7 @@ use crate::{
 		shift::{self, BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, OperatorData},
 		zero,
 	},
+	ring_switch::{self, RingSwitchVerifyOutput, eval_rs_eq},
 	verify_bitand_reduction,
 };
 
@@ -278,7 +282,7 @@ where
 	};
 
 	// Tie in the public values through the public-input consistency check.
-	// The shift evaluates them over the layout's power-of-two word count.
+	// The reduction reads them over the layout's power-of-two word count.
 	// Their count need not be a power of two, so they are passed unpadded.
 	{
 		let _guard = tracing::info_span!(
@@ -287,10 +291,11 @@ where
 			perfetto_category = "phase"
 		)
 		.entered();
+		let public_eval = verify_public_eval(cs, inout, public, &shift, channel)?;
 		shift::check_eval::<B128, _>(
 			cs,
 			inout,
-			public,
+			public_eval,
 			&zero,
 			&bitand,
 			&intmul,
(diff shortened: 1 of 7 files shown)


crates/verifier/src/reduction.rs
```diff
@@ -305,6 +307,117 @@ where
...
+	// indicator's challenges and the packed values are public, so the sum is run in plaintext and
+	// materialized as a single wire, the same way the shift's monster evaluation is.
+	let log_packing = <B128 as ExtensionField<B1>>::LOG_DEGREE;
+	let packed_public = pack_words(public);
+	let inputs = eval_point[log_packing..]
+		.iter()
+		.chain(&eq_r_double_prime)
+		.cloned()
+		.collect::<Vec<_>>();
+	let claim = channel.compute_public_value(
```
jimpo: It doesn't need to do compute_public_value or FieldFn. Just evaluate it directly here.

crates/verifier/src/reduction.rs
```diff
@@ -305,6 +307,117 @@ where
...
+
+	// The claimed evaluation's point: the bit within a word, then the words the segment spans.
+	let r_y = shift.r_y();
+	let eval_point = [shift.r_j(), &r_y[..log_packed_words]].concat();
+
+	let public_eval = channel.recv_one()?;
+	let RingSwitchVerifyOutput {
+		eq_r_double_prime,
+		sumcheck_claim,
+	} = ring_switch::verify(public_eval.clone(), &eval_point, channel)?;
```
jimpo, line 366: Actually do the `log_public_elems`-round sumcheck protocol here so you can get rid of `eval_rs_claim` and use `eval_rs_eq` instead.

jimpo (changes requested):

crates/verifier/src/reduction.rs
benji-potamus: Done — and you were right that this is the shape it wants.

The sumcheck reduces `sum_x P(x) A(x)` to one evaluation of each factor, so the verifier now evaluates the packed segment's multilinear at the sumcheck's point and the indicator with `eval_rs_eq` at the same point. `eval_rs_claim` is gone, and `crates/verifier/src/ring_switch.rs` is untouched by this PR.

I had gone the other way because I compared the two on the verifier's *native* cost, where summing against the indicator over the hypercube is the same order as the bit-level evaluation it replaces. What I missed is that the magnitude was never the point: the hypercube sum needs the indicator at every point, which is a tensor-algebra transpose per term, and that is exactly what forced the plaintext escape hatch in your other comment. One evaluation instead of `2^log_public_elems` of them removes the reason for it.

The prover reuses `ring_switch::prove` and runs the sumcheck on the shared `bivariate_product_prover`, so no new sumcheck machinery. The final multilinear evaluations need no message, since the verifier computes both.

Worth flagging: an M4 system's public half is the shared constants alone, so `log_public_elems` is 0 there and this runs a zero-round sumcheck. That path works and is covered by the M4 round trips.

crates/verifier/src/reduction.rs
benji-potamus: Done. `compute_public_value` and the `FieldFn` are gone; the closing check is now two lines of ordinary `Elem` arithmetic:

```rust
let packed_eval = evaluate_inplace_scalars(packed_public, &r_public);
let rs_eq_eval = eval_rs_eq(&eval_point[log_packing..], &r_public, &eq_r_double_prime);
channel.assert_zero(packed_eval * rs_eq_eval - eval)?;
```

This falls out of the sumcheck in your other comment. What had made me reach for the escape hatch was evaluating the indicator over the *whole* hypercube — a tensor-algebra transpose per point, which a wrapper channel would have built as a sub-circuit. At a single point there is no such thing to hide: it is one `eval_rs_eq` and one multilinear evaluation.

jimpo (changes requested):
Do this in the M4 protocol as well (ring-switching instead of direct public evaluation).

benji-potamus:
> Do this in the M4 protocol as well (ring-switching instead of direct public evaluation).

M4 already gets it — both sides go through the shared reduction, so there was never a second public-evaluation path to convert:

- **verifier** — `m4-verifier/src/verify.rs:125` calls `reduce_constraints`, and the ring-switch lives inside it at `verifier/src/reduction.rs:294` (`verify_public_eval`). The M4 verifier has no public-evaluation code of its own.
- **prover** — `m4-prover/src/prove.rs:328` calls `ring_switch::prove_public_eval`, right after the shift and before the trace's ring-switch.

I checked this by mutation rather than by reading: deleting that one call from the M4 prover fails **all 9** `protocol_round_trips*` tests in `m4-prover`, because the M4 verifier is already reading the messages it stops sending. Restored, they pass.

I think what makes this hard to see in the diff is that the M4 files show only two hunks — the prover call site, and a test. The M4-specific parts that *do* differ are handled by the shared code:

- the public segment is `InoutSegment::Hidden`, so it is the shared constants alone rather than constants + inout;
- that segment is often shorter than one field element, which is why the claim is padded up to a whole element — for M4 `log_public_elems` is typically 0, so it runs a **zero-round** sumcheck. That path is only exercised by the M4 tests.

Happy to be told I'm reading the request wrong — if you meant something else by "direct public evaluation" in M4, point me at it. The two remaining `evaluate_words_mle` calls near M4 are `m4-prover/src/shift.rs:380` (a test that drives the shift with no reduction around it) and `prover/src/protocols/shift/phase_2.rs:300` (the prover deriving `witness_eval`, shared by both protocols and not a verifier-side evaluation).

jimpo (approved):

PR #2115, opened 2026-08-12
Present the shift reduction as one sumcheck, phased in the prover

benji-potamus:
Closes [BINIUS-467](https://linear.app/jimpo/issue/BINIUS-467/shift-present-the-shift-reduction-as-one-sumcheck-phased-in-the-prover). Implements [binius.xyz#123](https://github.com/binius-zk/binius.xyz/pull/123), following #2113.

The shift reduction becomes one sumcheck over all `SHIFT_LOG_VARS + log_word_count` variables of the claim, degree 2 in each, with the rounds binding the shift variant, then the shift amount, then the bit position within a word, then the bit index the shift indicators read, and last the word index. The verifier runs a single sumcheck verification and splits the challenge point into its five runs; the phase structure lives entirely in the prover.

### Commits

1. **Return the g evaluation from the phase-1 sumcheck** — the phase-1 sumcheck folds both multilinears to a scalar and keeps only their product. The bit-index phase needs `g(r_j, r_s, r_v)` on its own, so return it. `run_phase_1_sumcheck` now splits the challenge point itself and returns `Phase1Output`, which every caller was building from it anyway.
2. **Bind the shift indicator bit index before the word index** — the protocol change.

### What moved

The bit-index rounds were a sumcheck of their own, running after the word-index rounds against a claim the prover sent as advice. They now sit in front of the word-index rounds, and:

- **`h_eval` stops being a prover message.** Phase 3 carries phase 1's `g` evaluation through its rounds as a constant, folded into the Lagrange weights so the pair sums to the claim phase 2 reduced to. A `debug_assert` ties that sum back to `gamma`.
- **Phase 4 scales its monster multilinear** by the evaluation phase 3 reduces the two bit-index factors to — exactly where `h_eval` used to sit, so `build_monster_segments` is unchanged but for the name of its scalar.
- **`check_eval` checks one equation instead of two**: the closing claim against the product of the sumcheck's four factors.

### Departure from the writeup

The writeup has phase 3 recover its carried constant by dividing the incoming claim by `h(r_j, r_s, r_v)`, with a fallback for when that is zero and a corresponding completeness caveat. That division is avoidable here: the phase-1 sumcheck already folds `g` down to a scalar and `prove_single` hands it back, so the constant is read off directly. No inversion, no exceptional case. jimpo confirmed this and is updating the writeup.

### Transcript

One field element shorter (no `h_eval`), same number of sumcheck rounds, same challenges drawn. Three `verify_sumcheck` calls collapse to one.

### Testing

`binius-verifier` and `binius-prover` shift tests, the `shift` prove/verify integration test, `binius-m4-prover`, `binius-recursion`, rustdoc, and clippy over the touched crates. The full workspace run is left to CI.

### DIFF
--- a/crates/prover/src/protocols/shift/prove.rs
+++ b/crates/prover/src/protocols/shift/prove.rs
@@ -126,10 +126,12 @@ impl<F: Field> PreparedOperatorData<F> {
 ///
 /// The result is a single multilinear evaluation claim on the witness.
 ///
-/// Two sumcheck phases run in sequence:
+/// One sumcheck runs, in four prover phases:
 ///
-/// 1. phase 1 proves the batched claims over the shift variants and operand positions;
-/// 2. phase 2 reduces those to a witness evaluation, against the monster multilinear.
+/// 1. phases 1 and 2 prove the batched claims over the shift variants, the shift amounts and the
+///    bit positions, sparsely and then densely;
+/// 2. phase 3 binds the bit index the shift indicators read;
+/// 3. phase 4 reduces what is left to a witness evaluation, against the monster multilinear.
 ///
 /// # Parameters
 /// - `key_collection`: the prover's key collection for the constraint system.
@@ -173,7 +175,8 @@ where
 		claims.prepare(|| channel.sample())
 	};
 
-	// Phase 1 outputs challenges `r_j || r_s || r_v`, and `gamma` as its evaluation (see paper).
+	// Phases 1 and 2 bind the shift variant, the shift amount and the bit position, outputting
+	// challenges `r_j || r_s || r_v` and the claim `gamma` (see paper).
 	let phase_1_output = prove_phase_1::<_, P, _, _>(
 		key_collection,
 		words,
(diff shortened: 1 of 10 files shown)


crates/prover/src/protocols/shift/prove.rs
```diff
@@ -183,33 +186,30 @@ where
...
+		phase_1_output.g_eval,
 	);
+	debug_assert_eq!(phase_3.beta(), phase_1_output.gamma);
+	let phase_3_output = phase_3.prove(channel, alloc);
 
-	// Phase 2 outputs challenges `r_y`, and the witness evaluation at the oblong point given by
+	// Phase 4 outputs challenges `r_y`, and the witness evaluation at the oblong point given by
 	// the univariate variable `r_j` and the multilinear variable `r_y`.
-	let output = prove_phase_2::<_, P, _, _>(
+	prove_phase_2::<_, P, _, _>(
```
jimpo, line 206: rename to prove_phase_4

jimpo (approved):

PR #2116, opened 2026-08-12
BINIUS-465: hand the statement to the Binius64 IOP as the channel's own words

benji-potamus:
Closes [BINIUS-465](https://linear.app/jimpo/issue/BINIUS-465/remove-wordipchannel-requirement-for-binius64-iopverifier), following [#2114](https://github.com/binius-zk/binius64/pull/2114) (BINIUS-466).

> **Revised.** The first version hoisted the observation to the callers so the `WordIPVerifierChannel` bound could come off `IOPVerifier::verify`. @jimpo asked for something better: the observation should *return* the channel's own words, and the IOP should verify against those. The description below is what landed.

```rust
fn observe_words(&mut self, words: &[Word]) -> Vec<Self::Word>;
```

The observation is the one place the protocol handles a 64-bit word as fixed data it is handed, so it is also the natural place for a channel to introduce its own representation. A channel over concrete values hands the same words back; one that builds a circuit allocates the wires there. `IOPVerifier::verify` takes what comes back, so everything downstream reads the statement **symbolically**.

That is what the recursive verifier needs. Until now `verify` lifted the statement through `From<Word>`, so `Binius64BuilderChannel` saw it already concrete and baked it in as constants — the recorded circuit verified one fixed instance rather than a statement.

The transcript is unchanged: the same words go into Fiat-Shamir at the same point.

## What changed

- **`WordIPVerifierChannel::observe_words`** takes concrete `&[Word]` and returns `Vec<Self::Word>`. The prover's is untouched — it has no symbolic words, so it returns nothing.
- **`WordIPVerifierChannel::pack_words`** is new. The public segment now reaches the reduction as `Self::Word`, so the packing #2114 added as a free function over concrete words cannot stay one. `pack_words_concrete` serves the channels carrying `Word`; the builder channel pairs wires up, which costs **no gates**, since a `SymbolicElem` is exactly the two halves of a 128-bit element. This is the method BINIUS-466's description anticipated.
- **`IOPVerifier::verify` and `reduce_constraints`** take `&[Channel::Word]`. The constraint system's constants lift into the channel's word type as themselves.
- **Callers observe before delegating**: the transparent verifier, the ZK verifier, the ZK prover's replay closure, and `recursion_end_to_end`.
- **`WordIPVerifierChannel<F>` now requires `F: BinaryField`** rather than `Field`, since packing words into an element needs its bit basis. Every instantiation was already `B128`.

## What this actually buys, and how it is checked

The round trip is green whether the statement is wires or constants — the filler run supplies the same values either way — so no existing test would have shown whether this did anything. `recursion_end_to_end` now asserts the property directly:

```rust
let statement_wires = recorded.inputs.iter()
    .filter(|input| input.kind == "observe_words")
    .count();
assert_eq!(statement_wires, witness.inout().len());
```

## Verification

Tests for the nine affected crates (27 suites: `binius-ip`, `binius-iop`, `binius-verifier`, `binius-prover`, `binius-recursion`, `binius-m4-prover`, `binius-m4-verifier`, `binius-spartan-prover`, `binius-spartan-verifier`), `cargo clippy --all --tests --benches --examples -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` are clean.

Every caller must observe, and missing one fails quietly — Fiat-Shamir diverges and surfaces much later as a bad assert. That is not hypothetical: an intermediate revision left the ZK verifier out and six `prove_verify` cases failed with `OuterVerification(InvalidAssert)`. The ZK and recursion paths both have end-to-end coverage, which is what caught it.

### DIFF
--- a/crates/verifier/src/verify.rs
+++ b/crates/verifier/src/verify.rs
@@ -114,7 +114,10 @@ impl IOPVerifier {
 	/// `verify`, rather than duplicating it here.
 	pub fn oracle_specs(&self, is_zk: bool) -> Vec<OracleSpec> {
 		let mut channel = OracleSetupChannel::new(is_zk);
-		let inout = vec![Word::ZERO; self.constraint_system.n_inout];
+		let inout = WordIPVerifierChannel::<B128>::observe_words(
+			&mut channel,
+			&vec![Word::ZERO; self.constraint_system.n_inout],
+		);
 		// The result is discarded: the setup channel performs no real verification (all `recv_*`
 		// return zero, `assert_zero` is a no-op), so we only read back the recorded oracle specs.
 		let _ = self.verify(&inout, &mut channel);
@@ -125,7 +128,16 @@ impl IOPVerifier {
 	///
 	/// This is the core verification logic, independent of the specific IOP compilation strategy.
 	/// For most users, [`Verifier::verify`] is the simpler interface.
-	pub fn verify<Channel>(&self, inout: &[Word], channel: &mut Channel) -> Result<(), Error>
+	///
+	/// `inout` is the statement as the channel carries it, which is what
+	/// [`WordIPVerifierChannel::observe_words`] returns: the caller observes the concrete words and
+	/// passes on what comes back. A channel that carries words as wires introduces them there, so
+	/// the verification below reads the statement symbolically rather than as fixed data.
+	pub fn verify<Channel>(
+		&self,
+		inout: &[Channel::Word],
+		channel: &mut Channel,
+	) -> Result<(), Error>
 	where
 		Channel: IOPVerifierChannel<B128> + WordIPVerifierChannel<B128>,
 		Channel::Elem: FieldOps<Scalar = B128> + From<B128>,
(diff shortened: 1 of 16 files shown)


jimpo (changes requested):
Put the observe_words back in the IOPVerifier/IOPProver. Too much unnecessary churn.

crates/verifier/src/verify.rs
```diff
@@ -114,7 +114,10 @@ impl IOPVerifier {
 	/// `verify`, rather than duplicating it here.
 	pub fn oracle_specs(&self, is_zk: bool) -> Vec<OracleSpec> {
 		let mut channel = OracleSetupChannel::new(is_zk);
-		let inout = vec![Word::ZERO; self.constraint_system.n_inout];
+		let inout = WordIPVerifierChannel::<B128>::observe_words(
```
jimpo, line 117: nit: no need to actually observe, just pass the words directly (because OracleSetupChannel uses concrete words)

jimpo (approved):

crates/verifier/src/verify.rs
benji-potamus: Agreed — `OracleSetupChannel::Word = Word`, so the observe round-trip buys nothing and the turbofished `WordIPVerifierChannel::<B128>::observe_words(&mut channel, ..)` (needed only because the element type is otherwise unconstrained there) goes away with it:

```rust
let inout = vec![Word::ZERO; self.constraint_system.n_inout];
```

Your approval enqueued the PR, though, and a queued branch can't be updated — so rather than dequeue and re-run everything for a nit, I'll land it as a small follow-up right after this merges. Shout if you'd rather I pull it out of the queue and fold it in here.

PR #2119, opened 2026-08-12
BINIUS-420: Allow verifying multiple linear relations on oracles

benji-potamus:
Closes [BINIUS-420](https://linear.app/jimpo/issue/BINIUS-420).

Lets one committed oracle be opened against several transparent multilinears in the same batched
BaseFold opening, and reshapes the channel interface as proposed in
https://github.com/binius-zk/binius64/pull/2061#issuecomment-5253986872.

Two commits, each independently green.

### 1. Take one oracle relation at a time

`prove_oracle_relations` bundled two responsibilities: stating claims about committed oracles, and
handing the channel ownership of each oracle's message buffer. Passing several relations at once is
a carryover from an earlier interface where the call consumed the channel.

```diff
-fn prove_oracle_relations(
-    &mut self,
-    oracle_relations: impl IntoIterator<Item = (Self::Oracle, FieldVec<P, A>, FieldVec<P, A>, P::Scalar)>,
-);
+fn prove_oracle_relation(&mut self, oracle: Self::Oracle, transparent: FieldVec<P, A>, claim: P::Scalar);
+
+/// Gives ownership of the oracle buffer to the channel.
+fn finalize_oracle(&mut self, oracle: Self::Oracle, buffer: FieldVec<P, A>);
```

The verifier side follows the same shape — `verify_oracle_relation(oracle, transparent, claim)`.
With the method singular, its parameters name what the `OracleLinearRelation` struct existed to
name, so that struct is gone; the BaseFold channels keep a private queue struct of their own.

The BaseFold prover channel now stores each message in its `CommittedOracleData`, which also makes
the ZK blinding `π_i' = (1-γ)π_i + γω_i` a per-oracle step rather than a per-relation one.

No protocol change in this commit: the transcript is byte-identical.

### 2. Allow several relations on one oracle

The opening asserted exactly one relation per committed oracle, with a `TODO` to lift it. That
blocks opening one commitment at several points, which the chip architecture needs directly: one
trace commitment holds the main circuit's words and every chip's block, and the main reduction and
each chip reduction each end with their own evaluation claim on it.

Both channels queue relations in a structure indexed by oracle, and batch each oracle's claims into
one before the opening runs, with a single batching challenge `λ` shared across oracles:

```
T_i = Σ_j λ^j · t_ij     the combined transparent
S_i = Σ_j λ^j · s_ij     the combined claim
```

The inner product is linear in the transparent, so `⟨π_i, T_i⟩ = S_i` holds exactly when every
`⟨π_i, t_ij⟩ = s_ij` does. The batched per-oracle claims are then combined again by the sumcheck's
own outer batching coefficient. Everything downstream — masking, sumcheck, FRI — sees one relation
per oracle, exactly as it did before.

**Soundness.** For oracle `i`, `P_i(X) = Σ_j (⟨π_i, t_ij⟩ - s_ij) X^j` is a nonzero polynomial of
degree `< k_i` whenever one of that oracle's claims is wrong, so a shared `λ` fails with probability
at most `Σ_i (k_i - 1) / |F|`. `λ` is drawn after every claim it combines is already bound to the
transcript, so no claim can be chosen as a function of it. It is sampled only when some oracle
carries more than one claim, matching how `γ` is drawn only when a ZK oracle is present.

**What the oracle-indexed queue buys.** The reconciliation between arrival order and oracle-index
order disappears: the masking values `σ`, the sumcheck provers and the reduced evaluations `α` all
run in oracle-index order, so the scatter of the `α`s is gone. It also settles an accounting
mismatch the assert was hiding — the prover sent one `σ` per ZK *relation* while the verifier read
one per ZK *oracle*, which agreed only while the assert held.

**Proofs move.** Relations are now ordered by oracle index rather than by arrival, and today's
callers do differ between the two: `prove.rs` commits the trace oracle first but opens it last,
after the logUp pushforwards. Nothing pins proof bytes — the committed example snapshots are circuit
statistics — and both sides derive the order the same way, but this is no longer a byte-identical
transcript.

### Not covered

The other half of the old `TODO` remains: an oracle with *no* relation still fails, because its `α`
would have to come from a plain multilinear evaluation rather than from a sumcheck prover. Both
channels now assert that requirement directly ("at least one relation per committed oracle").

### Verification

Locally, on the rebased stack:

- `cargo clippy --all --tests --benches --examples -- -D warnings` — clean
- nightly `fmt --all` — clean
- 206 tests across `binius-iop`, `binius-iop-prover`, `binius-prover`, `binius-verifier`,
  `binius-m4-prover`, `binius-m4-verifier`, `binius-spartan-prover`, `binius-spartan-verifier` —
  all pass, and commit 1 was checked out on its own for the same set before the rework
- the full `--workspace` run is left to CI

Four new channel tests cover the new capability: two claims on one ZK oracle; a mixed batch of a
non-ZK oracle with one claim and a ZK oracle with three, queued round-robin so relations arrive
interleaved across oracles; and tampered claims in a later position and on the non-ZK oracle.

### DIFF
--- a/crates/iop-prover/src/basefold/channel.rs
+++ b/crates/iop-prover/src/basefold/channel.rs
@@ -60,17 +60,11 @@ struct CommittedOracleData<P: PackedField, C, Data: Deref<Target = [P]>> {
 }
 
 /// A committed-oracle relation queued for the single batched opening.
-///
-/// Each entry records the transparent multilinear a committed message is opened against, and which
-/// committed oracle that message belongs to. That index reconciles the arrival-order and
-/// oracle-index views of the batch.
 struct QueuedRelation<P: PackedField, Data: Deref<Target = [P]>> {
-	/// Index of the committed oracle this relation opens.
-	oracle_index: usize,
-	/// The transparent multilinear `t_i` paired with the message, backed by the caller's
+	/// The transparent multilinear `t` the message is opened against, backed by the caller's
 	/// allocator.
 	transparent: FieldBuffer<P, Data>,
-	/// The claimed inner product `s_i = <pi_i, t_i>`.
+	/// The claimed inner product `s = <pi, t>`.
 	claim: P::Scalar,
 }
 
@@ -107,10 +101,10 @@ where
 	/// The combined FRI parameters over all committed oracles.
 	fri_params: FRIParams<F>,
 	committed_oracles: Vec<CommittedOracleData<P, Channel::Commitment, A::Vec<P>>>,
-	/// Oracle relations queued by [`IOPProverChannel::prove_oracle_relation`], opened together in
-	/// [`Self::finish`].
-	queue: Vec<QueuedRelation<P, A::Vec<P>>>,
-	next_oracle_index: usize,
+	/// Oracle relations queued by [`IOPProverChannel::prove_oracle_relation`], indexed by oracle
+	/// index and opened together in [`Self::finish`]. One entry per committed oracle, so its
+	/// length is also the number of oracles committed so far.
+	queue: Vec<Vec<QueuedRelation<P, A::Vec<P>>>>,
 	rng: StdRng,
 }
 
@@ -141,7 +135,6 @@ where
 			fri_params,
 			committed_oracles: Vec::new(),
 			queue: Vec::new(),
-			next_oracle_index: 0,
 			rng: StdRng::from_rng(&mut rng),
 		}
 	}
@@ -163,14 +156,13 @@ where
 			fri_params,
 			committed_oracles,
 			queue,
-			next_oracle_index,
 			rng: _,
 		} = self;
 
-		let n_remaining = oracle_specs.len() - next_oracle_index;
+		let n_remaining = oracle_specs.len() - queue.len();
 		assert!(n_remaining == 0, "finish called but {n_remaining} oracle specs remaining",);
 
-		if queue.is_empty() {
+		if queue.iter().all(Vec::is_empty) {
 			return;
 		}
 
@@ -194,19 +186,16 @@ where
 /// point `r`, then opens all committed oracles together with a single combined FRI. Mirrors
 /// [`binius_iop::basefold::channel::BaseFoldVerifierChannel::finish`].
 ///
-/// The masking inner products and the batched sumcheck process the `relations` in arrival order (so
-/// each reduced eval lines up with its batched-claim coefficient), while the per-oracle evaluations
-/// α_i and the FRI openings are emitted in oracle-index order. Each relation carries its oracle's
-/// index, so the two orders are reconciled by indexing rather than by sorting the relations; the
-/// per-oracle data (`oracle_specs`, `fri_params`, `committed_oracles`) is all indexed by oracle
-/// index.
+/// Everything runs in oracle-index order: `relations` arrives indexed by oracle, as do the
+/// per-oracle data (`oracle_specs`, `fri_params`, `committed_oracles`), the masking inner products
+/// σ_i, the sumcheck provers, the reduced evaluations α_i, and the FRI openings.
 fn prove_batch_zk_basefold<A, F, P, NTT, Channel>(
 	channel: &mut Channel,
 	ntt: &NTT,
 	oracle_specs: &[OracleSpec],
 	fri_params: &FRIParams<F>,
 	mut committed_oracles: Vec<CommittedOracleData<P, Channel::Commitment, A::Vec<P>>>,
-	relations: Vec<QueuedRelation<P, A::Vec<P>>>,
+	relations: Vec<Vec<QueuedRelation<P, A::Vec<P>>>>,
 	alloc: &A,
 ) where
 	A: Allocator,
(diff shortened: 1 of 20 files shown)


crates/iop-prover/src/basefold/channel.rs
```diff
@@ -105,8 +106,8 @@ where
 	oracle_specs: Vec<OracleSpec>,
 	/// The combined FRI parameters over all committed oracles.
 	fri_params: FRIParams<F>,
-	committed_oracles: Vec<CommittedOracleData<P, Channel::Commitment>>,
-	/// Oracle relations queued by [`Self::prove_oracle_relations`], opened together in
+	committed_oracles: Vec<CommittedOracleData<P, Channel::Commitment, A::Vec<P>>>,
+	/// Oracle relations queued by [`IOPProverChannel::prove_oracle_relation`], opened together in
 	/// [`Self::finish`].
 	queue: Vec<QueuedRelation<P, A::Vec<P>>>,
```
jimpo: Maybe better to just store the relations in a Vec<Vec<>> outer indexed by oracle ID to group them.

crates/iop-prover/src/basefold/channel.rs
```diff
@@ -444,6 +434,74 @@ fn prove_batch_zk_basefold<A, F, P, NTT, Channel>(
...
+///
+/// The inner product is linear in the transparent, so `⟨π_i, T_i⟩ = S_i` holds exactly when every
+/// `⟨π_i, t_j⟩ = s_j` does, except with probability at most `(k - 1) / |F|` over λ_i. Each λ_i is
+/// drawn after every claim it combines is already bound to the transcript, so no claim can be
+/// chosen as a function of it.
+///
+/// Oracles are returned in first-arrival order, and an oracle carrying a single claim draws no
+/// challenge at all, so a batch with one relation per oracle leaves the transcript untouched.
+/// Mirrors [`binius_iop::basefold::channel`]'s verifier-side batching.
+fn batch_relations_per_oracle<A, F, P, Channel>(
```
jimpo, line 421: We shouldn't need this, the channels should just queue the linear relations in a structured indexed by oracle ID

crates/iop-prover/src/basefold/channel.rs
```diff
@@ -444,6 +434,74 @@ fn prove_batch_zk_basefold<A, F, P, NTT, Channel>(
...
+		.into_iter()
+		.map(|group| {
+			let mut group = group.into_iter();
+			let mut batched = group.next().expect("groups are created non-empty");
+			let rest = group.collect::<Vec<_>>();
+			if rest.is_empty() {
+				return batched;
+			}
+
+			let lambda = channel.sample();
```
jimpo: Just sample one lambda to use to accumulate all groups. And then the batch_prove will sample a second, outer batching coeff.

jimpo (changes requested):

crates/iop-prover/src/basefold/channel.rs
benji-potamus: Done — the queue is `Vec<Vec<QueuedRelation<..>>>` indexed by oracle index now, on both channels. `send_oracle`/`recv_oracle` push an empty inner vec and `prove_oracle_relation`/`verify_oracle_relation` push into `queue[oracle.index]`, so `QueuedRelation` no longer carries an `oracle_index` at all.

crates/iop-prover/src/basefold/channel.rs
benji-potamus: Gone. With the queue indexed by oracle, the grouping pass and the whole arrival-order/oracle-index reconciliation go with it: `prover_oracle_indices` is deleted and the alphas are just `multilinear_evals` in order.

One consequence worth naming: relations are now processed in oracle-index order rather than arrival order, and today's callers do differ between the two — `prove.rs` commits the trace oracle first but opens it last, after the logUp pushforwards. So proofs move. Nothing pins proof bytes (the committed example snapshots are circuit statistics), and prover and verifier derive the order the same way, but the PR no longer produces byte-identical transcripts.

crates/iop-prover/src/basefold/channel.rs
benji-potamus: Switched to a single λ, sampled once and used to accumulate every group; `batch_prove` then draws its own outer coefficient over the per-oracle claims as you say.

Soundness holds with the shared challenge: for oracle `i`, `P_i(X) = Σ_j (⟨π_i, t_ij⟩ - s_ij) X^j` is a nonzero polynomial of degree `< k_i` whenever one of its claims is wrong, so the bad λ set is a union of at most `Σ_i (k_i - 1)` roots.

I kept it sampled only when some oracle carries more than one claim, matching how γ is drawn only when a ZK oracle is present.

crates/iop/src/basefold/channel.rs
```diff
@@ -245,6 +258,70 @@ where
...
+/// `binius_iop_prover::basefold::channel`.
+fn batch_relations_per_oracle<F, Channel>(
+	channel: &mut Channel,
+	relations: Vec<Vec<QueuedRelation<Channel::Elem>>>,
+) -> Vec<QueuedRelation<Channel::Elem>>
+where
+	F: BinaryField,
+	Channel: MerkleIPVerifierChannel<F, Elem: From<F> + 'static>,
+{
+	let lambda = relations
```
jimpo: Just sample lambda unconditionally here.

crates/iop/src/basefold/channel.rs
```diff
@@ -361,26 +438,27 @@ where
 			.recv_merkle_commitment(1 << fri_oracle.log_batch_size(), depth)?;
 
 		self.oracle_commitments.push(commitment);
+		self.queue.push(Vec::new());
 		self.next_oracle_index += 1;
```
jimpo: isn't next_oracle_index now redundant with self.queue.len()?

jimpo (changes requested):

crates/iop/src/basefold/channel.rs
benji-potamus: Done, both sides. Dropping the `Option` simplifies the prover loop too — no `rest` collection or early return, an oracle with one relation just folds nothing.

crates/iop/src/basefold/channel.rs
benji-potamus: It is — removed from both channels, replaced by `self.queue.len()` (`recv_oracle`/`send_oracle` push one inner vec per oracle, so it tracks the count exactly).

jimpo (approved):

PR #2121, opened 2026-08-12
Place a lift block narrower than a packed element

benji-potamus:
Closes [BINIUS-473](https://linear.app/jimpo/issue/BINIUS-473/basefold-repeat-placement-panics-when-a-lift-block-is-narrower-than).

The batched-witness placement in the BaseFold prover channel chunks the combined buffer by the lift block, and asserted that a block spans at least one whole packed element — the subtraction computing the chunk size needs it. A small committed oracle violates that and the prover panics.

```rust
assert!(n_i + log_lift >= P::LOG_WIDTH, "repeat placement requires whole-packed lift blocks");
let chunk_packed = 1usize << (n_i + log_lift - P::LOG_WIDTH);
```

### The fix

Below one packed element, several lift blocks share an element and no chunking can express the placement. The scalars of one element are the block pattern repeated, so the pattern is built once and added to every element; at or above one element the chunked path is untouched.

The placement moves out of the loop into `place_repeated`, which states what it computes:

> Adds `scalar · src` into the low `2^src.log_len()` scalars of every `2^log_block`-sized block of `dst`.

### Why it is a function with its own test

The first version of this PR clamped the chunk size instead, on the reasoning that lifting fills every oracle to `max_n`, so `log_repeat` is 0 and there is only ever one block. **That reasoning was wrong**, as jimpo suspected in review. With a large flexible oracle the optimizer picks a reduced dimension below `max_n`:

```
[(0, false), (8, false)]               n_i=0 log_lift=5 sum=5  max_n=8   → log_repeat = 3
[(0, false), (6, false), (12, false)]  n_i=0 log_lift=9 sum=9  max_n=12  → log_repeat = 3
```

`n_i + log_lift = max(n_i, reduced_log_dim)`, and `reduced_log_dim < max_n` whenever the largest oracle takes a flexible batch size. A clamp with `log_repeat > 0` and a sub-packed block writes the oracle into the low lanes of every *packed element* rather than every *lift block*, skipping the blocks that share an element — a wrong proof rather than a panic.

I could not make the FRI optimizer produce both conditions at once (it would need `reduced_log_dim ≤ 1` with `max_n ≥ 2`), but the placement should not depend on that. `place_repeated_matches_the_naive_placement` checks it against the definition over every `(log_src, log_block, log_dst)` shape up to 4 variables across `1x/2x/4x128b`, so the repeating sub-packed shapes are covered regardless. With the clamp-only version restored it fails at `LOG_WIDTH=1, log_src=0, log_block=0, log_dst=1`.

### Channel-level test

`run_zk_channel` and `run_mixed_channel` become generic over the packed field, and `batch_narrower_than_a_packed_element_proves` runs `[0, 1]` and `[1, 2]` under `PackedBinaryGhash4x128b`. Pinning the width is the point: the condition depends only on `P::LOG_WIDTH`, which `-Ctarget-cpu=native` derives from the host, so the existing 128-bit tests can never reach it and the CLI-artifacts failure only appears on AVX-512 runners.

`[0, 1]` reaches the case *and* lifts its first oracle (`log_lift = 1`); `[1, 2]` sits just the other side of the boundary — its blocks are exactly one packed element — and covers the chunked path the fix must leave alone. Both assert a full prove **and verify** cycle, so a mis-placed oracle fails rather than merely not panicking.

Without the fix, on any host:

```
panicked at crates/iop-prover/src/basefold/channel.rs:
repeat placement requires whole-packed lift blocks
```

### Left open

1. **Whether sub-packed batches should exist at all.** Direction (2) in the issue — padding small oracles up to one packed element when the FRI parameters are built — is still a live design question. This does not foreclose it.
2. **`-Ctarget-cpu=native` in CI.** It is what makes the `rust-checks-x86_64` leg's result depend on which runner it lands on. A CI-wide change with its own trade-offs, so it wants your call.

### DIFF
--- a/crates/iop-prover/src/basefold/channel.rs
+++ b/crates/iop-prover/src/basefold/channel.rs
@@ -5,7 +5,7 @@
 use std::ops::Deref;
 
 use binius_compute::Allocator;
-use binius_field::{BinaryField, PackedField};
+use binius_field::{BinaryField, Field, PackedField};
 use binius_iop::{channel::OracleSpec, fri::FRIParams};
 use binius_ip_prover::{
 	channel::{IPProverChannel, WordIPProverChannel},
@@ -358,22 +358,9 @@ fn prove_batch_zk_basefold<A, F, P, NTT, Channel>(
 
 			// Repeat placement: add scalar · π_i' into the first 2^{n_i} entries of each of the
 			// 2^{log_repeat} chunks of size 2^{n_i + log_lift}.
-			assert!(
-				n_i + log_lift >= P::LOG_WIDTH,
-				"repeat placement requires whole-packed lift blocks",
-			);
-			let scalar_broadcast = P::broadcast(eq_i);
-			let chunk_packed = 1usize << (n_i + log_lift - P::LOG_WIDTH);
 			// Borrow as a slice before the closure: the allocator's buffer type is only `Send`, so
 			// a closure capturing the owned buffer would not be `Sync` as `for_each` requires.
-			let witness_prime = witness_prime.to_ref();
-			combined
-				.as_mut()
-				.par_chunks_mut(chunk_packed)
-				.for_each(|chunk| {
-					let chunk_buf = FieldSliceMut::from_slice(n_i + log_lift, chunk);
-					accumulate_scaled_buffer(chunk_buf, witness_prime.to_ref(), scalar_broadcast);
-				});
+			place_repeated(combined.to_mut(), witness_prime.to_ref(), eq_i, n_i + log_lift);
 
 			// Repeat dims contribute 1; only the lift dims contribute an eq-to-zero factor.
 			s_prime += eq_i * alpha_i * eq_ind_zero(&point[n_i..][..log_lift]);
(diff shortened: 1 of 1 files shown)


crates/iop-prover/src/basefold/channel.rs
```diff
@@ -1009,7 +1013,34 @@ mod tests {
...
+	/// block is lifted to the combined dimension, so a block narrower than a packed element means
+	/// the whole buffer is — there is no whole-packed chunk to place into, and the placement has
+	/// to write into part of a single element instead.
+	///
+	/// Whether that happens is a function of the packed width alone, so this pins the width rather
+	/// than leaving it to `-Ctarget-cpu=native` and the host: with `PackedBinaryGhash4x128b` a
+	/// one-variable batch reaches the case on every machine, while the 128-bit type the other
+	/// tests use never does. The two-variable batch sits just the other side of the boundary.
+	#[test]
+	fn batch_narrower_than_a_packed_element_proves() {
```
jimpo, line 1128: I don't think this test exercises the right case. It should just `run_zk_channel::<PackedBinaryGhash4x128b>(&[1, 2], false)`.

It needs to have multiple oracles, so that the first one actually gets lifted.

jimpo (changes requested):

crates/iop-prover/src/basefold/channel.rs
benji-potamus: You're right that the single-oracle batch never lifts, and I've fixed that — but `[1, 2]` doesn't reach the case either. Instrumenting the placement under `PackedBinaryGhash4x128b` (`trips` = the old assert's condition, `n_i + log_lift < LOG_WIDTH`):

```
--- sizes [1, 2]
PLACEMENT n_i=1 log_lift=1 LOG_WIDTH=2 max_n=2 trips=false
PLACEMENT n_i=2 log_lift=0 LOG_WIDTH=2 max_n=2 trips=false
--- sizes [0, 1]
PLACEMENT n_i=0 log_lift=1 LOG_WIDTH=2 max_n=1 trips=true
PLACEMENT n_i=1 log_lift=0 LOG_WIDTH=2 max_n=1 trips=true
```

Lifting fills each oracle up to `max_n`, so `n_i + log_lift == max_n` for every oracle in the batch. `[1, 2]` therefore gives blocks of exactly one packed element and passes the old assert; with the fix reverted it proves fine, while `[0, 1]` panics.

So the batch has to be *smaller* than the boundary, not just heterogeneous. `[0, 1]` is the shape that satisfies both halves of your point: two oracles, the first genuinely lifted (`log_lift = 1`), and a sub-packed block. The test now runs that, and keeps `[1, 2]` alongside it as the just-above-the-boundary case, so the chunked path the clamp must leave alone is covered too.

Rebased onto #2119 while I was here; `binius-iop-prover` is green (43 tests) and clippy is clean.

benji-potamus: You were right, and the hole was bigger than the test. Chasing it found a real bug in my fix.

**`log_repeat > 0` is reachable** — my "it's always 0" claim came from fixtures that were too small. With a large flexible oracle the optimizer picks a reduced dimension below `max_n`:

```
[(0, false), (8, false)]               n_i=0 log_lift=5 sum=5  max_n=8   → log_repeat = 3
[(0, false), (6, false), (12, false)]  n_i=0 log_lift=9 sum=9  max_n=12  → log_repeat = 3
```

`n_i + log_lift = max(n_i, reduced_log_dim)`, and `reduced_log_dim < max_n` whenever the largest oracle takes a flexible batch size.

**So my clamp was wrong.** It is only correct when there is one block. With `log_repeat > 0` *and* a sub-packed block, `chunk_packed` clamps to 1 and the placement writes the oracle into the low lanes of every *packed element* instead of the low part of every *lift block* — the blocks sharing an element past the first are silently skipped. That is a wrong proof rather than a panic, which is worse than what I was fixing.

The two conditions are independent, and I could not get the FRI optimizer to produce both at once (it would need `reduced_log_dim ≤ 1` with `max_n ≥ 2`). But the placement should not depend on the optimizer avoiding a shape, so I fixed it properly rather than arguing about reachability:

- The placement is now `place_repeated`, which states what it computes. Below one packed element it builds the repeating lane pattern once and adds it to every element; at or above, the chunked path is untouched.
- `place_repeated_matches_the_naive_placement` checks it against the definition over every `(log_src, log_block, log_dst)` shape up to 4 variables, across `1x/2x/4x128b` — so it covers sub-packed blocks that repeat, whether or not the optimizer can currently ask for them. With the clamp-only version restored, that test fails at `LOG_WIDTH=1, log_src=0, log_block=0, log_dst=1`.

The channel-level test is `[0, 1]` and `[1, 2]` as described in my earlier reply: `[1, 2]` gives blocks of exactly one packed element and never tripped the old assert, while `[0, 1]` does and also lifts its first oracle.

`binius-iop-prover` is green (44 tests), clippy and rustdoc clean, rebased on #2119.

jimpo (approved):

PR #2129, opened 2026-08-12
Verify a composite M4 system on one BaseFold channel

benji-potamus:
[BINIUS-478](https://linear.app/jimpo/issue/BINIUS-478/m4-composite-setup-build-one-basefold-compiler-for-a) — second of three stacked PRs for [BINIUS-474](https://linear.app/jimpo/issue/BINIUS-474/m4-modify-protocol-to-prove-composite-circuits). **Stacked on #2127**; review that first.

Adds `IOPVerifierM4` and `VerifierM4`: setup and verification for a whole `ConstraintSystemM4`.

A composite proof runs one sub-proof per sub-system, on one channel, in one fixed order — the main chip, then the chips in ID order, which is the order `ConstraintSystemM4::verify` and witness generation already walk. Main is verified by the Binius64 verifier (it runs once; its inout values are the statement), each chip by this crate's data-parallel verifier (it runs once per call; its inout values are committed). One channel means one Fiat-Shamir transcript and one combined FRI opening; the sub-proofs are otherwise independent.

Per your call on the Linear thread, the interfaces are named as if they prove the full relation, and the unconstrained-chip-calls gap is documented on `IOPVerifierM4` where a caller meets it.

### Two things worth your eye

**The plan's "FRI must be sized off the max oracle" turned out to be a non-issue** — and finding out why surfaced something. `BaseFoldVerifierCompiler::new` takes an `_arity_strategy` it never consults: `FRIParams::optimal_for_batch` batches the specs and picks arities itself. So the `log_code_len` / `with_optimal_arity` computation in `Verifier::setup` (both this crate's and `binius-verifier`'s) is dead — its result is discarded. I didn't want to copy that into new code, so the composite setup skips it and passes a placeholder strategy with a comment. **Cleaning up the two existing call sites is out of scope here; say the word and I'll file it.**

**`oracle_specs` concatenates the sub-verifiers' lists** rather than replaying the composite `verify` against an `OracleSetupChannel`. Each sub-list is still derived from its own `recv_oracle` calls, so nothing is hand-maintained. Replaying the composite would be tidier, but `binius_verifier::IOPVerifier::oracle_specs` discards its replay's `Result`, so an erroring main would silently truncate the list and the FRI parameters would be sized for part of the system. The reasoning is in the doc comment.

### Testing

`binius-frontend` becomes a **dev-dependency** of `binius-m4-verifier` — building a composite constraint system by hand is impractical, and the frontend depends only on `binius-core`, so it adds no cycle. The fixture is a small main chip that takes each mixed word from a hint and calls a much larger chip to constrain it, so the two sub-systems commit oracles of visibly different sizes.

Three tests: setup covers every sub-system and derives each chip's instance count from its active count; the codeword holds every sub-system rather than being sized off one; setup rejects an invalid system.

Verified green: `cargo test -p binius-m4-verifier` (6 passed), `cargo clippy -p binius-m4-verifier --tests -- -D warnings`, `cargo fmt --check`.

### DIFF
--- a/crates/m4-verifier/src/composite.rs
+++ b/crates/m4-verifier/src/composite.rs
@@ -1,0 +1,358 @@
+// Copyright 2026 The Binius Developers
+
+//! Verifying a whole [`ConstraintSystemM4`]: the main chip and every numbered chip.
+
+use binius_core::{
+	Word,
+	constraint_system::{InoutSegment, m4::ConstraintSystemM4},
+};
+use binius_field::FieldOps;
+use binius_hash::StdHashSuite;
+use binius_iop::{
+	basefold::compiler::BaseFoldVerifierCompiler,
+	channel::{IOPVerifierChannel, OracleSpec},
+	fri::{ConstantArityStrategy, calculate_n_test_queries},
+};
+use binius_ip::channel::WordIPVerifierChannel;
+use binius_transcript::{VerifierTranscript, fiat_shamir::Challenger};
+use binius_utils::checked_arithmetics::log2_ceil_usize;
+use binius_verifier::{Error, SECURITY_BITS, config::B128};
+
+use crate::{
+	commit::BatchCommitLayout,
+	verify::{IOPVerifier, Scheme},
+};
+
+/// IOP verifier for a composite M4 system.
+///
+/// The main chip runs once, so it is verified by the Binius64 verifier: its inout values are the
+/// statement, known to the verifier and not committed. Each numbered chip runs once per call that
+/// reaches it, so it is verified by this crate's data-parallel [`IOPVerifier`] over a batch of
+/// instances, with its inout values committed alongside the private ones.
+///
+/// The sub-proofs share one channel, and so one Fiat-Shamir transcript and one combined FRI
+/// opening. They are otherwise independent: no claim crosses between them.
+///
+/// ## What a composite proof does not prove
+///
+/// Chip calls are unconstrained. A proof therefore establishes that main's constraints hold and
+/// that each chip's constraints hold on its own committed instances — it does *not* establish that
+/// the instances are the ones main's calls asked for, nor that they were called at all.
+/// [`ConstraintSystemM4::verify`] checks strictly more than this argument does. Constraining chip
+/// calls is future work.
+#[derive(Debug, Clone)]
+pub struct IOPVerifierM4 {
+	/// The verifier for the main chip, which runs once.
+	main: binius_verifier::IOPVerifier,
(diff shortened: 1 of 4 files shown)


crates/m4-verifier/src/composite.rs
```diff
@@ -0,0 +1,358 @@
...
+
+		// The query count is fixed by the rate and the soundness target.
+		let n_test_queries = calculate_n_test_queries(SECURITY_BITS, log_inv_rate);
+
+		let iop_compiler = BaseFoldVerifierCompiler::new(
+			&Scheme::new(),
+			oracle_specs,
+			log_inv_rate,
+			n_test_queries,
+			&ConstantArityStrategy::new(1),
```
jimpo, line 190: Use the MinProofSizeStrategy

crates/m4-verifier/src/composite.rs
```diff
@@ -0,0 +1,358 @@
...
+		&self,
+		inout: &[Word],
+		transcript: &mut VerifierTranscript<Challenger_>,
+	) -> Result<(), Error>
+	where
+		Challenger_: Challenger,
+	{
+		let mut channel = self
+			.iop_compiler
+			.create_channel_from_transcript::<StdHashSuite, Challenger_, _>(transcript);
```
jimpo, line 234: follow-up: Parameterize the VerifierM4 by H: HashSuite, like binius_verifier::Verifier. Do the same for binius_m4_verifier::Verifier.

jimpo (approved):

crates/m4-verifier/src/composite.rs
benji-potamus: Done — swapped for `&MinProofSizeStrategy`, and dropped my comment's claim that the strategy argument is ignored, since that was an observation about the current `BaseFoldVerifierCompiler::new` rather than something new code should lean on.

**The change is on #2131, not here.** By the time I got to this, #2129 was already in the merge queue, so pushing to this branch would have ejected it. #2131 is stacked on this one and carries the fix in, right behind it.

crates/m4-verifier/src/composite.rs
benji-potamus: Filed as [BINIUS-480](https://linear.app/jimpo/issue/BINIUS-480/m4-parameterize-verifierm4-and-verifier-by-h-hashsuite), covering both `VerifierM4` and the single-chip `Verifier`, and noting that `Prover<P>`/`ProverM4<P>` need the parameter threaded too since they inherit the suite through `from_verifier_compiler`.

PR #2130, opened 2026-08-12
BINIUS-472: take the inner statement as the recursive circuit's public input

tcoratger:
Closes [BINIUS-472](https://linear.app/jimpo/issue/BINIUS-472).

Moves the inner statement from a private witness of the recursive circuit to its public interface.

### The problem

[BINIUS-465](https://linear.app/jimpo/issue/BINIUS-465) (#2116) made the statement reach the verifier symbolically, so it stopped being baked in as constants. But the wires it arrives on come from `Shared::input_wire`:

```rust
pub fn input_wire(&self, kind: &'static str) -> Wire {
    let wire = self.builder.add_witness();          // private
    self.inputs.borrow_mut().push(Input { wire, kind });   // filled by the replay
    wire
}
```

So the statement is a **private witness**, supplied by whoever fills the circuit.

The whole point of a recursive artifact is that the statement it verifies is visible to whoever checks the outer proof. As it stood, an outer proof could be satisfied with any statement the filler chose, and nothing a checker can see said which one.

### The idea

Two kinds of wire were being conflated. Split them:

```
input_wire      -> add_witness()  -> Recorded::inputs      the replay fills these
statement_wire  -> add_inout()    -> Recorded::statement   the caller fills these
```

`inputs` exists for values the circuit **cannot derive**. A statement is neither derived nor replayed — it is *given*. So it gets its own path, and its own list.

### The change

`observe_words` is the only call site:

```
- .map(|_| SymbolicWord::wire(&self.shared, self.shared.input_wire("observe_words")))
+ .map(|_| SymbolicWord::wire(&self.shared, self.shared.statement_wire()))
```

`Recorded` carries the wires, and a small helper writes them:

```rust
recorded.populate_statement(&mut filler, witness.inout());
```

The helper exists so a caller cannot silently under-fill by hand-rolling the zip. It **asserts** on a length mismatch rather than returning an error: the statement comes from the caller, who holds the `Recorded`, so a wrong length is a precondition violation and not proof data. That is deliberately unlike `ProofLayout::populate` in [BINIUS-471](https://linear.app/jimpo/issue/BINIUS-471), where the length is attacker-controlled and an error is right.

The filler stops filling the statement and only forwards it:

```
- for &word in words { self.fill_word("observe_words", word); }
  self.inner.observe_words(words)
```

### What the test now asserts

The recursive circuit's public values are **exactly** the inner statement:

```
inout [0x0123456789abcdef, 0xfedcba9876543210, 0x1, 0xffffffffffffffff, 0xeb589e6bfde39e6a]
```

That is the CRC-64 fixture's four message words plus its digest — the inner circuit's own inout, verbatim, and nothing else alongside it.

The assertion is equality rather than a subsequence search, because a search would pass vacuously. The exact layout was measured first, then pinned.

### Cost

| | before | after |
|---|---|---|
| recorded inputs | 3,663 | 3,658 |
| gates | 273,264 | 273,264 |
| AND | 16,698 | 16,698 |

The input list drops by exactly the five statement words. The circuit is otherwise identical — the wires changed segment, they did not disappear.

### Scope

- **changed:** `Shared` (one new allocation path), `observe_words` on the builder channel, `observe_words` on the filler, `Recorded`
- **new API:** `Recorded::statement` and `Recorded::populate_statement`
- **left untouched:** every other recorded kind (`sample`, `sample_bits`, `opening`, `committed_vector`, `recv_one`, `merkle_root`) — those really are fill-me wires

### What this is not

Being a **public input** and being **absorbed into the Fiat-Shamir state** are different properties, and a recursive verifier needs both.

This is the first. [BINIUS-469](https://linear.app/jimpo/issue/BINIUS-469) is the second, and the `UNCONSTRAINED` marker at that call site stays until it lands.

That is also why there is no negative test here. Tampering the statement and expecting rejection would exercise the arithmetic binding that already existed, not this change — and the property this change is about is exactly the one the assertion above states.

### Relationship to BINIUS-471

Independent. That one touches `recv_openings` and `recv_committed_vector`; this touches `observe_words`. Both are branched from `main`, so they merge in either order — whichever lands second needs a trivial rebase in `channel.rs` and `filler.rs`.

### Verification

- `cargo test -p binius-recursion` — 28 unit + 3 integration passed

### DIFF
--- a/crates/recursion/src/channel.rs
+++ b/crates/recursion/src/channel.rs
@@ -7,7 +7,7 @@ use std::{array, rc::Rc};
 use binius_circuits::multiplexer::multi_wire_multiplex;
 use binius_core::word::Word;
 use binius_field::{BinaryField128bGhash as B128, Field, FieldOps, util::FieldFn};
-use binius_frontend::{Circuit, CircuitBuilder};
+use binius_frontend::{Circuit, CircuitBuilder, Wire, WitnessFiller};
 use binius_hash::StdHashSuite;
 use binius_iop::{
 	merkle_channel::{self, MerkleIPVerifierChannel},
@@ -41,6 +41,31 @@ pub struct Recorded {
 	pub circuit: Circuit,
 	/// The wires the witness must supply, in the order the verifier reached them.
 	pub inputs: Vec<crate::shared::Input>,
+	/// The inout wires the inner statement enters on, in observation order.
+	///
+	/// These are the circuit's public interface, so the caller fills them rather than the replay.
+	/// See [`Self::populate_statement`].
+	pub statement: Vec<Wire>,
+}
+
+impl Recorded {
+	/// Writes the inner statement onto the wires it enters the circuit on.
+	///
+	/// # Panics
+	///
+	/// Panics unless the statement is as long as the one the verifier observed.
+	/// A shorter one would leave wires unset.
+	/// A longer one means the caller and the circuit disagree about what is being verified.
+	pub fn populate_statement(&self, w: &mut WitnessFiller, statement: &[Word]) {
+		assert_eq!(
+			statement.len(),
+			self.statement.len(),
+			"precondition: the statement must be the one the circuit verifies"
+		);
+		for (&wire, &word) in self.statement.iter().zip(statement) {
+			w[wire] = word;
+		}
+	}
 }
 
 /// A channel that records a verifier run as a Binius64 circuit.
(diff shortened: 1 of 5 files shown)


jimpo (changes requested):
Not sure about this. I actually like the mechanism where replay fills the inout wires when observing. Fully forwarding all inputs is not always the desired behavior -- we may want to only partially constrain the inputs to an inner proof, for example.

I'd rather have `replay` return the out wires, and then outside of it, use `assert_eq` to bind the inout wires of the outer proof with the inout wires of the inner proof.

jimpo (approved):

PR #2132, opened 2026-08-12
perf(circuits): read each msm window in one shift, not a band per bit

wstran:
## What

The fixed-window ECDSA scalar mul builds each window's table selector by pulling the exponent bits out one at a time: a `band` with `1` to isolate the bit, then a shift into place. That is one AND constraint per exponent bit. The bits are contiguous, so one masked shift pulls the whole window instead. When a window straddles a 64-bit limb boundary, the two halves land in disjoint bit ranges and XOR joins them before the mask.

The selector values are unchanged, only the constraints that build them: 3 ANDs saved per window over 32 windows per 128-bit subscalar. The msm examples run two points (4 subscalars after the endomorphism split); p2pkh verifies with a single point (2 subscalars), hence half the delta.

## Numbers

Baseline is current `main` (e4a2612d), snapshots re-blessed on top of it.

| example | AND | ZERO | committed trace |
|---|---|---|---|
| ethsign | 113,992 -> 113,608 (-384) | 21,342 -> 20,958 (-384) | 189,236 -> 188,468 (-768) |
| ec_msm | 112,581 -> 112,197 (-384) | 21,366 -> 20,982 (-384) | 187,577 -> 186,809 (-768) |
| bitcoin_p2pkh | 83,892 -> 83,700 (-192) | 16,088 -> 15,896 (-192) | 134,917 -> 134,533 (-384) |

repro:
```
cargo run --release --example ethsign -- stat
```

## Correctness

The selector wire carries the same value as before; the msm tests compare against the k256 reference for windows 1 through 3 (window 3 exercises the limb-straddling path) and the endomorphism test covers the default window 4. All pass unchanged. The three example snapshots are re-blessed and no other snapshot moves. clippy, fmt and typos are clean, and ethsign, ec_msm and bitcoin_p2pkh prove and verify end to end.

## Scope

Only the selector packing in `strauss_accumulate`. The table build and the point additions are untouched.

### DIFF
--- a/crates/circuits/src/ecdsa/scalar_mul.rs
+++ b/crates/circuits/src/ecdsa/scalar_mul.rs
@@ -208,7 +208,6 @@ fn strauss_accumulate(
 		.map(|table| table.iter().map(Vec::as_slice).collect())
 		.collect();
 
-	let one = b.add_constant_64(1);
 	let n_windows = exponent_bits.div_ceil(window);
 	let mut acc = Secp256k1Affine::point_at_infinity(b);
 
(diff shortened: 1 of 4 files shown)


crates/circuits/src/ecdsa/scalar_mul.rs
```diff
@@ -223,23 +222,28 @@ fn strauss_accumulate(
...
-			let mut sel = b.add_constant(Word::ZERO);
-			for j in 0..window {
-				let bit_index = base_bit + j;
-				if bit_index >= exponent_bits {
-					continue; // past the top of the exponent — contributes a zero bit
+			// Selector = this window's exponent bits, low bit first; bits at or past
+			// `exponent_bits` read as zero. `multi_wire_multiplex` reads bit `j` of `sel` as bit
+			// `j` of the table index, matching `table[x] = x · P`. One masked shift pulls the whole
+			// chunk, where the old code spent a band per bit (two limbs when the chunk straddles a
+			// 64-bit boundary).
```
benji-potamus: CONTRIBUTING.md: comments must explain current behavior without referencing how the code used to work. "where the old code spent a band per bit" belongs in the commit message, not here.

```suggestion
			// Selector = this window's exponent bits, low bit first; bits at or past
			// `exponent_bits` read as zero. `multi_wire_multiplex` reads bit `j` of `sel` as bit
			// `j` of the table index, matching `table[x] = x · P`. One masked shift pulls the
			// whole chunk, joining two limbs when it straddles a 64-bit boundary.
```

crates/circuits/src/ecdsa/scalar_mul.rs
```diff
@@ -223,23 +222,28 @@ fn strauss_accumulate(
...
+			let sel = if base_bit >= exponent_bits {
```
benji-potamus: This arm is unreachable. `base_bit` is `w_idx * window` with `w_idx <= n_windows - 1`, and `(exponent_bits.div_ceil(window) - 1) * window < exponent_bits` for every `window`, so `base_bit < exponent_bits` always. The top partial window is already handled by the `min` on the next line. Please drop the `if` and inline the `else` body.

crates/circuits/src/ecdsa/scalar_mul.rs
```diff
@@ -223,23 +222,28 @@ fn strauss_accumulate(
...
+				let mask = b.add_constant_64((1u64 << n_bits) - 1);
+				let offset = (base_bit % Word::BITS) as u32;
+				let lo = base_bit / Word::BITS;
+				let hi = (base_bit + n_bits - 1) / Word::BITS;
+				if lo == hi {
+					b.band(b.shr(subscalar[lo], offset), mask)
+				} else {
+					// The halves land in disjoint bit ranges, so XOR joins them before the mask.
+					let low = b.shr(subscalar[lo], offset);
+					let high = b.shl(subscalar[hi], Word::BITS as u32 - offset);
```
benji-potamus: This shift amount is in range only by a non-obvious argument: `offset == 0` would make it 64, which `shl` asserts against, and that cannot happen because `n_bits <= window < Word::BITS` (asserted in `msm_strauss_endo`) forces a chunk starting on a limb boundary to fit in `subscalar[lo]`, taking the `lo == hi` arm instead. Please add that as a comment so the next reader does not have to re-derive it.

benji-potamus (changes requested):
Request changes.

Good change: pulling the window out with one masked shift is both cheaper and easier to read than the bit-by-bit loop, and the selector values are unchanged — the straddling join and the clamp on the top partial window are both right. The deltas reproduce on current `main` after re-blessing: ethsign and ec_msm −384 AND / −384 ZERO / −768 committed trace, bitcoin_p2pkh −192 / −192 / −384.

Three things inline: the new comment breaks CONTRIBUTING.md's rule against describing old behavior, the `base_bit >= exponent_bits` arm is dead, and the straddling shift amount deserves a note on why it is always in range.

The branch also needs a rebase — the three snapshots conflict with `main` and their numbers are stale. Only the snapshots conflict; `scalar_mul.rs` merges clean, so a rebase plus `bless-snapshot` on the three examples is all it takes.

benji-potamus (approved):
Approve.

Good, focused perf change: replaces a per-bit AND/shift/XOR chain with one masked shift per window (two shifts + XOR when a window straddles a 64-bit limb boundary), producing the same selector value with fewer constraints. I traced the bit math by hand — the low/high shift split lands in disjoint bit ranges and reconstructs the window correctly, including at the limb boundary and the short top window — and it holds. The unreachable `base_bit >= exponent_bits` arm removed in the second commit was indeed dead (the `w_idx` loop bound already guarantees `base_bit < exponent_bits`), so that cleanup is correct too. Snapshot deltas (ethsign -384, ec_msm -384, bitcoin_p2pkh -192 AND constraints) match what I get rebasing locally, and no other snapshots moved. Scope is tight — only the selector-packing lines in `strauss_accumulate` changed.

jimpo (approved):

PR #2136, opened 2026-08-13
Add a keyed variant of the fixed-length BLAKE3 gadget

benji-potamus:
Closes [BINIUS-481](https://linear.app/jimpo/issue/BINIUS-481/circuits-blake3-keyed-fixed).

Adds `blake3_keyed_fixed`, the keyed-mode counterpart of `blake3_fixed`, taking the key as a
`ByteVec` of 0 to 32 bytes.

## Commits

1. **Share the 32-bit word splitting helpers through the util module** — moves `split_u32_words`
   and `zeroed_u32_words` out of `hash_based_sig::hashing::chain_blake3` into `util`, unchanged.
   The keyed gadget needs the same byte-vector → 32-bit-word conversion, and `zeroed_u32_words`
   closes a malleability gap worth having in one place rather than two.

2. **Add a keyed variant of the fixed-length BLAKE3 gadget** — the new entry point plus the
   sharing that makes it a variant rather than a copy.

## How it shares with `blake3_fixed`

Keying BLAKE3 is exactly two substitutions: the key replaces the `IV` as the chaining value every
chunk and parent node starts from, and every compression carries `KEYED_HASH`. Those two always
travel together, so they are threaded as one `Option<[Wire; 8]>` through `blake3_chunk`,
`blake3_parent`, `blake3_parent_2x` and `blake3_tree_root` — `None` is the unkeyed mode. There is
no way to set the flag without a key or supply a key without the flag.

Both public entry points are thin wrappers over one shared body, so the chunk/tree shape, the
2x-batching and the padding logic have a single implementation.

`blake3_chunk` is public and gains the `key` parameter; it has no callers outside this module.

## Key handling — please sanity-check this reading of the issue

The issue asks for a `key: ByteVec` "with length between 0 to 32", while BLAKE3 natively keys on
exactly 32 bytes. The gadget therefore zero-pads a shorter key to 32, and:

- **The key length must be fixed at circuit construction time** (a point `len_range`, as
  `ByteVec::new_const_len` builds). A runtime-varying key length is rejected by an assert. This
  matches the `_fixed` family and the const-length `ByteVec`s `chain_blake3` already passes around.
  A dynamic-length key is implementable — it needs a runtime-driven byte mask instead of the
  constant one — but it costs constraints for a capability nothing asks for yet, so I left it out.
  Happy to add it if that was the intent.
- **Bytes past the key length are masked to zero**, not merely assumed zero: `ByteVec` makes no
  guarantee about them and the key wires are typically witness, so an unmasked padding byte would
  let a prover steer the digest. `keyed_ignores_bytes_past_the_key_length` populates them with
  `0xff` and pins the digest.
- Zero-padding means the key length is not itself hashed, so `b"k"` and `b"k\0"` produce the same
  digest. Documented on the function.

## Tests

Checked against `blake3::keyed_hash` with the zero-padded key: full 32-byte keys across message
lengths spanning the block/chunk/tree-shape boundaries, short keys on both sides of the 8-byte wire
and 4-byte word boundaries, garbage padding, a proptest over random message and key lengths, and
the two rejected-key asserts. Both commits are independently green (build, tests, clippy).

### DIFF
--- a/crates/circuits/src/util.rs
+++ b/crates/circuits/src/util.rs
@@ -79,9 +79,8 @@ pub(crate) fn zeroed_u32_words(
 			} else if word_start >= valid_bytes {
 				zero
 			} else {
-				let keep = valid_bytes - word_start;
-				let mask = builder.add_constant_64((1u64 << (keep * 8)) - 1);
-				builder.band(raw[i], mask)
+				let keep_bits = (valid_bytes - word_start) * 8;
+				clear_high_bits(builder, raw[i], (64 - keep_bits) as u32)
 			}
 		})
 		.collect()
--- a/crates/circuits/src/util.rs
+++ b/crates/circuits/src/util.rs
@@ -26,6 +26,67 @@ pub fn clear_high_bits(builder: &CircuitBuilder, w: Wire, n: u32) -> Wire {
 	builder.shr(builder.shl(w, n), n)
 }
 
+/// Splits each 64-bit wire into two little-endian 32-bit word wires (low half, then high half).
+///
+/// Returns exactly `num_words` words, zero-padding when `data` runs out.
+pub(crate) fn split_u32_words(
+	builder: &CircuitBuilder,
+	data: &[Wire],
+	num_words: usize,
+) -> Vec<Wire> {
+	let mut words = Vec::with_capacity(num_words);
+	for &w in data {
+		if words.len() >= num_words {
+			break;
+		}
+		words.push(clear_high_bits(builder, w, 32));
+		if words.len() >= num_words {
+			break;
+		}
+		words.push(builder.shr(w, 32));
+	}
+	while words.len() < num_words {
+		words.push(builder.add_constant_64(0));
+	}
+	words
+}
+
+/// Splits a byte vector into `num_words` 32-bit words, forcing every byte at index `>= valid_bytes`
+/// to zero.
+///
+/// The zeroing closes a malleability gap:
+/// - a byte vector leaves bytes past its length unconstrained,
+/// - a hash compression mixes the whole block, including those bytes,
+/// - pinning them stops a prover from choosing the padding to alter the digest.
+///
+/// Each word is handled by its position relative to the content:
+/// - fully inside: passed through,
+/// - fully past: replaced by the zero constant,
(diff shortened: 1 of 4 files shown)


crates/circuits/src/util.rs
```diff
@@ -26,6 +26,67 @@ pub fn clear_high_bits(builder: &CircuitBuilder, w: Wire, n: u32) -> Wire {
...
+	(0..num_words)
+		.map(|i| {
+			let word_start = 4 * i;
+			if word_start + 4 <= valid_bytes {
+				raw[i]
+			} else if word_start >= valid_bytes {
+				zero
+			} else {
+				let keep = valid_bytes - word_start;
+				let mask = builder.add_constant_64((1u64 << (keep * 8)) - 1);
```
jimpo: Don't mask, use the clear_high_bits util function.

jimpo (changes requested):

crates/circuits/src/util.rs
benji-potamus: Done, in its own commit (`Clear a straddling word's high bytes with the shift pair`) so it's easy to drop on its own if you don't like the numbers below.

It isn't cost-neutral here, so flagging rather than quietly blessing the snapshot. `zeroed_u32_words` is on the `hashsign` chain-hashing path, and swapping the `band` for the shift pair moves that circuit:

| | before | after | Δ |
|---|---|---|---|
| AND constraints | 136,383 | 135,720 | **−663** |
| ZERO constraints | 116,975 | 118,640 | +1,665 |
| Committed trace | 254,949 (97.3% of 2^18) | 255,951 (97.6% of 2^18) | +1,002 |

Which is exactly what the `clear_high_bits` doc predicts — the pair commits its intermediate where the mask folded into an AND constraint. Cheaper in AND, dearer in ZERO and committed values.

The part I'd want your call on: `hashsign`'s committed trace was already at 97.3% of 2^18, so this eats 1,002 of the 7,195 spare and leaves 6,193. Fine if AND is the dimension you care about, less fine if that headroom is the one under pressure. Say the word and I'll drop the commit. No other snapshot moves.

jimpo (approved):

PR #2142, opened 2026-08-13
Split blake3_compress rounds across both 32-bit lanes

benji-potamus:
Closes [BINIUS-484](https://linear.app/jimpo/issue/BINIUS-484/circuits-blake3-compress-should-split-rounds-into-lanes).

`blake3_compress` ran its 7 rounds in the low 32 bits of each word, leaving the high half idle — the half where the parallel-halves `iadd_32` and `rotr32` gates do a second, independent 32-bit computation for free. `blake3_compress_2x` and `blake3_compress_2x_seq` already fill both lanes, but only when a second compression is available to fill them with.

## The change

Cut the round sequence in two instead, so a lone compression fills both lanes with itself:

```text
    high lane (bits [32:64]):  round 0 -> round 1 -> round 2 -> (idle)
    low  lane (bits  [0:32]):  round 3 -> round 4 -> round 5 -> round 6
```

Four packed rounds replace seven single-lane ones. The two halves meet at the state after round 2, which the low lane needs as its input and the high lane only produces later; the same shape `blake3_compress_2x_seq` and `sha256_compress_2x_seq` already have. A `Blake3RoundSplitHint` supplies that state off-circuit, and one equality per state word pins it: the word's high half against the caller's own inputs, its low half against what the high lane really computed. Both halves are constrained, so the hint has no freedom.

Seven rounds do not divide evenly in two, so the high lane spends the last packed round idle. Its result stays in the returned words' high halves rather than being masked off — `blake3_compress` promises only the low 32 bits, as `blake3_compress_2x_seq` already does for a pair's first compression. The four call sites that read a whole word clear it themselves: `blake3_parent`, `cv8_to_hash4` in `chain_blake3`, the `independent_hashes` example, and the `run_compress` test helper. `blake3_chunk` needed nothing — it already masks once on the way out.

`round` now takes its message words already in schedule order, since the two lanes run different rounds at once and a round can no longer pick its own schedule.

## Cost

AND constraints, measured with `CircuitStat::collect`:

| circuit | before | after |
| -- | --: | --: |
| `blake3_compress` | 336 | **192** |
| `blake3_fixed(64)` | 336 | **192** |
| `blake3_fixed(192)` | 672 | **528** |
| `blake3_fixed(1024)` | 2688 | 2688 |
| `blake3_fixed(2048)` | 5712 | **5568** |

A pair of block compressions through `blake3_compress_2x_seq` still costs 336 for the two, so `blake3_chunk` keeps pairing and only its odd trailing block gets cheaper — 1024 bytes is 16 blocks, all paired, hence unchanged. The comment there explaining why an odd block is not padded to a pair is updated: the reason is now the pair's 7 rounds against a lone compression's 4, not an idle half-word.

The `hashsign` circuit, whose Winternitz chain steps are single BLAKE3 compressions, is where this lands hardest — its snapshot is the one file that moves:

```
- ├─ ZERO constraints: 118,640 used (90.5% of 2^17)
+ ├─ ZERO constraints: 114,662 used (87.5% of 2^17)
- ├─ AND constraints: 135,720 used (51.8% of 2^18)
+ ├─ AND constraints: 118,455 used (90.4% of 2^17)
```

so its AND table halves. The other twelve committed snapshots are unchanged (the `blake3` example hashes a whole number of block pairs in one chunk, so it never reaches a lone compression).

## A note on #2144

`blake3_keyed_fixed_2x` landed while this was in review, and this change moves the ground under its cost claim. Its `keyed_2x_never_costs_more_than_two_single_hashes` asserted `2 * paired <= single`, which held only because a lone compression cost 336. It now costs 192, so for a one-block message two single keyed hashes cost **384** against the pair's **336** — the pair saves 12.5%, not half. The assertion is relaxed to `paired < single` (still true at every length it tests, with equality at even block counts), and the gadget's doc now says what it actually recovers: the 3 rounds a lone compression evaluates twice. Whether the two-lane keyed gadget still earns its complexity at that margin is worth a second look, but not in this PR.

## Testing

- The existing compression tests all pass unchanged: the two spec-trace known answers from draft-aumasson-blake3-00, the `compress_matches_reference` proptest, and the whole-hash tests against the `blake3` crate.
- `compress_ignores_dirty_input_high_halves` is new. The half a 32-bit input does not use is no longer inert padding — it is where half of the mixing now happens — and nothing range-constrains those wires, so a witness filling them must not steer the result. Message words are masked and every other input is lifted by a shift that drops its high half.
- `cargo test -p binius-circuits -p binius-examples`, clippy on both, and the `blake3_compress`-based `protocol_round_trips_with_non_power_of_two_constants` in `binius-m4-prover`, which also covers the new hint under batched witness population.

### DIFF
--- a/crates/circuits/src/blake3/compress.rs
+++ b/crates/circuits/src/blake3/compress.rs
@@ -8,6 +8,10 @@
 //!
 //! The structure mirrors the [reference implementation] from the BLAKE3 crate.
 //!
+//! Every gadget here fills both 32-bit lanes of a word. Where two compressions are available they
+//! run side by side, one per lane ([`blake3_compress_2x`], [`blake3_compress_2x_seq`]); a lone
+//! compression splits its own round sequence across the lanes instead ([`blake3_compress`]).
+//!
 //! [reference implementation]: https://github.com/BLAKE3-team/BLAKE3/blob/master/src/portable.rs
 
 use std::{array, iter};
@@ -18,7 +22,32 @@ use binius_frontend::{ChipGadget, CircuitBuilder, Hint, Wire};
 use super::{IV, MSG_SCHEDULE};
 use crate::util::clear_high_bits;
 
+/// Rounds [`blake3_compress`] runs in the high lane, the first half of the split.
+const HIGH_LANE_ROUNDS: usize = 3;
+
+/// Rounds [`blake3_compress`] runs in the low lane, the second half of the split.
+///
+/// The longer half, so it also counts the packed rounds: 8 lane-rounds are evaluated for 7 real
+/// ones, and the high lane spends the last of them idle.
+const LOW_LANE_ROUNDS: usize = 7 - HIGH_LANE_ROUNDS;
+
 /// BLAKE3 compression function.
+///
+/// A single compression, evaluated as two lanes of one packed core: the round sequence is cut in
+/// two and the halves run side by side rather than one after the other.
+///
+/// ```text
+///     high lane (bits [32:64]):  round 0 -> round 1 -> round 2 -> (idle)
+///     low  lane (bits  [0:32]):  round 3 -> round 4 -> round 5 -> round 6
+/// ```
+///
+/// Both halves of the split want the same thing: the state after round 2, which is the low lane's
+/// input and the high lane's output. That circular dependency is broken with a
+/// `Blake3RoundSplitHint` that computes the state off-circuit and feeds it in, then constrains it
+/// word-for-word against what the high lane really computed, so the hint cannot lie.
+///
+/// Four packed rounds replace seven single-lane ones. The high lane sits idle through the last of
+/// them — 7 rounds do not divide evenly in two — and its result is discarded.
 ///
 /// # Arguments
 ///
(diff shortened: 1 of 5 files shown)


crates/circuits/src/blake3/compress.rs
```diff
@@ -41,30 +77,84 @@ pub fn blake3_compress(
...
+		if step + 1 == HIGH_LANE_ROUNDS {
+			for (hinted, (init, split)) in iter::zip(&merged, iter::zip(init_up, v)) {
+				let expected = builder.bxor(init, builder.shr(split, 32));
+				builder.assert_eq("blake3_compress.split_state", *hinted, expected);
+			}
+		}
+	}
+
+	// The result is the low lane's; the high lane holds its discarded final round.
+	array::from_fn(|i| clear_high_bits(builder, builder.bxor(v[i], v[i + 8]), 32))
```
jimpo: Is this necessary? I don't think `blake3_compress` should promise that the top bits are clear.

benji-potamus: Dropped it — the returned words now carry the discarded lane in their high halves, and `blake3_compress` promises only the low ones. Pushing shortly.

It was not costing an AND constraint (the mask is two shifts, which stay linear definitions and fuse into the consumers), but it did cost zero constraints: a lone compression goes from 193 to 177, and `blake3_fixed(192)` from 467 to 451, with the AND count unchanged at 192.

Four call sites read a whole returned word and now clear it themselves:

- `blake3_parent` masks its result, since a parent's value is read whole — the level above merges it with a shift, and the root is the digest.
- `cv8_to_hash4` (`chain_blake3`) masks the even word it merges into.
- The `independent_hashes` example masks before its 64-bit equality.
- The `run_compress` test helper masks, so the compression tests compare what a caller would read.

`blake3_chunk`'s trailing lone compression needed nothing: it already masks once on the way out, for exactly this reason.

PR #2146, opened 2026-08-13
BINIUS-486: derive the outer-shift phase's h rows from the folded oblong table

benji-potamus:
Rebased onto main now that #2145 has merged. The one genuinely new prover component in the six-phase shift reduction of [binius.xyz#124](https://github.com/binius-zk/binius.xyz/pull/124) ([BINIUS-486](https://linear.app/jimpo/issue/BINIUS-486)): the rounds binding the outer slot of a shift sequence. Landed **unwired** — nothing calls `OuterShiftStage` yet, so the transcript is untouched — so that the protocol change in BINIUS-487 is wiring rather than wiring plus new mathematics.

## The problem

The reduction's `h` factor spans 24 variables once a term carries two shifts, so its value table would hold `2^24` entries — 268 MB, and `2^24` field operations to fill. It is never formed. Writing `T` for the shift operator and `d` for the oblong weights,

```
eta := T[d]                                          2^15 entries — this is today's `h`
h(J, s_2, o_2, s_1, o_1) = T[eta(., s_2, o_2)](J, s_1, o_1)
```

`eta`'s layout `(bit, amount, variant)` low-to-high already matches the binding order, so the stage folds it with `fold_highest_var_inplace`: the outer variant before the outer amount, nine rounds, `2^15` multiplications in total. Per round, each live shift quadruple's `h` row is one slice of `T` applied to one 64-stride of the folded `eta` — `O(2^6)` apiece via the `shift_operator_row` that #2145 exposes — so a round costs `O(2^6 · n_shift)` with no charge proportional to the `2^18` space the quadruples are drawn from.

Folding `eta` and then applying `T` gives the same answer as folding `h`, because `T` is linear in its weights. That is the property everything here rests on, and it is what the test pins.

The doc comment records **why the outer slot binds first**, both reasons from #124: the indicator matrices don't commute (`sra` is the obstruction), so only this order computes `h`; and the push-through is a *shift* only while the inner pair is still a cube index — under the opposite order the outer indicator arrives folded to a dense `2^6 × 2^6` matrix and each quadruple costs a matrix-vector product instead.

## Also here

- `SparseShiftRows::new`, which `from_segments` now goes through. The outer rounds need rows over an 18-bit quadruple space while `LOG_SHIFT_ROWS` is still 9, and BINIUS-487 needs the same door when it widens the axis.
- `LOG_SHIFT_COUNT` beside `SHIFT_COUNT` in the verifier, with `SHIFT_COUNT = 1 << LOG_SHIFT_COUNT` so the two can't drift. It's the width of one shift slot's index — what splits a quadruple into its slots.

## Scope note

The issue also put the `Phase1SumcheckProver` staging here. I left it out: with the row axis still 9 bits the stage is unreachable, so wiring it now means a `Stage` variant and a constructor that nothing can enter — dead structure ahead of its use.

Having since built BINIUS-487 on top, the staging turns out not to be the right shape at all: `prove_single` consumes its prover, and `psi` — the weights these rounds leave behind — has to outlive it, because the rounds binding the intermediate bit index run against them. So 487 drives these nine rounds directly from `run_phase_1_sumcheck`, the way `run_sumcheck` already drives its own sparse first round, and `Phase1SumcheckProver` keeps its two stages.

## Testing

`round_messages_match_a_directly_folded_reference` runs all nine rounds against a reference that folds a **fully derived** `h` directly, so the two paths — derive-then-fold and fold-then-derive — are compared round by round rather than the implementation being checked against itself.

The reference is definitional and builds no `2^24` table: `h(j) = sum_{i,k} d(i) · shift-ind(i,k,outer) · shift-ind(k,j,inner)`, read straight off the word operations, and evaluated only over the inner shifts `g` is supported on — exact, because the outer rounds never mix inner indices. Each indicator selects rather than scales, so it needs no multiplications either.

The fixture covers every variant and every case whose operator slice is not a plain move of the weights (`sra`/`sra32` in either slot), plus a sign-extension pair, a shift under a rotate, repeated indices that must add up, and an identity inner slot. It asserts the claim is non-zero first, so matching messages mean something. Swapping the two slots' roles in the implementation makes it fail, which I checked.

Full `binius-prover`, `binius-m4-prover` and `binius-verifier` suites green; workspace clippy clean.

### DIFF
--- a/crates/prover/src/protocols/shift/outer.rs
+++ b/crates/prover/src/protocols/shift/outer.rs
@@ -1,0 +1,408 @@
+// Copyright 2026 The Binius Developers
+
+//! The rounds binding the outer slot of a shift sequence.
+//!
+//! A shifted value index names a word together with two shifts applied in sequence, and the shift
+//! reduction peels them from the output end inward. These are the rounds that peel the outer one.
+
+use std::iter;
+
+use binius_compute::Allocator;
+use binius_core::{ShiftVariant, word::Word};
+use binius_field::{BinaryField, Field, PackedField};
+use binius_ip::sumcheck::RoundCoeffs;
+use binius_ip_prover::sumcheck::round_evals::RoundEvals;
+use binius_math::{FieldVec, multilinear::fold::fold_highest_var_inplace};
+use binius_verifier::protocols::shift::{LOG_SHIFT_COUNT, SHIFT_COUNT};
+
+use super::{
+	monster::{shift_operator_row, shift_operator_table},
+	phase_1::SparseShiftRows,
+};
+
+/// The `(variant, amount)` pair a shift index names.
+///
+/// This inverts [`Shift::index`](binius_core::constraint_system::Shift::index) over the reduction's
+/// index space rather than over well-formed shifts: the amount axis spans `Word::BITS` for every
+/// variant, so a half-word (`*32`) variant's index may carry an amount no `Shift` of that variant
+/// could hold. Such an index is a hypercube vertex the sumcheck ranges over all the same.
+///
+/// # Panics
+///
+/// Panics unless the index is below [`SHIFT_COUNT`].
+fn decode_shift(index: usize) -> (ShiftVariant, usize) {
+	assert!(index < SHIFT_COUNT, "a shift index names one slot's spelling");
+	let variant = ShiftVariant::from_u8((index >> Word::LOG_BITS) as u8)
+		.expect("an index below SHIFT_COUNT has a variant field below the variant count");
+	(variant, index % Word::BITS)
+}
+
+/// The sumcheck rounds binding the outer shift of a sequence, against the folded oblong table.
+///
+/// The reduction's `h` factor spans 24 variables — the bit position and both shift slots — so its
+/// value table would hold `2^24` entries. It is never formed. Instead, writing `T` for the shift
+/// operator ([`shift_operator_table`]) and `d` for the oblong weights,
+///
+/// ```text
+///     eta := T[d]                                    2^15 entries
+///     h(J, s_2, o_2, s_1, o_1) = T[eta(., s_2, o_2)](J, s_1, o_1)
+/// ```
+///
+/// so `h` is reached by applying `T` to the oblong weights, taking the outer slice, and applying
+/// `T` again. This stage holds `eta` and folds it as the rounds bind the outer slot; each round's
+/// `h` rows are derived from the folded table, one slice at a time. Folding `eta` first and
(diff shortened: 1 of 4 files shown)


crates/prover/src/protocols/shift/phase_1.rs
```diff
@@ -155,20 +155,45 @@ impl<P: PackedField> SparseShiftRows<P> {
...
 			indices,
 			values,
-			log_rows: LOG_SHIFT_ROWS,
+			log_rows,
 		}
 	}
 
 	/// The number of row-index variables the sumcheck has yet to bind.
-	const fn log_rows(&self) -> usize {
+	pub(super) const fn log_rows(&self) -> usize {
```
jimpo: See CONTRIBUTING.md about visibility

crates/prover/src/protocols/shift/outer.rs
```diff
@@ -0,0 +1,408 @@
...
+		);
+
+		// The bit this round binds is an outer one, since the outer slot sits above the inner one
+		// in a quadruple. Dropping the inner slot off it leaves the bit that indexes eta's strides.
+		let half = g.half();
+		let facing_half = half >> LOG_SHIFT_COUNT;
+
+		// One scratch row per side, rewritten for each stored row: `shift_operator_row` writes
+		// every cell, so nothing carries over between rows.
+		let mut own = vec![F::ZERO; Word::BITS];
```
jimpo: Is vec better than array for some reason here? Avoiding the allocation seems good

jimpo (changes requested):
only minor comments

crates/prover/src/protocols/shift/phase_1.rs
benji-potamus: Switched all four to `pub(crate)`.

`phase_1` is a `pub mod` (`#[doc(hidden)]`, for the bench), so it can't be made private and the ancestor-level escape hatch doesn't apply — which is the first bullet under "Visibility scope", and that bullet names `pub(crate)` as the tool. I'd reached for `pub(super)` because `outer` is a sibling and that's the narrowest marker that works, but the guidance calls for the documented one rather than the tightest one, and consistency across the crate is worth more than the extra narrowing.

crates/prover/src/protocols/shift/outer.rs
benji-potamus: No reason — changed both to `[F::ZERO; Word::BITS]`.

`Word::BITS` is a const, so the array is the obvious form and I just didn't reach for it. It matters more than it looks: these are per-round, so at nine rounds the `Vec`s were eighteen allocations per proof for scratch that never outlives the call.

jimpo (approved):

PR #2147, opened 2026-08-13
BINIUS-432: verify a BaseFold opening end to end inside a circuit

tcoratger:
Closes [BINIUS-432](https://linear.app/jimpo/issue/BINIUS-432). Supersedes #2107, which is closed.

A polynomial-commitment opening, proved natively and verified inside a Binius64 circuit. This is the acceptance test for [BINIUS-425](https://linear.app/jimpo/issue/BINIUS-425).

### Why this is a tenth the size of #2107

#2107 was 3,053 lines and was rightly called too big. It had to carry its own tape-reading channel — 693 lines of it, plus 1,018 lines of channel tests — because the channel on `main` constrained nothing at the time.

With #2128, #2130, #2138 and #2143 merged, the channel on `main` **is** the verifier. This just drives it:

| | #2107 | here |
|---|---|---|
| new channel | 693 lines | none |
| channel tests | 1,018 lines | none |
| test + source | 1,342 lines | 709 |

### The problem this had to fix first

The test was not merely unwritten, it was **impossible**. Corrupting a proof byte gave:

```
Verification(FRI(MerkleError(InvalidProof)))
FRI(Channel(IPChannel(InvalidAssert)))
```

The replay aborted before the circuit was ever judged. `WitnessFillerChannel` still verified in three places — `assert_zero`, the layer fold, and the vector rebuild — each of which the circuit now constrains itself. Those were leftovers from when the channel checked nothing.

So a bad proof surfaced as a Rust error out of a witness generator, rather than as an unsatisfied circuit. That is the one property a recursive artifact has to have.

### The idea

The filler stops judging. It reads the transcript and fills wires; the circuit decides.

```
- assert_zero  -> forwards to the native check, aborts the replay
+ assert_zero  -> Ok(()), because the build emitted a ZERO constraint for it

- recv_openings         -> verify_layer natively, then fill
+ recv_openings         -> fill; the circuit folds the layer to the root

- recv_committed_vector -> verify_vector natively, then fill
+ recv_committed_vector -> fill; the circuit rebuilds the tree
```

That is the whole source change: **−15 lines**.

### What the negative tests now show

Each corruption is caught by a *named* check, verified from the failure paths rather than asserted loosely:

| corrupted byte | caught by |
|---|---|
| the commitment root | `layer[0].verify_layer.root` |
| a round polynomial | 370 failures across `opening[*].verify_opening.layer_digest` |
| the terminal codeword | `vector[66].verify_vector.root` **and** `assert_zero[64]` |
| the statement | the transcript diverges from the first challenge on |

The round-polynomial case is the one worth reading. #2107 described this mechanism in prose; here it is demonstrated, because the challenger is real now:

```
before:  round value  -> challenge -> ... -> query indices -> the committed positions
after:   round value' -> a different challenge from that round on, indices included
```

The opened values and siblings on the tape are untouched. The indices addressing them are not, so all 96 openings climb to the wrong entry of the decommitted layer.

A sweep flips one byte at each of 32 evenly spread offsets and requires every one rejected, so there is no unchecked region between the three named cases.

### The statement is observed

The native round trip in `binius-iop-prover` hands the point and the claim to both halves out of band. It can, being a unit test rather than a protocol.

Out of band they would have to be build-time constants, which ties one circuit to one opening. So both are written to the transcript instead, by the prover and by each verifier alike, which makes them wires.

- The point is observed before the commitment, since it is known up front.
- The claim is observed after the masking challenge, being a claim about the masked polynomial.

`one_circuit_verifies_two_different_openings` asserts what that buys: one compiled circuit, two proofs, two different points.

**This is a deviation from the native round trip** and the most reviewable decision here.

### Only the proof and the statement are inputs

Asserted explicitly, not just claimed. The recorded kinds are exactly:

```
committed_vector  merkle_branch  merkle_layer  merkle_root  observe_words  opening  recv_one
```

No challenge, no query index, no digest the circuit could have recomputed.

### The measurement BINIUS-428 tunes against

Building a circuit needs no proof, so the whole surface is measured without proving anything.

| n_vars | rate | queries | AND | BMUL | inputs |
|---|---|---|---|---|---|
| 8 | 1 | 32 | 226,950 | 15,463 | 1,710 |
| 10 | 1 | 32 | 341,924 | 29,545 | 2,614 |
| 8 | 2 | 32 | 281,836 | 20,387 | 2,094 |
| 8 | 1 | 16 | 144,804 | 6,007 | 1,070 |
| 8 | 1 | 64 | 276,940 | 84,164 | 2,346 |

Both readings from #2107 reproduce on the merged channel:

1. **Queries are sublinear.** 16 to 32 costs +57%, but 32 to 64 only +21%, because the shared decommitment layer deepens as the query count grows and every climb gets shorter.
2. **Committed size is not flat.** Four times the committed data costs +51%, not a few percent. Growth is logarithmic with a steep constant.

### Trade-off worth flagging

A malformed transcript no longer fails fast with a targeted error. It produces a witness and an unsatisfied circuit instead.

That is correct for a recursive artifact — the circuit is the verifier — but it is a debuggability trade, and the part I would most want a second opinion on.

### Scope

- **new:** one integration test in `binius-recursion`, and three dev-dependencies for it
- **changed:** `filler.rs` stops verifying, −15 lines
- **left untouched:** the builder channel, the gadgets, and everything outside the crate

### Verification

- `cargo test -p binius-recursion` — 30 unit + 14 integration passed

Rebased onto `main` after #2139 reshaped the frontend module tree; the numbers above are from that base.

### DIFF
--- a/crates/recursion/Cargo.toml
+++ b/crates/recursion/Cargo.toml
@@ -20,7 +20,10 @@ binius-utils = { path = "../utils" }
 digest.workspace = true
 
 [dev-dependencies]
+binius-compute = { path = "../compute" }
 binius-iop-prover = { path = "../iop-prover" }
+binius-ip-prover = { path = "../ip-prover" }
+binius-math = { path = "../math", features = ["test-utils"] }
 binius-prover = { path = "../prover" }
 binius-verifier = { path = "../verifier" }
 proptest.workspace = true
--- a/crates/recursion/src/filler.rs
+++ b/crates/recursion/src/filler.rs
@@ -127,10 +127,11 @@ where
 		value
 	}
 
-	fn assert_zero(&mut self, val: B128) -> Result<(), binius_ip::channel::Error> {
-		// The real check still runs, so a bad proof is caught here rather than surfacing as an
-		// unsatisfiable circuit.
-		self.transcript.borrow_mut().assert_zero(val)
+	fn assert_zero(&mut self, _val: B128) -> Result<(), binius_ip::channel::Error> {
+		// Deliberately not checked here.
+		// The build emitted a ZERO constraint for this, so a bad proof surfaces as an
+		// unsatisfied circuit rather than as an error out of a witness generator.
+		Ok(())
 	}
 
 	fn compute_public_value(&mut self, inputs: &[B128], f: impl FieldFn<B128>) -> B128 {
@@ -201,10 +202,10 @@ where
 		})
 	}
 
-	/// Reads the same advice the native channel reads, and fills the wires the gadget checks.
+	/// Reads the same advice the native channel reads, and fills the wires the gadgets check.
 	///
-	/// The climb is not repeated natively: the circuit hashes each leaf and matches the layer.
-	/// A forged opening therefore surfaces as an unsatisfied constraint, not an error at this call.
+	/// Nothing is verified here: the circuit folds the layer to the root and climbs every opening.
+	/// A forged one therefore surfaces as an unsatisfied constraint, not an error at this call.
 	fn recv_openings(
 		&mut self,
 		commitment: &Self::Commitment,
@@ -232,10 +233,6 @@ where
 			}
 			(layer, openings)
 		};
-
-		// The layer folds to the root over concrete digests, so that half stays a native check.
-		self.scheme
-			.verify_layer(&commitment.commitment.root, layer_depth, &layer)?;
 
 		// Filled in the order the build allocated: the shared layer, then a leaf and a branch per
 		// query.
@@ -265,10 +262,7 @@ where
 			.borrow_mut()
 			.decommitment()
 			.read_scalar_slice::<B128>(len)?;
-		// The rebuild runs over concrete values too, so this stays a native check.
-		self.scheme
-			.verify_vector(&commitment.commitment.root, &data, commitment.leaf_size)?;
-
+		// Not verified here either: the circuit rebuilds the tree over these values.
 		for &value in &data {
 			self.fill_elem("committed_vector", value);
 		}
--- a/crates/recursion/tests/basefold_opening.rs
+++ b/crates/recursion/tests/basefold_opening.rs
@@ -1,0 +1,697 @@
+// Copyright 2026 The Binius Developers
+
+//! A polynomial-commitment opening proved natively, then verified inside a Binius64 circuit.
+//!
+//! ```text
+//!   prover:    native, writing one real transcript
+//!   verifier:  the same code over the builder channel, emitting gates instead of checking values
+//! ```
+//!
+//! A satisfied constraint system is then the statement "this transcript opens this claim".
+//! An outer proof can check that statement, which is what makes this one step of recursion.
+//!
+//! # What enters the circuit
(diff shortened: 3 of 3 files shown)


benji-potamus (approved):
Approve.

Rebased onto `main` (currently #2142) — applies clean. `cargo +nightly-2026-07-01 fmt --all --check`, `cargo clippy -p binius-recursion --all-targets -- -D warnings` and `cargo test -p binius-recursion` are all clean (30 unit + 14 integration, release).

On the source change: I checked each of the three removed native checks against what the build actually emits, and they line up. `assert_zero` → `channel.rs:199` emits a pair of `assert_zero` gates for the `Wires` case, and the `Constant` case can only arise from build-time constants, so a value the build folded to zero is the same value at replay. `verify_layer` → `merkle.rs:200` is `fold_to_root` + `assert_eq_v` against the root, the same predicate as the native `scheme.rs:205`. `verify_vector` → `merkle.rs:225` rehashes every leaf, folds, and matches the root. Removing them is what makes the replay run to completion on any same-length tape, which is what the byte sweep depends on. The debuggability trade you flag reads as the right call to me.

### Findings

**Nothing runs the native verifier over the transcript `prove()` writes** (non-blocking, but the one thing I'd add) — `crates/recursion/tests/basefold_opening.rs:187`. Because `prove()` observes the point and the claim on the transcript, it is no longer the native round trip from `iop-prover/src/basefold/opening.rs:148`, and `record()`/`check()` reproduce that same deviation. So the test is a closed loop: if `prove()` observed the claim in the wrong place, or serialized an element in the wrong word order, `check()` would observe it identically and everything would still pass. That's exactly the decision you called out as most reviewable, and it's currently asserted rather than checked. I tried it — `VerifierMerkleTranscriptChannel` over `VerifierTranscript::new(StdChallenger::default(), proof)`, `observe_words` the point, `recv_merkle_commitment`, `sample` gamma, `observe_words` the claim, then `verify_mlecheck_basefold` — and the native verifier accepts, in 0.01s. About 30 lines, no new dependency, and it turns "this deviation is still the protocol" into something the suite proves.

**`the_cost_surface_over_shapes` costs more than the rest of the file combined** (non-blocking) — `crates/recursion/tests/basefold_opening.rs:623`. Alone it is 56s in release; the other seven together are 27s. CI runs the workspace in the dev profile (`opt-level = 1`) on three arch legs, so it will be several minutes each. What it buys is a printed table plus two assertions that hold for any sane cost model. Suggest `#[ignore]` it, or move it to `benches/`, and keep the table in the PR body where it already is.

**Stale doc on the `scheme` field** — `crates/recursion/src/filler.rs:45` still reads "Reproduces the layer depth and the native checks the builder's scheme also picks", but `optimal_verify_layer` is now its only use. `channel.rs:68` already has the wording this wants: "Consulted only for the layer depth an opening decommits to".

Nit: `crates/recursion/tests/basefold_opening.rs:191` uses a type annotation where CONTRIBUTING.md asks for turbofish — `random_scalars::<B128>(&mut rng, shape.n_vars)`.

jimpo (approved):

PR #2148, opened 2026-08-13
State each gate kind once, and dispatch through one checked table

tcoratger:
Makes each gate kind state its shape, its constraints and its instruction in one impl, and dispatches through one table instead of five hand-written ones.

**This is maintainability, not speed.** A function-pointer table costs an indirect call where a `match` costs a jump table. Measured below: within noise.

### The problem

Adding a gate today means editing **five** tables keyed on the same concept.

| table | arms | compiler catches an omission? |
|---|---:|---|
| `Opcode` | 22 | — |
| `Opcode::shape` | 22 | yes |
| `gates::constrain` | 22 | yes |
| `gates::emit_gate_bytecode` | 22 | yes |
| `EvalOpcode` + `from_byte` | 21 | **no** — guarded by a unit test |

Each gate contributed three loose free functions with no trait tying them together, so a gate's declared shape and the code destructuring against that shape were unrelated pieces of text.

That is where the arity assertions came from:

```
pub const fn shape() -> OpcodeShape { .. n_in: 2, n_out: 1 .. }

pub fn constrain(data: &GateData, builder: &mut ConstraintBuilder) {
    let [x, y] = inputs else { unreachable!() };   // restates n_in, three lines later
    let [z]    = outputs else { unreachable!() };  // restates n_out
```

**86 of them**, across 21 modules, every one a fact the shape above already stated.

And the six assertion gates needed an extra `PathSpec` parameter that no other gate takes, so the dispatch match had to know which six they were.

### The idea

One trait states all three facts, so they live or die together:

```
impl GateKind for Band {
    const SHAPE: OpcodeShape = OpcodeShape::new(2, 1);

    fn constrain(gate: GateParam<'_>, cb: &mut ConstraintBuilder) {
        let [x, y] = gate.in_wires();
        let [z] = gate.out_wires();
        cb.and(x, y, z);                    // x ^ y = z
    }

    fn emit(gate: GateParam<'_>, ctx: EmitCtx<'_>, bc: &mut BytecodeBuilder) {
        let [x, y] = gate.in_wires();
        let [z] = gate.out_wires();
        bc.emit_band(ctx.reg(z), ctx.reg(x), ctx.reg(y));
    }
}
```

`EmitCtx` carries the register resolver and the assertion path, so every kind has the same `emit` signature and the assertion special case disappears.

### How the arity assertions died

`&[Wire; Self::N_IN]` is not expressible on stable Rust — that needs generic const expressions.

So the group accessor is const-generic and converts once, with `N` inferred from the destructuring pattern:

```
pub fn in_wires<const N: usize>(&self) -> [Wire; N] {
    fixed(self.inputs, "input")
}
```

The call site destructures irrefutably and names no arity. 86 fallible patterns became 6 accessors, each with one `expect` that emission already checks against the gate's own shape.

### Per gate kind

- **before:** 3 free functions, 2 rows across 2 hand-written matches, ~4 `unreachable!()`
- **after:** 1 impl, 1 table row

→ 65 free functions become **2** (the module's two entry points), and 86 `unreachable!()` become **0**.

### The table is checked — the old ones were not

R7 called these "the two tables the compiler cannot currently check". The new one is checked three ways:

- each row carries the opcode it serves, so a row and its slot are not independent facts;
- the lookup reads that back under `debug_assert_eq!`;
- two tests pin that every row sits at the slot its opcode indexes, and that the hint gate is the one opcode with no row.

A misordered table now fails a test instead of silently constraining one gate as another.

### Scope

- **changed:** all 21 gate modules, the dispatch module, the opcode module, and `GateParam`
- **new:** `GateKind`, `EmitCtx`, the dispatch table, `OpcodeShape` const builders, 2 tests
- **deleted:** two 22-arm matches, `Opcode::is_const_shape`, the assertion special case
- **untouched:** every gate's constraint algebra and every emitted instruction — the shapes, operands and bytecode are the same
- the hint gate keeps its own branch: its shape lives in the registry rather than in a type, which is a separate change

The declarative macro for `Opcode` / `EvalOpcode` was considered and skipped. The constraint side is where the interesting code lives and it stays greppable; the bytecode correspondence is still test-guarded, exactly as before.

### Verification

| gate | result |
|---|---|
| `cargo test -p binius-frontend -p binius-circuits` | **596 passed, 0 failed** |
| `cargo check --workspace --all-targets` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo +nightly-2026-07-01 fmt --all --check` | clean |
| `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items` | clean |

The frontend suite includes the gate-fusion `insta` snapshots, which pin the exact rendered constraint system operand by operand. Those passing unchanged is the evidence that the conversion is behaviour-preserving. Gate counts also match the base exactly: 66,240 and 264,960.

### Benchmark — compile time, where the dispatch runs

`build()` calls both dispatch paths once per gate and reads a gate's shape many times per gate. Interleaved against the merge base, best of 7, three rounds, on an Apple M1 Pro.

| fixture | before | after |
|---|---|---|
| 24 keccak-f1600 permutations, 66,240 gates | 55.3 / 53.3 / 55.2 ms | 54.1 / 52.5 / 53.4 ms |
| 96 keccak-f1600 permutations, 264,960 gates | 231.8 / 230.1 / 232.0 ms | 225.8 / 229.3 / 227.9 ms |

The table costs nothing observable. The consistent 1–3% edge to "after" has a cause worth naming rather than claiming as a win: the old shape lookup ran `assert_eq!(self.is_const_shape(), dimensions.is_empty())` in release on every call, and shape is looked up many times per gate. That is now a `debug_assert!` on the trait's default.

### DIFF
--- a/crates/frontend/src/gates/assert_eq.rs
+++ b/crates/frontend/src/gates/assert_eq.rs
@@ -15,37 +15,27 @@
 
 use crate::{
 	eval_form::BytecodeBuilder,
-	gates::opcode::OpcodeShape,
-	ir::{GateData, GateParam, Wire, path::PathSpec},
+	gates::{EmitCtx, GateKind, OpcodeShape},
+	ir::GateParam,
 	lower::{ConstraintBuilder, expr},
 };
 
-pub const fn shape() -> OpcodeShape {
-	OpcodeShape {
-		const_in: &[],
-		n_in: 2,
-		n_out: 0,
-		n_aux: 0,
-		n_scratch: 0,
-		n_imm: 0,
+/// That two words are equal.
+pub struct AssertEq;
+
+impl GateKind for AssertEq {
+	const SHAPE: OpcodeShape = OpcodeShape::new(2, 0);
+
+	fn constrain(gate: GateParam<'_>, cb: &mut ConstraintBuilder) {
+		let [x, y] = gate.in_wires();
+
+		// x ⊕ y = 0
+		cb.zero(expr::xor2(x, y));
 	}
-}
 
-pub fn constrain(data: &GateData, builder: &mut ConstraintBuilder) {
-	let GateParam { inputs, .. } = data.gate_param();
-	let [x, y] = inputs else { unreachable!() };
+	fn emit(gate: GateParam<'_>, ctx: EmitCtx<'_>, bc: &mut BytecodeBuilder) {
+		let [x, y] = gate.in_wires();
 
-	// Constraint: x ⊕ y = 0
-	builder.zero(expr::xor2(*x, *y));
-}
-
-pub fn emit_eval_bytecode(
-	data: &GateData,
-	assertion_path: PathSpec,
-	builder: &mut BytecodeBuilder,
-	wire_to_reg: impl Fn(Wire) -> u32,
-) {
-	let GateParam { inputs, .. } = data.gate_param();
-	let [x, y] = inputs else { unreachable!() };
-	builder.emit_assert_eq(wire_to_reg(*x), wire_to_reg(*y), assertion_path.as_u32());
+		bc.emit_assert_eq(ctx.reg(x), ctx.reg(y), ctx.path().as_u32());
+	}
 }
--- a/crates/frontend/src/gates/assert_eq_cond.rs
+++ b/crates/frontend/src/gates/assert_eq_cond.rs
@@ -1,4 +1,5 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 //! Conditional equality assertion.
 //!
 //! Enforces `x = y` when the MSB-bool value of `cond` is true, and no constraint otherwise.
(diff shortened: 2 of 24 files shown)


benji-potamus (approved):
Approve.

Rebased onto current `main` (#2150) — applies clean. `cargo +nightly-2026-07-01 fmt --all --check`, `cargo clippy -p binius-frontend -p binius-circuits --all-targets -- -D warnings` and `cargo test -p binius-frontend -p binius-circuits` (600 passed, 0 failed) are all clean. I spot-checked the conversions that carry more than inputs and outputs — `icmp_eq` (2 consts, 1 aux, 1 scratch), `assert_eq_cond` (path through `EmitCtx`), `bxor_multi` (the dynamic shape), `shift` (2 immediates) — and the operands and their order are identical to the originals. With the insta snapshots passing unchanged on top of that, behaviour preservation is well evidenced.

### Findings

**The table maps every unregistered opcode to "hint", not just `Hint`** (non-blocking) — `crates/frontend/src/gates/mod.rs:136`. `GATE_KINDS.get(self as usize)?` returns `None` for any index past the table, so `Opcode::Hint` and "an opcode with no row" are indistinguishable. The three matches this replaces were exhaustive with no catch-all, so a new opcode used to be a compile error in `Opcode::shape`, `constrain` and `emit_gate_bytecode` alike; now it compiles.

I checked what that costs rather than assuming. Adding a variant after `Hint`, both new tests still pass — `the_hint_is_the_one_opcode_with_no_row` asserts `Opcode::Hint as usize == GATE_KINDS.len()`, which stays true when the new variant lands at 22 rather than displacing Hint. The gate then contributes no constraints and is emitted through `emit_hint`, which reads `immediates[0]` as a hint id. The only thing that stops it compiling today is an exhaustive match in an unrelated module, `crates/frontend/src/eval_form/const_eval.rs:29` — an incidental guard, and one a future `_` arm would remove.

Three lines fix it, and I verified them:

```rust
fn vtable(self) -> Option<&'static GateVTable> {
    if matches!(self, Opcode::Hint) {
        return None;
    }
    let row = &GATE_KINDS[self as usize];
    ...
```

An unregistered opcode then panics on the index instead of silently becoming a hint, and both existing tests still pass. Worth doing here specifically because the PR argues the new table is better checked than the old ones — with this, against these two it actually is.

PR #2152, opened 2026-08-13
BINIUS-487: widen the shift reduction to 36 variables and admit doubly shifted terms

benji-potamus:
Rebased onto main now that #2146 has merged. The lockstep change that turns the shift reduction into the one [binius.xyz#124](https://github.com/binius-zk/binius.xyz/pull/124) specifies ([BINIUS-487](https://linear.app/jimpo/issue/BINIUS-487)). Prover and verifier move together — the transcript changes. **This is where doubly shifted value indices first verify.**

## The sumcheck

One sumcheck over `36 + ℓ_words` variables, degree 2 in each, binding

```
O₂, S₂, O₁, S₁, J, K, I, Y
```

and closing on five factors:

```
δ_D~(r_î, r_i) · Z_cons~(r'_x, r_y, r_s₁, r_o₁, r_s₂, r_o₂)
              · shift-ind~(r_i, r_k, r_s₂, r_o₂)      the outer slot
              · shift-ind~(r_k, r_j, r_s₁, r_o₁)      the inner slot
              · w~(r_j, r_y)
```

Every variable occurs in exactly two of the five, which is what holds the degree at 2. The two indicators chain through the bits of the intermediate word — the outer carries the output bit down to `r_k`, the inner carries `r_k` down to the witness bit — and that word is never materialized; `K` is just one more sumcheck variable.

## Prover

Phase 1's new rounds landed in #2146. The rest is remapping:

| #124 phase | binds | here |
|---|---|---|
| 1 | `O₂, S₂` | `OuterShiftStage`, driven from `run_phase_1_sumcheck` |
| 2 | `O₁, S₁` | `Phase1SumcheckProver::Stage::Rows`, against `T[psi]` |
| 3 | `J` | `Stage::Bits` |
| 4 | `K` | `ShiftIndSumcheck` with weights `psi`, indicator at `(r_j, r_s₁, r_v₁)` |
| 5 | `I` | `ShiftIndSumcheck` again, weights `d`, indicator at `(r_k, r_s₂, r_v₂)` |
| 6 | `Y` | `prove_phase_2`, unchanged but for the outer weights |

**The row axis widens from 9 bits to 18**, keyed on the quadruple **outer-major** (`outer.index() << LOG_SHIFT_COUNT | inner.index()`), so the high bits are the ones the first rounds bind. `SparseShiftRows::fold` is index-space agnostic and needed no change; after the nine outer rounds the surviving indices *are* the inner shift indices, which is exactly what phase 2 wants. `DOUBLE_SHIFT_UNSUPPORTED` and its `debug_assert` are gone.

**The outer rounds are driven directly** rather than as a third `Stage`, which is the one place I departed from the issue's sketch. `prove_single` consumes its prover, and `psi` — the weights those rounds leave behind — has to outlive it, since phase 4 runs against them. `run_sumcheck` already drives its own sparse first round this way, so it's the existing idiom rather than a new one.

**No divisions.** Every scalar a later phase needs is a terminal fold: `G` from `g`, `psi` from `η`, the inner indicator's evaluation from `χ`. The handoff between the two `ShiftIndSumcheck` runs is `psi(r_k) = Σ_i d(i)·shift-ind(i, r_k, r_s₂, r_o₂)` — the same quantity from either side — which is why phase 5's claim is phase 4's `eval` with no inversion in between. Both handoffs are `debug_assert`ed in `prove`.

## Verifier

Small, because the structure was already there:

- `SHIFT_LOG_VARS` 21 → 36.
- The reversed point splits into eight runs instead of five; `VerifyOutput` grows `r_k` and the outer slot's `r_s`/`r_v`. `r_j` and `r_y` keep their meaning, so ring-switching and everything downstream is untouched.
- `evaluate_shift_inds` is called twice, each contracted against its own slot's variant tensor.
- `MonsterEvalFn` takes the outer slot's challenges and builds its table from them, in place of the identity-selecting constant. The slot-factorized `ShiftScalars` split was already in place, so this is one closure over both slots rather than a rewrite.
- Both `assert!(!svi.is_doubly_shifted())` are gone.

## `DenseShiftEncoding` ordering

`shift_indices` stops being ascending: the sequences sort lexicographically by `[inner, outer]` while the index is outer-major. Distinctness — the property `from_segments` actually needs — survives, so per your call on BINIUS-408 the doc comment is relaxed rather than the sort key changed.

## Costs

Prover: one extra `2^15` table build, nine extra sparse rounds at `O(2^6 · n_shift)`, six extra bit rounds at `O(2^6)`. `build_g` and phase 6 — the two that scale with the circuit — are untouched, so this should be lost in the noise.

Proof: 15 extra degree-2 round polynomials. At two coefficients each that's 30 field elements, ~480 bytes, plus 15 verifier rounds — paid by every circuit whether or not it uses double shifts, which is the tax you accepted on BINIUS-408. BINIUS-452 measures what it buys once the frontend emits pairs.

## Testing

The backend is fully exercisable here without the frontend, which still lowers at depth 1.

- **`create_double_shift_cs_with_witness`** in `prover/tests/shift.rs` is a hand-built system with five genuinely non-collapsible sequences — sign extension, a shift under a rotate, a byte masked off one end and back, the half-word family's own sign extension, and `sar` in the inner slot — plus one operand mixing unshifted, singly shifted and doubly shifted terms, which is what a compiler emitting pairs would produce. It asserts each sequence really is a `Composition::Pair` rather than trusting the comment, and it goes through `cs.validate()`, `cs.verify(&value_vec)` and the full prove/verify round trip.
- The verifier's monster fixture now draws a third of its terms doubly shifted, and `evaluate_monster_scales_by_the_outer_slot_weight` was rewritten around it: scaling the *whole* outer table scales the evaluation, which holds however the terms are spread, and the identity-selecting table is separately checked to reproduce what the singly shifted terms alone would give.
- `sparse_rounds_match_the_dense_prover` retargets to `Phase1SumcheckProver` directly, since `run_phase_1_sumcheck` now runs the outer rounds ahead of it.

Green on the rebase: `binius-prover` (60 + 15 + 1), `binius-m4-prover` (42), `binius-verifier` (10), and before it `binius-m4-verifier` (2) and `binius-recursion` (30 + 4). The recursion suite is the one worth naming — it builds a circuit mirroring the verifier, so 15 extra rounds was exactly the kind of thing that could have broken it. Workspace clippy clean over `--tests --benches --examples`.

### DIFF
--- a/crates/verifier/src/protocols/shift/verify.rs
+++ b/crates/verifier/src/protocols/shift/verify.rs
@@ -14,19 +14,22 @@ use binius_ip::{
 };
 use binius_math::{
 	BinarySubspace,
-	inner_product::inner_product_scalars,
 	line::extrapolate_line,
-	multilinear::eq::{
-		eq_ind_partial_eval_scalars, eq_ind_zero, eq_one_var, scaled_eq_ind_partial_eval,
-		scaled_eq_ind_partial_eval_scalars,
+	multilinear::{
+		eq::{
+			eq_ind_partial_eval_scalars, eq_ind_zero, eq_one_var, scaled_eq_ind_partial_eval,
+			scaled_eq_ind_partial_eval_scalars,
+		},
+		evaluate::evaluate_inplace_scalars,
 	},
 	univariate::{evaluate_univariate, lagrange_evals_scalars},
 };
 use getset::Getters;
 
 use super::{
-	BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, OperationEvalFn, SHIFT_COUNT, SHIFT_LOG_VARS,
-	ShiftScalars, ZERO_ARITY, encode_operation_input, error::Error, shift_ind::evaluate_shift_inds,
+	BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, LOG_SHIFT_COUNT, OperationEvalFn, SHIFT_COUNT,
+	SHIFT_LOG_VARS, ShiftScalars, ZERO_ARITY, encode_operation_input, error::Error,
+	shift_ind::evaluate_shift_inds,
 };
 
 /// Evaluates the bit-level multilinear extension of a word slice at the point `r_j ++ r_y`.
@@ -108,15 +111,25 @@ pub struct VerifyOutput<F> {
 	intmul_lambda: F,
 	/// Random coefficient for batching BMUL constraint evaluations.
 	binmul_lambda: F,
-	/// Challenge point for the bit index variables (length `Word::LOG_BITS`).
+	/// Challenge point for the witness bit index (length `Word::LOG_BITS`).
 	pub r_j: Vec<F>,
-	/// Challenge point for the shift-amount variables (length `Word::LOG_BITS`).
-	pub r_s: Vec<F>,
-	/// Challenge point for the shift-variant variables (length `LOG_SHIFT_VARIANT_COUNT`).
-	pub r_v: Vec<F>,
+	/// Challenge point for the inner shift's amount variables (length `Word::LOG_BITS`).
+	pub r_s_inner: Vec<F>,
+	/// Challenge point for the inner shift's variant variables (length
+	/// `LOG_SHIFT_VARIANT_COUNT`).
+	pub r_v_inner: Vec<F>,
+	/// Challenge point for the outer shift's amount variables (length `Word::LOG_BITS`).
+	pub r_s_outer: Vec<F>,
+	/// Challenge point for the outer shift's variant variables (length
+	/// `LOG_SHIFT_VARIANT_COUNT`).
+	pub r_v_outer: Vec<F>,
 	/// Challenge point for the word index variables (length `log_segment_words`).
 	pub r_y: Vec<F>,
-	/// Challenge point for the bit index the shift indicators read (length `Word::LOG_BITS`).
+	/// Challenge point for the bit index of the intermediate word, where the two shift indicators
+	/// meet (length `Word::LOG_BITS`).
+	pub r_k: Vec<F>,
+	/// Challenge point for the output bit index the oblong weights attach to (length
+	/// `Word::LOG_BITS`).
 	pub r_i: Vec<F>,
 	/// Challenge for the witness's segment selector variable.
 	pub r_segment: F,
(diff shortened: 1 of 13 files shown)


crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -308,37 +324,49 @@ where
...
-	// Two of the sumcheck's four factors depend on the bit index the middle rounds bound: the
-	// Lagrange weights of the univariate challenge, and the shift indicators interpolated over the
-	// variant axis. Both are evaluated at `r_i` alone.
+	// Three of the sumcheck's five factors are the verifier's to evaluate from the bit-index
+	// challenges alone: the Lagrange weights of the univariate challenge at `r_i`, and the two
+	// shift indicators, one per slot of a term's shift sequence. The indicators chain through the
+	// intermediate word — the outer one carries the output bit down to `r_k`, the inner one carries
+	// `r_k` down to the witness bit — which is what makes a sequence of two shifts one index entry.
 	let l_tilde = lagrange_evals_scalars(subspace, r_zhat_prime);
 	let l_tilde_eval = inner_product_scalars(l_tilde, eq_ind_partial_eval_scalars(r_i));
```
jimpo: Use evaluate_inplace_scalars for these three calls.

jimpo (changes requested):

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done — all three now go through `evaluate_inplace_scalars`.

```rust
let l_tilde_eval = evaluate_inplace_scalars(lagrange_evals_scalars(subspace, r_zhat_prime), r_i);
let outer_ind_eval = evaluate_inplace_scalars(&mut evaluate_shift_inds(r_i, r_k, r_s_outer)[..], r_v_outer);
let inner_ind_eval = evaluate_inplace_scalars(&mut evaluate_shift_inds(r_k, r_j, r_s_inner)[..], r_v_inner);
```

Both of the indicator ones were mine and I'd written them by pattern-matching the `l_tilde` line above rather than asking what it should be — so the third call is the one that shows the habit was already there. Folding in place is the better form on both counts: it drops the tensor allocation, and it halves the multiplications (`n·2^{n-1}` against `2^n` for the expansion plus `2^n` for the inner product).

The two indicator calls take `&mut …[..]` because `evaluate_shift_inds` returns `[E; 8]` and arrays don't implement `DerefMut`; `lagrange_evals_scalars` returns a `Vec`, so that one passes by value.

`inner_product_scalars` is no longer used in the file and its import is gone.

jimpo (approved):

PR #2153, opened 2026-08-13
BINIUS-419: skip the logUp* pushforward sumcheck for transparent tables

tcoratger:
@jimpo This one is a bit big because it took the approach of adding couple of new things; not sure if you had a better idea in mind for the approach.

Closes [BINIUS-419](https://linear.app/jimpo/issue/BINIUS-419/logup-special-handling-for-transparent-lookup-tables).

When a logUp* lookup table is transparent, the final pushforward sumcheck is unnecessary work: the two claims it reduces are already linear relations on the one committed pushforward, so the channel opening can carry both. Builds on [BINIUS-420](https://linear.app/jimpo/issue/BINIUS-420) (#2119), which let an oracle hold several relations.

### The problem

A logUp* reduction ends with two claims per table, and both read the **same** multilinear — that table's pushforward $Y_t$:

```
<Y_t, eq_{z_t}> = Y_t(z_t)      the fractional-addition leaf claim
<Y_t, T_t>      = e_t           the product claim
```

Today a batched degree-2 sumcheck over $\max_t m_t$ variables collapses the pair to one shared point.
It ends in one evaluation of $Y_t$ and one of $T_t$, and the caller must then check that table claim itself.

That is the right thing to do when $T_t$ is committed.
It is pure overhead when $T_t$ is **transparent** — a table the verifier evaluates directly from a succinct formula.

### The idea

Both claims are linear in $Y_t$, and $Y_t$ is a committed oracle.
So hand both to the channel as oracle relations and let the opening do the work:

```
relation 1:  <Y, eq_z> = Y(z)      transparent = the equality indicator at the leaf point
relation 2:  <Y, T>    = e         transparent = the table's own MLE
```

BaseFold folds an oracle's relations into one against a batching challenge $\lambda$, then opens once.
No sumcheck, no table claim, no extra FRI opening — $\lambda$ is a sampled challenge, so the second relation is free on the wire.

The saving is only available for a transparent $T_t$: relation 2 asks the verifier to weigh $Y_t$ against $T_t$ itself. A committed table still needs `verify_reduction`.

### Per table

For a table over $m$ variables, dropping the sumcheck removes:

- **before:** $m$ rounds $\times$ 2 round-polynomial coefficients, plus 2 evaluations sent at the end
- **after:** nothing on the wire; one more relation queued into the existing opening

→ $2m + 2$ field elements of proof, and $m$ rounds of folding two $2^m$-element columns of prover work.

### The change

Four paired entrypoints, one pair per layer. Each pair shares every step up to the pushforward checks.

| layer | committed table | transparent table |
|---|---|---|
| `binius-ip` | `verify_reduction` | `verify_reduction_transparent` |
| `binius-ip-prover` | `prove_reduction` / `prove` | `prove_reduction_transparent` / `prove_transparent` |
| `binius-iop` | `verify` | `verify_transparent` |
| `binius-iop-prover` | `prove` | `prove_transparent` |

The transparent function **is** the shared core; the committed one is a thin wrapper over it:

```diff
 pub fn verify_reduction(...) -> Result<LogupOutput<C::Elem>, Error> {
-    // ... the whole reduction, inline ...
-    let table_claims = /* per-table leaf + product claims */;
+    let LogupTransparentOutput { index_eval_point, tables } =
+        verify_reduction_transparent::<F, C>(gamma, tables, channel)?;
+
+    let table_claims = tables.iter().map(|table| TableClaim { ... });
     let Pushforward { .. } = verify_pushforward::<F, C>(table_claims, channel)?;
 }
```

So the two paths cannot drift, and the transcript prefix is bit-identical up to the point where they diverge.

The IOP verifier takes a `TransparentEvalFn` per table and queues both relations in a fixed order, which the prover mirrors:

```
verify_oracle_relation(Y, eq_z, Y(z))     // relation 1
verify_oracle_relation(Y, T,    e)        // relation 2
```

### Soundness

The logUp identity that pins $Y_t$ to the pushforward is checked exactly as before — the GKR, the leaf interpolation, the root denominator, the per-table challenges $c_t$ are all untouched.

What changes is where the product claim is enforced. `verify_reduction_transparent` returns it **unchecked**, because it never reads a table:

- a lookup claim is proved only once the caller opens `<Y_t, T_t> = e_t`;
- the channel's own $\lambda$-batching of an oracle's relations adds at most $\sum_i (k_i - 1)/|\mathbb{F}|$, with $k_i = 2$ here;
- $\lambda$ is drawn after every claim it combines is bound to the transcript, so no claim can be chosen as a function of it.

The contract is documented on the function and on the `product_claim` field, and pinned by two tests:

- `test_transparent_reduction_leaves_the_product_claim_unchecked` — a perturbed looker claim passes the IP reduction and comes back as a `product_claim` that is not the honest inner product;
- `test_basefold_transparent_rejects_wrong_eval_claim` — the real BaseFold opening rejects that same instance.

### Scope

- **new:** `verify_reduction_transparent` + `LogupTransparentOutput` / `LogupTransparentTableOutput` (`binius-ip`); `prove_reduction_transparent` + `prove_transparent` (`binius-ip-prover`); `verify_transparent` + `TransparentTableLookup` + `LogupTransparentProof` (`binius-iop`); `prove_transparent` + `LogupTransparentProof` (`binius-iop-prover`).
- **changed:** `verify_reduction` and `prove_reduction` are now wrappers over the transparent core — same signature, same transcript, same output.
- **changed:** intmul phase 5 (prover and verifier) switches to the transparent path.
- **untouched:** the pushforward sumcheck itself, on both sides. It is still the only path for a committed table.
- **not attempted:** mixing transparent and committed tables in one reduction. Each entrypoint takes one kind.

### intmul, the one transparent-table caller

intmul reads its fixed-base power table $i \mapsto g^i$ through logUp*, and that table is succinct — the verifier already evaluated its MLE as a product of selects. So phase 5 moves over, and the old `assert_zero(eval_power_table_mle(point) - table_claim)` disappears: the table's MLE is now the transparent of relation 2.

Its `LIMB_BITS` is 16, so the removed sumcheck is $16 \times 2 + 2 = 34$ field elements.

### Benchmark

M4 proof size for a circuit with IMUL constraints, measured end-to-end through the real FRI path:

| batch | before | after | saving |
|---|---|---|---|
| `log_instances = 0` | 200096 B | 199552 B | 544 B |
| `log_instances = 6` | 210752 B | 210208 B | 544 B |

Exactly the predicted $34 \times 16$ bytes, independent of batch size — the saving is per table, not per row. Prover-side, 16 rounds of folding two $2^{16}$-element columns also go away.

### Verification

- `cargo clippy` on the six touched crates, `--all-targets` — clean.
- nightly `cargo fmt --all` — applied.
- `binius-ip`, `binius-ip-prover`, `binius-iop-prover` logUp* tests: 34 pass, including new transparent round-trips over the naive channel **and** real BaseFold (multi-table, multi-looker, unequal-length, one-variable-table shapes), plus the two contract tests above.
- `binius-prover` intmul: 7 pass. `binius-m4-prover` IMUL end-to-end: 1 pass.
- logUp* and intmul sets re-run with `--features rayon`.

One thing worth a reviewer's eye: the recursive-verifier channels (`IronSpartanBuilderChannel`, `zk_wrapped_channel`) materialize one inout wire per relation, so each pushforward oracle now allocates one more. Both sides do it symmetrically per call, and `binius-verifier` compiles for those channels — but no test in the repo drives a recursive circuit with IMUL constraints, so that path is as unexercised as it was before this change.

### DIFF
--- a/crates/iop-prover/src/logup_star.rs
+++ b/crates/iop-prover/src/logup_star.rs
@@ -12,6 +12,9 @@
 //! unchanged. The pushforwards are the only oracles this protocol introduces, per
 //! [Soukhanov25, Section 3].
 //!
+//! [`prove_transparent`] is the variant for tables the verifier evaluates itself.
+//! It opens each `Y` against both `eq_z` and the table, and so needs no pushforward sumcheck.
+//!
 //! [Soukhanov25]: <https://eprint.iacr.org/2025/946>
 
 use binius_compute::Allocator;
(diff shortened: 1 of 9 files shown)


benji-potamus (approved):
Approve.

Traced the split through all four layers (`binius-ip` → `binius-ip-prover` → `binius-iop` → `binius-iop-prover`) plus the intmul integration. `verify_reduction`/`prove_reduction` are now thin wrappers over the `_transparent` core followed by `verify_pushforward`/`prove_pushforward`, which are byte-for-byte untouched — so the committed path is structurally guaranteed to produce the same transcript as before, and the two paths can't drift out of sync. The transparent path's relation order (Y against eq_z, then Y against T) matches exactly between `verify_transparent` and `prove_transparent`, and the "unchecked product claim" contract is pinned by a real negative test at the right layer: `test_transparent_reduction_leaves_the_product_claim_unchecked` shows the bare IP reduction lets a wrong claim through, and `test_basefold_transparent_rejects_wrong_eval_claim` shows the actual BaseFold oracle opening catches it. That's the soundness argument demonstrated, not just asserted. The intmul integration is a clean swap — dropping the old `assert_zero` against the table's MLE in favor of passing it as the second relation's transparent closure.

One thing worth a follow-up look, per your own note in the PR body: the recursive-verifier channels (`IronSpartanBuilderChannel`, `zk_wrapped_channel`) now materialize one more inout wire per relation on every pushforward oracle, but nothing in the repo drives a recursive circuit over IMUL constraints, so that path is genuinely unexercised — worth a test once recursion has an IMUL-bearing circuit to point at, not a blocker here.

jimpo (approved):

PR #2161, opened 2026-08-13
Reimplement hash_based_sig against the XMSS reference

benji-potamus:
The old `hash_based_sig` did not verify the XMSS scheme it was named for. This replaces it with the scheme leanVM-b's `xmss` crate implements, keeping BLAKE3 as the hash under the tweakable hash — per your call, the tweakable hash is the only intended difference.

## What was wrong

Audited against [`leanEthereum/leanVM-b`'s `crates/xmss`](https://github.com/leanEthereum/leanVM-b/tree/main/crates/xmss) (at `b672d106`). BLAKE3-vs-BLAKE2s was the smallest of the differences — every concrete parameter and the shape of the tweakable hash differed too:

| | reference | old local |
|---|---|---|
| digest | 16 B | 32 B |
| public param | 16 B | 18 B |
| randomness | 24 B | 32 B |
| chains `v` | 42 | 72 / 36 |
| Winternitz `w` | 3 bits, chain length 8 | 2 or 4 bits |
| target sum | 195 | 119 / 297 |
| encoding | 3-bit digits, 2 leftover bits ground to zero | power-of-two widths only, no leftover concept |
| tree height | 32 | caller's choice |

Its tweakable hash was also two unrelated constructions — a bare compression with the chain value in the chaining value for chain steps, a keyed hash with the tweak as the key everywhere else — neither of which is a hash of the byte string the scheme defines. A signature produced by either implementation would not verify under the other.

## What this does

Seven commits, each building on its own so the sequence bisects:

1. **Delete the legacy module** and its two consumers.
2. **The tweakable hash** — BLAKE3 of `tweak(16) ‖ pp(16) ‖ payload`, truncated to 128 bits, with the reference's `[type | sub_position | index | zeros]` tweak.
3. **WOTS** — the target-sum encoding, the chain walk, the public-key hash.
4. **XMSS** — the 32-level authentication path, verification, and a signer for witness generation.
5. **Aggregation** — independent signers, one message, one epoch.
6. **The example and the m4-prover test**, rebuilt on the new API.
7. **Two-lane chain steps.**

BLAKE3 and BLAKE2s both compress 64 bytes at a time, so swapping the primitive changes no byte string and no compression count: a native verification is the same constant 144 compressions the reference documents.

## Two things worth your attention

**The circuit walks 294 chain steps, not 99.** A step's tweak is a circuit constant, so it cannot be indexed by the digit — every chain evaluates all 7 of its steps and takes the hashed value only past its digit. The target sum therefore buys encoding validity here rather than verifier work. That is inherent to the constant-tweak approach, not a regression; the old code did the same.

**Commit 7 is worth ~23% and not the 50% the lane count suggests.** Measured over the example at three signers:

```
AND constraints   195,600 -> 151,104   -23%
committed trace   375,244 -> 275,358   -27%
gates             685,542 -> 476,508   -31%
```

A lone `blake3_compress` already uses both lanes, splitting its own seven rounds across them for four packed rounds, so on rounds alone a pair only trades eight for seven. Most of the saving is that a lone compression hints its mid-round state and then constrains that hint word for word, and a pair has no split to pin.

## Scope I did not take on

- **No keygen or signing.** `xmss::generate_signature` builds only the authentication path: siblings are drawn at random and the root is whatever climbing from the leaf produces, which is what the reference's secret key does for out-of-range nodes. Enough to generate witnesses, and deliberately no secret key to sign with twice. The reference's PRF-derived keys and sqrt-memory secret key are prover-side machinery a circuits crate never exercises.
- **`xmss::xmss_verify` / `wots::wots_encode` keep the reference's names**, which duplicates the module name against CONTRIBUTING's namespacing guidance. Sibling modules in this crate already do the same (`blake3::blake3_fixed`, `sha256::sha256_fixed`), and matching the reference helps a future audit. Say the word and I'll rename to `xmss::verify` / `wots::encode`.

## Gates

- `cargo test -p binius-circuits hash_based_sig` — 26 passed, plus 1 doctest
- `cargo test --release -p binius-m4-prover --test prove_hash_based_sig -- --ignored` — passes, with the chain steps as `Blake3Compress2x` chip instances
- `cargo run --example hashsign -- prove` — proves and verifies end to end, 250 KiB at three signers
- fmt clean; clippy clean on every crate this touches

One pre-existing clippy failure is untouched by this branch: `crates/examples/src/cli.rs:598` trips `needless_borrow` under the local toolchain, on a line with no changes here.


crates/circuits/src/hash_based_sig/mod.rs
```diff
@@ -1,9 +1,124 @@
...
+//! xmss::xmss_verify(&public_key, &message, &signature, epoch).unwrap();
+//!
+//! // The same signature at any other epoch is not a signature.
+//! assert!(xmss::xmss_verify(&public_key, &message, &signature, epoch + 1).is_err());
+//! ```
+//!
+//! [`xmss::circuit_xmss_verify`] is the in-circuit form of that check, and
+//! [`aggregate::circuit_xmss_multisig`] runs it for several signers over one message.
+//!
+//! CREDIT: <https://github.com/leanEthereum/leanVM-b> (XMSS construction).
```
jimpo, line 58: Add `Copyright (c) 2026 leanEthereum` copyright notices to all files that use code from that codebase

crates/circuits/src/hash_based_sig/wots.rs
```diff
@@ -0,0 +1,430 @@
...
+	// what would otherwise let a word carry a slack term the digits do not account for.
+	for (k, &word) in digest.iter().enumerate() {
+		builder.assert_zero(format!("encoding_leftover_bit[{k}]"), builder.shr(word, 63));
+	}
+
+	let mask = builder.add_constant_64((1 << W) - 1);
+	let digits: [Wire; V] = std::array::from_fn(|i| {
+		let word = digest[i / DIGITS_PER_WORD];
+		let shift = (W * (i % DIGITS_PER_WORD)) as u32;
+		builder.band(builder.shr(word, shift), mask)
```
jimpo: Don't mask with band. Do shl followed by shr to select out bits.

jimpo (changes requested):
There are several places where you use band for masking. Prefer to use clear_high_bits util or shl -> shr to select bits from the middle of a word.

crates/circuits/src/hash_based_sig/mod.rs
benji-potamus: Added in `893f7235` to the four files carrying ported code: `mod.rs` (parameters and module structure), `hashing.rs` (`make_tweak` and the tweak hash), `wots.rs` (the encoding and chain walking), `xmss.rs` (the Merkle node and verification order).

I left `aggregate.rs` out on the grounds that the reference has no aggregation in its `xmss` crate — it aggregates in the VM guest — so the multisig loop isn't derived from it. Say the word if you'd rather it carry the notice too.

crates/circuits/src/hash_based_sig/wots.rs
benji-potamus: Done in `fffc8c4f`. There were three `band` masks; all are gone and there are none left in the module.

- The digit extraction here is now `shl` to lift the digit's `W` bits to the top of the word, then `shr` to drop them back to the bottom.
- The tweak's byte 8 (`hashing.rs`) does the same to select the index's top byte.
- The tweak's low three bytes needed no selection at all: shifting the index up to byte 5 places them and carries everything above them off the top of the word in one instruction, so that `band` was pure overhead.

jimpo (approved):

PR #2165, opened 2026-08-14
BINIUS-497: spend only the chain hashes a verifier walks

benji-potamus:
Closes [BINIUS-497](https://linear.app/jimpo/issue/BINIUS-497/xmss-more-hash-efficient-wots-verification).

A chain's tail is `CHAIN_LENGTH - 1 - digit` hashes, and the target sum fixes the total across all chains at `NUM_CHAIN_HASHES` = 99 however the digits fall. The circuit gave every chain room for its longest possible tail — `V * (CHAIN_LENGTH - 1)` = 294 hashes, two thirds of them selected away — because a step's tweak was a circuit constant and could not be indexed by the digit. The tweak now takes its sub-position as a wire, so the tails concatenate into one hinted list of exactly 99 and the constraints pin it.

Two commits: the flat list, then @jimpo's suggestion from the issue thread to bound each lookup's range.

## Measured

Over the example at three signers, against `main`:

```
                   main       flat list   + windowed lookup
AND constraints    132,030 -> 61,197   -> 62,085     -53%
ZERO constraints   137,343 -> 68,549   -> 69,140     -50%
committed trace    271,958 -> 192,577  -> 173,935    -36%
gates              490,569 -> 300,828  -> 281,886    -43%
```

The padded sections fall a power of two: **AND from 2^18 to 2^16, the committed trace from 2^19 to 2^18.** Wall-clock is in [this comment](https://github.com/binius-zk/binius64/pull/2165#issuecomment-5288377115) — **prove 311 → 161 ms (−48%)**, verify 44.4 → 27.6 ms, proof 250 → 215 KiB, three runs a side with matching `RUSTFLAGS`.

The thing that makes this pay is not obvious, and I nearly wrote the idea off over it: **a multiplexer costs no AND constraints at all** — selects lower to committed words, not AND. Measured marginals, which is what the design was chosen on:

| | AND | committed |
|---|---|---|
| paired chain-step core (2 hashes) | 243 | 511 |
| multiplexer, 99 entries × 4 wires | **0** | 399 |

So the variable layout is paid for in committed words rather than hashes, and one table lookup per entry is far cheaper than the 97 paired cores it removes.

**Worth a decision before the next change lands here**: AND now sits at 62,085 of 2^16, with 3,451 spare. The windowed lookup spends ~900 AND to buy 19k committed words — the right trade today, and also what ate most of the remaining headroom. Anything much added to this circuit tips that section into 2^17 and doubles it.

## What pins the list

Each entry carries its input digest, its chain and its step. Its **output is the hash**, not a hinted value, so nothing has to check it; and its **input is hinted** rather than carried from the entry before it, so the 99 hashes are independent and still pair two to a core.

- `chain` never decreases, so a chain's entries are one contiguous run.
- Within a run the step advances by one and each input is the previous output.
- A run starts at its chain's digit, from that chain's signature tip, and ends at `CHAIN_LENGTH - 2` with its output as that chain's end.

A run's length is therefore exactly the tail length. Since the list is exactly `NUM_CHAIN_HASHES` long — which the target sum makes the sum of those lengths — **no chain that owes hashes can be missing a run**, or the entries would not add up. That counting step is the one part worth a careful read.

The windowed lookup adds that each entry can only belong to a bounded span of chains, since a chain holds at most `CHAIN_LENGTH - 1` entries. That span is **asserted**, not left implied: it does follow from monotonicity and the run lengths, but the mux reads only the low bits of its selector, so a chain outside the window would alias onto another chain's row — and the constraints that would catch that are the ones the derivation leans on.

## Where I deviated from the issue, and why

The approach is the issue's; four details in the sketch needed repair, and one mechanism I changed.

1. **Empty chains.** A digit of `CHAIN_LENGTH - 1` owes no hashes, and with the mean digit at 4.64 that is roughly a fifth of chains — not an edge case. "Mux out the last item of each chain" has no such item for them; their end is the tip the signature already gave. A test fails loudly if a fixture ever stops covering both that path and a walked chain.
2. **`hash_items[0].chain == 0`** is false whenever chain 0 is empty. Dropped; the per-run start check subsumes it.
3. **`item_1.chain == item_0.chain + 1`** is false when an intervening chain is empty — the jump exceeds 1. Replaced by monotonicity.
4. **Nothing bound a run's first input to the signature tip.** Without it the circuit still could not be forged (a prover would have to invert the chain to reach the committed public key), but `chain_tips` would be decorative for every non-empty chain and a witness disagreeing with the hinted inputs would go undetected. Bound.
5. **Boundaries are detected locally** — a run starts where the `chain` field changes — rather than via cumulative sums indexing the 99-entry list. No index clamping for empty chains, and cheaper: the lookup it needs is one table of chains per entry rather than a 99-entry lookup per chain. Happy to swap in the cumulative-sum version if you prefer it as written.

## Gates

- `cargo test -p binius-circuits hash_based_sig` — 27 passed, plus 1 doctest
- `cargo test --release -p binius-m4-prover --test prove_hash_based_sig -- --ignored` — passes, chain steps still landing as `Blake3Compress2x` chip instances
- snapshot re-blessed; fmt and clippy clean; all 8 CI checks green

### DIFF
--- a/crates/circuits/src/hash_based_sig/wots.rs
+++ b/crates/circuits/src/hash_based_sig/wots.rs
@@ -331,6 +331,21 @@ pub fn circuit_recover_public_key(
 		.collect();
 	let table_rows = table.iter().map(|row| row.as_slice()).collect::<Vec<_>>();
 
+	// Which chains an entry can possibly belong to. A chain contributes at most
+	// `CHAIN_LENGTH - 1` entries, of which at most `CHAIN_LENGTH - 2` can fall on one side of any
+	// entry of its own. So of the `k` entries before this one, all but `CHAIN_LENGTH - 2` need
+	// chains strictly below its own, and likewise above for the entries after it. That confines
+	// each entry to about two thirds of the chains, and only that window has to be muxed over.
+	let window = |k: usize| -> (usize, usize) {
+		let per_chain = CHAIN_LENGTH - 1;
+		let own_side = CHAIN_LENGTH - 2;
+		let before = k.saturating_sub(own_side).div_ceil(per_chain);
+		let after = (NUM_CHAIN_HASHES - 1 - k)
+			.saturating_sub(own_side)
+			.div_ceil(per_chain);
+		(before, V - 1 - after)
+	};
+
 	// The hashes. Every entry's input is hinted rather than carried from the entry before it, so
 	// they are independent and pair two to a core.
 	let sub_position = |k: usize| builder.bxor(builder.shl(chain_of(k), W as u32), step_of(k));
@@ -362,7 +377,6 @@ pub fn circuit_recover_public_key(
 
 	let one = builder.add_constant_64(1);
 	let last_step = builder.add_constant_64((CHAIN_LENGTH - 2) as u64);
-	let all_chains = builder.add_constant_64(V as u64);
 	let never = builder.add_constant(Word::ZERO);
 	let always = builder.add_constant(Word::ALL_ONE);
 
@@ -370,7 +384,16 @@ pub fn circuit_recover_public_key(
 		let b = builder.subcircuit(format!("chain_hash[{k}]"));
 		let (chain, step, input) = (chain_of(k), step_of(k), input_of(k));
 
-		let row = multi_wire_multiplex(&b, &table_rows, chain);
+		// The window is asserted rather than merely implied: the mux reads only the low bits of
+		// its selector, so a chain outside the window would alias onto a row that is not its own.
+		let (lowest, highest) = window(k);
+		let lowest_wire = b.add_constant_64(lowest as u64);
+		b.assert_true("chain_at_least_window", b.icmp_ule(lowest_wire, chain));
+		b.assert_true("chain_at_most_window", b.icmp_ule(chain, b.add_constant_64(highest as u64)));
+
+		let zero = b.add_constant_64(0);
+		let offset = b.isub_bin_bout(chain, lowest_wire, zero).0;
+		let row = multi_wire_multiplex(&b, &table_rows[lowest..=highest], offset);
 		let tip: [Wire; DIGEST_WIRES] = std::array::from_fn(|w| row[w]);
 		let digit = row[DIGEST_WIRES];
 		let end: [Wire; DIGEST_WIRES] = std::array::from_fn(|w| row[DIGEST_WIRES + 1 + w]);
@@ -423,13 +446,6 @@ pub fn circuit_recover_public_key(
 			);
 		}
 	}
-
-	// The table is indexed by a hinted chain, so the chain has to name a real one. Monotonicity
-	// carries the bound to every entry.
-	builder.assert_true(
-		"chain_in_range",
-		builder.icmp_ult(chain_of(NUM_CHAIN_HASHES - 1), all_chains),
-	);
 
 	// A chain at the last digit owes no hashes and so has no entry to close it: its end is the
 	// tip the signature already gave.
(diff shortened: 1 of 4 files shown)


crates/circuits/src/hash_based_sig/wots.rs
```diff
@@ -191,50 +194,274 @@ pub fn circuit_wots_encode(
...
+					b.select(continues, input[w], previous),
+					previous,
+				);
+			}
+		}
+
+		// Opening a chain: start at the digit, from the signature's tip for that chain.
+		let starts = b.bnot(continues);
+		b.assert_eq("starts_at_digit", b.select(starts, step, digit), digit);
+		for w in 0..DIGEST_WIRES {
```
jimpo, line 406: We don't need to assert that it starts at the tip. The tip is just a hint. Verification just requires that there exists a preimage that begins a chain of the right length.

crates/circuits/src/hash_based_sig/wots.rs
```diff
@@ -191,50 +194,274 @@ pub fn circuit_wots_encode(
...
+	// A chain at the last digit owes no hashes and so has no entry to close it: its end is the
+	// tip the signature already gave.
+	let empty_digit = builder.add_constant_64((CHAIN_LENGTH - 1) as u64);
+	for i in 0..V {
+		let empty = builder.icmp_eq(digits[i], empty_digit);
+		for w in 0..DIGEST_WIRES {
+			builder.assert_eq(
+				format!("empty_chain_end[{i}][{w}]"),
+				builder.select(empty, chain_ends[i][w], chain_tips[i][w]),
+				chain_tips[i][w],
```
jimpo: No assertion against chain_tips is necessary. If the chain is empty, then no preimage must be revealed.

jimpo (changes requested):
This change to the multiplexing logic doesn't seem good to me.

The ticket specifies that outside of the combined hash chain, we look up into it once per chain. This implements the opposite, that each entry in the combined chain looks up into a table, of length V.

In a new commit, flip it and gives the comparison in # of gates. Instead of the step going from `encoding[i]` to `CHAIN_LEN - 1`, the step should store the amount of hashing so far in that chain. So in each link between two adjacent items, either: 1) the chain remains the same or, 2) the new chain is strictly greater and the step resets to 0. We additionally constrain that step on the very first entry is 0. Then, outside of the hash chain construction, every i in 0..V, we lookup the chain end (the offsets into the windows for lookup can even be hinted, not cumulative sums). The chain end should have step as expected, chain as expected, and pull out the output digest. If the chain is empty, we select on that condition and instead use chain_tips directly.

crates/circuits/src/hash_based_sig/wots.rs
benji-potamus: Removed. You're right — a run's first input is free now, and verification only needs some preimage that walks a chain of the right length onto the committed public key.

Worth noting what it cost: this assertion was the only reason each entry had to know its chain's tip, so dropping it removes two of the five columns the per-entry lookup was reading. That is most of why the flip to one lookup per chain comes out ahead rather than level.

`chain_tips` is now read only for chains that owe no hashes.

crates/circuits/src/hash_based_sig/wots.rs
benji-potamus: Gone — it's a `select` now, not an assertion: an empty chain's end is just the value the signature revealed, and nothing about its lookup is asserted.

That also removed the last reason for `chain_ends` to be hinted. The per-chain lookup produces them, so the hint dropped 84 output words.

jimpo (approved):

PR #2167, opened 2026-08-14
fix(spartan-verifier): pack the statement into wires, not circuit constants

diegode:
An honest ZK proof fails verification once the constraint system's public segment passes a
power of two. For a circuit built from `hash_based_sig::aggregate::circuit_xmss_multisig`, a
statement of **512** public words verifies and one of **513** does not, the two circuits
differing only by one inert padding word — an inout constrained to the constant zero that
nothing else reads.

```
ZKVerifier::verify → Err(OuterVerification(IPChannel(InvalidAssert)))
```

`Prover`/`Verifier` (non-ZK) accept both, so this is the ZK composition, not the core IOP.

The first commit adds the two tests, the second fixes the bug, the third pins the rule it turned
on. Bisected to `156b0bd6` ("BINIUS-466: ring-switch the public segment's evaluation", #2114).

## The bug

`IronSpartanBuilderChannel::pack_words` packed the words it was handed into *constants*:

```rust
fn pack_words(&mut self, words: &[Word]) -> Vec<Self::Elem> {
    // The words are concrete, so the packed elements are settled while building.
    pack_words_concrete::<F, Self::Elem>(words)
}
```

But those words are the statement, and the circuit they are being packed into is built once —
`ZKVerifier::setup` symbolically executes the inner verifier against a placeholder of all-zero
inout words — and then reused for every real statement. So the circuit was built around zeros
that no real statement shares.

Zeros are not inert in that circuit. `CircuitElem`'s multiplication folds `x * 0` to a constant,
so wherever a placeholder zero made a product vanish that a real statement does not, the
symbolic run allocated one wire fewer than verification then did. Nothing catches that:
`InstanceGenerator::write_public` drops wires the layout has no slot for, which is also how
pruned intermediates are handled, so the run continues with every later derived value written
into the previous one's slot, and the outer proof fails.

Crossing a power of two is what exposes it. `verify_public_eval` pads the packed public segment
up to `1 << log_public_elems` with zeros, so at 513 words the fold that evaluates it pairs each
of the statement's own elements against a padding zero: `delta = 0 - w`, which is zero for the
placeholder and non-zero for a real statement. At exactly 512 words there is no padding, every
element is paired with another real element, and the deltas are non-zero either way — which is
why that case passes, and why the threshold moves with the circuit rather than sitting at a
fixed word count.

Two probes confirm the path, both on the 513-word circuit:

- seeding `setup`'s symbolic run with the real statement instead of zeros makes the unfixed
  prover and verifier agree — the proof verifies;
- the circuits the zero statement and the real one induce differ, by exactly the wires the
  vanished products would have allocated.

## The fix

The packed elements enter as inout wires. `ZKWrappedVerifierChannel` and `ReplayChannel` fill
them with the concrete packing, which both sides compute from words they already hold, so
nothing is added to the transcript and no new value has to be agreed on.

The point is not that the zeros are gone but that the statement is: the wrapper circuit no
longer holds any of it, so it is now *identical* whatever statement the symbolic run was handed.
Building it against an all-zero statement and against a real one gives the same constants, the
same mul constraints and the same wire counts, which was not true before — and is what makes the
placeholder in `setup` sound rather than lucky.

It costs nothing. On the aggregate circuit above:

| | constants | inout | derived | mul constraints | `log_public` |
|---|---|---|---|---|---|
| before | 431 | 1,019 | 594,499 | 396,630 | 20 |
| after | 255 | 1,276 | 594,499 | 396,630 | 20 |

176 constants traded for 257 inout wires — the packed segment has to live somewhere — with the
derived wires, the constraints and the padded public size all unchanged, so the outer proof is
the same size.

One nearby hazard of the same kind, left alone here since nothing wrapped today reaches it:
`subset_sum` and `select` shape the circuit from a concrete word, and the builder channel's
`sample_bits` returns `Word::ZERO`. A verifier that reached either of those with sampled bits
would build its circuit around that zero and run it against another word.

## Testing

- The two tests in `crates/prover/tests/prove_verify.rs` fail (513) / pass (512) on the first
  commit, and both pass on the second.
- `test_pack_words_allocates_inout_wires`, in the third commit, states the rule where the
  packing is: the elements come back as wires, for words that include a zero.
- `cargo test --release -p binius-spartan-verifier -p binius-spartan-prover -p binius-verifier
  -p binius-prover -p binius-ip` — the crates that depend on the changed code, plus the
  end-to-end tests: 122 passed, 0 failed.
- `cargo +nightly-2026-07-01 fmt`: clean.
- Full workspace suite and clippy: 

## Bisect

Bisected over `eaa3b6d0..029b44d8` with a different circuit (a blind-signature aggregate of our
own, which crosses **256** public words at 22 signers):

| rev | 22-signer circuit |
|---|---|
| `eaa3b6d0` — "Bind the shift indicator bit index with a sumcheck" (#2113) | verifies |
| `156b0bd6` — "BINIUS-466: ring-switch the public segment's evaluation" (#2114) | `OuterVerification(IPChannel(InvalidAssert))` |

Every later rev tested (`1c8a1e68`, `029b44d8`, `7d457ae1`, `a7ef8201`) fails. #2114 did not
introduce the packing into constants — it introduced the zero-padded public fold that reaches
the statement's own words with a multiplication.

## Why it went unnoticed

`test_prove_verify_public_wider_than_hidden` covers a wide-public circuit at 301 public words
and passes: with no ZK wrapper in that test, nothing is built against a placeholder. The
aggregate tests in `hash_based_sig::aggregate` check only `ConstraintSystem::verify`, never a
prove/verify round trip.

Circuits whose public segment happens to land on a power of two, or whose statement words are
paired only with other statement words, verify fine — which is most of what the suite covers.

### DIFF
--- a/crates/ip/src/channel.rs
+++ b/crates/ip/src/channel.rs
@@ -186,6 +186,14 @@ pub trait WordIPVerifierChannel<F: Field>: IPVerifierChannel<F> {
 	fn pack_words(&mut self, words: &[Self::Word]) -> Vec<Self::Elem>;
 }
 
+/// The number of elements [`WordIPVerifierChannel::pack_words`] returns for `n_words` words.
+///
+/// A channel that packs the words into wires rather than computing them needs the count on its
+/// own, so the layout stays in one place.
+pub fn n_packed_elems<F: BinaryField>(n_words: usize) -> usize {
+	n_words.div_ceil(F::N_BITS / Word::BITS)
+}
+
 /// [`WordIPVerifierChannel::pack_words`] over concrete words, for channels carrying [`Word`].
 ///
 /// A set bit contributes the basis element at its position in the packed layout, which is what
--- a/crates/ip/src/channel.rs
+++ b/crates/ip/src/channel.rs
@@ -190,7 +190,7 @@ pub trait WordIPVerifierChannel<F: Field>: IPVerifierChannel<F> {
 ///
 /// A channel that packs the words into wires rather than computing them needs the count on its
 /// own, so the layout stays in one place.
-pub fn n_packed_elems<F: BinaryField>(n_words: usize) -> usize {
+pub const fn n_packed_elems<F: BinaryField>(n_words: usize) -> usize {
 	n_words.div_ceil(F::N_BITS / Word::BITS)
 }
 
--- a/crates/prover/tests/prove_verify.rs
+++ b/crates/prover/tests/prove_verify.rs
@@ -671,24 +671,37 @@ fn xmss_aggregate_circuit(num_signers: usize, n_pad: usize) -> (ConstraintSystem
 	(cs, witness)
 }
 
+/// The padding that takes a `num_signers` circuit's public segment to exactly `n_public` words.
+///
+/// The unpadded width is whatever the XMSS circuit currently needs, and it moves whenever that
+/// circuit does, so the tests below state the width they mean to exercise and derive the padding
+/// from it rather than hard-coding a count that drifts out from under them. The probe pads by one
+/// rather than none so that the zero constant the padding introduces is already inside the
+/// measured base.
+fn pad_for_public_words(num_signers: usize, n_public: usize) -> usize {
+	let (cs, _) = xmss_aggregate_circuit(num_signers, 1);
+	let base = cs.n_public_words(InoutSegment::Public) - 1;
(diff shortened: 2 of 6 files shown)


jimpo (approved):
Amazing!

jimpo (approved):

PR #2168, opened 2026-08-14
BINIUS-496: fold the Merkle tree in subtrees, not one layer at a time

tcoratger:
Closes [BINIUS-496](https://linear.app/jimpo/issue/BINIUS-496/merkle-tree-fold-the-tree-in-subtrees-not-one-layer-at-a-time).

Groups several tree levels into one rayon round instead of one, so a synchronization point lands every few levels instead of every level.

### The problem

`build_from_iterator` folds a Merkle tree one layer at a time: 17 parallel loops for a 2^17-leaf tree, each with its own barrier.
The upper layers are wide enough to split across every core, but the lower layers are not.
Past a few thousand nodes, a layer is smaller than one task's floor, so it runs on a single thread — no matter how many cores sit idle.

Measured on a 2^17-leaf, 16xB128-per-leaf, SHA-256 tree on 4 cores:

| | 1 thread | 2 threads | 4 threads |
|---|---|---|---|
| node fold, all layers | 2.470 ms | 1.550 ms | 1.365 ms |
| widest layer alone | 1.235 ms | 0.620 ms | 0.354 ms |
| one 2^12-output layer | 77.06 µs | 77.08 µs | 77.07 µs |

The widest layer parallelizes fine (3.49x at 4 threads).
The narrow layer is identical at every thread count — it never splits at all.

### The idea

Fix a subtree depth of 6 levels.
Instead of folding one full layer at a time, fold every 6 consecutive levels as `2^(n - 6)` independent subtrees, each folded to its single root entirely inside one rayon task.

```
one barrier per layer:        L0 -> L1 -> L2 -> ... -> L16 -> L17     (17 barriers)
one barrier per 6 levels:     L0 ---------------------> L6 -> ... -> L17   (3 barriers)
```

Barriers drop from `n` to `ceil(n / 6)`.
Each task now carries `2^6 - 1 = 63` compressions instead of one, comfortably above the per-task floor even where a single narrow layer used to fall below it.
A subtree's intermediates also stay small enough (126 digests) to fit in cache for its whole fold, instead of round-tripping through a full-width layer between rounds.

The per-level compression call inside each subtree still goes through the hash suite's existing parallel/SIMD-batched path (e.g. SHA-256's 4-way interleaved kernel), so this only removes the barrier — it does not fall back to a slower single-lane compression.

### The change

- `crates/hash/src/binary_merkle_tree.rs`: `from_leaves` now loops over rounds of up to 6 levels instead of one level, carving out each round's layers and handing them to a new `fold_subtrees` helper.
- `fold_subtrees` partitions every layer in a round into `num_subtrees` equal stripes, transposes those per-layer chunks into one stripe set per subtree, and folds each subtree inside a single rayon task.
- `min_subtrees_per_task` scales the existing per-task time budget by a subtree's real cost (`2^depth - 1` compressions), not a single compression.
- `HashSuite::ParCompression` gained a `Sync` bound, needed because one instance is now shared across the new per-round rayon closure. Every existing `HashSuite` impl already satisfied it.

### Soundness

Purely a scheduling change to the prover's tree construction.
The sequence of compressions and their inputs are unchanged, so every layer's digests — and the root — are bit-identical to the one-layer-at-a-time construction.

### Scope

- **changed:** `crates/hash/src/binary_merkle_tree.rs` only
- **left untouched:** the `HashSuite`/`ParallelPseudoCompression` trait surface used by every hash family (SHA-256, BLAKE3), the tree's public layout (`layer`/`branch`/`root`), and every caller (FRI, basefold, logup*)

### Verification

- `cargo clippy -p binius-hash -p binius-iop-prover --all-targets` (with and without `--features rayon`) — clean
- `cargo test -p binius-hash` and `-p binius-iop-prover` (with and without `--features rayon`) — all pass, including FRI/basefold end-to-end tests that build much larger trees
- added a test that builds trees spanning one round, one round plus a partial round, and two full rounds, and checks every layer against a naive single-layer-at-a-time reference fold — the existing tests never built a tree taller than depth 5, so they never exercised more than one round

### Benchmark

Minimal before/after using the existing `slow/merkle_tree/SHA-256` group (2^17 leaves, 16xB128 per leaf), same machine, same run:

| variant | before | after | speedup |
|---|---|---|---|
| 16xB128 leaf, 128b, global alloc | 5.53 ms | 4.26 ms | 1.30x |
| 16xB128 leaf, 128b, pooled | 4.99 ms | 4.62 ms | 1.08x |
| 16xB128 leaf, 1x128b, global alloc | 5.25 ms | 4.73 ms | 1.11x |
| 16xB128 leaf, 1x128b, pooled | 4.92 ms | 4.04 ms | 1.22x |

Smaller than the issue's back-of-envelope 2.1x estimate — this machine has fewer cores than the issue's 4-core reference, and this implementation keeps the SIMD-batched compression path inside each subtree rather than falling back to single-lane compression, trading some of the theoretical barrier-elimination gain for that throughput.

### DIFF
--- a/crates/hash/src/binary_merkle_tree.rs
+++ b/crates/hash/src/binary_merkle_tree.rs
@@ -1,13 +1,17 @@
 // Copyright 2024-2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::fmt;
+use std::{fmt, mem::MaybeUninit};
 
 use binius_compute::{Allocator, GlobalAllocator, VecLike};
 use binius_field::Field;
 use binius_utils::{
 	checked_arithmetics::checked_log_2,
-	rayon::{prelude::*, slice::ParallelSlice},
+	rayon::{
+		prelude::*,
+		slice::ParallelSlice,
+		task_size::{WorkPerItem, min_len_for_work},
+	},
 };
 use digest::{
 	Digest, FixedOutputReset, Output, OutputSizeUser,
@@ -35,8 +39,11 @@ pub trait HashSuite {
 	/// Parallel counterpart of [`Self::LeafHash`] used during proving.
 	type ParLeafHash: ParallelDigest<Digest = Self::LeafHash> + Default;
 	/// Parallel counterpart of [`Self::Compression`] used during proving.
+	///
+	/// `Sync` because one instance is shared across threads while folding the tree.
 	type ParCompression: ParallelPseudoCompression<Output<Self::LeafHash>, 2, Compression = Self::Compression>
-		+ Default;
+		+ Default
+		+ Sync;
 }
 
 /// Reason a Merkle tree operation rejects its arguments.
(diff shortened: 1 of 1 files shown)


crates/hash/src/binary_merkle_tree.rs
```diff
@@ -233,6 +260,92 @@ impl<N: ArraySize, A: Allocator> BinaryMerkleTree<Array<u8, N>, A> {
...
+			let chunk_len = layer.len() / num_subtrees;
+			layer.chunks_mut(chunk_len)
+		})
+		.collect();
+	let subtree_layers: Vec<Vec<&mut [MaybeUninit<D>]>> = (0..num_subtrees)
+		.map(|_| {
+			layer_chunks
+				.iter_mut()
+				.map(|chunks| chunks.next().expect("every layer has num_subtrees stripes"))
+				.collect()
```
jimpo: I'd like to avoid these collects

jimpo (changes requested):
I had in mind a strategy that would only synchronize in one place, after the leaf hashers. Then build the entire rest of the tree recursively as subtrees, but without synchronizing. (Basically build top down instead of bottom up). The tricky part is getting references to the nodes from the shared mutable buffer. But I think we could just accept unsafe indexing against a mut pointer.

crates/hash/src/binary_merkle_tree.rs
tcoratger: Fixed

jimpo (approved):

PR #2208, opened 2026-08-18
perf(prover): build the key collection as a counting sort, in parallel

pauldelucia:
Re-expressed on top of the `key_collection` restructure (#2242, #2240, #2231): the counting sort now sits behind `KeyCollection::build`, the machinery stays private to `builder.rs`, and the one-pass path (`BuilderKeyLists` + `KeySegment::build`) moved wholesale into the equivalence tests' module, where it serves as the byte-identity oracle; the production files carry only the counting sort. Algorithm and figures are unchanged from the first round of this PR.

**What.** `Prover::setup` spends most of its time building the key collection when the constraint system is large and set up once per proof. This rewrites `KeyCollection::build` as a stable counting sort by word, behind the same signature: one pass counts each word's references and records the shift sequences in use, one pass scatters the references in walk order into a flat array indexed by word, and the per-word grouping into keys runs over disjoint word ranges in parallel, with no allocation per key. The final concatenation consumes each chunk as it is appended, so the segment never holds two copies of the reference array.

**Numbers** (measured on our circuit in the first round; unchanged). About 3.6M AND, 78k ZERO and 154k BMUL constraints, 88M operand terms over 4.1M hidden words, 29.7M keys, on an M2: the build goes from 9.5 s to 1.6 s, of which the two serial passes are about 0.6 s and the rest is the parallel grouping. `size_of::<ConstraintIndex>()` is 8, so the duplicate reference array the consuming concatenation avoids is on the order of 700 MB there.

**Testing** (byte-for-byte `SerializeBytes` equality against the one-pass oracle, both `InoutSegment`s throughout):
- the module's existing shifted system;
- random systems with all four operations, multi-term operands, two-slot shifts and a few hot words, seeds 1/2/3 at 64/1,000/20,000 AND constraints;
- the same shape spanning 5 and 74 chunks, so cross-chunk rebasing and sequence-unioning run off the edge cases;
- the empty system, plus key-count and range-shape asserts.
The module's other tests run unchanged and now exercise the counting sort through `KeyCollection::build`.

**Costs.** The constraint system is walked twice (count, then scatter) instead of once, and the old one-pass builder stays in-tree, inside the test module, as the oracle — which is most of the diff's size. Release builds compile none of it.

### Why
We build a fresh Binius64 circuit for almost every Ethereum block (proving the block's receipts trie is fully indexed), so `Prover::setup` runs once per proof for us rather than once per circuit. For circuits that are set up once and proved many times the difference is a one-off.

Happy to adjust the chunk size, the oracle's placement, or the test shapes to your preference.

### Checks
`cargo +nightly-2026-07-01 fmt -- --check`; `cargo clippy -p binius-prover --all-targets --all-features -- -D warnings`; `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-prover --document-private-items --no-deps`; `cargo test -p binius-prover --lib` and `--test shift`; `tools/check_copyright_notice.py`.


crates/prover/src/protocols/shift/key_collection/key_segment.rs
```diff
@@ -17,9 +17,10 @@ use bytes::{Buf, BufMut};
 use itertools::izip;
 use tracing::instrument;
 
+#[cfg(test)]
+use super::builder::BuilderKey;
```
tcoratger: Stuffs like this when used only for the tests should be in the test module directly I think in order not to pollute the core of the file.

crates/prover/src/protocols/shift/key_collection/key_segment.rs
```diff
@@ -177,6 +178,9 @@ impl KeySegment {
 	}
 
 	/// Builds the segment's keys from the builder keys lists of its words.
+	///
+	/// The reference layout the counting-sort builder is checked against.
+	#[cfg(test)]
```
tcoratger: Stuffs like this when used only for the tests should be in the test module directly I think in order not to pollute the core of the file.

crates/prover/src/protocols/shift/key_collection/builder.rs
```diff
@@ -110,3 +140,366 @@ impl BuilderKeyLists {
...
+
+#[inline]
+const fn operation_from_code(code: u8) -> Operation {
+	match code {
+		0 => Operation::Zero,
+		1 => Operation::BitwiseAnd,
+		2 => Operation::IntegerMul,
+		_ => Operation::BinMul,
+	}
+}
```
tcoratger: Should be in Operation impl directly no?

crates/prover/src/protocols/shift/key_collection/builder.rs
```diff
@@ -20,8 +45,13 @@ pub(super) struct BuilderKey {
...
 
 /// One builder key list per word of the constraint system, indexed by word position.
+///
+/// Kept under `cfg(test)`: [`KeyCollection::build`] runs the counting sort below, and this
+/// one-pass path remains as the byte-identity oracle its tests compare against.
+#[cfg(test)]
 pub(super) struct BuilderKeyLists(Vec<Vec<BuilderKey>>);
 
+#[cfg(test)]
 impl BuilderKeyLists {
```
tcoratger: Stuffs like this when used only for the tests should be in the test module directly I think in order not to pollute the core of the file.

PR #2209, opened 2026-08-18
BINIUS-507: Merge same-round oracle commitments into one

tcoratger:
Closes [BINIUS-507](https://linear.app/jimpo/issue/BINIUS-507/iop-channel-wrapper-for-merging-same-round-commitments).

Adds a channel decorator that commits a round's oracles as one combined oracle instead of one commitment each.

### The problem

IOP verifier and prover channels commit oracles one at a time.
Several oracles often get committed in the same interaction round, before the same challenge is sampled.
Each one still pays for its own physical commitment (a Merkle tree, under BaseFold/FRI), even though they're about to be opened together right after.

### The idea

Buffer a round's oracles and commit them together, as one combined oracle, instead.

Sort largest to smallest before concatenating.
That ordering guarantees every oracle lands on a boundary aligned to its own size, which is exactly what lets it be recovered exactly at opening time.

On the verifier side, recovering one oracle's claim against the combined oracle means extending its transparent-polynomial check with an equality indicator that selects its own block.
On the prover side, which holds real buffers rather than formulas, the same idea is just zero-padded placement of the same data.

```
extended_transparent(x) = transparent(low bits of x) * is_this_block(high bits of x)
```

Flushing happens exactly when the next challenge is sampled — never later.
That matches how a real Fiat-Shamir transcript must work: a challenge can only be derived once every commitment it depends on has been absorbed.

Everything merged into one round must share the same witness-dependence.
A physical commitment is masked as a whole, never partly, so zero-knowledge and non-zero-knowledge oracles cannot share one round.

A round of a single oracle needs no combining at all — it is forwarded unchanged, at zero cost.

### The change

- `crates/iop/src/channel/merge.rs` — `MergeVerifierChannel`, the verifier-side decorator.
- `crates/iop-prover/src/channel/merge.rs` — `MergeProverChannel`, its prover-side mirror.

Both wrap any existing channel implementation, so they compose with the crate's naive test channel or the real BaseFold/FRI channel without changes to either.

### Scope

- **new:** the two decorator modules above, each with its own unit and round-trip tests
- **changed:** one `pub mod merge;` line added to each crate's `channel/mod.rs`
- **left untouched:** no existing channel, protocol, or call site — this is purely additive, nothing yet wires it into a real protocol driver

### Verification

- `cargo clippy -p binius-iop -p binius-iop-prover --all-targets --all-features` — clean
- nightly `cargo fmt --all -- --check` — clean
- `cargo test -p binius-iop -p binius-iop-prover` (default features and `--features rayon`) — 66 tests pass, including a proptest sweep over random round shapes

### Benchmark

Not shipped as code in this PR — a one-off script drove the real BaseFold/FRI channel with and without the decorator, over the same witness data, and verified both proofs:

| | commitments | proof size |
|---|---|---|
| without merging | 6 | 41840 bytes |
| with merging | 2 | 28320 bytes |

2 rounds of 3 oracles each, merging cuts proof size by 32.3% on this shape.

<details>
<summary>Reproduce this: drop into <code>crates/iop-prover/examples/</code> and run with <code>cargo run --release --example merge_channel_proof_size -p binius-iop-prover</code></summary>

```rust

//! Compares real BaseFold/FRI proof size with and without same-round oracle merging.
//!
//! Both scenarios commit the exact same witness data over the exact same round shape.
//! The only difference is whether each round's oracles get one Merkle commitment each, or
//! one combined commitment per round.
//! Both proofs are also verified, so the size numbers below are for proofs that actually
//! check out.
//!
//! Run with: cargo run --release --example merge_channel_proof_size -p binius-iop-prover

use binius_compute::GlobalAllocator;
use binius_field::{BinaryField128bGhash, PackedBinaryGhash1x128b};
use binius_hash::{StdDigest, StdHashSuite};
use binius_iop::{
	basefold::compiler::BaseFoldVerifierCompiler,
	channel::{IOPVerifierChannel, OracleSpec, merge::MergeVerifierChannel},
	fri::MinProofSizeStrategy,
	merkle_tree::BinaryMerkleTreeScheme,
};
use binius_iop_prover::{
	basefold::compiler::BaseFoldProverCompiler,
	channel::{IOPProverChannel, merge::MergeProverChannel},
};
use binius_ip::channel::IPVerifierChannel;
use binius_ip_prover::channel::IPProverChannel;
use binius_math::{
	FieldBuffer,
	inner_product::inner_product_buffers,
	multilinear::eq::eq_ind_partial_eval,
	ntt::{NeighborsLastSingleThread, domain_context::GaoMateerOnTheFly},
};
use binius_transcript::{ProverTranscript, VerifierTranscript, fiat_shamir::HasherChallenger};
use binius_utils::checked_arithmetics::log2_ceil_usize;

type F = BinaryField128bGhash;
type P = PackedBinaryGhash1x128b;
type StdChallenger = HasherChallenger<StdDigest>;

// A rate-1/2 code with 32 test queries, the same parameters the crate's own BaseFold tests
// use.
const LOG_INV_RATE: usize = 1;
const SECURITY_BITS: usize = 32;

const fn n_test_queries() -> usize {
	SECURITY_BITS.div_ceil(LOG_INV_RATE)
}

fn make_ntt(log_domain_size: usize) -> NeighborsLastSingleThread<GaoMateerOnTheFly<F>> {
	NeighborsLastSingleThread::new(GaoMateerOnTheFly::generate(log_domain_size))
}

fn make_merkle_scheme() -> BinaryMerkleTreeScheme<F, StdHashSuite> {
	BinaryMerkleTreeScheme::new()
}

/// A random witness for one oracle, together with a transparent polynomial and the claim
/// their inner product produces.
fn generate_oracle_data(rng: &mut StdRng, n_vars: usize) -> (FieldBuffer<P>, FieldBuffer<P>, F) {
	let buffer = random_field_buffer::<P>(&mut *rng, n_vars);
	let point = random_scalars::<F>(&mut *rng, n_vars);
	let transparent = eq_ind_partial_eval::<P>(&point);
	let claim = inner_product_buffers(&buffer, &transparent);
	(buffer, transparent, claim)
}

/// Commits, proves, and verifies every oracle in `rounds` as its own commitment.
///
/// Returns the finalized proof size in bytes.
fn run_unmerged(rounds: &[&[usize]]) -> usize {
	let sizes: Vec<usize> = rounds
		.iter()
		.flat_map(|round| round.iter().copied())
		.collect();
	let data: Vec<(FieldBuffer<P>, FieldBuffer<P>, F)> = sizes
		.iter()
		.map(|&n| generate_oracle_data(&mut rng, n))
		.collect();
	let oracle_specs: Vec<OracleSpec> = sizes.iter().map(|&n| OracleSpec::new_zk(n)).collect();

	let verifier_compiler = BaseFoldVerifierCompiler::new(
		&make_merkle_scheme(),
		oracle_specs,
		LOG_INV_RATE,
		n_test_queries(),
		&MinProofSizeStrategy,
	);
	let ntt = make_ntt(verifier_compiler.max_log_domain_size());
	let prover_compiler =
		BaseFoldProverCompiler::<P, _>::from_verifier_compiler(&verifier_compiler, ntt);

	// Prover side: one commitment per oracle, in arrival order.
	let mut prover_transcript = ProverTranscript::new(StdChallenger::default());
	let mut prover_channel = prover_compiler
		.create_channel_from_transcript::<StdHashSuite, StdChallenger, _, _>(
			&mut prover_transcript,
			StdRng::seed_from_u64(1),
			GlobalAllocator,
		);

	let oracles: Vec<_> = data
		.iter()
		.map(|(buffer, _, _)| prover_channel.send_oracle(buffer.to_ref()))
		.collect();
	for (&oracle, (_, transparent, claim)) in oracles.iter().zip(&data) {
		prover_channel.prove_oracle_relation(oracle, transparent.clone(), *claim);
	}
	for (oracle, (buffer, _, _)) in oracles.iter().copied().zip(data.iter().cloned()) {
		prover_channel.finalize_oracle(oracle, buffer);
	}
	prover_channel.finish();
	let proof = prover_transcript.finalize();
	let proof_size = proof.len();

	// Verifier side: replay the exact same commitments and relations.
	let mut verifier_transcript = VerifierTranscript::new(StdChallenger::default(), proof);
	let mut verifier_channel = verifier_compiler
		.create_channel_from_transcript::<StdHashSuite, StdChallenger, _>(&mut verifier_transcript);

	let v_oracles: Vec<_> = sizes
		.iter()
		.map(|&n| verifier_channel.recv_oracle(n, true).unwrap())
		.collect();
	for (oracle, (_, transparent, claim)) in v_oracles.into_iter().zip(data) {
		verifier_channel
			.verify_oracle_relation(
				oracle,
				Box::new(move |point: &[F]| {
					let eq = eq_ind_partial_eval::<P>(point);
					inner_product_buffers(&transparent, &eq)
				}),
				claim,
			)
			.unwrap();
	}
	verifier_channel
		.finish()
		.expect("unmerged proof must verify");

	proof_size
}

/// Commits, proves, and verifies every round in `rounds` as one combined commitment.
///
/// Returns the finalized proof size in bytes.
fn run_merged(rounds: &[&[usize]]) -> usize {
	let fine_sizes: Vec<usize> = rounds
		.iter()
		.flat_map(|round| round.iter().copied())
		.collect();
	let fine_specs: Vec<OracleSpec> = fine_sizes.iter().map(|&n| OracleSpec::new_zk(n)).collect();
	let data: Vec<(FieldBuffer<P>, FieldBuffer<P>, F)> = fine_sizes
		.iter()
		.map(|&n| generate_oracle_data(&mut rng, n))
		.collect();

	// One coarse spec per round, sized to fit that round's total.
	let coarse_specs: Vec<OracleSpec> = rounds
		.iter()
		.map(|sizes| {
			let total: usize = sizes.iter().map(|&n| 1usize << n).sum();
			OracleSpec::new_zk(log2_ceil_usize(total))
		})
		.collect();

	let verifier_compiler = BaseFoldVerifierCompiler::new(
		&make_merkle_scheme(),
		coarse_specs,
		LOG_INV_RATE,
		n_test_queries(),
		&MinProofSizeStrategy,
	);
	let ntt = make_ntt(verifier_compiler.max_log_domain_size());
	let prover_compiler =
		BaseFoldProverCompiler::<P, _>::from_verifier_compiler(&verifier_compiler, ntt);

	// Prover side: every round is sent, then sampled once, so the merging decorator commits
	// each round as a single oracle underneath.
	let mut prover_transcript = ProverTranscript::new(StdChallenger::default());
	let base_channel = prover_compiler
		.create_channel_from_transcript::<StdHashSuite, StdChallenger, _, _>(
			&mut prover_transcript,
			StdRng::seed_from_u64(1),
			GlobalAllocator,
		);
	let mut merge_channel = MergeProverChannel::new(base_channel, &fine_specs, GlobalAllocator);

	let mut oracles = Vec::new();
	let mut index = 0;
	for sizes in rounds {
		for _ in *sizes {
			let (buffer, _, _) = &data[index];
			oracles.push(merge_channel.send_oracle(buffer.to_ref()));
			index += 1;
		}
		merge_channel.sample();
	}
	for (&oracle, (_, transparent, claim)) in oracles.iter().zip(&data) {
		merge_channel.prove_oracle_relation(oracle, transparent.clone(), *claim);
	}
	for (oracle, (buffer, _, _)) in oracles.iter().copied().zip(data.iter().cloned()) {
		merge_channel.finalize_oracle(oracle, buffer);
	}
	merge_channel.into_inner().finish();
	let proof = prover_transcript.finalize();
	let proof_size = proof.len();

	// Verifier side: the same round boundaries, wrapped the same way.
	let mut verifier_transcript = VerifierTranscript::new(StdChallenger::default(), proof);
	let base_verifier_channel = verifier_compiler
		.create_channel_from_transcript::<StdHashSuite, StdChallenger, _>(&mut verifier_transcript);
	let mut merge_verifier_channel = MergeVerifierChannel::new(base_verifier_channel, &fine_specs);

	let mut v_oracles = Vec::new();
	for sizes in rounds {
		for &n in *sizes {
			v_oracles.push(merge_verifier_channel.recv_oracle(n, true).unwrap());
		}
		merge_verifier_channel.sample();
	}
	for (oracle, (_, transparent, claim)) in v_oracles.into_iter().zip(data) {
		merge_verifier_channel
			.verify_oracle_relation(
				oracle,
				Box::new(move |point: &[F]| {
					let eq = eq_ind_partial_eval::<P>(point);
					inner_product_buffers(&transparent, &eq)
				}),
				claim,
			)
			.unwrap();
	}
	merge_verifier_channel
		.into_inner()
		.expect("merged proof must verify")
		.finish()
		.expect("merged proof must verify");

	proof_size
}

fn main() {
	// Two rounds, three oracles each: sizes are log2 variable counts.
	let rounds: &[&[usize]] = &[&[8, 7, 7], &[9, 6, 6]];
	let n_oracles: usize = rounds.iter().map(|round| round.len()).sum();
	let n_rounds = rounds.len();

	let unmerged_bytes = run_unmerged(rounds);
	let merged_bytes = run_merged(rounds);

	let reduction = 100.0 * (1.0 - merged_bytes as f64 / unmerged_bytes as f64);

	println!("Same-round oracle merging: proof-size comparison");
	println!("Rounds (log2 oracle sizes): {rounds:?}");
	println!();
	println!("Without merging: {n_oracles} commitments -> {unmerged_bytes} bytes");
	println!("With merging:    {n_rounds} commitments -> {merged_bytes} bytes");
	println!("Reduction: {reduction:.1}% (both proofs verified)");
}
```

</details>

### Follow-ups

Rebased on `main`. The port to the current APIs is part of this branch: `eq_ind` / `eq_ind_partial_eval` are now `Hypercube::One` methods, `FieldBuffer::to_ref` is `as_view`, `clone_from_slice` is `from_view_in`, the indexed mutable chunk accessor is gone (`place_block` writes scalars directly), and `BinaryField128bGhash` is `Ghash128b`. The benchmark script in the collapsed section above predates those renames.

Left for follow-up PRs, per review:

- an `OracleSchedule` carrying the specs *and* their grouping into interaction rounds, so both decorators can map every placement up front instead of discovering rounds as they go
- surfacing a flush error rather than dropping it in `sample`
- flushing in `sample_bits`
- deriving the combined `is_witness_dependent` as the OR of the round's oracles instead of asserting they agree
- the prover-side buffer work: drop `zeros_in` in favour of `extend_from_slice`, drop the `send_oracle` copy, and batch the openings at this layer instead of proving each constituent against the combined oracle

### DIFF
--- a/crates/iop-prover/src/channel/merge.rs
+++ b/crates/iop-prover/src/channel/merge.rs
@@ -1,0 +1,744 @@
+// Copyright 2026 The Binius Developers
+
+//! A channel decorator that merges oracles committed within the same interaction round.
+//!
+//! This is the prover-side half of a matching pair.
+//! Whatever it commits here must be received by a matching verifier-side decorator.
+
+use std::{cmp::Reverse, ops::DerefMut};
+
+use binius_compute::Allocator;
+use binius_field::{Field, PackedField};
+use binius_iop::channel::OracleSpec;
+use binius_ip_prover::channel::{IPProverChannel, WordIPProverChannel};
+use binius_math::{FieldBuffer, FieldSlice, FieldVec};
+use binius_utils::checked_arithmetics::log2_ceil_usize;
+
+use crate::channel::IOPProverChannel;
+
+/// A handle to an oracle sent through the merging decorator.
+#[derive(Debug, Clone, Copy)]
+pub struct MergeOracle {
+	index: usize,
+}
+
+/// Records where one constituent oracle lives inside its round's combined oracle.
+#[derive(Clone, Copy)]
+struct Mapping {
+	/// Which round's combined oracle this constituent belongs to.
+	group: usize,
+
+	/// The constituent's position among its round-mates, as a block index.
+	///
+	/// Its own `2^n` scalars begin at scalar `block_index * 2^n`.
+	block_index: usize,
+
+	/// The base-2 logarithm of the combined oracle's length.
+	combined_log_len: usize,
+}
+
+/// Tracks one oracle from the moment it is sent until its round is flushed.
+struct Record {
+	/// The base-2 logarithm of the oracle's own length.
+	log_msg_len: usize,
+
+	/// Where this oracle ends up inside its round's combined oracle.
+	///
+	/// Filled in once that round is flushed.
+	mapping: Option<Mapping>,
+}
+
+/// The combined oracle actually committed for one round.
+///
+/// Kept alive until every constituent oracle is finalized.
+struct Group<P: PackedField, A: Allocator, Oracle> {
+	/// The handle the underlying channel returned for the combined oracle.
+	outer: Oracle,
+
+	/// The combined witness data.
+	///
+	/// Kept until every constituent has been handed back.
+	/// Then forwarded to the underlying channel exactly once, in their place.
+	buffer: Option<FieldVec<P, A>>,
+
+	/// How many constituent oracles make up this round.
+	n_members: usize,
+
+	/// How many of those constituents have been handed back so far.
+	n_finalized: usize,
+}
+
+/// A prover channel decorator that merges one round's oracles into one combined oracle.
+///
+/// # Overview
+///
+/// An interaction round is the run of oracles sent between two challenge samples.
+///
+/// Committing each oracle separately costs one commitment per oracle.
+/// One Merkle tree per oracle, for example.
+///
+/// This decorator buffers a round's oracles instead.
+/// It commits them together as one larger oracle.
+/// That cuts the cost to one commitment per round.
+///
+/// A round's oracles are sorted from largest to smallest, then laid out end to end.
+/// That ordering makes every oracle's position exact.
+///
+/// Every earlier oracle is at least as large as the current one.
+/// So their combined space is a whole multiple of the current oracle's size.
+///
+/// A round of a single oracle needs no combining.
+/// It is forwarded unchanged, at zero cost.
+///
+/// Every oracle in a round must declare the same witness-dependence.
+///
+/// A commitment is masked as a whole, never partly.
+/// So mixing witness-carrying and structural oracles in one round is not supported.
+///
+/// # Timing
+///
+/// A round's oracles are committed the moment a challenge is sampled.
+/// Not the moment they arrive.
+///
+/// A real Fiat-Shamir transcript works the same way.
+/// A challenge can only be derived after its commitments are absorbed.
+/// So committing cannot wait past the sample that follows a round.
+///
+/// # Opening
+///
+/// A verifier only ever holds a formula for a transparent polynomial.
+///
+/// This side holds the actual coefficients instead.
+/// A constituent's own transparent polynomial becomes one for the combined oracle.
+/// Write its values into a zero buffer, at the constituent's own position.
+/// Every other position stays zero.
+///
+/// That placement is the same polynomial a verifier reaches by formula.
+/// One side holds it as explicit values.
+/// The other evaluates it on demand.
+pub struct MergeProverChannel<'a, P, A, C>
+where
+	P: PackedField,
+	A: Allocator,
+	C: IOPProverChannel<P, A>,
+{
+	/// The underlying channel every oracle, challenge, and opening passes through.
+	inner: C,
+
+	/// Fine-grained specs, one per oracle this channel's caller will send.
+	///
+	/// Not the coarser, one-per-round specs the underlying channel uses.
+	oracle_specs: &'a [OracleSpec],
+
+	/// The allocator this channel draws its combined and padded buffers from.
+	alloc: A,
+
+	/// Buffers received for the current round, not yet sent to the underlying channel.
+	pending: Vec<FieldVec<P, A>>,
+
+	/// Every oracle sent so far, in arrival order.
+	records: Vec<Record>,
+
+	/// Every round committed so far, in commit order.
+	groups: Vec<Group<P, A, C::Oracle>>,
+}
+
+impl<'a, P, A, C> MergeProverChannel<'a, P, A, C>
+where
+	P: PackedField,
+	A: Allocator,
+	C: IOPProverChannel<P, A>,
+{
+	/// Creates a new merging prover channel over an underlying channel.
+	///
+	/// # Arguments
+	///
+	/// * `inner` — the channel every combined oracle is committed to, already configured with the
+	///   coarser, one-per-round spec list this decorator will produce.
+	/// * `oracle_specs` — the fine-grained specs for every oracle this channel's caller will pass
+	///   through, in arrival order.
+	/// * `alloc` — where this channel draws its combined and padded buffers from.
+	pub const fn new(inner: C, oracle_specs: &'a [OracleSpec], alloc: A) -> Self {
+		Self {
+			inner,
+			oracle_specs,
+			alloc,
+			pending: Vec::new(),
+			records: Vec::new(),
+			groups: Vec::new(),
+		}
+	}
+
+	/// Commits the current round's queued buffers as one combined oracle.
+	///
+	/// Does nothing if every sent oracle is already committed.
+	fn flush(&mut self) {
+		// Nothing new has arrived since the last flush.
+		if self.pending.is_empty() {
+			return;
+		}
+		let first_index = self.records.len() - self.pending.len();
+
+		// Order this round largest to smallest.
+		//
+		// This lets every position be a whole number of block sizes.
+		// That holds once every earlier buffer is at least as large.
+		let mut order: Vec<usize> = (0..self.pending.len()).collect();
+		order.sort_by_key(|&k| Reverse(self.pending[k].log_len()));
+
+		// Size the combined oracle to fit every buffer end to end.
+		let total_len: usize = order
+			.iter()
+			.map(|&k| 1usize << self.pending[k].log_len())
+			.sum();
+		let combined_log_len = log2_ceil_usize(total_len);
+
+		// Lay every buffer into the combined data.
+		// Record where each one landed.
+		//
+		// Every step adds a whole multiple of the next block's size.
+		// So each offset divides evenly by that buffer's own size.
+		let mut combined = FieldBuffer::zeros_in(&self.alloc, combined_log_len);
+		let mut block_indices = vec![0usize; self.pending.len()];
+		let mut offset = 0usize;
+		for &k in &order {
+			let n_k = self.pending[k].log_len();
+			let block_index = offset >> n_k;
+			block_indices[k] = block_index;
+			place_block(&mut combined, self.pending[k].as_view(), block_index);
+			offset += 1 << n_k;
+		}
+
+		// Commit the whole round as one oracle on the underlying channel.
+		//
+		// The combined data stays alive in its own round record.
+		// It is still needed once every constituent is handed back.
+		let outer = self.inner.send_oracle(combined.as_view());
+		let group_index = self.groups.len();
+		let n_members = self.pending.len();
+		self.groups.push(Group {
+			outer,
+			buffer: Some(combined),
+			n_members,
+			n_finalized: 0,
+		});
+
+		// Record where each constituent oracle landed.
+		for k in 0..n_members {
+			self.records[first_index + k].mapping = Some(Mapping {
+				group: group_index,
+				block_index: block_indices[k],
+				combined_log_len,
+			});
+		}
+
+		self.pending.clear();
+	}
+
+	/// Commits any oracles still queued and returns the underlying channel.
+	///
+	/// # Panics
+	///
+	/// Panics if any declared oracle has not yet been sent.
+	pub fn into_inner(mut self) -> C {
+		self.flush();
+		let n_remaining = self.oracle_specs.len() - self.records.len();
+		assert!(n_remaining == 0, "into_inner called but {n_remaining} oracle specs remaining",);
+		self.inner
+	}
+}
+
+/// Writes one buffer into a fixed-size block of another.
+///
+/// Every other position of the destination is left untouched.
+///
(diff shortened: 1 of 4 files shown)


crates/iop/src/channel/merge.rs
```diff
@@ -0,0 +1,550 @@
...
+	fn subset_sum(&mut self, elems: &[Self::Elem], word: &Self::Word) -> Self::Elem {
+		self.inner.subset_sum(elems, word)
+	}
+
+	fn select(&mut self, elems: &[Self::Elem], word: &Self::Word) -> Self::Elem {
+		self.inner.select(elems, word)
+	}
+
+	fn sample_bits(&mut self, bits: usize) -> Self::Word {
+		self.inner.sample_bits(bits)
```
jimpo, line 316: Probably need flush here

crates/iop/src/channel/merge.rs
```diff
@@ -0,0 +1,550 @@
...
+		//
+		// This lets every position be a whole number of block sizes.
+		// That holds once every earlier oracle is at least as large.
+		let mut order: Vec<usize> = (self.flushed..self.records.len()).collect();
+		order.sort_by_key(|&i| Reverse(self.records[i].log_msg_len));
+
+		// One commitment is masked as a whole, never partly.
+		//
+		// Every oracle in the round must agree on its witness-dependence.
+		let is_witness_dependent = self.records[order[0]].is_witness_dependent;
```
jimpo, line 192: No need to assert. This is just the logical OR of all the is_witness_dependents in the batch.

crates/iop/src/channel/merge.rs
```diff
@@ -0,0 +1,550 @@
...
+		self.inner.recv_public_claim()
+	}
+
+	fn sample(&mut self) -> Self::Elem {
+		// Commit this round before deriving its challenge.
+		//
+		// Sampling cannot return an error.
+		// A flush failure here has nowhere to go.
+		// It stays queued for the next call that can return one.
+		let _ = self.flush();
```
jimpo, line 279: Hmm, this is a problem that the error gets dropped.

I can think of two possible solutions:
1. Store any errors during flushing and surface them later in `into_inner()` or `verify_oracle_relation()`. They could be stored on the mapping.
2. We restructure this so that the MergeVerifierChannel gets more information about which commitment round each OracleSpec is in. This is a bigger change, but probably a stronger one. It would then be able to validate its own `remaining_oracle_specs` against those of the inner channel.

crates/iop-prover/src/channel/merge.rs
```diff
@@ -0,0 +1,746 @@
...
+			.map(|&k| 1usize << self.pending[k].log_len())
+			.sum();
+		let combined_log_len = log2_ceil_usize(total_len);
+
+		// Lay every buffer into the combined data.
+		// Record where each one landed.
+		//
+		// Every step adds a whole multiple of the next block's size.
+		// So each offset divides evenly by that buffer's own size.
+		let mut combined = FieldBuffer::zeros_in(&self.alloc, combined_log_len);
```
jimpo, line 201: We shouldn't need zeros_in. Should be able to allocate a Vec with capacity and use `VecLike::extend_from_slice`. And only at the end wrap with `FieldBuffer`. Some special handling needed for tiny oracles at the very at that are less than P::WIDTH. (you could even just assert that's not the case right now, it would never happen in practice, only tests maybe)

crates/iop-prover/src/channel/merge.rs
```diff
@@ -0,0 +1,746 @@
...
+		let remaining = self.remaining_oracle_specs();
+		assert!(!remaining.is_empty(), "send_oracle called but no remaining oracle specs");
+		debug_assert_eq!(buffer.log_len(), remaining[0].log_msg_len);
+
+		// Copy the data into a buffer this channel owns.
+		//
+		// The caller's buffer is only borrowed for this call.
+		// Its round may not commit until a later challenge sample.
+		self.pending
+			.push(FieldBuffer::clone_from_slice(&self.alloc, buffer));
```
jimpo: This clone is probably unnecessary. Could we just do the copy into the combined buffer immediately here? I suppose that would be more efficient if the channel new the full schedule of oracle specs ahead of time.

crates/iop-prover/src/channel/merge.rs
```diff
@@ -0,0 +1,746 @@
...
+			combined_log_len,
+		} = record.mapping.expect("flushed above");
+
+		// Place the constituent's transparent polynomial into a zero buffer.
+		// Use the width of the combined oracle, at this oracle's own block.
+		//
+		// Everywhere outside that block reads as zero.
+		// So the produced inner product equals the original claim exactly.
+		let mut padded = FieldBuffer::zeros_in(&self.alloc, combined_log_len);
+		place_block(&mut padded, transparent.to_ref(), block_index);
```
jimpo: This is pretty inefficient to copy a small chunk into a big buffer. It'll be more efficient to do the batching at this layer instead of proving multiple oracles.

We'll do that as an optimization in a future PR.

jimpo (commented):
This is a good start, and I think we should merge it now and do various fixes/optimizations as follow-ups (since this is so isolated).

The main this is that this seems fragile since the channels don't know the full schedule of oracles & interaction rounds ahead of time. I'm thinking that we should introduce an OracleSchedule struct or something that encodes both the list of OracleSpecs, but also groups them into interaction rounds. There's a few ways to represent that, but something like this seems like it would fit nicely:

```rust
/// Represents the sequence of oracles transferred in a protocol, grouped by interaction round.
struct OracleSchedule {
    pub specs: Vec<OracleSpec>,
    /// A list of oracles per interaction round. An interaction round begins when the verifier receives an oracle, and ends when the verifier samples the next challenge.
    pub oracles_per_round: Vec<usize>,
}
```

The `OracleSetupChannel` can create this, and it can be passed to the Merge channels on construction. They can then fully map out the merged oracle placements.

jimpo (approved):
Let's merge and improve with follow-up PRs.

PR #2213, opened 2026-08-20
BINIUS-508: defer the GHASH reduction in ring-switch's sumcheck-claim inner product

tcoratger:
Closes [BINIUS-508](https://linear.app/jimpo/issue/BINIUS-508/defer-the-ghash-reduction-in-ring-switchs-sumcheck-claim-inner-product).

Sums 128 wide products before reducing, instead of reducing each one and then summing. Output is bit-identical — pure reassociation, no protocol change.

### The problem

Ring-switching computes a sumcheck claim as an inner product of two 128-element $\mathbb{F}_{2^{128}}$ sequences.
The naive way multiplies each pair, reduces the wide product immediately, then XORs 128 already-reduced values.
That's 128 reduce operations to produce one scalar.

### The idea

GF(2^128)'s reduction map is linear over $\mathbb{F}_2$.
So `reduce(a) XOR reduce(b)` equals `reduce(a XOR b)`.

```
naive:     reduce(a1*b1) ^ reduce(a2*b2) ^ ... ^ reduce(a128*b128)   // 128 reduces
deferred:  reduce( (a1*b1) ^ (a2*b2) ^ ... ^ (a128*b128) )           // 1 reduce
```

Sum all 128 wide (unreduced) products first, then reduce once at the end.

### Per call

- **before:** 128 reduces
- **after:** 1 reduce

→ the reduce step is amortized 128x, though the multiplies still dominate total cost.

### The change

```diff
- inner_product(s_hat_u, eq_r_double_prime.as_ref().iter().copied())
+ inner_product_packed::<B128, B128>(
+     log_packing,
+     s_hat_u.into_iter(),
+     eq_r_double_prime.as_ref().iter().copied(),
+ )
```

One call site, in `crates/prover/src/ring_switch.rs`.

- binius64 already has this deferred pattern for buffer-shaped inputs (`inner_product_par`, `inner_product_buffers`), proven sound there by the same linearity argument.
- This wires ring-switch's specific inner product onto the existing `inner_product_packed` machinery — no new kernel.
- Distinct from the deferred-reduce family already measured dead for block-local row folds (`fold_word::accumulate_word_chunk`, `ring_switch::fold_1b_rows_for_b128_split`): those fold exactly one multiply per output column, so there's nothing to defer within a chunk. This call collapses 128 terms into one scalar — a genuinely different shape, where deferring pays off.
- Verifier-side call sites with the same N-terms-to-one-scalar shape (`binmul.rs`, `reduction.rs`) were considered and declined: they're generic over `FieldOps`, a trait deliberately weaker than `WideMul` so it can also represent symbolic constraint-system field types. Adding the bound there would break that generality for a tiny, non-hot-loop win.

### Soundness

Pure reassociation of an XOR sum, justified by linearity of the reduction map.
No protocol change, no new soundness assumption, output is bit-identical by construction.
The verifier independently recomputes this exact claim, so a wrong value would fail every proof rather than silently diverging.

### Scope

- **changed:** one call site in `crates/prover/src/ring_switch.rs`
- **new:** an in-binary A/B bench group in `crates/prover/benches/ring_switch.rs`

### Verification

- `cargo check -p binius-prover --all-targets` — clean
- `cargo clippy -p binius-prover --all-targets` — clean
- `cargo +nightly fmt -p binius-prover -- --check` — clean
- `cargo test -p binius-prover --lib` — 60 passed
- `cargo test -p binius-prover --test '*'` — 17 + 1 passed (full prove/verify round trips)

### Benchmark

| variant | time | throughput |
|---|---|---|
| naive per-term reduce | 153.46 ns | 834 Melem/s |
| deferred reduce | 112.08 ns | 1.14 Gelem/s |

1.37x faster at N=128, the shape this call always runs at. Measured in-binary (both paths in one criterion binary) on a quiet machine via `cargo bench -p binius-prover --bench ring_switch -- sumcheck_claim_inner_product`.

### DIFF
--- a/crates/prover/benches/ring_switch.rs
+++ b/crates/prover/benches/ring_switch.rs
@@ -3,8 +3,9 @@
 use std::mem::size_of;
 
 use binius_compute::BufferPool;
-use binius_field::{ExtensionField, arch::OptimalPackedB128};
+use binius_field::{ExtensionField, Random, arch::OptimalPackedB128};
 use binius_math::{
+	inner_product::{inner_product_packed, inner_product_subfield},
 	multilinear::eq::eq_ind_partial_eval,
 	test_utils::{random_field_buffer, random_scalars},
 };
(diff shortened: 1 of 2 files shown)


crates/prover/benches/ring_switch.rs
```diff
@@ -86,5 +87,47 @@ fn bench_suffix_tensor_pipeline(c: &mut Criterion) {
 	group.finish();
 }
 
-criterion_group!(ring_switch, bench_fold_1b_rows_split, bench_suffix_tensor_pipeline);
+/// The sumcheck-claim inner product at its one real shape: `N = 2^log_packing = 128` terms.
+///
+/// In-binary A/B against `inner_product_subfield` (naive per-term reduce).
+/// `inner_product_packed` (deferred reduce) is the path `ring_switch::prove` uses in production.
+fn bench_sumcheck_claim_inner_product(c: &mut Criterion) {
```
jimpo, line 94: Remove this benchmark please, it's for a totally insignificant operation.

PR #2214, opened 2026-08-20
BINIUS-509: bulk-copy Word/BinaryField128bGhash slices instead of serializing one element at a time

tcoratger:
Closes [BINIUS-509](https://linear.app/jimpo/issue/BINIUS-509/bulk-copy-wordbinaryfield128bghash-slices-instead-of-serializing-one).

Replaces a per-element serialization loop with one bulk memory copy, for the two types where a slice's raw bytes already equal its serialized form. Output is bit-identical, proven by proptest.

### The problem

`Word` and `BinaryField128bGhash` are both `Pod` and little-endian on every supported target.
Each type's `serialize` call already writes exactly that type's raw memory bytes, in the wire format's own byte order.
So a whole slice's memory already equals the concatenation of every element's serialized bytes.

But every serialization path — the blanket `SerializeBytes for [T]` impl, and the transcript's `write_slice`/`write_scalar_slice` — looped element by element regardless, calling one small bounds-checked write per item.
For a slice that's already byte-identical to its own serialized form, that loop is pure overhead: like translating a French sentence into French, one word at a time via a dictionary, instead of just copying the sentence.

### The idea

Add one trait method, `SerializeBytes::serialize_slice`, with a default that's exactly today's loop — zero behavior change for every type that doesn't override it.
A type overrides it only when its slice's byte layout already matches its serialized form, collapsing the whole call into one bulk copy.

```
default:              for item in slice { item.serialize(&mut buf)? }   // N small writes
Word / B128 override: buf.put_slice(bytemuck::cast_slice(slice))        // 1 bulk copy
```

### Per slice

- **before:** N calls to `serialize`, one bounds-checked write each
- **after:** 1 `bytemuck::cast_slice` + 1 `put_slice`

→ bigger slices benefit more, since the old loop's per-item overhead scales with `N` while the bulk copy barely notices size.

### The change

```diff
 impl<T: SerializeBytes> SerializeBytes for [T] {
     fn serialize(&self, mut write_buf: impl BufMut) -> Result<(), SerializationError> {
         SerializeBytes::serialize(&self.len(), &mut write_buf)?;
-        self.iter()
-            .try_for_each(|item| SerializeBytes::serialize(item, &mut write_buf))
+        T::serialize_slice(self, write_buf)
     }
 }
```

Plus overrides on `Word` and `BinaryField128bGhash`, and `TranscriptWriter::write_slice`/`write_scalar_slice` routed through the same trait method so every call site benefits with no call-site changes.

### Soundness

Every override is proven byte-identical to the default loop by a proptest sweeping 0–300 random elements, plus a round-trip read-back through `deserialize`.
No wire format changed — this only changes how the existing bytes get produced.

### Scope

- **changed:** `crates/utils/src/serialization.rs` (new trait method + default), `crates/transcript/src/transcript.rs` (route through it)
- **new:** `serialize_slice` overrides on `Word` (`crates/core/src/word.rs`) and `BinaryField128bGhash` (`crates/field/src/ghash.rs`), plus their equivalence proptests
- deliberately **not attempted**: overlapping a prefix pre-encode with the PCS open on a helper thread (part of the same upstream idea) — skipped without first confirming binius64's transcript is Fiat-Shamir-final before the open runs

### Verification

- `cargo clippy -p binius-core -p binius-field -p binius-transcript -p binius-utils --all-targets` — clean
- `cargo +nightly fmt` on the touched crates — clean
- `cargo test --lib` on all 4 touched crates — 142 + 229 + 5 + 41 passed

### Benchmark

| n | loop | bulk | speedup |
|---|---|---|---|
| 1,024 | 1.39 µs | 0.43 µs | 3.2x |
| 16,384 | 40.3 µs | 4.69 µs | 8.6x |
| 262,144 | 648 µs | 94.9 µs | 6.8x |

Measured in-binary (both paths in one criterion binary) on a quiet machine, with a throwaway bench that isn't part of this diff — source below for anyone who wants to reproduce it.

<details>
<summary>Throwaway bench source (not committed — save as <code>crates/field/benches/serialize_slice.rs</code> and add a matching <code>[[bench]]</code> entry to <code>crates/field/Cargo.toml</code> to run it)</summary>

```rust

//! Benchmark serializing a slice of field elements two ways.
//!
//! Every `SerializeBytes` implementor gets a per-element loop by default.
//! `BinaryField128bGhash` overrides it with a bulk byte copy instead.
//!
//! Every supported target here is little-endian.
//! `BinaryField128bGhash` is also `Pod`, so its serialized form is just its raw memory bytes.
//! The bulk path copies a whole slice at once instead of writing each element separately.

use std::hint::black_box;

use binius_field::BinaryField128bGhash;
use binius_utils::SerializeBytes;
use criterion::{Criterion, Throughput, criterion_group, criterion_main};

fn bench_serialize_slice(c: &mut Criterion) {
	let mut group = c.benchmark_group("ghash_serialize_slice");

	for &n in &[1 << 10, 1 << 14, 1 << 18] {
		let values: Vec<BinaryField128bGhash> = (0..n)
			.map(|_| BinaryField128bGhash::from(rand::random::<u128>()))
			.collect();

		group.throughput(Throughput::Bytes((n * 16) as u64));

		group.bench_function(format!("loop/n={n}"), |b| {
			b.iter(|| {
				let mut buf = Vec::with_capacity(n * 16);
				for value in black_box(&values) {
					value.serialize(&mut buf).unwrap();
				}
				black_box(buf)
			})
		});

		group.bench_function(format!("bulk/n={n}"), |b| {
			b.iter(|| {
				let mut buf = Vec::with_capacity(n * 16);
				BinaryField128bGhash::serialize_slice(black_box(&values), &mut buf).unwrap();
				black_box(buf)
			})
		});
	}

	group.finish();
}

criterion_group!(benches, bench_serialize_slice);
criterion_main!(benches);
```

Add to `crates/field/Cargo.toml`:

```toml
[[bench]]
name = "serialize_slice"
harness = false
```

</details>


crates/field/src/ghash.rs
```diff
@@ -99,6 +100,19 @@ impl SerializeBytes for BinaryField128bGhash {
...
 	}
+
+	fn serialize_slice(
+		slice: &[Self],
+		mut write_buf: impl BufMut,
+	) -> Result<(), SerializationError> {
+		// `Self` is `Pod` and little-endian on every supported target.
+		// So each element's serialized bytes are already its raw memory representation.
+		// One `put_slice` over the whole slice replaces `slice.len()` separate 16-byte writes.
+		let bytes: &[u8] = bytemuck::cast_slice(slice);
```
jimpo, line 111: This is implemented at the wrong layer. `serialize_slice` should be specialized at the Underlier level, then this delegates to the Underlier impl (like serialize does).

crates/core/src/word.rs
```diff
@@ -362,6 +364,19 @@ impl SerializeBytes for Word {
...
 	}
+
+	fn serialize_slice(
+		slice: &[Self],
+		mut write_buf: impl BufMut,
+	) -> Result<(), SerializationError> {
+		// `Word` is `#[repr(transparent)]` and little-endian on every supported target.
+		// So each word's serialized bytes are already its raw memory representation.
+		// One `put_slice` over the whole slice replaces `slice.len()` separate 8-byte writes.
+		let bytes: &[u8] = bytemuck::cast_slice(slice);
```
jimpo, line 375: Same comment as below.

PR #2220, opened 2026-08-21
BINIUS-513: overlap Merkle leaf-hashing with the tail of the Reed-Solomon/NTT encode

tcoratger:
Closes [BINIUS-513](https://linear.app/jimpo/issue/BINIUS-513/overlap-merkle-leaf-hashing-with-the-tail-of-the-reed-solomonntt).

### The problem

Committing to an encoded codeword is two strict, sequential steps today: transform the entire codeword with the additive NTT, and only once that is fully done, start hashing leaves into the Merkle tree.
But `ReedSolomonCode::encode_batch`'s NTT already splits into `2^log_num_shares` independent chunks, one per thread — each one finishes well before its siblings, and well before the whole codeword is done.
Measured on a real commit: the encode and the leaf-hash are both a substantial fraction of the combined phase (roughly two-thirds/one-third), so overlapping them is a real, not marginal, opportunity.

### The idea

Hash each independent NTT chunk's leaves the instant that chunk finishes transforming, concurrently with the NTT still transforming the remaining chunks, instead of waiting for the whole codeword.

```
sequential: [-------- encode (all chunks) --------][-- hash all leaves --]
pipelined:  [chunk 0][chunk 1][chunk 2][chunk 3]
                [hash 0][hash 1] [hash 2] [hash 3]
```

Each independent chunk is large enough, at realistic `log_num_shares`, to keep the existing SIMD-batched leaf hasher fully saturated — this matters, because folding Merkle parents in small increments instead of one contiguous sweep per finished region would fragment large batched hash calls into many small underfilled ones and quietly give the win back.

### The change

- `NeighborsLastMultiThread::forward_transform_with_callback` — same transform as `forward_transform`, but invokes a callback the instant each independent chunk finishes; `forward_transform` now delegates to it with a no-op callback.
- `ReedSolomonCode::encode_batch_with_callback` — threads that callback through the encode.
- `BinaryMerkleTree::from_leaves_pipelined` + `LeafRangeWriter` — lets a caller hash disjoint leaf ranges as they become ready, from any thread, in any order, instead of hashing the whole leaf layer from one contiguous input.
- `commit_pipeline::encode_and_commit_pipelined` — wires the two together, hashing each finished NTT chunk via a zero-copy `FieldSlice::par_chunk_scalars` view straight over the packed codeword.

The zero-copy step matters on its own: an earlier version of this change materialized each chunk into an owned `Vec<F>` before hashing it, and that copy — running serially on the very thread that had just finished the chunk's NTT work — was enough overhead to turn the whole thing into a **9% regression**. Hashing straight from the zero-copy view instead is what makes the overlap a net win.

### Soundness

The result is byte-identical to the sequential path: same codeword, same committed Merkle root, same bytes at every tree layer.
Only the scheduling of leaf-hashing relative to the NTT changes — no hashed or transformed value differs.
A dedicated equivalence test proves this directly (not just "should be" — it asserts codeword, root, and every layer's bytes match), across every `log_num_shares` from 0 to 3 and two packing widths.
It caught one real bug while it was being written: an early draft passed the codeword's scalar log-length to the tree builder instead of the leaf-count log-length, and the root mismatch caught it immediately.

### Scope

- **new:** `crates/iop-prover/src/commit_pipeline.rs` (`encode_and_commit_pipelined`, `PipelinedCommit`, equivalence test)
- **new:** `BinaryMerkleTree::from_leaves_pipelined`, `LeafRangeWriter` in `crates/hash/src/binary_merkle_tree.rs`
- **new:** `NeighborsLastMultiThread::forward_transform_with_callback` in `crates/math/src/ntt/neighbors_last.rs` (existing `forward_transform` now a thin wrapper over it)
- **new:** `ReedSolomonCode::encode_batch_with_callback` in `crates/math/src/reed_solomon.rs` (existing `encode_batch` untouched)
- **new:** a permanent `sequential` vs `pipelined` bench group added to the existing `crates/iop-prover/benches/binary_merkle_tree.rs`, timed in one binary so both arms meet identical machine conditions

### Verification

- `cargo clippy -p binius-math -p binius-hash -p binius-iop-prover --all-targets` — clean
- `cargo +nightly fmt` on the touched crates — clean
- `cargo test --lib` on all three touched crates, with and without `--features rayon` — 144 + 29 + 46 passed every way, including the new equivalence test

### Benchmark

Full commit phase, 2^21 B128 scalars, in-binary A/B (both arms in one criterion binary, alternated) — 5 back-to-back runs on a quiet machine:

| run | sequential | pipelined | delta |
|---|---|---|---|
| 1 | 10.869 ms | 11.598 ms | −6.7% |
| 2 | 11.685 ms | 10.945 ms | +6.3% |
| 3 | 11.454 ms | 10.833 ms | +5.4% |
| 4 | 11.429 ms | 10.864 ms | +4.9% |
| 5 | 12.453 ms | 12.147 ms | +2.5% |

Pipelined faster in 4 of 5 runs; median delta ≈ **+4.9%**.
Reported as measured — the one slower run is included, not dropped.

### Context

Surfaced while mining `Layr-Labs/flock-challenge-multi` (a new x86-only Flock speed-challenge repo) for portable optimization ideas — see the "Round 2" section of `FLOCK_MULTI_X86_IDEAS.md` at the repo root. A prior session on this same question (worktree `merkle-ntt-pipeline`, not pushed) confirmed the encode/hash split was real but declined to build the pipelining under contended machine conditions; this is the follow-up that actually builds and verifies it.

### DIFF
--- a/crates/hash/src/binary_merkle_tree.rs
+++ b/crates/hash/src/binary_merkle_tree.rs
@@ -1,7 +1,7 @@
 // Copyright 2024-2025 Irreducible Inc.
 // Copyright 2026 The Binius Developers
 
-use std::{fmt, mem::MaybeUninit, slice};
+use std::{fmt, marker::PhantomData, mem::MaybeUninit, slice};
 
 use binius_compute::{Allocator, GlobalAllocator, VecLike};
 use binius_field::Field;
@@ -233,6 +233,113 @@ impl<N: ArraySize, A: Allocator> BinaryMerkleTree<Array<u8, N>, A> {
 			log_len,
 			inner_nodes,
 		}
+	}
+
+	/// Commits leaves written through a [`LeafRangeWriter`] instead of hashed from one
+	/// contiguous input all at once.
+	///
+	/// `populate` must write every leaf in `0..2^log_len` exactly once through the writer it is
+	/// handed, from any thread, in any order, before it returns.
+	/// That is what lets a caller start hashing an early leaf range while a later one is still
+	/// being produced, instead of waiting for all of it before hashing starts.
+	///
+	/// # Panics
+	///
+	/// Panics unless `log_len` is nonzero.
+	pub fn from_leaves_pipelined<F, H>(
+		log_len: usize,
+		n_items_per_input: usize,
+		alloc: &A,
+		populate: impl FnOnce(&LeafRangeWriter<'_, F, H>),
+	) -> Self
+	where
+		F: Field,
+		H: HashSuite<LeafHash: OutputSizeUser<OutputSize = N>>,
+	{
+		assert!(log_len > 0, "a tree over a single leaf has no nodes to build");
+
+		let total_length = (1 << (log_len + 1)) - 1;
+		let mut inner_nodes = alloc.alloc::<Array<u8, N>>(total_length);
+
+		let (leaf_layer, remaining) = inner_nodes.spare_capacity_mut().split_at_mut(1 << log_len);
+
+		{
+			let _span = tracing::debug_span!("hash_leaves_pipelined").entered();
+			let writer = LeafRangeWriter {
+				leaves: SendPtr(leaf_layer.as_mut_ptr()),
+				n_items_per_input,
+				_marker: PhantomData,
+			};
+			populate(&writer);
+		}
+
+		let leaves: &[Array<u8, N>] = unsafe {
+			// SAFETY: `populate`'s contract guarantees every leaf was written before it returned.
+			leaf_layer.assume_init_ref()
+		};
+
+		let parallel_compression = H::ParCompression::default();
+		let ctx = BuildContext {
+			leaves,
+			log_len,
+			nodes: SendPtr(remaining.as_mut_ptr()),
+			compression: &parallel_compression,
+		};
+		ctx.build_node(0, log_len);
+
+		unsafe {
+			// SAFETY: the leaf layer is fully written by `populate`'s contract, and `build_node`
+			// just wrote every remaining node above it.
+			inner_nodes.set_len(total_length);
+		}
+		Self {
+			log_len,
+			inner_nodes,
+		}
+	}
+}
+
+/// Lets [`BinaryMerkleTree::from_leaves_pipelined`]'s caller write disjoint leaf ranges from any
+/// thread, in any order, instead of hashing the whole leaf layer from one contiguous input.
+///
+/// Every scalar range handed to [`Self::write_range`] must be disjoint from every other range
+/// written through the same writer.
+/// Together, the ranges must cover every leaf exactly once before the `populate` closure that
+/// received this writer returns — that return is the only signal
+/// [`BinaryMerkleTree::from_leaves_pipelined`] has that every leaf is ready to read.
+pub struct LeafRangeWriter<'a, F, H: HashSuite> {
+	leaves: SendPtr<MaybeUninit<Output<H::LeafHash>>>,
+	n_items_per_input: usize,
+	_marker: PhantomData<&'a [F]>,
+}
+
+impl<F: Field, H: HashSuite> LeafRangeWriter<'_, F, H> {
+	/// Hashes `scalars` into the leaves starting at `leaf_start`, `n_items_per_input` scalars
+	/// per leaf.
+	///
+	/// # Panics
+	///
+	/// Panics unless `scalars.len()` is a multiple of `n_items_per_input`.
+	pub fn write_range(&self, leaf_start: usize, scalars: &[F]) {
+		assert_eq!(
+			scalars.len() % self.n_items_per_input,
+			0,
+			"a leaf range must cover a whole number of leaves"
+		);
+		let n_leaves = scalars.len() / self.n_items_per_input;
+
+		let dst = unsafe {
+			// SAFETY: `Self`'s contract guarantees every range written through this writer is
+			// disjoint from every other one, so this range does not alias any other write.
+			slice::from_raw_parts_mut(self.leaves.0.add(leaf_start), n_leaves)
+		};
+		H::ParLeafHash::default().digest_with_const_len(
+			self.n_items_per_input,
+			scalars
+				.par_chunks(self.n_items_per_input)
+				.map(|chunk| chunk.iter().copied()),
+			dst,
+		);
 	}
 }
 
--- a/crates/hash/src/binary_merkle_tree.rs
+++ b/crates/hash/src/binary_merkle_tree.rs
@@ -235,13 +235,16 @@ impl<N: ArraySize, A: Allocator> BinaryMerkleTree<Array<u8, N>, A> {
 		}
 	}
 
-	/// Commits leaves written through a [`LeafRangeWriter`] instead of hashed from one
-	/// contiguous input all at once.
+	/// Commits leaves written through a [`LeafRangeWriter`].
 	///
-	/// `populate` must write every leaf in `0..2^log_len` exactly once through the writer it is
-	/// handed, from any thread, in any order, before it returns.
-	/// That is what lets a caller start hashing an early leaf range while a later one is still
-	/// being produced, instead of waiting for all of it before hashing starts.
+	/// Unlike [`Self::from_leaves`], leaves don't come from one contiguous input all at once.
+	///
+	/// `populate` must write every leaf in `0..2^log_len` exactly once, through the writer.
+	/// It may write from any thread, in any order.
+	/// Every leaf must be written before `populate` returns.
+	///
+	/// That lets a caller start hashing an early leaf range while a later one is still produced.
+	/// Without this, hashing would have to wait for the whole leaf layer first.
 	///
 	/// # Panics
 	///
@@ -288,8 +291,8 @@ impl<N: ArraySize, A: Allocator> BinaryMerkleTree<Array<u8, N>, A> {
 		ctx.build_node(0, log_len);
 
 		unsafe {
-			// SAFETY: the leaf layer is fully written by `populate`'s contract, and `build_node`
-			// just wrote every remaining node above it.
+			// SAFETY: the leaf layer is fully written, by `populate`'s contract.
+			// `build_node` just wrote every remaining node above it.
 			inner_nodes.set_len(total_length);
 		}
 		Self {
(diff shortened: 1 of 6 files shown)


benji-potamus (commented):
Post-merge review (this landed without one, so the findings below are follow-up work rather than a merge gate).

Mechanical pass on current `main` is clean: `cargo clippy -p binius-math -p binius-hash -p binius-iop-prover --all-targets --all-features -- -D warnings`, `cargo +nightly-2026-07-01 fmt --check`, and `typos` all pass on the touched files.

**Correctness.** I ran a differential test of `encode_batch_with_callback` against `encode_batch` outside the range the equivalence test covers, and they diverge.

**Blocking — `crates/math/src/reed_solomon.rs:196`: `encode_batch_with_callback` skips the NTT entirely when `data.log_len() < P::LOG_WIDTH`.**

```rust
if data.log_len() < P::LOG_WIDTH {
    on_chunk_ready(0, output.as_ref());
    return output;   // <- never reaches forward_transform_with_callback
}
```

It returns the bit-reversed, repeated message rather than a codeword. `encode_batch` has no such early return — at this PR's base commit its `data.log_len() < P::LOG_WIDTH` branch built the same buffer and then *fell through* to `ntt.forward_transform(...)`; the copy here turned that into a `return`. So the doc's "Same encoding as `encode_batch`" is false whenever `log_dim + log_batch_size < P::LOG_WIDTH` and `log_dim > 0`.

Repro — `PackedBinaryGhash4x128b` (`LOG_WIDTH = 2`), `log_dim = 1`, `log_inv_rate = 2`, `log_num_shares = 0`:

```
assertion `left == right` failed: encode_batch_with_callback diverges from encode_batch
  left:  [Packed4x128([0xc6017c94...,0x958b761d...,0xc6017c94...,0x958b761d...]), ...]   <- repeated message
  right: [Packed4x128([0xc6017c94...,0x538a0a89...,0x95f49331...,0x007fe52c...]), ...]   <- codeword
```

The fix is to delete the `if` block outright: `forward_transform_with_callback` already fires `on_chunk_ready(0, ..)` in its own `log_d <= P::LOG_WIDTH` fallback, so falling through covers that case correctly. The equivalence test only exercises `log_dim = 10`, which is why it never sees this — worth adding a case with `log_dim + log_batch_size < P::LOG_WIDTH`.

**Blocking — `crates/hash/src/binary_merkle_tree.rs:253` and `:321`: two safe `pub fn`s whose memory safety rests on unchecked caller contracts.**

`from_leaves_pipelined` calls `leaf_layer.assume_init_ref()` on the strength of "`populate` must write every leaf". A safe caller passing `|_writer| {}` reads uninitialized `Array<u8, N>` — UB with no `unsafe` typed anywhere. Likewise `write_range` does `slice::from_raw_parts_mut(self.leaves.0.add(leaf_start), n_leaves)` with no bounds check, so any `leaf_start` past the layer is an out-of-bounds write from safe code.

Either make both `unsafe fn` with a `# Safety` section, or make them checked: assert `leaf_start + n_leaves <= 1 << log_len` in `write_range`, and have the writer accumulate an `AtomicUsize` of leaves written that `from_leaves_pipelined` asserts equals `1 << log_len` after `populate` returns. The second is cheap at this granularity — one atomic per range, not per leaf — and it turns the contract into something the compiler-visible API actually enforces.

**Not blocking — `crates/iop-prover/src/commit_pipeline.rs:46` documents a guard that doesn't exist.** "an assertion inside the leaf writer catches it" — `write_range` has no assertion. The case it refers to is real: at `commit_pipeline.rs:87`, `n_leaves_per_chunk = (1 << chunk_log_len) / leaf_scalars` is `0` when a leaf spans more than one chunk, so every chunk computes `leaf_start = 0` and they all write over each other. Adding the assertion above makes the doc true; otherwise drop the claim.

**Not blocking — nothing calls `encode_and_commit_pipelined`.** Its only users are its own test and the bench; the prover's commit path still runs `encode_batch` then `commit_field_buffer`. As it stands the PR adds public API in three crates plus the unsafe leaf writer for a win nothing collects. Worth either wiring it into the prover commit path or reverting it until something needs it — especially since the posted A/B has the sequential arm alone ranging 10.869–12.453 ms, a spread wider than the ~4.9% median effect being claimed.

crates/math/src/reed_solomon.rs
```diff
@@ -169,6 +169,79 @@ impl<F: BinaryField> ReedSolomonCode<F> {
...
+		let log_output_len = self.log_dim() + log_batch_size + self.log_inv_rate;
+		let output_data = if data.log_len() < P::LOG_WIDTH {
+			let mut scalars = data.iter_scalars().collect::<Vec<_>>();
+			bit_reverse_indices(&mut scalars);
+			let elem_0 = P::from_scalars(scalars.into_iter().cycle());
+			let len = 1 << log_output_len.saturating_sub(P::LOG_WIDTH);
+			let mut output = alloc.alloc::<P>(len);
+			output.resize(len, elem_0);
+			on_chunk_ready(0, &output);
+			return FieldBuffer::new(log_output_len, output);
```
benji-potamus, line 226: **Blocking (correctness).** This `return` skips the NTT entirely, so `encode_batch_with_callback` hands back the bit-reversed, repeated *message* rather than a codeword.

`encode_batch` right above has no such early return: its `data.log_len() < P::LOG_WIDTH` branch (line 141) builds `output_data` and then *falls through* to `ntt.forward_transform(...)`. Copying the branch into this function turned the fall-through into a `return`, so the doc's "Same encoding as [`Self::encode_batch`]" is false whenever `log_dim + log_batch_size < P::LOG_WIDTH` and `log_dim > 0`.

Repro on `PackedBinaryGhash4x128b` (`LOG_WIDTH = 2`), `log_dim = 1`, `log_inv_rate = 2`, `log_num_shares = 0`:

```
assertion `left == right` failed: encode_batch_with_callback diverges from encode_batch
  left:  [Packed4x128([0xc6017c94..,0x958b761d..,0xc6017c94..,0x958b761d..]), ..]   <- repeated message
  right: [Packed4x128([0xc6017c94..,0x538a0a89..,0x95f49331..,0x007fe52c..]), ..]   <- codeword
```

Fix: drop the `on_chunk_ready` + `return` and let the branch fall through to `forward_transform_with_callback` like `encode_batch` does. The callback still fires — `forward_transform_with_callback` invokes `on_chunk_ready(0, ..)` in its own `log_d <= P::LOG_WIDTH` fallback.

The equivalence test only covers `log_dim = 10`, so it never reaches this branch; worth adding a case with `log_dim + log_batch_size < P::LOG_WIDTH`.

Separately, and the reason this slipped: this whole function is a copy of `encode_batch`'s body rather than `encode_batch` delegating to it with a no-op callback — which is exactly the shape you used for `forward_transform` in `neighbors_last.rs`. Same treatment here would have made the divergence impossible to write.

crates/hash/src/binary_merkle_tree.rs
```diff
@@ -234,6 +234,111 @@ impl<N: ArraySize, A: Allocator> BinaryMerkleTree<Array<u8, N>, A> {
...
+				leaves: SendPtr(leaf_layer.as_mut_ptr()),
+				n_items_per_input,
+				_marker: PhantomData,
+			};
+			populate(&writer);
+		}
+
+		let leaves: &[Array<u8, N>] = unsafe {
+			// SAFETY: `populate`'s contract guarantees every leaf was written before it returned.
+			leaf_layer.assume_init_ref()
```
benji-potamus, line 281: **Blocking (soundness).** `from_leaves_pipelined` is a safe `pub fn`, but this `assume_init_ref` rests entirely on a doc contract the caller is trusted to honour. `BinaryMerkleTree::from_leaves_pipelined(log_len, n, alloc, |_writer| {})` compiles without `unsafe` anywhere and reads uninitialized `Array<u8, N>` — UB reachable from safe code.

Two ways out:

- make it `unsafe fn` with a `# Safety` section instead of `# Panics`, pushing the obligation onto the caller where the compiler can see it; or
- make it checked: have `LeafRangeWriter` accumulate an `AtomicUsize` of leaves written, and assert it equals `1 << log_len` after `populate` returns.

I'd lean toward the second — it's one atomic add per *range*, not per leaf, so it's free at this granularity, and it converts "every leaf must be written exactly once" from a comment into something actually enforced.

crates/hash/src/binary_merkle_tree.rs
```diff
@@ -234,6 +234,111 @@ impl<N: ArraySize, A: Allocator> BinaryMerkleTree<Array<u8, N>, A> {
...
+	pub fn write_range<ParIter>(&self, leaf_start: usize, leaves: ParIter)
+	where
+		ParIter: IndexedParallelIterator<Item: IntoIterator<Item = F, IntoIter: Send>>,
+	{
+		let n_leaves = leaves.len();
+		let dst = unsafe {
+			// SAFETY: `Self`'s contract guarantees every range written through this writer
+			// is disjoint from every other one.
+			// So this range does not alias any other write.
+			slice::from_raw_parts_mut(self.leaves.0.add(leaf_start), n_leaves)
```
benji-potamus, line 338: **Blocking (soundness), same theme as the `assume_init_ref` above.** `write_range` is safe and takes an arbitrary `leaf_start`, but there's no bounds check before `from_raw_parts_mut`. Any `leaf_start` past the leaf layer — or a `leaf_start + n_leaves` that runs off its end — is an out-of-bounds write from safe code.

The SAFETY comment only argues disjointness between ranges, and even that is asserted by the caller rather than checked. Nothing here establishes that the range lies inside the layer at all.

An `assert!(leaf_start + n_leaves <= 1 << log_len)` costs nothing next to the hashing that follows — it needs `log_len` threaded onto the writer, which is worth it. That assert plus the coverage counter above would make the whole contract enforced rather than documented.

crates/iop-prover/src/commit_pipeline.rs
```diff
@@ -0,0 +1,173 @@
...
+				ntt,
+				message,
+				log_batch_size,
+				alloc,
+				|block, chunk: &[P]| {
+					// A leaf is `leaf_scalars` scalars.
+					// Every chunk this NTT split produces is the same size.
+					// So the chunk index alone gives its leaf offset.
+					let chunk_log_len = chunk.len().ilog2() as usize + P::LOG_WIDTH;
+					let n_leaves_per_chunk = (1usize << chunk_log_len) / leaf_scalars;
```
benji-potamus, line 88: When a leaf spans more than one chunk, this integer division is `0`, so `leaf_start` is `0` for *every* chunk and they all write over each other — the aliasing that `write_range`'s SAFETY comment assumes away.

Line 47 says this case is caught: "an assertion inside the leaf writer catches it". There is no assertion in `write_range`. Either add the bounds/coverage checks suggested on that function — which would genuinely catch it — or drop the claim from the doc, because right now it reads as a guarantee that doesn't exist.

crates/iop-prover/src/commit_pipeline.rs
```diff
@@ -0,0 +1,173 @@
...
+/// ## Preconditions
+///
+/// * Same as [`ReedSolomonCode::encode_batch`].
+/// * `log_leaf_len` must be at most the codeword's log length.
+/// * Every independent NTT chunk (see
+///   [`NeighborsLastMultiThread::forward_transform_with_callback`]) must span at least one whole
+///   leaf.
+/// * That holds unless `log_leaf_len` is unusually large relative to `ntt.log_num_shares`, in which
+///   case an assertion inside the leaf writer catches it.
+pub fn encode_and_commit_pipelined<F, P, DC, H, N, A>(
```
benji-potamus, line 48: Not blocking, but worth settling: nothing calls this outside its own test and the bench. The prover's commit path still runs `encode_batch` then `commit_field_buffer`, so as merged this adds public API in three crates plus the unsafe leaf writer for a speedup nothing actually collects.

Either wire it into the prover commit path so the win is real, or revert it until something needs it. Leaning on the second unless the former is imminent — the posted A/B has the sequential arm alone ranging 10.869–12.453 ms, a spread wider than the ~4.9% median effect being claimed, so the measurement doesn't yet establish the win it's carrying this API surface for.

benji-potamus (commented):
Inline anchors for the findings in my summary comment above — one blocking correctness bug (`encode_batch_with_callback` skipping the NTT), two blocking soundness issues in the new `binius-hash` API, and two non-blocking notes.

Mechanical pass on current `main` is clean: clippy with `-D warnings`, nightly fmt `--check`, and `typos` all pass on the touched files.

PR #2222, opened 2026-08-21
BINIUS-515: Price the FRI proof-size versus encoding trade-off across Reed-Solomon rates

tcoratger:
@jimpo Let me know what you think about this — it makes the rate trade-off measurable before we get into Ligerito/WHIR.

Closes [BINIUS-515](https://linear.app/jimpo/issue/BINIUS-515).

### In one line

Sweep the Reed–Solomon rate, price the proof at each one, and benchmark what each one costs the prover, so the trade-off is a table instead of a guess.

### The problem

`log_inv_rate` is the single biggest lever on FRI proof size, and it was a fixed input.

A *rate* of 1/2 means the codeword is twice as long as the message. Lowering the rate to 1/4 makes it four times as long. A longer codeword is further from every other codeword, so a corrupted one is caught by fewer random samples — fewer queries, smaller proof.

But the prover has to *build* that longer codeword. Every step down doubles its encoding work.

Neither half of that trade was measurable without doing it by hand.

### The change

Price both halves from code already in the tree, and put them side by side.

- **Proof bytes:** sweep the rate; for each candidate re-derive the query count and re-optimize the fold arities, then price the result with the repo's own `proof_size`.
- **Prover cost:** benchmark `ReedSolomonCode::encode_batch` at each rate, through the real commit-path NTT.

```
crates/iop/src/fri/size_estimation.rs
+ pub struct RateEstimate { log_inv_rate, n_test_queries, proof_size }
+ RateEstimate::for_rate(merkle_scheme, oracles, log_inv_rate, n_test_queries) -> Self
+ RateEstimate::sweep(merkle_scheme, oracles, n_queries, log_inv_rates)        -> Vec<Self>
+ RateEstimate::best(..)                                                       -> Option<Self>
```

These hang off `RateEstimate` rather than standing as free functions, matching the crate around them: `FRIParams::optimal_for_batch` is an associated constructor and `FRIParams::proof_size` a method. `for_rate` is the primitive, `sweep` maps it over candidates, and `best` takes the minimum — returning `Option` rather than panicking on an empty candidate list.

The query count arrives as a rate-indexed closure rather than being derived inside. Rate is not the only lever on it — the proximity-testing regime is another — and a caller pricing a different regime should not have to edit this module to do it. Pass `calculate_n_test_queries` for the unique-decoding regime this repo ships.

`best` breaks ties toward the first candidate, which with ascending candidates is the cheapest to encode. `FRIParams` is untouched — this only *reports*.

### The measurement

At 100 bits, one non-ZK oracle of `log_msg_len` 20, against encode cost at the matching dimension. Encode is single-threaded, normalized to rate 1/2.

| rate | queries | proof | vs 1/2 | encode | vs 1/2 |
|---|---|---|---|---|---|
| 1/2 | 241 | 311.25 KiB | 1.00x | 38.6 ms | 1.00x |
| 1/4 | 148 | 234.38 KiB | 1.33x | 71.5 ms | 1.85x |
| 1/8 | 121 | **217.00 KiB** | **1.43x** | 140.7 ms | 3.64x |
| 1/16 | 110 | 220.09 KiB | 1.41x | 280.0 ms | 7.24x |
| 1/32 | 105 | 232.31 KiB | 1.34x | 605 ms | 15.7x |
| 1/64 | 103 | 249.50 KiB | 1.25x | 1489 ms | 38.5x |

**Proof size bottoms out around rate 1/8 at 1.3x-1.5x, then gets worse.** Query counts flatten as the rate falls — 121 at 1/8 against 103 at 1/64 — while every step doubles the codeword and its Merkle tree. Rates 1/32 and below are dominated on *both* axes.

**Rate 1/4 is the value pick**, buying roughly 75-90% of the achievable win for under 2x encode.

At `log_msg_len` 17 the proof optimum is also 1/8 (156.25 KiB, 1.32x); at 24 it moves to 1/16 (310.06 KiB, 1.54x).

### One caveat on reading the table

The proof column is indexed by oracle `log_msg_len` and the encode column by code `log_dim`, which differ by the chosen batch size. The trends are comparable; the absolute milliseconds are not the encode cost of that exact proof. The bench also times `encode_batch` including allocation, so it measures encode-plus-alloc. The ratios across rates are the reliable part.

### Scope

- **new:** the three items above, plus `crates/math/benches/reed_solomon.rs` and its `[[bench]]` registration
- **changed:** the `pub use` line in `crates/iop/src/fri/mod.rs`
- `FRIParams` and every caller are untouched

### Verification

- `cargo clippy -p binius-iop -p binius-math --all-targets --all-features -- -D warnings` — clean
- `cargo test -p binius-iop` — 14 passed; `cargo test -p binius-math` — 150 passed plus doctests
- `RUSTDOCFLAGS="-D warnings" cargo doc` — clean
- `cargo check -p binius-iop-prover` — clean, confirming the widened `pub use` breaks no glob import

Tests pin the proof-size table per rate, the query counts 241/148/121/110/105, agreement between `sweep` and a hand-written `optimal_for_batch` + `proof_size` loop over three batch shapes, a ZK batch (which pins `log_batch_size = 1` and takes a different path), the candidate-list boundaries, and that halving the query count shrinks the proof at every rate.

`binius-iop` has no `rayon` feature, so that gate is inapplicable there; `binius-math` does, and both modes pass.

Rebased on `main`.

### DIFF
--- a/crates/iop/src/fri/mod.rs
+++ b/crates/iop/src/fri/mod.rs
@@ -27,8 +27,10 @@ mod batch;
 mod common;
 mod error;
 pub mod fold;
+mod size_estimation;
 pub mod verify;
 
 pub use common::*;
 pub use error::*;
 pub use fold::FRIFoldVerifier;
+pub use size_estimation::RateEstimate;
--- a/crates/math/benches/reed_solomon.rs
+++ b/crates/math/benches/reed_solomon.rs
@@ -1,0 +1,62 @@
+// Copyright 2026 The Binius Developers
+
+use binius_compute::GlobalAllocator;
+use binius_field::{Ghash128b as B128, arch::OptimalPackedB128};
+use binius_math::{
+	ntt::{NeighborsLastMultiThread, domain_context::GaoMateerOnTheFly},
+	reed_solomon::ReedSolomonCode,
+	test_utils::random_field_buffer,
+};
+use binius_utils::{checked_arithmetics::log2_ceil_usize, rayon::current_num_threads};
+use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+
+/// Message dimensions to sweep, matching the oracle sizes the FRI rate sweep prices.
+const LOG_DIMS: [usize; 2] = [17, 20];
+
+/// Candidate inverse rates, matching the sweep `binius_iop::fri::RateEstimate::sweep` prices.
+const LOG_INV_RATES: [usize; 6] = [1, 2, 3, 4, 5, 6];
+
+/// Benchmarks [`ReedSolomonCode::encode_batch`] at each candidate Reed-Solomon rate.
+///
+/// This is the prover-cost half of the trade-off `binius_iop::fri::RateEstimate::sweep` prices.
+/// That function reports how many proof bytes a rate saves; this reports what the rate costs.
+/// One step lower doubles the codeword, so encode time should roughly double with it.
+///
+/// The NTT is the multi-threaded one the prover's commit path uses, at one share per core.
+/// The domain context is regenerated per rate, since the code length changes with the rate.
+fn bench_encode_by_rate(c: &mut Criterion) {
+	let mut rng = rand::rng();
+	let log_num_shares = log2_ceil_usize(current_num_threads());
+
+	let mut group = c.benchmark_group("reed_solomon_encode_by_rate");
+	group.sample_size(10);
+
+	for log_dim in LOG_DIMS {
+		let message = random_field_buffer::<OptimalPackedB128>(&mut rng, log_dim);
+
+		for log_inv_rate in LOG_INV_RATES {
+			let rs_code = ReedSolomonCode::<B128>::new(log_dim, log_inv_rate);
+			let domain_context = GaoMateerOnTheFly::<B128>::generate(rs_code.log_len());
+			let ntt = NeighborsLastMultiThread::new(domain_context, log_num_shares);
+
+			// Codeword bytes, so throughput reads as encode bandwidth across rates.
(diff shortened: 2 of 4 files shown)


crates/math/benches/reed_solomon.rs
```diff
@@ -0,0 +1,62 @@
...
+	let mut rng = rand::rng();
+	let log_num_shares = log2_ceil_usize(current_num_threads());
+
+	let mut group = c.benchmark_group("reed_solomon_encode_by_rate");
+	group.sample_size(10);
+
+	for log_dim in LOG_DIMS {
+		let message = random_field_buffer::<OptimalPackedB128>(&mut rng, log_dim);
+
+		for log_inv_rate in LOG_INV_RATES {
```
jimpo, line 37: encode time is just linear in the inv rate, right? I don't think we really need to bench across different rates. Even `log_inv_rate = 0` would get the whole picture (I'm not totally sure ReedSolomonCode allows rate = 1, but it could).

crates/iop/src/fri/mod.rs
```diff
@@ -27,8 +27,10 @@ mod batch;
 mod common;
 mod error;
 pub mod fold;
+mod size_estimation;
 pub mod verify;
 
 pub use common::*;
 pub use error::*;
 pub use fold::FRIFoldVerifier;
+pub use size_estimation::RateEstimate;
```
jimpo, line 36: Does this need to be pub? This just looks like testing code to me.

jimpo (commented):
This is good to measure, but I'm not sure how this is exposed or whether it needs to be.

My recommendation is that this is be a private test module with an `#[ignore]` test that could pin some proof size values and print out some numbers.

PR #2240, opened 2026-08-22
BINIUS-526: Move build_g and build_monster_segments onto their owning types

tcoratger:
Closes [BINIUS-526](https://linear.app/jimpo/issue/BINIUS-526).

Turns two "free function that is really a method in disguise" cases into actual methods. Pure refactor — no behavior change.

### The problem

Two functions in the shift-reduction prover take one type as their main argument, but live as free functions instead of methods on that type.

- One function folds a key segment's own keys into the witness-and-batching multilinear, taking the segment as an argument instead of being called on it.
- The other builds a key collection's two constraint-matrix segments, taking the collection as an argument instead of being called on it.

A caller has to know which module holds the right free function, instead of finding the operation on the type it operates on.

### The idea

Move each function onto the type it already treats as primary.

```
build_g(words, segment, prepared)               ->  segment.build_g(words, prepared)
build_monster_segments(alloc, collection, ...)   ->  collection.build_monster_segments(alloc, ...)
```

Both functions keep their bodies unchanged; only the receiver moves from an explicit argument to `self`.

### The change

- `crates/prover/src/protocols/shift/phase_1.rs`: the row-folding function moves onto the key-segment type in `crates/prover/src/protocols/shift/key_collection/key_segment.rs`, next to the type's other methods. `phase_1.rs` keeps only the row-length helper the method still needs, now visible to its new home.
- `crates/prover/src/protocols/shift/monster.rs`: the segment-building function moves onto the key-collection type in `crates/prover/src/protocols/shift/key_collection/collection.rs`. The scalar-table type and the outer-slot-weight helper it depends on are exposed from `monster.rs` at the visibility its new caller needs.
- Three call sites update to the method form: the single-instance prover (`crates/prover/src/protocols/shift/prove.rs`), the batched M4 prover (`crates/m4-prover/src/shift.rs`), and the shift-reduction benchmark (`crates/prover/benches/shift_reduction.rs`).

### Scope

- **changed:** two functions become methods; their imports and visibility follow them; three call sites updated across two crates.
- **left untouched:** every other function in the touched files, including the batched-witness variant of the row-folding function in the M4 crate, which stays a free function since it does not take a key segment as its main argument.

### Verification

- `cargo check -p binius-prover -p binius-m4-prover --all-targets` — clean.
- `cargo clippy -p binius-prover -p binius-m4-prover --all-targets -- -D warnings`, with and without `--features rayon` — clean, no `too_many_arguments` needed on the moved method.
- `cargo +nightly fmt -p binius-prover -p binius-m4-prover` — applied.
- `cargo test -p binius-prover -p binius-m4-prover`, with and without `--features rayon` — all passing.

### DIFF
--- a/crates/m4-prover/src/shift.rs
+++ b/crates/m4-prover/src/shift.rs
@@ -16,8 +16,7 @@ use binius_prover::{
 	protocols::shift::{
 		KeyCollection, KeySegment, OperatorClaims, PreparedOperatorClaims, ShiftChallengePoint,
 		ShiftIndOutput, ShiftIndSumcheck, ShiftOutput,
-		monster::build_monster_segments,
-		phase_1::{Phase1Output, SparseShiftRows, build_g},
+		phase_1::{Phase1Output, SparseShiftRows},
 		phase_2::run_sumcheck,
 	},
 };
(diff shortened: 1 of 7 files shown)


jimpo (commented):
I somewhat prefer it the old way (with standalone functions), but not a big deal.

PR #2257, opened 2026-08-23
Give the word-axis fold a parallel driver

tcoratger:
Lets a caller fold several word columns against one shared point without choosing between reusing the folder's tables and using all the cores. No protocol change — the folded values are identical.

### The problem

A `&[Word]` list is a matrix over GF(2): rows are words, columns are bit positions.
The word-axis fold contracts the row axis, giving one field element per bit position.

The reusable folder that does this only had a **sequential** driver.
So a caller had two options, and both are wrong:

- reuse the folder's tables, and fold on one core
- rebuild the tables per column via the one-shot function, and fold on all of them

Both real callers landed on the wrong side of that:

- binary multiplication folds **six** columns against one point, sequentially
- integer multiplication folds **four**, rebuilding the tables each time

Table construction is not the expensive part — it is ~40 us against a ~6.6 ms fold at from^{20}$ words.
The sequential driver is.

### The idea

Add a second driver that divides the **chunk axis** of a single list.
A chunk is 64 words, so a list of from^{20}$ words is 16384 independent chunks — plenty to divide.

The two drivers then cover the two shapes a caller can be in:

- **many** lists share the point: parallelize across lists, fold each sequentially
- **few** lists share the point: fold each one across its own chunks

Six columns cannot fill a ten-core machine on their own, which is exactly the second case.

### The change

```
binmul: six columns, one shared point
- let a_lo_evals = folder.fold(a_lo);          // one core
+ let a_lo_evals = folder.fold_par(a_lo);      // all of them

intmul: four columns, one shared point
- let per_bit_evals = |e: &[Word]| fold_across_words::<_, P>(e, r_out);   // rebuilds tables x4
+ let folder = WordFolder::<F>::new(r_out);                               // built once
+ let a_evals = folder.fold_par(a_exponents);
```

The one-shot free function becomes a thin wrapper over the new driver.
That deletes the chunk-and-tail driver it had duplicated from the sequential one, so the trailing partial chunk is now handled in exactly one place.

Its unused `P` type parameter goes too — it was never read.

### Why the split needs a floor

Left unbounded, rayon divides down to one chunk per task.
A chunk folds in well under a microsecond, which is about what a handoff costs, so at from^{12}$ words the unbounded parallel fold is **slower than sequential** (29.7 us against 25.5 us).

A floor also caps how far a loop *can* divide: `n` chunks split into at most `n / floor` tasks.
Set it too high and a short list starves the cores; too low and a long list drowns in handoffs.
Sixteen chunks — 1024 words — loses at neither end:

| words | sequential | no floor | floor 16 | floor 64 |
|---|---|---|---|---|
| from^{12}$ | 0.0255 ms | 0.0297 ms | **0.0198 ms** | 0.0325 ms |
| from^{16}$ | 0.4094 ms | 0.1085 ms | **0.0862 ms** | 0.0850 ms |
| from^{20}$ | 6.6203 ms | 1.1231 ms | **1.1281 ms** | 1.0989 ms |

### Benchmark

Apple M1 Pro, 8P + 2E, 10 rayon threads, min of many runs. One column, both drivers:

| words | sequential | parallel | speedup |
|---|---|---|---|
| from^{12}$ | 0.0255 ms | 0.0198 ms | 1.29x |
| from^{14}$ | 0.1032 ms | 0.0356 ms | 2.90x |
| from^{16}$ | 0.4094 ms | 0.0862 ms | 4.75x |
| from^{18}$ | 1.6408 ms | 0.2966 ms | 5.53x |
| from^{20}$ | 6.6203 ms | 1.1281 ms | **5.87x** |

The binary-multiplication shape — six columns, one shared point — through the in-repo criterion bench:

| words per column | sequential | parallel | speedup |
|---|---|---|---|
| from^{16}$ | 2.455 ms | 0.832 ms | 2.95x |
| from^{20}$ | 39.5 ms | 7.71 ms | **5.12x** |

The single-column ratio is the better estimate of the ceiling; the six-column figure is measured on a contended machine and is the conservative one.

A follow-up that changes the merge strategy of this same loop is worth a further ~1.2x on top.

### Soundness

Field addition is associative and exact, so any grouping of the per-chunk sums gives the same element:

```
sequential: (((c_0 + c_1) + c_2) + c_3)
parallel  : ((c_0 + c_1) + (c_2 + c_3))
```

The transcript is byte-identical to before. Nothing about the challenge distribution changes.

### Scope

- **new:** the parallel driver, a private helper for the trailing partial chunk, and a test pinning the two drivers equal across every chunk regime including the empty list
- **changed:** six call sites in binary multiplication, four in integer multiplication
- **left untouched:** the sequential driver, which is still the right one for the batched instance fold in `binius-m4-prover` — that caller is already parallel over thousands of columns
- the bench now runs both drivers side by side, at one column and at six

### Verification

- `cargo check -p binius-prover -p binius-m4-prover --all-targets` — clean
- `cargo clippy -p binius-prover --all-targets --all-features -- -D warnings` — clean
- nightly `cargo fmt` — clean
- `cargo test -p binius-prover --lib fold_word` — 8 passed, and again with `--features rayon`
- `cargo test -p binius-prover protocols::binmul` — 3 passed; `protocols::intmul` — 7 passed; both with and without `--features rayon`

The no-rayon shim is the default test path in this repo, so every test above was run in both configurations.

### DIFF
--- a/crates/prover/src/protocols/intmul/prove.rs
+++ b/crates/prover/src/protocols/intmul/prove.rs
@@ -44,7 +44,7 @@ use super::{
 	error::Error,
 	witness::{Witness, limb_index, two_valued_field_buffer},
 };
-use crate::fold_word::{fold_across_words, fold_words};
+use crate::fold_word::{WordFolder, fold_words};
 
 /// Proves the integer multiplication (IntMul) reduction over the four operand columns.
 ///
@@ -400,11 +400,13 @@ where
 		// verifier binds the stacked-index claim via the GF(2)-linearity of the embedding, the `b`
 		// evals via sum_i eq(r_I^b, i) * b(i, r_out) = B(r_out), and the parity bits directly.
 		let output_guard = tracing::debug_span!("Compute output bit evals").entered();
-		let per_bit_evals = |exponents: &[Word]| fold_across_words::<_, P>(exponents, r_out);
-		let a_evals = per_bit_evals(a_exponents);
-		let c_lo_evals = per_bit_evals(c_lo_exponents);
-		let c_hi_evals = per_bit_evals(c_hi_exponents);
-		let b_evals = per_bit_evals(b_exponents);
+		// All four columns fold against the same point, so the lookup tables and the per-chunk
+		// weights are built once and shared.
+		let folder = WordFolder::<F>::new(r_out);
+		let a_evals = folder.fold_par(a_exponents);
+		let c_lo_evals = folder.fold_par(c_lo_exponents);
+		let c_hi_evals = folder.fold_par(c_hi_exponents);
+		let b_evals = folder.fold_par(b_exponents);
 		drop(output_guard);
 
 		self.channel.send_many(&a_evals);
--- a/crates/prover/benches/fold_across_words.rs
+++ b/crates/prover/benches/fold_across_words.rs
@@ -1,34 +1,99 @@
 // Copyright 2026 The Binius Developers
 use binius_core::word::Word;
-use binius_field::arch::OptimalPackedB128;
 use binius_math::test_utils::random_scalars;
-use binius_prover::fold_word::fold_across_words;
+use binius_prover::fold_word::WordFolder;
 use binius_verifier::config::B128;
 use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
 use rand::prelude::*;
 
-fn bench_fold_across_words(c: &mut Criterion) {
+/// Word columns a single proof folds against one shared point.
+///
+/// The integer-multiplication protocol sends four, and binary multiplication sends six.
+/// Six is the wider case, so it is the one benched.
+const N_COLUMNS: usize = 6;
+
+/// One column folded against one point, with the two drivers side by side.
+///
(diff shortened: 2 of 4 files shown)


crates/prover/src/protocols/intmul/prove.rs
```diff
@@ -400,11 +400,13 @@ where
...
 		let output_guard = tracing::debug_span!("Compute output bit evals").entered();
-		let per_bit_evals = |exponents: &[Word]| fold_across_words::<_, P>(exponents, r_out);
-		let a_evals = per_bit_evals(a_exponents);
-		let c_lo_evals = per_bit_evals(c_lo_exponents);
-		let c_hi_evals = per_bit_evals(c_hi_exponents);
-		let b_evals = per_bit_evals(b_exponents);
+		// All four columns fold against the same point, so the lookup tables and the per-chunk
+		// weights are built once and shared.
+		let folder = WordFolder::<F>::new(r_out);
+		let a_evals = folder.fold_par(a_exponents);
```
jimpo: We could even do a rayon loop over these 4 operations.

PR #2267, opened 2026-08-24
Split the field buffer module by type, and name its chunk iterators

tcoratger:
Reorganises `binius-math`'s central container and gives its chunk machinery real types. No behaviour change — the 143 tests are byte-identical to before and all still pass.

### The problem

`field_buffer.rs` was one file of 1716 lines: 871 of code, 845 of tests.

It held four public types with no separation between them — the buffer, the store behind a shared view, and two write-back guards. Finding the guard that makes a sub-word split work meant scrolling past every constructor and every chunk accessor.

Three smaller things were tangled up in that:

- The four chunk iterators returned anonymous `impl Iterator`, so a caller could not name one, store one in a struct field, or rely on its length.
- Locating a chunk smaller than a packed word is a two-step index split. That arithmetic was written out **five times**: both single-chunk accessors, both parallel iterators, and the write-back guard's `Drop`.
- All 845 test lines sat in one flat `mod tests`.

### The idea

Split by **type**, not by concern. This is the shape `std` uses for `Vec`: `vec/mod.rs` is 4543 lines and carries `Vec` with every inherent impl, while each sibling file is a distinct type — `IntoIter`, `Drain`, `Splice`, `SetLenOnDrop`.

```
field_buffer/
    mod.rs         the buffer, every impl on it, and its tests
    chunks.rs      the chunk iterators and the sub-word chunk location
    view.rs        the borrowed aliases and the store behind a shared one
    write_back.rs  the two guards
```

The alternative — splitting the buffer's own inherent impls across files by concern (`build.rs`, `chunks.rs`, `split.rs`) — was tried first and is wrong. `clippy::multiple_inherent_impl`, which this workspace sets to `warn` and CI escalates with `-D warnings`, rejects it. That lint was right: a reader looking for `chunk_mut` should not have to guess which concern bucket it landed in. **This PR needs no `#[allow]` anywhere**, which is the signal the grouping is correct.

### Named chunk iterators

```diff
- pub fn chunks(&self, log_chunk_size: usize) -> impl Iterator<Item = FieldSlice<'_, P>> + Clone
+ pub fn chunks(&self, log_chunk_size: usize) -> Chunks<'_, P>

- pub fn chunks_mut(&mut self, log_chunk_size: usize) -> impl Iterator<Item = FieldSliceMut<'_, P>>
+ pub fn chunks_mut(&mut self, log_chunk_size: usize) -> ChunksMut<'_, P>
```

Both now carry `ExactSizeIterator`. `Chunks` keeps a hand-written `Clone`, so cloning it does not demand a `Clone` element type. This is the `slice::Chunks` / `slice::ChunksMut` shape.

The two **parallel** iterators deliberately keep `impl IndexedParallelIterator`. Naming them means hand-writing a rayon `Producer` with its own `split_at`, roughly 120 lines each. That is a bad trade here: a subtle bug in a producer split yields a wrong sum rather than a crash, and the gain would be file organisation only.

### One value object for the sub-word chunk location

A chunk smaller than a packed word does not get a word to itself — several share one, so a chunk is a run of lanes inside it:

```
WIDTH = 4, log_chunk_size = 1, so 2 chunks per word

chunk 0 -> word 0, lane 0     chunk 2 -> word 1, lane 0
chunk 1 -> word 0, lane 2     chunk 3 -> word 1, lane 2
```

That split was open-coded at five sites. It is now one type:

```rust
pub(super) struct SubWordChunk<P> {
    word_index: usize,    // which word holds the chunk
    lane_offset: usize,   // which lane its first element sits at
    log_len: usize,
    packing: PhantomData<P>,
}

impl<P: PackedField> SubWordChunk<P> {
    const fn new(log_chunk_size: usize, chunk_index: usize) -> Self;
    fn scalars(self, word: P) -> impl Iterator<Item = P::Scalar> + Send + Clone;
    fn repack(self, words: &[P]) -> P;
    fn merge_into(self, word: &mut P, chunk: &P);
}
```

The `PhantomData<P>` is load-bearing, not decoration. The arithmetic depends on `P::LOG_WIDTH`, so putting the packing in the type means a location can only ever be read against the packing it was computed for. The five open-coded copies could not prevent that mismatch.

Two consequences worth noting:

- The write-back guard now stores a `location` instead of loose `log_len` and `chunk_offset` fields, so its `Drop` is one call rather than a hand-rolled lane loop.
- After moving `scalars` and `merge_into` onto the type, the `lane_offset` accessor became dead and was deleted. Nothing outside the type needs to know a lane offset exists.

### Scope

- **new:** `chunks.rs` (`Chunks`, `ChunksMut`, `SubWordChunk`), `view.rs`, `write_back.rs`.
- **changed:** `field_buffer.rs` becomes `field_buffer/mod.rs`; git tracks it as a rename, so `--follow` keeps working.
- **public surface:** unchanged except that two return types gained names. The 37 public items are the same set, and `Chunks` / `ChunksMut` are additions.
- **tests:** moved inline with the impls they exercise, and renamed to drop the redundant `test_` prefix that `#[test]` already conveys.
- **left alone:** the two parallel iterators keep `impl` return types; no method was renamed; no signature changed beyond those two return types.

### What holds this to the old behaviour

This is a refactor of production code, so a line-for-line move proof does not apply. What does:

- **The 487 test lines are identical** — no test was changed, added, or weakened to accommodate the refactor. Verified by comparing the multiset of test-body lines against the original, modulo the `test_` prefix rename.
- The public item set is unchanged: 37 before, 37 after, diffed and identical.
- The partition proptest is the load-bearing one here. It sweeps `log_len` in `0..=6` and every chunk size within it, cross-checking the parallel scalar iterator against the serial accessor and asserting the chunks concatenate back to the buffer. That covers both sides of the packing width, which is exactly where `SubWordChunk` does its work.

### Verification

- `cargo clippy -p binius-math --all-targets --all-features -- -D warnings` — clean, no `#[allow]` added.
- `cargo test -p binius-math` — 143 passed, 0 failed, plus 1 doctest.
- Same again with `--features rayon` — 143 passed, 0 failed. Both configs matter, since the default build uses the hand-written no-rayon shim.
- `cargo test -p binius-ip-prover` — 97 passed, 0 failed.
- `cargo doc -p binius-math --no-deps --all-features` — zero warnings. This one earned its place: it caught module docs linking to private items.

### DIFF
--- a/crates/math/src/field_buffer/chunks.rs
+++ b/crates/math/src/field_buffer/chunks.rs
@@ -1,0 +1,196 @@
+// Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
+
+//! Iterators over the aligned chunks of a field buffer.
+//!
+//! A chunk of `2^k` elements starts at a multiple of its own size.
+//! Chunks of one size therefore tile the buffer exactly, with none left over.
+//!
+//! A chunk takes one of two shapes, decided by its size against the packing width:
+//!
+//! ```text
+//! chunk >= one packed word  ->  a run of whole words, borrowed from the store
+//! chunk <  one packed word  ->  some lanes of one word
+//! ```
+//!
+//! The iterators here cover the first shape, where a chunk is a borrowed run of words.
+//! The second shape gets a type of its own, since locating those lanes is shared work.
+
+use std::{iter, marker::PhantomData, slice};
+
+use binius_field::PackedField;
+
+use super::{FieldBuffer, FieldSlice, FieldSliceData, FieldSliceMut};
+
+/// The location of one chunk that is smaller than a packed word.
+///
+/// A chunk this small does not get a word to itself.
+/// Several of them share one word, so a chunk is a run of lanes inside it:
+///
+/// ```text
+/// WIDTH = 4, log_chunk_size = 1, so 2 chunks per word
+///
+/// chunk 0 -> word 0, lane 0     chunk 2 -> word 1, lane 0
+/// chunk 1 -> word 0, lane 2     chunk 3 -> word 1, lane 2
+/// ```
+///
+/// The chunk index splits in two to get there.
+/// High bits pick the word, low bits pick the lanes within it.
+///
+/// The packing width is part of the type.
+/// So a location can only ever be read against the packing it was computed for.
+#[derive(Debug, Clone, Copy)]
+pub(super) struct SubWordChunk<P> {
+	/// Which word of the backing store holds the chunk.
+	word_index: usize,
+	/// Which lane of that word the chunk's first element sits at.
+	lane_offset: usize,
+	/// Element count of the chunk, as a base-2 logarithm.
+	log_len: usize,
+	/// Ties the arithmetic above to one packing width.
+	packing: PhantomData<P>,
+}
+
+impl<P: PackedField> SubWordChunk<P> {
+	/// Locates the chunk at `chunk_index` among chunks of `2^log_chunk_size` elements.
+	///
+	/// The size must be below the packing width, which is what makes several chunks share a word.
+	#[inline]
+	pub(super) const fn new(log_chunk_size: usize, chunk_index: usize) -> Self {
+		let log_chunks_per_word = P::LOG_WIDTH - log_chunk_size;
+		let chunk_subindex = chunk_index & ((1 << log_chunks_per_word) - 1);
+		Self {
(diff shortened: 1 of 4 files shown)


jimpo (commented):
I'm not opposed to splitting up modules, necessary, but all the `pub(super)` is a code smell indicating that the split is not cohesive. I consider `pub(super)` an anti-pattern (it's discouraged in the CONTRIBUTING.md).

Is there a better way to organize this module so that the pub(super) stuff just becomes private?

PR #2275, opened 2026-08-24
Give the field buffer the standard iteration surface

tcoratger:
Gives the field buffer the iteration surface a `std`-grade container is expected to have. Additive — no existing signature changes, no behaviour changes.

### The problem

`FieldBuffer` is the central container in `binius-math`, and three standard things were missing:

- **No `IntoIterator`.** A scalar iterator existed as a named method, but the type could not be used in a `for` loop, in `collect`, or in anything generic over iterables.
- **No `FromIterator`.** Building one from scalars meant materialising a slice first and handing that over.
- **No packed-word iterator.** Callers reached for `as_ref().iter()` and `as_mut().iter_mut()`, which leaks the backing store one level out of the abstraction.

### The idea

```rust
pub struct Scalars<'a, P: PackedField> { .. }        // nameable, ExactSizeIterator, Clone

impl<..> IntoIterator for &FieldBuffer<P, Data> { type IntoIter = Scalars<'_, P>; }
impl<P: PackedField> FromIterator<P::Scalar> for FieldBuffer<P> { .. }

pub fn iter_packed(&self) -> slice::Iter<'_, P>;
pub fn iter_packed_mut(&mut self) -> slice::IterMut<'_, P>;
```

`FromIterator` packs into words as elements arrive, so no intermediate scalar vector is built, and it sizes the store from the iterator's lower size hint.

### Why iteration is reference-only

There is no by-value `IntoIterator`, and that is deliberate.

A scalar is `Copy` and is extracted out of a lane, so a consuming impl would yield exactly the copies the shared one yields while destroying the buffer — no gain, and it would make `for x in buf` silently move an owned buffer where `for x in &buf` does not. It is also meaningless for the borrowed-store aliases, where there is nothing to consume.

There is no `&mut` impl either: a packed lane cannot be lent out as a mutable reference to a scalar.

### Why there are now two scalar iterators

The existing named method goes through the packed field's own slice iterator. The new nameable type indexes the store instead. That duplication is forced, and worth stating plainly rather than leaving as a puzzle:

`IntoIterator::IntoIter` is an associated type, and an associated type cannot name an opaque `impl Iterator` return. So the nameable iterator cannot wrap the other one.

Collapsing them the other way would cost real speed. The packed field's slice iterator is genuinely specialized — the portable packed types walk their underliers directly, and scalar fields reduce to copying straight out of the slice. Both avoid a lane extraction per element, and the named method has roughly 45 hot call sites.

The clean fix is upstream: give that trait method a **named** return type instead of `impl Iterator`, and the new iterator becomes a thin newtype over it. That touches a widely implemented trait, so it belongs in its own change. Until then the tests cross-check that the two agree, above and below the packing width.

### Panics

`FromIterator` cannot be fallible, so it panics when the element count is not a power of two — empty included, matching the existing slice constructor. Both cases are documented under `# Panics` and pinned by a `#[should_panic]` test each.

### Scope

- **new:** `field_buffer/scalars.rs` — the scalar iterator type, its iterator impls, the borrowed-iteration impl, and their test. It is a distinct type, so it gets its own file, matching how the module already separates the chunk iterators, the views, and the write-back guards.
- **new:** the collect impl and the two packed-word accessors in `field_buffer/mod.rs`, in the impl blocks whose bounds already match, so no new inherent impl block and no interaction with `multiple_inherent_impl`.
- **changed:** 6 call expressions across 3 files inside `binius-math` migrated off `as_ref().iter()`, so the new methods land with users. None remain in the crate.
- **left alone:** other crates. A workspace-wide sweep is its own change.
- **left alone:** one `as_ref().par_iter()` site — a parallel packed iterator would be a separate addition.
- **left alone:** the existing named scalar method keeps its body and its ~45 call sites.

### Verification

- `cargo clippy -p binius-math --all-targets --all-features -- -D warnings` — clean. Two real hits were caught and fixed on the way: a redundant clone in a proptest assertion, and a move-after-use in the same test.
- `cargo test -p binius-math` — 150 passed, plus 1 doctest. Same again with `--features rayon`, since the default build uses the hand-written no-rayon shim. Baseline is 143, so the 7 new tests are all present.
- No doc comment sits above any `impl` block in the module — checked mechanically, since an `impl` block is a grouping mechanism and the prose belongs on the type or the method.
- `cargo doc -p binius-math --no-deps --all-features` — zero warnings.

### Tests

Seven added, each covering a sub-packing-width case, which is where this container's bugs live — below one word the store is a single word whose lanes past the logical length are not elements.

- Borrowed iteration in a `for` loop, through a generic consumer, and via the exact-size length; agreeing with the named method sub-width and at four words, including through a borrowed view.
- `FromIterator` round-tripped against the slice constructor as a proptest over 0 to 6 variables.
- `FromIterator` sub-width, asserting the three dead lanes come out zero.
- Panics on a non-power-of-two count, and on empty.
- The packed iterator yielding only live words — checked against a buffer with eight words of capacity but one live word, and again after truncation.
- The mutable packed iterator writing through, at two words and sub-width.


crates/math/src/field_buffer/scalars.rs
```diff
@@ -0,0 +1,134 @@
...
+	#[inline]
+	fn size_hint(&self) -> (usize, Option<usize>) {
+		let remaining = self.len - self.index;
+		(remaining, Some(remaining))
+	}
+}
+
+impl<P: PackedField> ExactSizeIterator for Scalars<'_, P> {}
+
+impl<'a, P: PackedField, Data: Deref<Target = [P]>> IntoIterator for &'a FieldBuffer<P, Data> {
```
jimpo, line 59: Please remove the `IntoIterator`. The `iter_scalars` method is fine. IntoIterator doesn't match the behavior on a Slice -- a slice's IntoIterator returns element references, not copied elements.

crates/math/src/field_buffer/mod.rs
```diff
@@ -705,6 +730,47 @@ impl<P: PackedField, Data: DerefMut<Target = [P]>> AsMut<[P]> for FieldBuffer<P,
...
+	/// The count is known only once the iterator runs dry, so a bad one can only panic.
+	#[track_caller]
+	fn from_iter<I: IntoIterator<Item = P::Scalar>>(iter: I) -> Self {
+		let iter = iter.into_iter();
+		// The lower bound is the whole count whenever the iterator knows its own length.
+		let mut values = Vec::with_capacity(iter.size_hint().0 >> P::LOG_WIDTH);
+
+		let mut word = P::default();
+		let mut len = 0usize;
+		for scalar in iter {
```
jimpo, line 749: This is inefficient. You can use `Itertools::chunks` and `PackedField::from_scalars` instead.

PR #2276, opened 2026-08-24
Carry the hypercube expansions as provided methods on the trait

tcoratger:
Deletes a 501-line module that was pure forwarding, and closes a footgun that could silently change which polynomial basis a computation uses. Pure refactor — no behaviour change.

### The problem

`crates/math/src/multilinear/hypercube.rs` is generic over which hypercube the coefficients are indexed by. Two instances exist: the Boolean cube with basis $(1 - X, X)$, and the infinity cube with basis $(1, X)$.

`multilinear/eq.rs` existed only to specialize the generic one to the Boolean cube. Its 179 non-test lines were **eleven public functions, every one a single-line forward**:

```
pub fn eq_ind_partial_eval<P>(point: &[P::Scalar]) -> FieldBuffer<P> {
    hypercube::eq_ind_partial_eval::<OneCube, P>(point)
}
```

Each carried a full doc comment that had to be kept in sync with the generic version's. The generic module's own header admitted the arrangement: *"The routines here mirror those in `super::eq`, which specialize them to `OneCube`."*

**The real cost is not the duplication.** The two modules exported eleven *identically named* functions, differing only in which basis they use. An unqualified import of the wrong one silently changes the basis a polynomial is expressed in — from $(1 - X, X)$ to $(1, X)$. That is a wrong proof, not a compile error, and nothing in the type system or the test suite would flag it.

### The idea

`std` specializes a generic through **provided methods on a trait**, and this trait already had one — the single-variable equality indicator carried a default body.

So the routines move up, and the specialization becomes free:

```
- multilinear::eq::eq_ind_partial_eval::<P>(point)              // one of two same-named functions
- multilinear::hypercube::eq_ind_partial_eval::<OneCube, P>(..) // two-parameter turbofish
+ OneCube::eq_ind_partial_eval::<P>(point)                      // the cube names itself
```

The basis is now part of the call, so there is no wrong import to make. The two-parameter turbofish is gone from the codebase.

### What stayed a free function, and why

One routine did **not** move: the private reserved-capacity form of the tensor expansion.

Every method of a public trait is public. Putting a private, precondition-bearing internal on the trait would leak it into the public API. It keeps its own generics and is called from two default bodies.

The trait also gained a `Sized` supertrait. A default body that passes `Self` as a type argument to that free function needs it, and the trait was never dyn-compatible anyway — every method is generic.

### Scope

- **deleted:** `multilinear/eq.rs`, 501 lines to zero.
- **changed:** `multilinear/hypercube.rs`, 693 to 961 lines — it absorbs the routines and the folded tests.
- **changed:** 63 files in total, across `binius-math`, `iop`, `iop-prover`, `ip`, `ip-prover`, `prover`, `verifier`, `m4-prover`, `recursion`, `spartan-prover`, `spartan-verifier`.
- **net:** 904 insertions, 1016 deletions.

Two call sites were not plain imports and needed hand work:

- One in the batch FRI code called a routine fully qualified rather than importing it.
- One in the shift verifier passes a routine **as a function value** to a higher-order helper. The trait's associated function works as a function item there, so no wrapper closure was needed.

One stale intra-doc link pointing at the deleted module was rewritten as plain English.

Every call site now needs the trait in scope for method resolution, so imports name the trait alongside the cube. That is the one ergonomic cost of the design.

### Soundness

No formula changed — the routines moved verbatim from free functions into default bodies, and the arithmetic is identical.

The risk in a change like this is not a crash but a **basis swap**: a call site migrating to the wrong cube would compile and produce a wrong result. Audited directly:

- The infinity cube appears nowhere outside its own module and one bench, so no migrated call site picked it up.
- Every migrated site came mechanically from the deleted module, which was the Boolean cube by definition.

### Tests

The deleted module's tests folded into the generic module, keeping the two cubes' coverage adjacent. 13 of 16 kept and prefixed by cube; three dropped as strictly subsumed, verified rather than assumed:

- A consistency check at one random index — the surviving version loops **every** index at the same arity.
- A scalar-versus-packed consistency check — subsumed by a proptest covering both cubes over 0 to 8 variables.
- A scaled-expansion check for the Boolean cube — merged into the same-named proptest, which now loops **both** cubes where it previously covered only the infinity one.

So coverage is strictly wider than before, and the count moves 143 to 140 exactly as −16 + 13.

### Verification

- `cargo clippy` with `--all-targets --all-features -- -D warnings` over all eleven touched crates — clean.
- `cargo test -p binius-math` — 140 passed, plus 1 doctest. Same again with `--features rayon`, since the default build uses the hand-written no-rayon shim.
- Downstream: `binius-ip` 46 + 2, `binius-verifier` 13, `binius-prover` 67 + 17 + 1, `binius-iop-prover` 10 — all 0 failed.
- `cargo doc --no-deps --all-features` on `binius-math`, `binius-ip`, `binius-verifier` — zero warnings. This gate earned its place: a broken intra-doc link to the deleted module shows up here and nowhere else.


jimpo (commented):
Please revert, this makes all the calling code look worse.

PR #2283, opened 2026-08-24
Make bit-reversal a method on the mutable buffer view

tcoratger:
Turns a free function whose only argument is the thing it mutates into a method on that thing. Pure refactor — the permutation is untouched.

### The problem

```rust
pub fn bit_reverse_packed<P: PackedField>(buffer: FieldSliceMut<P>)
```

One argument, mutated in place, nothing returned. The buffer is the subject, not a parameter — this is a method that never got written.

Every one of the ten call sites reads the same way:

```
bit_reverse_packed(tensor.to_mut())
```

### The change

```diff
- bit_reverse_packed(tensor.to_mut())
+ tensor.to_mut().bit_reverse()
```

The name also loses a word it no longer needs to carry. "Packed" distinguished this from the index-permuting variant on a plain slice; as a method on the packed-buffer view, the receiver already says it.

### Why a method and not a trait

This is a **container permutation**, not polynomial algebra, so it belongs as an inherent method rather than on an extension trait.

The test that settles it: the operation is meaningful for a buffer that is not a multilinear at all. Two of its callers are the Reed–Solomon encoder, permuting a codeword.

### Where the impl block lives

In `bit_reverse.rs`, next to the tiled algorithm it dispatches to — **not** in `field_buffer/view.rs`, which already carries an inherent impl on the same view type.

Folding it into that existing block would separate the dispatch from what it dispatches to. The entry point's whole job is choosing a tile width from cache-line reasoning; the arms it chooses between are twenty lines below it in this file. It would also force two private helpers to become `pub(crate)` for no gain.

### On `multiple_inherent_impl` — checked, not assumed

This adds `impl<P: PackedField> FieldSliceMut<'_, P>` while `field_buffer/view.rs` already has an identically parameterised block. The workspace sets `multiple_inherent_impl = "warn"` and CI runs clippy with `-D warnings`, so the obvious worry is that this fails the build.

It does not — and the silence was verified rather than trusted. A throwaway non-generic struct with two inherent impls was added to the same file, and clippy did emit `multiple implementations of this structure`. So the lint is genuinely live in this crate; it simply does not apply to generic inherent impls. The probe was removed before committing.

### Scope

- **changed:** `bit_reverse.rs` — the free function becomes an inherent method; the private helpers keep their names and signatures, since "packed" still describes packed field elements there rather than the deleted public name.
- **changed:** 10 call sites across `reed_solomon.rs`, `multilinear/hypercube.rs`, `ip-prover/src/sumcheck/switchover.rs`, this file's own tests, and the bench.
- **changed:** the bench's case label, since the name it quoted no longer exists.
- **not kept:** no free-function form remains. Every caller took method syntax.

Two smaller decisions:

- The receiver is `&mut self`, not `self`. Every other mutating method on the buffer takes `&mut self`, and by-value would consume a view the caller may still want. Each dispatch arm reborrows internally.
- The private helpers still take their view by value, which is what the recursion wants.

### Soundness

Bit reversal is a permutation used inside the NTT and FRI paths, so a wrong index corrupts a codeword silently rather than panicking. The existing tests are the guard, and they still run:

- Tiled-versus-naive equivalence across nine lengths, covering every tile branch: 0, 1, 3, 4, 7, 8, 9, 12, 13.
- A proptest that the tiled path matches the naive reference.
- A proptest that the permutation is an involution.
- The tile transpose against a scalar transpose.

The naive reference is untouched by this change, so it remains an independent oracle: the method under test is the permuting side, and a wrong index shows up as a mismatch.

### Verification

- `cargo clippy -p binius-math -p binius-ip-prover --all-targets --all-features -- -D warnings` — clean.
- `cargo test -p binius-math` — 147 passed, plus 1 doctest. Same again with `--features rayon`, since the default build uses the hand-written no-rayon shim. Baseline is 147, and no test was added or removed.
- `cargo test -p binius-ip-prover` — 93 passed.
- `cargo check -p binius-math --benches` — clean, since the bench is a call site.
- `cargo doc -p binius-math --no-deps --all-features` — zero warnings.

### DIFF
--- a/crates/ip-prover/src/sumcheck/switchover.rs
+++ b/crates/ip-prover/src/sumcheck/switchover.rs
@@ -9,7 +9,6 @@ use std::{
 use binius_field::{Field, PackedField};
 use binius_math::{
 	FieldBuffer, FieldSlice,
-	bit_reverse::bit_reverse_packed,
 	multilinear::{
 		fold::{binary_fold_high, fold_highest_var_inplace},
 		hypercube::{Hypercube, OneCube},
@@ -126,9 +125,9 @@ where
 			// Prepend the new variable via bit-reverse + append + bit-reverse. This does not need
 			// to be fast: it runs once per pre-switchover round on a small tensor (see
 			// BINIUS-327).
-			bit_reverse_packed(tensor.to_mut());
+			tensor.to_mut().bit_reverse();
 			let mut tensor = OneCube::tensor_prod_eq_ind(tensor, &[challenge]);
-			bit_reverse_packed(tensor.to_mut());
+			tensor.to_mut().bit_reverse();
 			self.tensor = tensor;
 
 			if self.tensor.log_len() == self.switchover {
--- a/crates/math/benches/bit_reverse.rs
+++ b/crates/math/benches/bit_reverse.rs
@@ -4,7 +4,7 @@ use binius_field::{
 	BinaryField, PackedBinaryGhash1x128b, PackedBinaryGhash2x128b, PackedBinaryGhash4x128b,
 	PackedField,
 };
-use binius_math::{bit_reverse::bit_reverse_packed, test_utils::random_field_buffer};
+use binius_math::test_utils::random_field_buffer;
 use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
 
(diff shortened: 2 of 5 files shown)


jimpo (commented):
Please revert -- general guidance about cohesive struct interfaces. standalone function is better.

PR #2297, opened 2026-08-24
Bring the field buffer's names up to std convention

tcoratger:
Renames seven pieces of the field buffer's surface to match `std` convention, and establishes one vocabulary rule. Pure refactor — no behaviour change, no signature change beyond the names.

### The problem

Several names said something the code does not do.

- **`to_ref` / `to_mut`.** In `std`, `to_*` is a **clone-producing** conversion — `to_owned`, `to_vec`, `to_string`. A cheap borrow is `as_*`. Worse, `Cow::to_mut` genuinely clones, so `to_mut` actively reads as "may allocate" on a method that only borrows.
- **`take_data`.** In `std`, `take` means "replace with `Default` and return" — `Option::take`, `mem::take`. This consumes `self`, which is `into_inner`.
- **`clone_from_slice`.** Collides head-on with `slice::clone_from_slice`, which copies *into* an existing slice. This is a constructor.
- **`chunks_par`.** Its sibling is `par_chunk_scalars`, and rayon's convention is the `par_` prefix. Two orderings for one idea.
- **`FieldBufferSplitMut`.** The prefix restates the module: `field_buffer::FieldBufferSplitMut`.
- **`split_half_ref` / `split_half` / `split_half_mut`.** Three names where `_ref` marks the `&self` method, inverting the `std` convention that the bare name *is* the shared one.
- **The `values` field.** The store holds packed words. `values` is what the scalar constructors take.

### The vocabulary rule this settles

The obvious rename of `to_ref` is `as_slice`, and it is **wrong** — that was caught during review rather than after.

`FieldBuffer` already implements `AsRef<[P]>`, so `as_ref()` returns `&[P]`. An `as_slice()` returning a `FieldSlice` would put two different return types behind one word, and break `std`'s promise that `as_slice` yields `&[T]`. The rename would have read *less* `std`-like than what it replaced, while appearing to fix it. The same collision showed up a second time: `from_slice_in(alloc, src: FieldSlice)` next to `FieldSlice::from_slice(log_len, &[P])`, where "slice" means a view in one and raw words in the other.

So the rule is: **"slice" always means a raw `&[P]`; "view" always means a `FieldSlice`.** No name means both.

```
to_ref()            ->  as_view()          returns FieldSlice
to_mut()            ->  as_mut_view()      returns FieldSliceMut
as_ref()            ->  unchanged          returns &[P]
clone_from_slice()  ->  from_view_in()     takes a FieldSlice
FieldSlice::from_slice()  ->  unchanged    takes a raw &[P]

take_data()         ->  into_inner()
chunks_par()        ->  par_chunks()
FieldBufferSplitMut ->  SplitMut
values (field)      ->  words

split_half_ref(&self)     ->  split_half(&self)
split_half(self)          ->  into_split_half(self)
split_half_mut(&mut self) ->  unchanged
```

The rule is now visible in the code: one line reads `FieldBuffer::new(4, packed_values.as_slice())` — a `Vec<P>` handing over raw words — a few hundred lines from `buffer.as_mut_view()` handing over a borrowed buffer. Those used to look like the same operation.

### The halving trio, and how the risky rename was fenced

Renaming `split_half_ref` to `split_half` while a consuming `split_half` still exists means calls can silently resolve to the wrong method. Rather than one simultaneous edit, the order does the policing:

1. Rename the **consuming** method first. That leaves *no* inherent `split_half` on the buffer at all, so the compiler must flag the one remaining call site — inside `split_half_mut`.
2. `cargo check --workspace --all-targets --all-features` at that intermediate point. Clean.
3. Then rename the borrowing method into the freed name.

Same guarantee as an atomic edit, with a machine check in the middle.

**A textual count would have been misleading.** `split_half` appears on three unrelated types in the sumcheck store, plus two `split_half_claim` helpers. The discriminator is the signature: the buffer's method returns a tuple destructured `(a, b)`, while all three foreign ones return `[Self; 2]` destructured `[a, b]`. Final tally at the intermediate point — 9 foreign mentions, 2 pre-existing intra-doc links, 1 test name, and zero buffer ones, which is what forced the compiler's hand.

### One rename deliberately skipped

The plan also proposed `chunk(log_size, i)` -> `chunk_at(i, log_size)`, renaming **and** swapping the argument order. Not done, on purpose.

Both parameters are `usize`, so a swapped call compiles and silently returns the wrong chunk — a wrong witness, not a panic. Putting the one change that can fail silently inside ~270 mechanical changes that cannot makes it invisible to review. If the argument order is worth fixing, it deserves a PR where it is the only thing under review.

`pub fn chunk(&self, log_chunk_size: usize, chunk_index: usize)` is untouched, and no diff line touches its signature.

### Scope

Seven commits, one rename group each, 56 files, +327/−323. Every commit compiles and is individually `fmt`-clean; the rustfmt reflows a longer name causes sit in the commit that causes them.

The `values` -> `words` rename was *almost* module-private: the field and its locals are, but the name also appeared in a public constructor's parameter, its rustdoc, and two panic messages, so those moved too. `values` is deliberately kept wherever it genuinely names a scalar slice — `from_values`, `from_values_in` — since that contrast is what makes the pair readable.

### Verifying that no rename is half-applied

Combined hits for every removed name — `to_ref`, `to_mut`, `clone_from_slice`, `from_slice_in`, `take_data`, `chunks_par`, `FieldBufferSplitMut`, `split_half_ref`: **zero**.

`as_slice` and `as_mut_slice` cannot be zero, since `Vec::as_slice` predates this work. So the stronger property was checked instead: their workspace counts are byte-identical to base, and the diff touches **zero** lines containing either token. No pre-existing `Vec` caller was collateral damage.

Each new name has exactly one definition, all three on the buffer, and none existed at base.

### Verification

- `cargo clippy` with `--all-targets --all-features -- -D warnings` across the nine touched crates — clean.
- `cargo test -p binius-math` — 150 passed, plus 1 doctest; same with `--features rayon`, since the default build uses the hand-written no-rayon shim. Matches the baseline exactly.
- The other eight touched crates — all green.
- `cargo check --workspace --all-targets --all-features` — clean. Justified here, since public method names change workspace-wide.
- `cargo test -p binius-math --target wasm32-wasip1 --no-run` — compiles. CI runs wasm with default features, which an `--all-features` gate misses.

### DIFF
--- a/crates/math/src/field_buffer/mod.rs
+++ b/crates/math/src/field_buffer/mod.rs
@@ -296,7 +296,7 @@ impl<P: PackedField, Data: Deref<Target = [P]>> FieldBuffer<P, Data> {
 	}
 
 	/// Borrows the whole buffer as a shared view.
-	pub fn to_ref(&self) -> FieldSlice<'_, P> {
+	pub fn as_view(&self) -> FieldSlice<'_, P> {
 		FieldSlice::from_slice(self.log_len, self.as_ref())
 	}
 
@@ -546,7 +546,7 @@ impl<P: PackedField, Data: Deref<Target = [P]>> FieldBuffer<P, Data> {
 
 impl<P: PackedField, Data: DerefMut<Target = [P]>> FieldBuffer<P, Data> {
 	/// Borrows the whole buffer as a mutable view.
-	pub fn to_mut(&mut self) -> FieldSliceMut<'_, P> {
+	pub fn as_mut_view(&mut self) -> FieldSliceMut<'_, P> {
 		FieldSliceMut::from_slice(self.log_len, self.as_mut())
 	}
 
@@ -645,7 +645,7 @@ impl<P: PackedField, Data: DerefMut<Target = [P]>> FieldBuffer<P, Data> {
 	/// * `self.log_len()` must be greater than 0.
 	#[track_caller]
 	pub fn split_half_mut(&mut self) -> FieldBufferSplitMut<P, &'_ mut [P]> {
-		self.to_mut().split_half()
+		self.as_mut_view().split_half()
 	}
 }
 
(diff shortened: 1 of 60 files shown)


crates/math/src/field_buffer/mod.rs
```diff
@@ -83,8 +83,8 @@ pub use write_back::FieldBufferSplitMut;
 pub struct FieldBuffer<P: PackedField, Data: Deref<Target = [P]> = Vec<P>> {
 	/// log2 the number over elements in the buffer.
 	log_len: usize,
-	/// The packed values.
-	values: Data,
+	/// The packed words.
+	words: Data,
```
jimpo, line 87: This is super confusing. Rename to `data` or `packed_values`. "words" has a completely different meaning in this codebase.

PR #2350, opened 2026-08-26
BINIUS-507: Build the combined round buffer by concatenation

tcoratger:
Follow-up to [BINIUS-507](https://linear.app/jimpo/issue/BINIUS-507) and #2209, addressing [this review comment](https://github.com/binius-zk/binius64/pull/2209#discussion_r3866834008).

Builds a round's combined buffer by appending each oracle's packed words once, instead of allocating a zeroed buffer and writing it back scalar by scalar.

### The problem

`MergeProverChannel::flush` did two avoidable things per round:

- `FieldBuffer::zeros_in` zeroed the whole combined buffer, all $2^N$ elements of it
- `place_block` then wrote every oracle back into it one scalar at a time, through `get`/`set`

So each element of the round was touched twice, the second time through the slowest available path — a per-scalar lane extract and insert, rather than a word copy.

### The idea

The sort that makes the merge work already makes the copy trivial.

Oracles are laid out largest to smallest, and every size is a power of two. So the oracles at least one packed word wide form a **prefix** of the round, and each of those occupies whole words and nothing else:

```
combined store, in packed words:

  [ oracle 1's words | oracle 2's words | ... | zero padding ]
    <-- one extend_from_slice each -->
```

Their word runs therefore concatenate into exactly the combined layout. That means: allocate with capacity, `extend_from_slice` once per oracle, and wrap the store in a `FieldBuffer` at the end. No prior zero fill, no scalar loop, no element written twice.

### Why a tail remains

Oracles narrower than one packed word **share** a word, so appending their runs would put each in a word of its own — at the wrong offset.

Sorted largest first, those oracles are exactly the tail of the round. So the tail keeps the scalar placement, and only the tail:

```
n_whole = oracles with log_len >= P::LOG_WIDTH      // the appendable prefix
words.resize(n_words, P::zero())                    // padding + the words the tail shares
place_block(...) for order[n_whole..]               // scalar path, tail only
```

Under `PackedBinaryGhash1x128b` the width is one element, so the tail is always empty and every oracle takes the fast path.

### The change

```
-let mut combined = FieldBuffer::zeros_in(&self.alloc, combined_log_len);
-for &k in &order {
-    place_block(&mut combined, self.pending[k].as_view(), block_index);
-}
+let mut words = self.alloc.alloc::<P>(n_words);
+for &k in &order[..n_whole] {
+    words.extend_from_slice(self.pending[k].as_ref());
+}
+words.resize(n_words, P::zero());
+let mut combined = FieldBuffer::new(combined_log_len, words);
+for &k in &order[n_whole..] {
+    place_block(&mut combined, self.pending[k].as_view(), block_indices[k]);
+}
```

Placement bookkeeping is unchanged — the block indices are still computed from the same running offset, in the same sorted order.

### Scope

- **changed:** `flush` in `crates/iop-prover/src/channel/merge.rs`
- **kept:** `place_block`, now reached only by the sub-word tail
- **changed test:** `round_trip_proptest` also runs under `PackedBinaryGhash4x128b`, where random round shapes mix word-aligned oracles with sub-word ones and the two paths meet in the same round
- **left untouched:** the verifier side, and the copy in `send_oracle` — dropping that one needs the round schedule known up front, which is [the separate restructuring](https://github.com/binius-zk/binius64/pull/2209#discussion_r3866854716)

### Verification

- `cargo check -p binius-iop -p binius-iop-prover --all-targets` — clean
- `cargo test -p binius-iop-prover channel::merge` — 7 tests pass, including the round trip under both packing widths

No benchmark: the copy is not on a measured path yet, since nothing wires the decorator into a real protocol driver. The win is a zero fill and a per-scalar loop removed per round, replaced by one `memcpy` per oracle.


benji-potamus (approved):
Approve.

Mechanical pass clean after rebasing onto `main`: nightly `cargo fmt --check`, `cargo clippy -p binius-iop-prover --all-targets --all-features -- -D warnings`, and `typos` all pass with no issues. `cargo nextest run -p binius-iop-prover -E 'test(channel::merge)'` is 7/7 green, including the round-trip proptest under both `PackedBinaryGhash1x128b` and the newly-added `PackedBinaryGhash4x128b`.

I traced the offset arithmetic by hand to check the concatenation lines up with what `place_block` would have written: since `order` is sorted descending by `log_len` and every size is a power of two, the running scalar `offset` before placing block `k` is always an exact multiple of `2^n_k` (existing invariant), so `block_index_k * 2^n_k == offset`. Restricting to the whole-word prefix (`n_k >= P::LOG_WIDTH`), dividing through by the packed width gives the word-offset `block_index_k * 2^(n_k - LOG_WIDTH)` equal to the sum of word-counts of the buffers appended before it — exactly where `extend_from_slice` places it. The tail (buffers narrower than a packed word) is untouched and still goes through `place_block`, so the two paths meet correctly at `n_whole`. `FieldBuffer::new`'s length assert (`words.len() == 1 << log_len.saturating_sub(LOG_WIDTH)`) backstops this at runtime, and the resize-then-construct ordering means it can only fire on an actual bug, not a benign case.

No blocking findings.

jimpo (approved):

PR #2352, opened 2026-08-26
BINIUS-552: Prove and verify the recursive circuit

tcoratger:
Closes [BINIUS-552](https://linear.app/jimpo/issue/BINIUS-552).

Adds `RecursiveCircuit`: builds the circuit that verifies a Binius64 proof, fills its witness, and — new here — lets that circuit be proved and verified like any other. Also measures what one more level costs.

### The problem

`crates/recursion` stopped one step short of recursion.

It built a circuit that verifies a proof, replayed the proof to fill the witness, and checked the circuit was satisfied. Then it stopped.

Nothing ever proved that circuit. So the pipeline produced a *satisfied circuit*, never a *proof of a verification* — and recursion is exactly the second thing.

Two consequences:

- No depth-2 anything. Nothing verified a proof of the recursive circuit, so nobody knew whether a level composes.
- The glue lived inside `tests/recursion_end_to_end.rs`. A caller outside the crate had to reimplement channel wrapping, statement observation, public binding, and replay by hand.

### The idea

Split the pipeline along the seam the protocol already has.

```
inner shape      ->  build    ->  circuit
circuit + proof  ->  witness  ->  values that satisfy it
```

The seam matters: **`build` needs only the shape, never a proof.** A shape is the constraint system, the FRI parameters, the oracle specs — everything fixed before a proof exists. No length in the protocol depends on a received value, so one circuit verifies *every* proof of that shape.

And what `build` returns is an ordinary circuit, so it has an ordinary `Verifier`. Building again over that verifier is one level deeper, for free:

```
Verifier(inner) -> build -> circuit_1 -> Verifier(circuit_1) -> build -> circuit_2
```

That is what makes the depth-2 measurement a short test rather than a project.

### The change

```
- let mut channel = compiler.create_channel(Binius64BuilderChannel::new());   // in a test
- let statement = channel.observe_words(witness.inout());                     // by hand
- ... verify, check_symbolic, finish, bind_public, new_witness_filler, replay ...
+ let recursive = RecursiveCircuit::build(verifier)?;
+ let witness = recursive.witness(inout, proof)?;
```

### What it measures

Depth 1 proves and verifies: inner proof 34,688 bytes to outer proof 286,816 bytes, both carrying the same statement.

Depth 2 builds:

| level | AND | BMUL | ZERO | rows |
|---|---|---|---|---|
| inner CRC-64 | 256 | 0 | 508 | 764 |
| depth 1 | 403,996 | 365,235 | 109,708 | 878,939 |
| depth 2 | 2,876,045 | 8,258,066 | 1,850,080 | 12,984,191 |

**One level costs 14.77x the constraint rows of the level it verifies.**

That is divergence, and it is expected. The wiring ("monster") claim is discharged in-circuit, at a cost proportional to the inner system — measured separately at 6.0 BMUL per inner constraint row, stable to 1% over a 256x range. A tower closes only when the ratio sits at or below one, so no fixed point exists while the claim is paid inline.

The test pins a **floor**, not a ceiling (`growth > 5.0`), so the day deferral lands it fails and says why. It is `#[ignore]`d because it builds a 50M-gate circuit.

### Soundness

No protocol change. Three points the new API relies on:

- **`build`'s placeholder statement is sound.** On the builder channel, `observe_words` reads its argument's *length* and never its values — it allocates one witness wire per word and absorbs the wires. The statement enters as circuit input, not a baked constant, which is what keeps one circuit good for every proof of the shape. A test asserts the recorded `observe_words` count equals the inout length.
- **The statement's length is not guessed.** It is the inner system's `n_inout`, the same value `IOPVerifier::verify` rejects a mismatch against, so a wrong length fails in `build` rather than silently.
- **`fill` returns an unchecked filler by design**, so negative tests can perturb what the replay wrote. Its docs state the caller owes `populate_wire_witness`; `witness` is the checked path.

### Scope

- **new:** `crates/recursion/src/circuit.rs` — `RecursiveCircuit` with `build`, `witness`, `fill`, and accessors
- **changed:** `binius-verifier` moves from `dev-dependencies` to `dependencies`. Recording the verifier is what this crate is for. Acyclic and contained: nothing in the workspace depends on `binius-recursion`
- **changed:** the end-to-end test drives the public API, so no crypto glue is duplicated between library and test
- **new tests:** prove-and-verify, depth-2 growth (ignored), wrong-length statement, malformed proof
- **untouched:** the channels, the challenger, the Merkle gadgets, and every existing negative test

### Drive-by, and why it is here

The second commit adds one semicolon in `crates/iop-prover/src/channel/merge.rs` — a file this PR otherwise does not touch.

`clippy::semicolon_if_nothing_returned` fires on a `const { assert!(..) }` block in `multi_round_round_trip_narrow_packing`, and CI lints with `-D warnings`. It is pre-existing on `main` (this branch is based on `0c1bfac5`) and it fails clippy for **every** PR against main, this one included. The lint only appears under `--all-targets`, since the block sits in a test module, which is how it reached main.

Happy to split it into its own PR if you would rather keep this one single-purpose — say so and I will.

### Verification

- nightly `cargo fmt --all --check`, `cargo clippy -p binius-recursion --all-targets`, `cargo doc -p binius-recursion --no-deps` — all clean
- `cargo test -p binius-recursion` — 31 tests pass (7 in the end-to-end suite, 1 ignored by design)
- the ignored depth-2 test run explicitly — passes, reporting the 14.77x above
- `cargo clippy -p binius-recursion --all-targets --all-features -- -D warnings` — clean, i.e. this crate passes under CI's exact flags
- **Not verifiable locally:** `--workspace` anything, and clippy on `binius-iop-prover`. No C compiler is installed here, so `alloca`'s build script fails and those targets cannot build. The semicolon fix above therefore rests on CI, not on a local run — it is the exact lint, at the exact line, CI reported, and it is the only occurrence of the pattern in the workspace. Nothing depends on `binius-recursion`, so the dependency move cannot reach another crate.

### DIFF
--- a/crates/recursion/src/circuit.rs
+++ b/crates/recursion/src/circuit.rs
@@ -1,0 +1,265 @@
+// Copyright 2026 The Binius Developers
+
+//! The circuit that verifies a Binius64 proof, built once per inner shape.
+//!
+//! ```text
+//!   inner shape      ->  build    ->  circuit
+//!   circuit + proof  ->  witness  ->  values that satisfy it
+//! ```
+//!
+//! A shape is everything the protocol fixes before any proof exists: the constraint system, the
+//! FRI parameters, the oracle specs. No length in the protocol depends on a received value, so
+//! one circuit verifies every proof of that shape.
+//!
+//! [`RecursiveCircuit::build`] therefore reads a [`Verifier`] and never a proof.
+//! A proof is needed only to fill a witness.
+//!
+//! # Going deeper
+//!
+//! What `build` returns is an ordinary circuit, so it has an ordinary [`Verifier`].
+//! Building again over that verifier is one level deeper.
+//!
+//! ```text
+//!   Verifier(inner) -> build -> circuit_1 -> Verifier(circuit_1) -> build -> circuit_2
+//! ```
+//!
+//! Every level pays for the level below it in full, because the wiring claim is discharged
+//! in-circuit rather than deferred. Measured over a CRC-64 inner circuit, a level costs about
+//! fifteen times the constraint rows of the level it verifies, so the tower diverges instead of
+//! closing. Deferring that claim is what makes a fixed point possible; see `RECURSION_PLAN.md`.
+
+use std::iter;
+
+use binius_core::{constraint_system::ValueVec, word::Word};
+use binius_frontend::{Circuit, PopulateError, Wire, WitnessFiller};
+use binius_hash::StdHashSuite;
+use binius_ip::channel::WordIPVerifierChannel;
+use binius_transcript::VerifierTranscript;
+use binius_verifier::{Pcs, Verifier, config::StdChallenger};
+
+use crate::{Binius64BuilderChannel, Recorded, WitnessFillerChannel};
+
+/// Why a recursive circuit could not be built, or its witness could not be filled.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+	/// The inner trace is opened with a scheme the in-circuit gadgets do not express.
+	#[error("the inner trace is opened with {pcs:?}, and only FRI is expressed in-circuit")]
+	UnsupportedPcs {
+		/// The scheme the inner verifier was set up with.
+		pcs: Pcs,
+	},
+
+	/// The statement handed to [`RecursiveCircuit::witness`] is the wrong length.
+	#[error("the inner statement holds {expected} words, and {actual} were supplied")]
+	StatementLength {
+		/// The inner constraint system's inout count.
+		expected: usize,
+		/// What the caller supplied.
+		actual: usize,
+	},
+
+	/// Running the verifier over a recursion channel failed.
+	#[error("verifier error: {0}")]
+	Verifier(#[from] binius_verifier::Error),
+
+	/// The oracle layer rejected the run.
+	#[error("IOP channel error: {0}")]
+	IOPChannel(#[from] binius_iop::channel::Error),
+
+	/// The replayed values leave the recorded circuit unsatisfied.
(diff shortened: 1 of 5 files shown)


crates/recursion/src/circuit.rs
```diff
@@ -0,0 +1,265 @@
...
+	/// rather than paying for a clone this constructor does not need.
+	///
+	/// # Errors
+	///
+	/// Returns [`Error::UnsupportedPcs`] when the inner trace is not opened with FRI.
+	///
+	/// Returns [`Error::Verifier`] when the shape itself is unsatisfiable, which a build-time
+	/// constant assertion decides before any proof exists.
+	pub fn build(verifier: Verifier<StdHashSuite>) -> Result<Self, Error> {
+		let Some(compiler) = verifier.iop_compiler().as_basefold() else {
```
tcoratger, line 114: We can later check for Ligerito etc but for now basefold to simplify

benji-potamus (approved):
Approve.

`RecursiveCircuit` is the right cut: splitting `build` (shape only) from `witness`/`fill` (shape + proof) is exactly what lets one circuit serve every proof of a shape and makes depth-2 a couple dozen lines instead of a project. The API replaces the hand-rolled channel wiring in the end-to-end test with a single type, and the new tests (wrong-length statement, malformed/truncated proof, disagreeing public input, actual prove-and-verify round trip) cover the seams that matter. The depth-2 divergence measurement is a nice touch — pinning a floor rather than a ceiling means a future fix that closes the tower fails loudly instead of silently.

The drive-by semicolon fix in `merge.rs` is moot after rebase — that line's already on `main`.

jimpo (approved):
Nice!

PR #2353, opened 2026-08-26
BINIUS-553: Transpose the Blake3 message with an AVX-512 shuffle network

tcoratger:
Closes [BINIUS-553](https://linear.app/jimpo/issue/BINIUS-553).

Moves the Blake3 multi-lane message transpose from scalar loads into an AVX-512 shuffle network.
Output is bit-identical to `blake3::hash` — no protocol change, no new soundness assumption.

### The problem

The kernel holds the message as 16 words of `N` lanes, one lane per message.
The input arrives the other way round: `N` blocks of 16 words, one block per lane.
Converting between the two layouts is a 16x16 transpose of 32-bit words.

`load_block_words` does that transpose one word at a time.
Per 64-byte block it reads 256 words and scatters them across 16 separate rows.

That runs **once per compression** — on every leaf block and every Merkle inner node.
On an AMD Zen 5 it is about a fifth of the kernel's runtime.

### The idea

At 16 lanes the square is exactly 16 rows of one 512-bit register.
So a fixed shuffle network can move it entirely in registers, with no memory round trip.

The network walks the transpose one power of two at a time:

```
stage 1: 32-bit words,  rows 1 apart    vpunpckldq  / vpunpckhdq
stage 2: 64-bit pairs,  rows 2 apart    vpunpcklqdq / vpunpckhqdq
stage 3: 128-bit lanes, rows 4 apart    vshufi32x4
stage 4: 128-bit lanes, rows 8 apart    vshufi32x4
```

### Per 64-byte block

- **before:** 256 scalar loads, then 256 scattered stores
- **after:** 16 vector loads, 64 shuffles, 16 vector stores

### How this relates to `zooko/blake3-sme2` — and what it is *not*

The idea is borrowed from that repo, where SME2 gets this same transpose **for free**: it writes 16 blocks as rows into a ZA tile and reads the columns straight back out.

- This PR ports the *observation* — that the transpose is a real cost in a multi-buffer Blake3 — not the code.
- That code is ARM SME2 assembly for Apple M4; AVX-512 has no ZA tile, so the shuffle network is the closest equivalent.
- Nothing about the compression core changes here. Only the message load does.

### The change

```
- fn load_block_words(..)          // byte-wise loader, the only path
+ fn load_block_words(..)          // dispatches on lane count
+ fn load_block_words_portable(..) // byte-wise loader, now the tested reference
```

The new module is gated on `target_feature = "avx512f"` and claims only 16 lanes, mirroring the existing aarch64 NEON hook.
16 is the width `Blake3HashSuite` ships, and the only lane count that makes the block square.

### Soundness

A transpose only permutes words, so the compression input is unchanged.
Three tests pin the network equal to the byte-wise loader, and the pre-existing `blake3::hash` equality tests now route through it.

### Scope

- **new:** `crates/hash/src/blake3_avx512.rs`
- **changed:** `load_block_words` in `blake3_portable.rs` becomes a dispatcher
- **left untouched:** the compression core, the aarch64 NEON path, and every lane count other than 16 — all still take the byte-wise loader

### Verification

- `cargo clippy -p binius-hash --all-targets`, run **both** with and without `avx512f` — clean
- `cargo +nightly-2026-07-01 fmt --all --check` and `tools/check_copyright_notice.py` — clean
- `cargo test -p binius-hash --lib`: 30/30 with AVX-512; the 8 Blake3 tests still pass with the module cfg'd out
- New tests: a fixed-random equality check, a coordinate-counting block that catches a swapped row or lane, and a proptest

### Benchmark

The repo's own `blake3_portable` bench, single-threaded and pinned to one core so the kernel is not masked by memory bandwidth, built with `-C target-cpu=native`.
`portable_16` is the path this PR touches; `portable_4` is untouched and acts as a control.

| leaf (elems) | before | after | change |
|---|---|---|---|
| 1 | 812.2 MiB/s | 964.1 MiB/s | +18.6% |
| 2 | 1.535 GiB/s | 1.805 GiB/s | +17.7% |
| 3 | 2.208 GiB/s | 2.583 GiB/s | +17.4% |
| 4 | 2.802 GiB/s | 3.218 GiB/s | +14.9% |
| 8 | 3.163 GiB/s | 3.657 GiB/s | +15.7% |
| 16 | 3.405 GiB/s | 3.989 GiB/s | +17.1% |
| 32 | 3.514 GiB/s | 4.116 GiB/s | +17.1% |
| 64 | 3.577 GiB/s | 4.186 GiB/s | +17.2% |

The control moved +0.1% to +5.7% over the same runs, so read the real effect as roughly +12% to +17%.

Two honest caveats:
- On the default multi-threaded bench the win largely washes out at ~20 GiB/s, which reads as bandwidth-bound rather than as the change not working.
- Machine was an AMD Ryzen 9 9950X3D (Zen 5). Numbers on Intel AVX-512 parts will differ.

### DIFF
--- a/crates/hash/src/blake3_avx512.rs
+++ b/crates/hash/src/blake3_avx512.rs
@@ -1,0 +1,117 @@
+// Copyright 2026 The Binius Developers
+
+//! x86-64 AVX-512 message transpose for the multi-lane Blake3 kernel.
+//!
+//! The lane loops hold the message as 16 words of `N` lanes, one lane per message.
+//! The input arrives the other way round: `N` blocks of 16 words, one block per lane.
+//! Turning one layout into the other is a 16x16 transpose of 32-bit words.
+//!
+//! At 16 lanes that square is exactly four 512-bit registers wide and 16 rows tall,
+//! so a fixed shuffle network moves it with no memory round trip.
+
+use std::arch::x86_64::{
+	__m512i, _mm512_loadu_si512, _mm512_setzero_si512, _mm512_shuffle_i32x4, _mm512_storeu_si512,
+	_mm512_unpackhi_epi32, _mm512_unpackhi_epi64, _mm512_unpacklo_epi32, _mm512_unpacklo_epi64,
+};
+
+use blake3::BLOCK_LEN;
+
+/// Lane count the shuffle network is built for.
+///
+/// A 64-byte block holds 16 words, so only 16 lanes make the message a square.
+/// A non-square block would need a different network per shape, which the lane loops already cover.
+const LANES: usize = 16;
+
+/// Reports whether this module has a transpose for the given lane count.
+#[inline(always)]
+pub const fn handles_lanes(n: usize) -> bool {
+	n == LANES
+}
+
+/// Transposes 16 blocks of 16 words into 16 words of 16 lanes.
+///
+/// Both buffers are 1024 contiguous bytes, read and written as 16 rows of one 512-bit register.
+///
+/// # Safety
+///
+/// - `src` must be readable for 16 rows of `BLOCK_LEN` bytes.
+/// - `dst` must be writable for 16 rows of `LANES` words.
+/// - The two regions must not overlap.
+#[inline(always)]
+unsafe fn transpose_16x16(src: *const u8, dst: *mut u32) {
+	unsafe {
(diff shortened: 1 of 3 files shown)


jimpo (approved):
As a follow-up, please make a `binius_hash::blake3` submodule with the arch-specific optimized versions as private submodules of that.

PR #2354, opened 2026-08-26
BINIUS-551: Record the full oracle schedule in the setup channel

tcoratger:
Closes [BINIUS-551](https://linear.app/jimpo/issue/BINIUS-551/record-full-oracleschedule-in-oraclesetupchannel).

Teaches the setup channel to record *where the commit rounds fall*, not just which oracles were committed.

No caller reads the schedule yet. This is the record two follow-ups need.

### The problem

The setup channel is a dry run: drive a verifier with it, and it writes down the shape of every oracle the protocol commits. What it writes down is a flat list.

That list loses the one thing the merge decorator cares about — the **round boundaries**. A round is the run of oracles between two challenge samples, and it is what gets committed as a single oracle.

```
recv, recv, sample, recv, sample, recv, recv, recv
\__________/        \__/          \______________/
   round 0         round 1           round 2
```

Two open questions on #2209 both bottom out here:

- [prover side](https://github.com/binius-zk/binius64/pull/2209#discussion_r3866854716) — `send_oracle` copies each incoming buffer into a holding buffer, because the round's layout is unknown until the round closes. Knowing the schedule up front means writing straight into the combined buffer instead.
- [verifier side](https://github.com/binius-zk/binius64/pull/2209#discussion_r3866764499) — the decorator cannot cross-check its own remaining specs against the inner channel's, because it has no way to say which round a given spec belongs to.

### The idea

Record the boundaries alongside the specs, in the one place that already sees both.

```rust
pub struct OracleSchedule {
    specs: Vec<OracleSpec>,
    ends: Vec<usize>,     // exclusive end of each closed round, indexing `specs`
}
```

Flat storage plus round ends, so the flat view stays a zero-cost `&[OracleSpec]` and the grouped view is a slice per round.

The setup channel closes a round wherever a challenge could be drawn — which is exactly where the merge decorator stops taking members into a round:

| site | why it closes a round |
|---|---|
| `sample` | a challenge follows the commitments it is derived from |
| `sample_bits` | a sampled word is a challenge like any other |
| `verify_oracle_relation` | opening needs every commitment already in place |
| end of the dry run | the trailing round has no challenge after it |

`end_round` is a no-op on an empty round, so consecutive challenges never manufacture one, and mirroring every flush site is always safe.

### What it derives

`merged_specs` produces the coarse, one-per-round list that a merge decorator's *underlying* channel has to be configured with. Nothing computed that before.

```
round [2^4, 2^2, 2^2]  ->  total 24  ->  2^5, one bit of padding
```

A round is masked as a whole, so its oracle is zero-knowledge as soon as any member is.

The rule that sizes a merged oracle now lives in one place. Both `merged_specs` and the verifier decorator's `flush` call it, instead of each spelling out `log2_ceil(sum of 2^n)`.

### Scope

- **new:** `OracleSchedule` in `crates/iop/src/channel/mod.rs`, and `into_oracle_schedule` on the setup channel
- **changed:** the setup channel stores a schedule instead of a bare `Vec`; `oracle_specs` and `into_oracle_specs` are unchanged projections of it, so the three existing callers are untouched
- **changed:** the verifier decorator's `flush` calls the shared sizing rule
- **left untouched:** the prover decorator's `flush`, which #2350 is currently rewriting. It adopts the shared rule in the follow-up rather than conflicting here.
- **left untouched:** every consumer. Wiring the schedule into the decorators is the next step, not this one.

### Tests

The load-bearing one is a **proptest that pins the schedule against the decorator**. Two independent routes to the same coarse list, over random round shapes:

```
decorator: merges each round, and its inner channel records what was committed
schedule : records where the rounds fall, then merges them itself
```

They must agree, or the schedule is not a faithful model of what the decorator does. Only the sizing rule is shared between the two paths; the sort order and the zero-knowledge derivation are reached independently.

Alongside it, the boundary cases: a challenge closes the open round, a sampled word does too, the trailing round closes without a challenge, back-to-back challenges make no empty round, a run with no oracles has no rounds, dropping the boundaries leaves the flat sequence untouched, and a round is masked when any single member is.

### Verification

| check | result |
|---|---|
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | clean |
| `cargo test -p binius-iop -p binius-iop-prover` | 190 passed, 0 failed |
| `cargo +nightly-2026-07-01 fmt --all --check` | clean |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-iop --document-private-items --no-deps` | clean |

### The second commit

`cargo clippy --workspace --all-targets --all-features -- -D warnings` already fails on `main` at `0c1bfac5`, before anything in this branch: an unterminated `assert!` inside a `const` block at `crates/iop-prover/src/channel/merge.rs:619`.

`main` keeps a green check because the clippy step is skipped on `push`, where the workflow only warms the build cache — so nothing re-lints `main` after a merge.

#2350 and #2351 each carry the same one-line fix. Whichever lands first makes this commit a no-op on rebase.

### DIFF
--- a/crates/iop-prover/src/channel/merge.rs
+++ b/crates/iop-prover/src/channel/merge.rs
@@ -619,7 +619,7 @@ mod tests {
 			assert!(
 				PackedBinaryGhash4x128b::LOG_WIDTH > 0,
 				"the fixture needs sub-packed-width oracle sizes to appear"
-			)
+			);
 		};
 		run_merge_round_trip::<PackedBinaryGhash4x128b>(&[&[3, 3], &[4, 2, 2], &[1]], false);
 	}
--- a/crates/iop/src/channel/merge.rs
+++ b/crates/iop/src/channel/merge.rs
@@ -8,9 +8,10 @@ use binius_core::word::Word;
 use binius_field::{BinaryField, Field, field::FieldOps};
 use binius_ip::channel::{IPVerifierChannel, WordIPVerifierChannel};
 use binius_math::multilinear::hypercube::Hypercube;
-use binius_utils::checked_arithmetics::log2_ceil_usize;
 
-use crate::channel::{Error, IOPVerifierChannel, OracleSpec, TransparentEvalFn};
+use crate::channel::{
+	Error, IOPVerifierChannel, OracleSpec, TransparentEvalFn, merged_log_msg_len,
+};
 
 /// A handle to an oracle received through the merging decorator.
 #[derive(Debug, Clone, Copy)]
@@ -194,11 +195,8 @@ where
 		let is_witness_dependent = order.iter().any(|&i| self.records[i].is_witness_dependent);
 
 		// Size the combined oracle to fit every oracle end to end.
-		let total_len: usize = order
-			.iter()
-			.map(|&i| 1usize << self.records[i].log_msg_len)
-			.sum();
-		let combined_log_len = log2_ceil_usize(total_len);
+		let combined_log_len =
+			merged_log_msg_len(order.iter().map(|&i| self.records[i].log_msg_len));
 
 		// Commit the whole round as one oracle on the underlying channel.
 		let outer = self
@@ -415,9 +413,11 @@ where
 #[cfg(test)]
 mod tests {
 	use binius_field::Ghash128b;
+	use binius_utils::checked_arithmetics::log2_ceil_usize;
+	use proptest::prelude::*;
 
 	use super::*;
-	use crate::channel::oracle_setup::OracleSetupChannel;
+	use crate::channel::{OracleSchedule, oracle_setup::OracleSetupChannel};
 
 	type F = Ghash128b;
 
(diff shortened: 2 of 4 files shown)


benji-potamus (approved):
Approve.

Mechanical pass clean: `cargo +nightly-2026-07-01 fmt --check`, `cargo clippy -p binius-iop -p binius-iop-prover --all-targets --all-features -- -D warnings`, `cargo doc -p binius-iop --no-deps --document-private-items`, and `typos` on the changed files all pass with no findings. `cargo nextest run -p binius-iop -p binius-iop-prover` is 190/190 green, matching the PR description.

The design is sound: `OracleSchedule` keeps the flat spec list and round boundaries in one structure, and `merged_log_msg_len` is factored out so the setup channel's `merged_specs()` and the merge decorator's `flush` compute round sizing from the same function instead of two copies that could drift. The proptest cross-checking the schedule's `merged_specs()` against the decorator's actual output over random round shapes is exactly the right test for that invariant, and the boundary-case unit tests (challenge closes a round, trailing round with no challenge, back-to-back challenges, empty run, ZK masking when any member is witness-dependent) cover the edges well.

Second commit (`Terminate the const assert...`) is now a no-op after rebase onto current `main`, as the PR description anticipated — nothing to do there.

No blocking findings.

jimpo (approved):

PR #2356, opened 2026-08-27
Fold the Blake3 arch kernels into a blake3 submodule

tcoratger:
Follow-up to @jimpo's review on #2353: make `binius_hash::blake3` a real submodule and hang the arch-specific kernels off it as private children.

Rebased on top of #2355, so this sits on the post-`MSG_SCHEDULE` kernel.

Pure reorganization — no logic, no lane counts, no dispatch conditions change.

### The problem

The Blake3 code was four files sitting side by side at the crate root.

Two were public modules of the crate. The other two were not modules of anything you could name — `blake3_portable` reached them through `#[path]` attributes:

```
#[cfg(all(target_arch = "aarch64", target_feature = "neon"))]
#[path = "blake3_neon.rs"]
mod neon;
```

So the file tree said "four peers at the crate root" while the module tree said "two of these are children of a third". Reading either one told you the wrong thing about the other.

It also put the arch kernels one level too deep. They are alternatives to the portable kernel, not parts of it, so they belong beside it rather than inside it.

### The change

```
- src/blake3.rs             pub mod blake3
- src/blake3_portable.rs    pub mod blake3_portable
- src/blake3_neon.rs        #[path] child of blake3_portable
- src/blake3_avx512.rs      #[path] child of blake3_portable

+ src/blake3/mod.rs         pub mod blake3
+ src/blake3/portable.rs    pub mod blake3::portable
+ src/blake3/neon.rs        private mod blake3::neon
+ src/blake3/avx512.rs      private mod blake3::avx512
```

`neon` and `avx512` are private, so nothing outside `blake3` can name them. `portable` stays public because the `blake3_portable` bench drives it directly.

The `#[path]` attributes are gone — each file now sits where its module path says it sits.

### Where the shared constants went

`IV`, `MSG_PERMUTATION`, `N_ROUNDS`, and `MSG_SCHEDULE` used to live in `blake3_portable`, and `neon` borrowed `IV` and `MSG_SCHEDULE` from it.

They move up into `blake3/mod.rs`, alongside `CHUNK_START`, `CHUNK_END`, and `ROOT`. They are facts about the hash function, not about the portable kernel, and every kernel reads them.

That placement also means no visibility annotations are needed: a private item in a parent module is visible to all of its descendants, so `portable`, `neon`, and `avx512` each reach them as `super::IV` while they stay invisible outside `blake3`.

`MSG_SCHEDULE` keeps its `const` block verbatim, including the `while` loops that compose the permutation at compile time.

### Scope

- **moved:** the four files, with `git` recording all four as renames
- **changed:** the module declarations, three import blocks, one bench import, and one line of `lib.rs`
- **left untouched:** every kernel body, every lane count, every `cfg` condition, and every test assertion

The only prose edits are to two module docs, saying which entry point hands off to which arch module.

### Verification

- `cargo build --workspace --lib --bins --tests --examples` — clean
- `cargo test -p binius-hash` — 27/27
- `cargo +nightly-2026-07-01 fmt --all --check` and `tools/check_copyright_notice.py` — clean

Both gated paths were exercised, not just the default build:

- **AVX-512:** rebuilt with `-C target-feature=+avx512f,+avx512vl,+avx512bw` — 30/30, including the three `avx512` transpose tests that the default build cfgs out.
- **NEON:** cross-compiling needs an `aarch64` C toolchain this machine does not have, so instead the `cfg` gates were temporarily removed to force `neon` and its tests into an x86-64 build. The only error was `unresolved import std::arch::aarch64` — expected on x86-64. Nothing else failed to resolve, which is what pins `mod neon;`, `use super::neon;`, `use super::{IV, MSG_SCHEDULE};`, and the `super::neon::compress_block` / `super::neon::load_block_words` test paths as correct. CI's `rust-checks-arm64` job is the real check.

### One thing deliberately left alone

`PortableBlake3ParallelDigest` now reads as `binius_hash::blake3::portable::PortableBlake3ParallelDigest`, which stutters twice over.

Dropping the prefix would collide with the `ParallelDigest` trait that the same module imports, so the fix is not free. Left for a separate PR rather than mixed into a file move.


benji-potamus (approved):
Approve.

Clean reorganization — the file tree now matches the module tree, the `#[path]` indirection is gone, and the shared spec constants (`IV`, `MSG_PERMUTATION`, `N_ROUNDS`, etc.) landing in `blake3/mod.rs` is the right home since every kernel reads them and `super::` visibility falls out for free. Diff matches the description exactly: no logic, lane counts, or `cfg` gates changed.

jimpo (approved):

PR #2357, opened 2026-08-27
BINIUS-555: Export the wiring claim instead of discharging it

tcoratger:
Closes [BINIUS-555](https://linear.app/jimpo/issue/BINIUS-555).

Lets a recursive circuit put the inner proof's wiring claim on public wires instead of evaluating it as constraints. Per-level growth falls from **14.77x to 5.65x**.

### The problem

[#2352](https://github.com/binius-zk/binius64/pull/2352) measured a tower that diverges: one level costs **14.77x** the constraint rows of the level it verifies.

One term causes it. The circuit discharges the inner wiring claim — the "monster" — as constraints, and evaluating that multilinear walks every constraint of the inner system. Measured, **6.0 BMUL per inner constraint row**, stable to 1% over a 256x range.

```
R(C) = H(C) + 6.0 * C
       ^^^^   ^^^^^^^
       replay   the claim
```

A tower closes only if some `C*` has `R(C*) <= C*`. The claim term alone is `6.0 * C`, so `R(C) > 6C > C` always. **No fixed point exists** while the claim is paid inline, and the replay cannot rescue it: it is *added to*, not multiplied by, a term with slope 6.

### The idea

Don't evaluate the claim. Export it.

The claim is a point and a value — "the wiring multilinear at `inputs` equals `claimed`" — and both are elements the verifier already derived. Put them on public wires, unchecked.

That is a win because the claim is **short**. Every section of `inputs` is a challenge vector over a padded log-sized index, so its length is logarithmic in the inner system: measured, **51 elements**. Exporting 51 elements replaces a cost proportional to 878,939 rows.

The check moves rather than vanishing. Whoever holds the outer proof owes it, and it is one native evaluation against a *public* constraint system — ordinary work outside a circuit, paid once however many levels deferred into it.

Same trade `succinctlabs/flock` makes with `MatrixAssertion`, arrived at independently by `leanEthereum/leanVM-b`.

### Per level

| discharge | depth 1 | depth 2 | growth |
|---|---|---|---|
| in-circuit | 878,939 | 12,984,191 | 14.77x |
| deferred | 873,977 | 4,939,643 | **5.65x** |

**Deferral cuts depth 2 by 62%** — 8,044,548 rows.

At depth 1 the drop is 4,884 BMUL, exactly the claim's cost at that size, landing BMUL on 360,351 — the value #2352 measured with the claim removed. The depth-2 saving *exceeds* the direct `6.0 * C` term because a smaller circuit also needs smaller commitments, so the replay shrinks with it.

**The tower still does not close.** 5.65x is well above one; the replay term dominates at these sizes. Closure is the merge node's checkpoint, with its own FRI parameters, not this seam's. Stated so nobody reads 5.65x as done.

### What this does not do: claims accumulate

Exporting a claim gets it out of the circuit. It does not get it out of the *statement*.

A level reads its child's whole statement as public input, carries it forward, and appends its own claim. Measured over two levels:

| | statement | added |
|---|---:|---|
| inner | 5 words | — |
| level 1 | 107 words | claim of 51 elements |
| level 2 | 305 words | claim of 99 elements |

So a level never checks its child's claim, it only copies it along. At the root, `N` levels means `N` separate claims to settle, and each is larger than the last because a claim's length tracks the log of the system it is about.

A tree over many leaves would carry one claim per leaf, which is a to-do list rather than aggregation. Folding claims into one is the next PR, and it is what makes the statement constant-size regardless of tree width or depth.

Worth being explicit that folding needs more than this seam: claims combine only when they are about the *same* polynomial, and the claims above are about different constraint systems. That is what forces a uniform node shape later.

### A follow-up this leaves open

A deferred claim carries no identifier of which constraint system it is about; settling it takes the system as an argument and trusts the caller to pair them correctly. Documented, and sufficient while there is one claim and one obvious system.

It stops being sufficient once claims from different systems meet in one accumulator, where the requirement is that a mismatched claim is *rejected* rather than silently folded. That wants a real key on the claim, which is best added alongside the accumulator that needs it.

### The change

```
- claim.check_symbolic(&mut channel)?;                  // constraints over every inner row
+ match discharge {
+     Discharge::InCircuit => claim.check_symbolic(&mut channel)?,
+     Discharge::Deferred  => bind the claim to public wires, unchecked,
+ }
```

- `WiringEvalFn` carries a `WiringEvalShape` — the section lengths — instead of seven loose fields. `Copy`, fixed by the constraint system and the reduction, never read off a prover message. So it is build-time data travelling with the circuit, not with each proof, which is what lets a claim be read back away from the run that raised it.
- `WiringEvalClaim::defer()` returns a `#[must_use]` `DeferredWiringClaim`, settled by `check(&constraint_system)`.
- `Discharge::{InCircuit, Deferred}` picks the behaviour. `RecursiveCircuit::build` is unchanged and still discharges in-circuit.
- `RecursiveCircuit::check_deferred` settles the claim from the outer statement; `verify_outer` does proof-then-claim in one call.

### Soundness

Deferral swaps a checked constraint for an owed one, so what matters is that the claim cannot be lied about, and cannot be quietly skipped.

- **The claim is bound, not reported.** Each element keeps the wires that derived it and gains public wires constrained equal to them, so a prover cannot publish a claim the circuit did not derive. A test flips one claim word and pins the failure to `bind_public_elem`.
- **The check is a real evaluation.** A test moves the first, second, and last claim words and requires `check_deferred` to reject each. Otherwise deferral would export an unchecked constraint and call it settled.
- **The binding doubles as a cross-check.** The replay computes the claim concretely and supplies the public half, so a build-versus-replay desync fails at the binding rather than surfacing later.
- **Settling needs no secret.** The constraint system is public and the shape is build-time data.

**The obligation this creates, stated plainly.** A deferred circuit whose claim nobody checks attests to a verification *minus its wiring constraint*, and the proof does not say so. `#[must_use]` cannot reach that, because the claim lives inside the circuit. Mitigated by making the safe path the short one (`verify_outer`) and documenting the obligation at every entry point; making it structurally impossible needs a sink that consumes the claim, which is the accumulator in the next PR.

### Scope

- **changed:** `crates/verifier/src/protocols/shift/verify.rs` — the shape refactor plus `defer`/`DeferredWiringClaim`
- **changed:** `crates/recursion` — `Discharge`, `bind_public_elems`, `check_deferred`, `verify_outer`
- **untouched:** the replay itself. Hashing, Merkle and sumcheck costs are the same; this removes only the term proportional to the inner system
- **behaviour-preserving by default:** `build` still discharges in-circuit, so every existing caller is unaffected

### Verification

- `cargo clippy -p binius-recursion -p binius-verifier --all-targets --all-features -- -D warnings` — clean, CI's exact flags
- nightly `cargo fmt --all --check`, `cargo doc` on both crates — clean
- `cargo test -p binius-recursion` — 70 tests pass (12 in the end-to-end suite, 1 ignored by design)
- `cargo test -p binius-verifier` — 13 pass
- `cargo test -p binius-prover --test prove_verify` — 17 pass, which is what exercises `WiringEvalFn` natively after the shape refactor
- the ignored growth test run explicitly — reports the table above
- **Not verifiable locally:** anything `--workspace`. No C compiler here, so `alloca`'s build script fails. CI covers it.

### DIFF
--- a/crates/recursion/src/channel.rs
+++ b/crates/recursion/src/channel.rs
@@ -73,6 +73,8 @@ pub struct Binius64BuilderChannel {
 	n_merkle_checks: usize,
 	/// Words bound to public inputs so far, so each binding gets a distinct name.
 	n_public: usize,
+	/// Elements bound to public inputs so far, so each binding gets a distinct name.
+	n_public_elems: usize,
 	/// Proofs of work checked so far, so each gets a distinct name.
 	n_grinds: usize,
 }
@@ -90,6 +92,7 @@ impl Binius64BuilderChannel {
 			scheme: BinaryMerkleTreeScheme::new(),
 			n_merkle_checks: 0,
 			n_public: 0,
+			n_public_elems: 0,
 			n_grinds: 0,
 		}
 	}
@@ -113,6 +116,37 @@ impl Binius64BuilderChannel {
 			.map(|(i, word)| {
 				let public = builder.add_inout();
 				builder.assert_eq(format!("{}", first + i), word.to_wire(&builder), public);
+				public
+			})
+			.collect()
+	}
+
+	/// Binds elements to fresh public inputs, returning the two inout wires each occupies.
+	///
+	/// An element is its low and high half, so each contributes a `(lo, hi)` pair, in that order.
+	/// The returned wires therefore run twice the length of `elems`.
+	///
+	/// As with [`bind_public`](Self::bind_public), each element keeps the wires that derived it
+	/// and gains public wires equal to them. The replay supplies the public half, so the equality
+	/// is also a cross-check: the value the circuit derived and the value the replay computed have
+	/// to agree, and a desync between the two runs fails here rather than silently later.
+	///
+	/// This is how a claim leaves the circuit unchecked but pinned: whoever reads the outer proof
+	/// sees the values the verifier derived, and owes whatever check was skipped inside.
+	pub fn bind_public_elems(&mut self, elems: &[SymbolicElem]) -> Vec<Wire> {
+		// Numbered across calls, so a failing binding names which element it was.
+		let first = self.n_public_elems;
+		self.n_public_elems += elems.len();
+
+		// One named subcircuit, so a broken binding is traceable to this gadget.
+		let builder = self.shared.builder().subcircuit("bind_public_elem");
+		elems
+			.iter()
+			.enumerate()
+			.flat_map(|(i, elem)| {
+				let (lo, hi) = elem.to_wires(&builder);
+				let public = [builder.add_inout(), builder.add_inout()];
+				builder.assert_eq_v(format!("{}", first + i), [lo, hi], public);
 				public
 			})
 			.collect()
--- a/crates/recursion/src/channel.rs
+++ b/crates/recursion/src/channel.rs
@@ -126,13 +126,14 @@ impl Binius64BuilderChannel {
 	/// An element is its low and high half, so each contributes a `(lo, hi)` pair, in that order.
 	/// The returned wires therefore run twice the length of `elems`.
 	///
-	/// As with [`bind_public`](Self::bind_public), each element keeps the wires that derived it
-	/// and gains public wires equal to them. The replay supplies the public half, so the equality
-	/// is also a cross-check: the value the circuit derived and the value the replay computed have
-	/// to agree, and a desync between the two runs fails here rather than silently later.
+	/// Each element keeps the wires that derived it, and gains public wires equal to them.
 	///
-	/// This is how a claim leaves the circuit unchecked but pinned: whoever reads the outer proof
-	/// sees the values the verifier derived, and owes whatever check was skipped inside.
+	/// The replay supplies the public half.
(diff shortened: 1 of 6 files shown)


jimpo (commented):
Nice, I agree on the premise of the issue

benji-potamus (approved):
Approve.

Clean seam: `Discharge::{InCircuit, Deferred}` picks whether the wiring claim is checked as
constraints or exported to public wires, and `build` stays behavior-preserving so no existing
caller is affected. The binding in `bind_public_elems` doubles as the build-vs-replay cross-check,
and the test suite backs up every soundness claim in the description — tampering with a claim word,
disagreeing with the derived value, and `verify_outer` rejecting a moved claim after the raw proof
still verifies. Traced the `element_words`/`check_deferred` low/high-half round trip by hand; it's
consistent.

One non-blocking nit:

- `crates/verifier/src/protocols/shift/verify.rs:615` — `WiringEvalShape::n_inputs()` isn't called
  anywhere in this PR (only its own definition and doc comment reference it). Either wire it into
  `check`/`check_deferred` as a length assertion on `inputs`, or drop it until something needs it.

PR #2363, opened 2026-08-27
BINIUS-556: Split the hash crate along the prover/verifier line

tcoratger:
Closes [BINIUS-556](https://linear.app/jimpo/issue/BINIUS-556).

Splits `binius-hash` into the sequential side a verifier links against and the batched side only
a prover needs. Pure refactor — no protocol change, no behavior change.

### The problem

The SIMD kernels and the threaded Merkle tree builder lived in the same crate a verifier
links against.

Reading the verifier's trust base therefore meant reading past code it never calls:

```
    binius-hash (before)
      compress.rs, serialization.rs        <- the verifier actually uses this
      binary_merkle_tree.rs                <- rayon, tree construction
      parallel_digest.rs, parallel_*.rs    <- rayon
      sha256/{portable,avx512,sha_ni,neon} <- 70 target_feature sites, 68 unsafe
      blake3/{portable,avx512,neon}        <- same
```

Nothing enforced the separation, so the next kernel would land in the same place.

### The idea

Two crates, split on what verification actually needs.

Verification walks a single Merkle path, so it needs one sequential leaf hash and one
sequential compression. Proving folds whole layers, so it needs a batched counterpart for each.

```
    binius-hash          leaf hash, two-to-one compression, serialization, the suite trait
    binius-hash-prover   the multi-lane kernels, the parallel traits, the tree builder
```

### The suite trait splits with the crates

The trait bundled all four hashes, which is what tied the two sides together:

```diff
- trait HashSuite { LeafHash; Compression; ParLeafHash; ParCompression }

+ // binius-hash
+ trait HashSuite { LeafHash; Compression }
+ // binius-hash-prover
+ trait ParallelHashSuite: HashSuite { ParLeafHash; ParCompression }
```

Because the batched trait has the sequential one as a supertrait, prover-side code keeps
reading `H::LeafHash` and `H::Compression` unchanged. Only the bound and the import move:

```diff
- use binius_hash::binary_merkle_tree::HashSuite;
+ use binius_hash_prover::ParallelHashSuite;
-     H: HashSuite,
+     H: ParallelHashSuite,
```

I upgraded only the bounds the compiler proved needed it, rather than renaming all of them.

### The separation is enforced, not documented

`binius-hash` now denies unsafe code outright.

The one exception is the buffer trait a hash writer has to implement, and it is marked where it
stands rather than waved through crate-wide:

| | `binius-hash` | `binius-hash-prover` |
|---|---|---|
| `unsafe` sites | 1, marked | 68 |
| `target_feature` sites | 0 | 70 |
| rayon uses | 0 | 15 |

Five dependencies drop off the verifier crate as a result: `binius-compute`, `binius-field`,
`itertools`, `thiserror`, `tracing`. `rand` becomes dev-only.

### Soundness

- The sequential compression is unchanged, including its native word order, so committed
  Merkle roots and paths are bit-identical.
- Its known-answer vector moves to the verifier crate and still pins that byte order.
- Every batched path keeps the proptest holding it equal to its sequential counterpart, which
  is what lets both sides of a proof commit to the same tree.
- One real consequence, and the point of the exercise: the verifier's compression now calls the
  reference block function directly instead of the one-lane kernel. Same bytes, no shared code.

### Scope

- **new:** `binius-hash-prover`, and the two suite modules
- **moved:** every kernel, the parallel traits, the tree builder, both benches — as renames, so
  `git log --follow` still works
- **changed:** 8 dependent crates, mostly one import line each
- **left alone:** the kernels themselves, byte for byte

### Verification

- nightly `cargo fmt --all --check`, `RUSTDOCFLAGS="-D warnings" cargo doc`, copyright — clean
- `cargo test --workspace` — 461 passed, 0 failed
- `cargo build -p binius-hash` on its own, to confirm the verifier crate stands alone


benji-potamus (approved):
Approve.

Clean split along a real boundary: `binius-hash` keeps only the sequential leaf hash and
compression a verifier walks a single Merkle path with, `binius-hash-prover` gets the batched
kernels, `rayon`, and the tree builder. `#![deny(unsafe_code)]` on the verifier crate with the one
exception marked at its definition (`serialization.rs`) makes the separation load-bearing rather
than just documented. The trait split (`HashSuite` sequential, `ParallelHashSuite: HashSuite`
batched) let every dependent crate's bound get upgraded only where the compiler actually required
it — checked that no stray `binary_merkle_tree::HashSuite` or unqualified bound was left behind,
and every consumer that needs batching now pulls `ParallelHashSuite` from the new crate.

jimpo (approved):

PR #2370, opened 2026-08-27
BINIUS-561: Let the sparse-dense sumcheck take any foldable weight

tcoratger:
Closes [BINIUS-561](https://linear.app/jimpo/issue/BINIUS-561).

Rebased onto `main` now that #2369 has merged, so this is a single commit with no stacked base.

**Updated per review** (thanks @jimpo): the trait abstraction is gone — the prover holds `FactoredMultilinear` directly. The diff dropped from three files (169/40) to **one file (91/17)**, and both `common.rs` and `factored_multilinear.rs` are now untouched.

### The stack

Three PRs build the fold that batches deferred wiring claims. Each is independently reviewable; only the last one folds anything.

```
  #2369   3b-i    the weight, stored as pieces instead of one table   (merged)
  #2370   3b-ii   teach the sparse-dense sumcheck to take one      <- this PR
  #2371   3b-iii  the fold itself
```

Design note, with the reduction and its soundness argument: [Folding deferred wiring claims](https://linear.app/jimpo/document/folding-deferred-wiring-claims-design-99d3c050a1d5).

### The problem

The sparse-dense product sumcheck is already the right shape for the fold's column sumcheck. Its **sparse** side matches exactly: entries are `(index, value)` pairs whose repeats add without deduplication — which is what the wiring tensor's characteristic-two multiplicities need — and per-round work tracks the entry count, not the hypercube size.

Its **dense** side did not. It took a full buffer, so its variable count was whatever that buffer could hold. The fold's column weight spans 47 variables: `2^47` elements, 2 PiB.

#2369 stores such a weight as a product of factors, at about `2^26` elements. There was no way to hand one over.

### The idea

The fold's column weight is already held as a product of factors — one per axis — costing their sum rather than their product. So take that type directly.

```diff
- pub struct SparseDenseProductSumcheckProver<P, Data: BufferData<P>> { dense: FieldBuffer<P, Data> }
+ pub struct SparseDenseProductSumcheckProver<P>                      { dense: FactoredMultilinear<P> }
```

Nothing is given up. A weight that really is one table is a weight of **one factor**, so the existing tests wrap their buffers and read the same:

```rust
// One factor over every variable is exactly the table it holds.
let weight = FactoredMultilinear::new([dense.clone()]);
```

Nothing about the protocol changes either. The weight is the same multilinear; only its storage differs.

### Why this is safe

The prover has **no callers** outside its own tests — a recent primitive nothing has picked up yet — so the generalization cannot disturb anything.

Worth stating the adjacent risk explicitly: the shift reduction's first phase runs a sparse-dense sumcheck too, but through its own prover, not this one. Untouched.

### Soundness

Storage is invisible to the protocol, and that is what the test asserts, the strongest way available.

The same claim is proved twice, once over a factored weight and once over the table it multiplies out to, and the two **transcripts are compared byte for byte**. Identical transcripts mean identical round polynomials in *every* round, which is much stronger than agreeing on the final value. The shared claim is then verified, so neither form can be consistently wrong in the same way.

The sum is asserted non-zero too, since a vacuous claim would let a broken weight pass unnoticed.

### Scope

- **changed:** the prover's dense side becomes the factored weight, in one file
- **no new abstraction**, and the factored weight itself is untouched
- **no production caller.** This makes the fold expressible; it does not perform one

### Verification

- `cargo clippy -p binius-ip-prover --all-targets --all-features -- -D warnings` — clean, CI's exact flags
- nightly `cargo fmt --all --check`, `cargo doc` — clean
- `cargo test -p binius-ip-prover` — **101 pass**, including the new equivalence test
- **Not verifiable locally:** anything `--workspace`. No C compiler here, so `alloca`'s build script fails.

### DIFF
--- a/crates/ip-prover/src/sumcheck/sparse_dense_product.rs
+++ b/crates/ip-prover/src/sumcheck/sparse_dense_product.rs
@@ -2,16 +2,17 @@
 
 //! Sumcheck prover for the product of a sparse and a dense multilinear.
 
-use binius_compute::BufferData;
 use binius_field::{Field, PackedField};
 use binius_ip::sumcheck::RoundCoeffs;
-use binius_math::{FieldBuffer, multilinear::fold::fold_highest_var_inplace};
 use binius_utils::rayon::{
 	prelude::*,
 	task_size::{IndexedParallelIteratorExt, WorkPerItem},
 };
 
-use super::{common::SumcheckProver, round_evals::RoundEvals, round_state::RoundState};
+use super::{
+	common::SumcheckProver, factored_multilinear::FactoredMultilinear, round_evals::RoundEvals,
+	round_state::RoundState,
+};
 
 /// One entry of a sparse multilinear: a hypercube index and the value carried there.
 ///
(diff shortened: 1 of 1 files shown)


crates/ip-prover/src/sumcheck/sparse_dense_product.rs
```diff
@@ -62,18 +63,23 @@ pub type SparseEntry<F> = (usize, F);
 /// Entries thus stay a flat list of the same length for the whole protocol, and collapse to the
 /// evaluation of $A$ at the challenge point only in [`SumcheckProver::finish`], which emits the
 /// sparse multilinear's evaluation before the dense one's.
-pub struct SparseDenseProductSumcheckProver<P: PackedField, Data: BufferData<P>> {
+pub struct SparseDenseProductSumcheckProver<P: PackedField> {
```
jimpo, line 66: Follow-up: restore Data type param so that FactoredMultilinear can hold allocator-backed buffers

jimpo (approved):

crates/ip-prover/src/sumcheck/sparse_dense_product.rs
tcoratger: Done in #2381.

Restored on both the weight and the prover that reads it, with `BufferData` as the bound rather than plain `Deref` — folding truncates a factor in place as its variables bind, and `truncate` is what `BufferData` adds.

The parameter defaults to the heap, so no caller changes and no signature reads differently; the diff is two files and every existing test compiles untouched.

Two tests pin that the storage never reaches the arithmetic, across folding rather than only at construction: one folds a heap-backed and an arena-backed weight in lockstep comparing the whole table after every challenge, the other proves the same claim over both and compares transcripts byte for byte.

PR #2371, opened 2026-08-27
BINIUS-565: Fold evaluation claims on a sparse tensor into one claim

tcoratger:
Closes [BINIUS-565](https://linear.app/jimpo/issue/BINIUS-565).

Rebased onto `main` now that #2370 has merged, so this is a **single commit** with no stacked base.

### The stack

```
  #2369   3b-i    the weight, stored as pieces instead of one table   (merged)
  #2370   3b-ii   teach the sparse-dense sumcheck to take one         (merged)
  #2371   3b-iii  the fold itself                                     <- this PR
```

Both PRs below have landed, so this is now a plain single-commit PR against `main`.

Design note, with the reduction and its soundness argument: [Folding deferred wiring claims](https://linear.app/jimpo/document/folding-deferred-wiring-claims-design-99d3c050a1d5).

### The problem

[#2357](https://github.com/binius-zk/binius64/pull/2357) let a recursive circuit export its wiring claim rather than discharge it. Per-level growth fell from **14.77x to 5.65x** — but the claims then accumulate. A level copies its child's claim forward and appends its own:

| level | statement | added |
|---|---:|---|
| inner | 5 words | — |
| 1 | 107 words | claim of 51 elements |
| 2 | 305 words | claim of 99 elements |

A tree over `N` leaves carries `N` claims to the root, each larger than the last. That is a to-do list, not aggregation.

### The idea

Fold them.

```
    k claims at k points  ->  one claim at one point
```

Because the claim that comes out has the **same shape** as the ones that went in, a node folds again at the next level, and the state it passes upward never grows. That shape-preservation is the whole purpose.

The reduction is one sumcheck over the tensor's index space, weighted by a combination of the claims' equality indicators. The verifier derives every weight itself and needs exactly one value from the prover: the tensor's evaluation at the point the sumcheck lands on.

### Per review

Two rounds of feedback, both addressed.

**1. Not a new protocol, and not in the general sumcheck module.** The verifier is a many-point-to-one-point evaluation reduction, so it now lives in `binius_ip::batch_eval::verify`, and `sumcheck/batch.rs` is untouched. It takes `impl IntoIterator<Item = MultilinearEvalClaim<C::Elem>>` and returns `MultilinearEvalClaim<C::Elem>` — the type the crate already defines.

Reducing over **flat** points simplified more than the naming: axes mean nothing to the verifier, so all per-axis structure moved to the prover, and the shape-checking ceremony disappeared. Over the batched sumcheck it adds one reconstruction, needed because the weight is a *combination* of indicators rather than a single one (so `mlecheck` cannot absorb it):

```
weight = sum_j batch^j * eq(point_j, point)
assert  eval * weight == output.eval
```

**2. The entries were folded once per claim.** A real bug: the prover held `entries.to_vec()` per claim and folded every copy each round, so memory *and* folding work scaled with the claim count.

`SparseMultiDenseProductSumcheckProver` now holds one shared sparse column and one dense column per claim, emitting a round polynomial per column. `drive::batch_with_coeff` extends across each prover's `execute()` output, so one prover emitting `k` polynomials batches exactly as `k` provers would — no driver change.

Two side benefits: `finish()` returns `[sparse_eval, dense_0, …]`, so the tensor evaluation is structurally single (removing an `assert!` that `k` provers agreed on it); and a new test pins the multi-column prover against the single-column one at one column, byte for byte, since the round math is now written twice.

### One sumcheck, where the design says two

The design note takes its two-sumcheck reduction from `succinctlabs/flock`. That second sumcheck exists to collapse the row side so every claim reads against one shared residual — which is only necessary when the claims' weights are **not** indicators.

Every weight here is an indicator, so a claim already *is* an evaluation of the extension, and the batch reduces directly. Recording this as a simplification of the design rather than a departure: the two-sumcheck form is still what the first level needs, where the operand axis carries batching powers instead of an indicator.

### It runs in a circuit — measured, not asserted

The verifier names no concrete field element. It reads claims off a channel and asserts one relation, so over a channel carrying wires that assertion becomes a constraint.

A test drives it over the recursion builder channel and measures the result. For two claims over five variables:

| | |
|---|---:|
| AND | 5,324 |
| BMUL | 21 |
| ZERO | 1,468 |
| recorded inputs | 46 |

Dominated by absorbing the fold's own transcript, and **independent of the tensor's entry count**. That independence is the property an aggregation tree needs.

### Soundness

A fold does not check its inputs. It produces a claim that is false whenever any input was, so that settling the output **once**, at the root, catches an error anywhere below. That is the accumulation property, and the tests are built around it.

- **Truth and shape survive.** Three true claims fold to a claim that holds against the tensor read directly, spanning the same axes.
- **One false input poisons the output.** Each of three claims is corrupted in turn; the fold must either reject or return a claim that does not hold. A true output from a false input is the one outcome that must never occur.
- **A single claim folds.** The degenerate arity works rather than being a case a caller must avoid.
- **Mismatched axes are rejected**, since folding them would combine claims about two different polynomials.

Error bound from the design note's argument: about `(k - 1 + 3n) / |F|` for `k` claims over `n` variables.

### Scope

- **new:** `binius_ip::batch_eval` (the verifier); `SparseMultiDenseProductSumcheckProver`; the fold prover; the in-circuit test
- **reuses** the crate's existing evaluation-claim type and error type
- **nothing existing changed** beyond two module declarations

**What is still not connected:** the wiring claim's own weights are not indicators, and its value axis is segmented. Bridging those into this fold is the accumulator's work, and it is where the two-sumcheck form earns its place.

### Verification

- `cargo clippy --all-targets --all-features -- -D warnings` on all three crates — clean, CI's exact flags
- nightly `cargo fmt --all --check`, `cargo doc` — clean
- `binius-ip` 15 pass, `binius-ip-prover` 105 pass, `binius-recursion` all suites pass
- **Not verifiable locally:** anything `--workspace`. No C compiler here, so `alloca`'s build script fails.

### DIFF
--- a/crates/ip-prover/src/claim_fold.rs
+++ b/crates/ip-prover/src/claim_fold.rs
@@ -1,0 +1,307 @@
+// Copyright 2026 The Binius Developers
+
+//! Proving a fold of evaluation claims on one sparse tensor.
+
+use binius_field::{Field, PackedField};
+use binius_ip::MultilinearEvalClaim;
+use binius_math::multilinear::eq::{eq_ind_partial_eval, eq_ind_partial_eval_scalars};
+
+use crate::{
+	channel::IPProverChannel,
+	sumcheck::{
+		batch::batch_prove, factored_multilinear::FactoredMultilinear,
+		sparse_dense_product::SparseMultiDenseProductSumcheckProver,
+	},
+};
+
+/// An evaluation claim whose point is cut into one run per tensor axis.
+///
+/// The verifier reduces claims over a flat point, since the axes mean nothing to it.
+/// A prover needs the cut, because it builds one weight factor per axis.
+///
+/// Axes run lowest first: the first run owns the lowest index bits.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AxisClaim<F> {
+	/// The point, one coordinate run per axis, lowest axis first.
+	pub point: Vec<Vec<F>>,
+	/// The value the extension is claimed to take there.
+	pub value: F,
+}
+
+impl<F: Clone> AxisClaim<F> {
+	/// The width of each axis, in variables, lowest axis first.
+	pub fn axes(&self) -> Vec<usize> {
+		self.point.iter().map(Vec::len).collect()
+	}
+
+	/// The same claim over a flat point, as the verifier reduces it.
+	pub fn flatten(&self) -> MultilinearEvalClaim<F> {
+		MultilinearEvalClaim {
+			eval: self.value.clone(),
+			point: self.point.concat(),
+		}
+	}
+}
+
+/// Cuts a flat point into one run per axis, lowest axis first.
+///
+/// This is how a reduced claim regains the structure its inputs had, so a caller may reduce again.
+pub fn split_axes<F: Clone>(point: &[F], axes: &[usize]) -> Vec<Vec<F>> {
+	let mut runs = Vec::with_capacity(axes.len());
+	let mut rest = point;
+	for &width in axes {
+		let (run, tail) = rest.split_at(width);
+		runs.push(run.to_vec());
+		rest = tail;
+	}
+	runs
+}
+
+/// One nonzero of a sparse tensor: a flat index over every axis, and the value carried there.
+///
+/// Axes run lowest first, so the first axis owns the lowest index bits.
+/// Entries at a repeated index add, so a caller need not deduplicate them.
+pub type TensorEntry<F> = (usize, F);
+
+/// Proves a fold of evaluation claims on one tensor, returning the claim they folded to.
+///
+/// The claims all name the same tensor at different points.
+/// One sumcheck reduces them to one point.
+///
+/// What comes out has the same shape as what went in, so a caller can fold again.
+///
+/// # Cost
+///
+/// Per round, one pass over the entries per claim.
+///
+/// Nothing is materialized over the product of the axes.
+/// So the axes may span a space far larger than the entry list.
+///
+/// # Panics
+///
+/// Panics if no claim is given, or if the claims disagree about the axis widths.
+pub fn prove<F, P>(
+	entries: &[TensorEntry<F>],
+	claims: &[AxisClaim<F>],
+	channel: &mut impl IPProverChannel<F>,
+) -> AxisClaim<F>
+where
+	F: Field,
+	P: PackedField<Scalar = F>,
+{
+	let first = claims
+		.first()
+		.expect("precondition: a fold needs at least one claim");
+	let axes = first.axes();
+	assert!(
+		claims.iter().all(|claim| claim.axes() == axes),
+		"precondition: every claim must span the same axes"
+	);
+
+	// One weight per claim, all riding a single copy of the tensor's entries.
+	//
+	// The weight is an equality indicator over every axis at once, which factorizes across them.
+	// Holding it as one factor per axis is what keeps it off the product of their lengths.
+	let weights = claims
+		.iter()
+		.map(|claim| {
+			FactoredMultilinear::new(claim.point.iter().map(|run| eq_ind_partial_eval::<P>(run)))
+		})
+		.collect::<Vec<_>>();
+	let sums = claims.iter().map(|claim| claim.value).collect::<Vec<_>>();
+
+	// One prover over all of them, so the entry list is stored once and folded once.
+	// A prover per claim would hold a copy each, and fold every copy every round.
+	let prover = SparseMultiDenseProductSumcheckProver::new(entries.to_vec(), weights, &sums);
+	let output = batch_prove(vec![prover], channel);
+
+	// The evaluations lead with the tensor's, shared by every claim, then one weight's per claim.
+	let tensor_eval = output.multilinear_evals[0][0];
+
+	// The verifier derives its own weight evaluations, so the tensor's is the one thing it needs.
+	channel.send_one(tensor_eval);
+
+	// The rounds bind the highest variable first, so the point reads back in reverse.
+	let mut point = output.challenges;
+	point.reverse();
+
+	// Cut the point into one run per axis, matching the shape the claims came in with.
+	let mut runs = Vec::with_capacity(axes.len());
+	let mut rest = point.as_slice();
+	for width in axes {
+		let (run, tail) = rest.split_at(width);
+		runs.push(run.to_vec());
+		rest = tail;
+	}
+
+	AxisClaim {
+		point: runs,
+		value: tensor_eval,
+	}
+}
+
+/// The tensor's multilinear extension at one point, evaluated directly from its entries.
+///
+/// This is the reference a folded claim is settled against.
+/// It reads the entries and the point, and nothing from the run that raised the claim.
+pub fn evaluate<F: Field>(entries: &[TensorEntry<F>], point: &[Vec<F>]) -> F {
+	let flat = point.concat();
+	let indicator = eq_ind_partial_eval_scalars(&flat);
+	entries
+		.iter()
+		.map(|&(index, value)| value * indicator[index])
+		.sum()
+}
+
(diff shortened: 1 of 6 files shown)


benji-potamus (approved):
Approve.

Reviewed only the second commit (`74de5c34`, the BINIUS-565 fold itself), per the PR description — #2370 is still open, so I treated its commit as already-reviewed base.

This is a clean, well-isolated addition: three new files, no existing code touched beyond two `pub mod` lines. The verifier/prover split is a direct translation of the design note's single-sumcheck simplification, and the batching math checks out — `output.eval` from `batch_verify` is `tensor(point) * sum_i batch_coeff^i * weight_i(point)`, which is exactly what `verify()` reconstructs and asserts against the prover-supplied tensor evaluation. Axis handling (concat for the flat point, split-by-width to rebuild per-axis runs) is consistent between `prove`, `verify`, and the reference `evaluate`, and is well documented ("axes run lowest first"). The soundness-focused tests (truth/shape preservation, per-claim corruption, single-claim degeneracy, axis-mismatch rejection) directly exercise the accumulation property the design relies on, and the in-circuit test confirms the verifier only asserts relations rather than naming field elements.

No blocking findings.

crates/ip/src/claim_fold.rs
```diff
@@ -0,0 +1,171 @@
...
+/// The terms are the batching challenge, the sumcheck itself at degree two, and the chance the
+/// batched weight vanishes at the point the sumcheck lands on.
+///
+/// # Errors
+///
+/// Returns an error if no claim is given, or if the claims disagree about the tensor's shape.
+///
+/// The reduced claim is asserted over the channel rather than compared, so a false one surfaces
+/// as a channel error here and as an unsatisfied constraint in a circuit.
+pub fn verify<F, C>(
```
jimpo: Why do we need a special verification routine? Isn't this just a regular batch sumcheck?

jimpo (changes requested):

crates/ip/src/claim_fold.rs
tcoratger: You're right — it is just a regular batch sumcheck, and the special routine is gone.

Concretely it's the batching for **several points on one multilinear**, which is the exact complement of `batch_verify_mle`'s **one point across several multilinears**. So it now sits beside it in `sumcheck/batch.rs` as `batch_verify_multi_point`, reusing `sumcheck::Error`.

What's left over the plain batching is one reconstruction, because the weight is a combination of indicators rather than a single one (so `mlecheck` can't absorb it):

```
weight = sum_j batch^j * eq(point_j, point)
assert  value * weight == eval
```

That deleted `crates/ip/src/claim_fold.rs` entirely — the module, its bespoke `Error` enum, and its shape-checking ceremony. The claim struct moved to the prover side where it's the fold's own vocabulary, and the axis split became a caller concern.

Behaviour is unchanged: the in-circuit measurement is identical to the constraint (5,324 AND / 21 BMUL / 1,468 ZERO), which is what tells me the restructure didn't move any arithmetic.

While re-auditing I also found my doc check had been passing vacuously on these files (I'd diffed against `HEAD` after committing, so it compared nothing) — the docs are now genuinely one-sentence-per-line throughout.

tcoratger: As per the previous response of the agent, just fixed, we should merge #2370 before and then rebase this one :)

crates/ip/src/sumcheck/batch.rs
```diff
@@ -59,6 +59,74 @@ where
...
+/// The reduced point, in variable order rather than the round order the challenges arrive in.
+/// And the multilinear's evaluation there.
+///
+/// # Errors
+///
+/// Returns an error if the sumcheck fails, or if the reconstruction does not hold.
+///
+/// The reconstruction is asserted over the channel rather than compared.
+/// So it becomes a constraint on a channel that carries wires.
+pub fn batch_verify_multi_point<F, C>(
```
jimpo: This shouldn't go in the general sumcheck module. This is the verification routine for a many-point-to-one-point multilinear evaluation reduction. It should take an `impl IntoIterator<Item=MultilinearEvalClaim<C::Elem>>` and output a MultilinearEvalClaim<C::Elem>.

Put it in another module (this is logic around `batch_verify`). I'd put it in `binius_ip::batch_eval` and name it `verify`. (So callers would call `batch_eval::verify`).

crates/ip-prover/src/claim_fold.rs
```diff
@@ -0,0 +1,317 @@
...
+	//
+	// The weight is an equality indicator over every axis at once, which factorizes across them.
+	// Holding it as one factor per axis is what keeps it off the product of their lengths.
+	let provers = claims
+		.iter()
+		.map(|claim| {
+			let weight = FactoredMultilinear::new(
+				claim.point.iter().map(|run| eq_ind_partial_eval::<P>(run)),
+			);
+			SparseDenseProductSumcheckProver::new(entries.to_vec(), weight, claim.value)
```
jimpo: This looks like it's going to redundantly fold the entries once per evaluation point. That needs to be shared for efficiency. The easiest way to handle this would be to write a new, custom SumcheckProver that has one sparse column and many FactoredMultilinear columns.

jimpo (changes requested):

crates/ip/src/sumcheck/batch.rs
tcoratger: Done, exactly as suggested — it's `binius_ip::batch_eval::verify` now, in its own module, and `sumcheck/batch.rs` is back to untouched.

It takes `impl IntoIterator<Item = MultilinearEvalClaim<C::Elem>>` and returns `MultilinearEvalClaim<C::Elem>`, reusing the type the crate already defines rather than a new one. Reducing over **flat** points turned out to simplify things beyond the naming: the axes mean nothing to the verifier, so all the per-axis structure moved to the prover (which needs it, to build one weight factor per axis) and the shape-checking ceremony disappeared.

The prover keeps a small `AxisClaim` with a `flatten()` into the shared type, plus a `split_axes` helper to give a reduced point its structure back.

crates/ip-prover/src/claim_fold.rs
tcoratger: Good catch, that was a real bug — it held `entries.to_vec()` **per claim** and folded every copy each round, so memory and folding work both scaled with the claim count.

Fixed the way you suggested: `SparseMultiDenseProductSumcheckProver` holds one shared sparse column and a `Vec<FactoredMultilinear>`, emits one round polynomial per column, and folds the entry list **once** per round. `drive::batch_with_coeff` extends across each prover's `execute()` output, so a single prover emitting *k* polynomials batches exactly as *k* provers would — no driver change needed.

Two side benefits:
- `finish()` now returns `[sparse_eval, dense_0, …]`, so the tensor evaluation is structurally single. That removed an `assert!` I had checking that *k* independent provers agreed on it.
- I added a test pinning the multi-column prover against the single-column one at one column, comparing transcripts byte for byte. The round math is written twice now, so it can drift once.

That pinning test earned its place immediately: it failed first time. I'd compared `prove_single` against `batch_prove`, but the batching driver samples a coefficient *before* the rounds, so the challenger diverges from round two. Both sides now go through the same driver.

crates/ip/src/batch_eval.rs
```diff
@@ -0,0 +1,115 @@
...
+///
+/// Panics if no claim is given, or if the claims do not all span the same number of variables.
+pub fn verify<F, C>(
+	claims: impl IntoIterator<Item = MultilinearEvalClaim<C::Elem>>,
+	channel: &mut C,
+) -> Result<MultilinearEvalClaim<C::Elem>, Error>
+where
+	F: Field,
+	C: IPVerifierChannel<F>,
+	C::Elem: FieldOps<Scalar = F>,
```
jimpo:
```suggestion
```

Redundant trait bound

jimpo (approved):

PR #2374, opened 2026-08-27
Fold the field buffer's helper types into its own file

tcoratger:
Answers the review on #2267 — `pub(super)` all over the field buffer's submodules — by removing the reason it existed. **14 `pub(super)` → 0, and no `pub(crate)` or `#[allow]` in their place.** Pure refactor: no behavior change, no public API change.

### The diagnosis

Every annotation traced back to one arrangement:

```text
chunks.rs      defines Chunks, ChunksMut     built only by mod.rs
write_back.rs  defines SplitMut, ChunkMut    built only by mod.rs
sub_word.rs    defines SubWordChunk          used by both of the above
```

Each of those types exists only to be returned by one buffer method, and each is constructed in exactly one place — a different file from the one defining it. A module boundary drawn there separates a type from its only construction site, and `pub(super)` is what pays for the crossing.

So the split was not carving out independent units. It was cutting through the middle of one.

### The change

The three modules fold into the buffer's own file, next to the methods that produce them. Nothing else moves.

Everything then becomes a private item of that module:

```diff
-pub(super) fn new(words: &'a [P], log_chunk_size: usize, chunk_count: usize) -> Self
+fn new(words: &'a [P], log_chunk_size: usize, chunk_count: usize) -> Self

-pub struct SubWordChunk<P>          // `pub` inside a private module
+struct SubWordChunk<P>              // simply private
```

The locator is the clearest case. It had been `pub` within a private module — technically unreachable from outside, but still asking a reader to work out why a `pub` type never appears in the public API. Now it is private and there is nothing to work out.

`view.rs` stays a module. Its contents are the borrowed type aliases, their public constructors, and the store enum. They *name* the buffer rather than being produced by it, callers spell them out directly, and nothing in that file was ever annotated — so the boundary there is real.

### On the alternative

Reopening `FieldBuffer` and splitting the method definitions across the existing files is the other way to close the same gap, and it was tried first.

It does not build here. Defining `FieldBuffer::chunks` in `chunks.rs` puts a second `impl<P: PackedField, Data: Deref<Target = [P]>> FieldBuffer<P, Data>` block in the crate while `mod.rs` already has one, and the workspace sets `multiple_inherent_impl = "warn"` with `-D warnings` in CI. Three errors, one per duplicated bound set.

Worth recording, since #2283 concluded the opposite: that lint **does** fire on generic inherent impls. It keys on the self type together with the generics, so `mod.rs`'s five existing blocks coexist only because their bounds all differ. Splitting a bound set across files is what trips it.

That leaves `#[allow(clippy::multiple_inherent_impl)]` as the price of keeping the files, which is trading one annotation for another. Folding the types in costs nothing and removes the question entirely.

### Cost

`field_buffer/mod.rs` goes from 2027 to 2450 lines, of which about 1130 are tests. That is a big file for a big struct interface.

### Scope

- **deleted:** `chunks.rs`, `sub_word.rs`, `write_back.rs`
- **changed:** `mod.rs` — the three modules' contents, their constructors made private, and a module doc describing the layout and the two chunk shapes
- **untouched:** `view.rs`, every public signature, and every crate outside `binius-math`
- **out of scope:** one `pub(super)` remains in `multilinear/hypercube/expansion.rs`, a file #2368 deletes

### Verification

| check | result |
|---|---|
| `cargo build --workspace --lib --bins --tests --examples` | pass |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | pass |
| `cargo test --workspace` | pass, 0 failed |
| `cargo test -p binius-math` | 176 passed, 1 doctest |
| `cargo test --doc --workspace` | pass |
| `cargo +nightly-2026-07-01 fmt --all --check` | pass |
| `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items --no-deps` | pass |
| `python3 tools/check_copyright_notice.py` | 0 wrong, 0 missing |

Test count is unchanged at 176. Nothing about how a chunk is located, lent, or merged back changed — only which file it is written in and who can name it.


jimpo (commented):
I still don't really see it. You have structs whose sole purpose is to be a return type in a different file from the one place they are constructed? Perhaps you want to reopen FieldBuffer and split the method definitions across different files?

But I'm also not that opposed to the single big file; it's a big struct interface anyway.

crates/math/src/field_buffer/mod.rs
```diff
@@ -77,6 +76,9 @@ pub use write_back::{ChunkMut, SplitMut};
 ///
 /// Truncation additionally zeros the dead lanes it creates.
 /// Other constructors may leave whatever the caller's store held there.
+// The buffer's methods are grouped by topic, one module per group, so its inherent impls span
+// more than one file.
+#[allow(clippy::multiple_inherent_impl)]
```
jimpo, line 81: This is the downside of this approach. I think it's warranted here.

jimpo (approved):
I pushed a commit that splits the implementation more cohesively. The downside is that the inherent impls are split over multiple files. In general that's discouraged and smaller struct interfaces are preferable. But FieldBuffer just is a big struct, so I think splitting the files is warranted.

As a follow-up, you could move the `split` methods into a `split.rs` if you want to break it up more.

Or if you don't like this, drop the commit and just merge the one big file.

PR #2375, opened 2026-08-27
BINIUS-559: Batch shift-reduction operand claims with tensor weights

benji-potamus:
Closes [BINIUS-559](https://linear.app/jimpo/issue/BINIUS-559/batch-shift-reduction-operand-claims-with-tensor-weights-instead-of).

Replaces the shift reduction's four per-operation lambdas with two challenge vectors, and makes a
claim's batching weight the product of their equality indicators:

```
batched_eval = sum_op sum_{m < arity(op)}  eq(operation_batch_challenges, op)
                                         * eq(operand_batch_challenges, m) * e_{op,m}
```

The operand weights then no longer differ between operations, so the table that fans them into the
shift scalars is built **once** instead of four times, and becomes a plain tensor expansion rather
than a bespoke fan-out. The operation weight is a single scalar, which rides for free in the
constraint-index expansion's seed.

**Transcript change**: 4 samples become 5 (2 for the operation axis, then 3 for the operand axis),
so recorded proofs need regenerating. Soundness is unchanged in practice — the old scheme is a
degree-≤6 polynomial in four challenges (error ≤ 6/|F|), the new one a multilinear in five
(error ≤ 5/|F|).

### Commits

**1. `Hand the shift-scalar tables to the monster evaluation by reference`**

`OperationEvalFn`'s two evaluations now read [`WiringWeights`] — the borrowed per-axis bundle
#2361 introduced for `contract` — instead of a flat `FieldFn` input they had to be copied into.
`ShiftScalars`, `encode_operation_input`, `split_input` and the three segment-length fields all
delete; `call` and `contract` end up reading the same four tables, which is what lets the shared
ones be built once and lent to all four operations. Net −360/+250 lines, no behaviour change.

*This is the one design call worth a look.* The issue proposed a new borrowed field on
`OperationEvalFn` alongside a shrunken flat input; routing through `WiringWeights` instead converges
with #2361 rather than standing up a second bundle beside it. Happy to split this back out into its
own PR if you'd rather review it separately from the protocol change.

**2. `Batch shift-reduction operand claims with tensor weights`**

The protocol change itself, verifier and prover in lockstep (the transcript order is shared, so the
two cannot land apart).

- `mod.rs` — `OPERATION_COUNT`, `LOG_OPERATION_COUNT`, `LOG_MAX_ARITY`, hardcoded with `const _`
  asserts against the four arities.
- `verify.rs` — samples the two vectors in place of four scalars; `OperatorData::batched_eval`
  becomes an inner product against the operand tensor prefix, scaled by the operation weight; the
  wiring claim's input header goes from four lambdas to the five challenges, so `n_inputs` is the
  only shape change.
- `monster.rs` — `operand_shift_scalar_table` deletes; the inner table is indexed
  `(shift_index << LOG_MAX_ARITY) | operand_id`, which is one expansion of one concatenated point.
- prover `claims.rs` / `prove.rs` — `prepare` samples the two vectors in the verifier's order;
  `PreparedOperatorClaims` gains the shared `operand_weights`; `PreparedOperatorData` loses
  `lambda_powers` and folds its operation weight into the constraint tensor, which is renamed
  `weighted_r_x_prime_tensor` so the name doesn't hide the scaling (that was the issue's open
  question — folding it in is where it's free, and the rename is what keeps it honest).
- prover `key_collection/` — `build_scalars` collapses from four loops at strides 1/3/4/6 to one at
  stride 8, `ScalarTables` collapses to a single `Vec<F>`, and the per-key `match key.operation` for
  arity goes away with it.

### Expected impact

|  | now | after |
| -- | -- | -- |
| Verifier weight-table setup, muls / eval | ~8,350 | ~4,600 (−45%) |
| Prover `build_scalars` muls / key segment | ~30·D | ~12·D (−60%) |
| `ScalarTables` size / key segment | 14·D elems | 8·D elems |
| Transcript samples | 4 | 5 |

The table's stride is the padded `1 << LOG_MAX_ARITY = 8` rather than the true max arity of 6:
stride 6 would save ~1k multiplications but is not a tensor expansion, so it would mean keeping the
hand-rolled helper.

### Verification

`cargo nextest run --workspace` — 2075 passed, 15 skipped. Doc tests, `cargo doc
--document-private-items`, and `cargo clippy --workspace --all-targets -- -D warnings` all clean.
Each commit is independently green. The recursion end-to-end tests cover the deferred-claim round
trip through the changed `WiringEvalShape`.

### DIFF
--- a/crates/prover/src/protocols/shift/claims.rs
+++ b/crates/prover/src/protocols/shift/claims.rs
@@ -19,7 +19,11 @@
 use std::ops::Index;
 
 use binius_field::Field;
-use binius_verifier::protocols::shift::{BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, ZERO_ARITY};
+use binius_ip_prover::channel::IPProverChannel;
+use binius_math::multilinear::eq::eq_ind_partial_eval_scalars;
+use binius_verifier::protocols::shift::{
+	BINMUL_ARITY, BITAND_ARITY, INTMUL_ARITY, LOG_MAX_ARITY, LOG_OPERATION_COUNT, ZERO_ARITY,
+};
 
 use super::{Operation, OperatorData, PreparedOperatorData};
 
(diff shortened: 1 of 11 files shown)


crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -376,10 +389,10 @@ where
 		let r_v_len = r_v_inner.len();
 		let r_y_len = r_y.len();
 
-		let inputs: Vec<C::Elem> = iter::once(zero_lambda.clone())
-			.chain(iter::once(bitand_lambda.clone()))
-			.chain(iter::once(intmul_lambda.clone()))
-			.chain(iter::once(binmul_lambda.clone()))
+		let inputs: Vec<C::Elem> = operation_batch_challenges
```
jimpo: nit: use chain! macro from itertools

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -204,15 +212,24 @@ where
...
+		.map(|_| channel.sample())
+		.collect::<Vec<_>>();
+
+	// A claim's batching weight is the product of the two axes' equality indicators, so the two
+	// expansions factor it: one weight per operation, one per operand position. The operation
+	// weights are indexed in the order the four operator arguments are declared.
+	let operation_weights = eq_ind_partial_eval_scalars(&operation_batch_challenges);
+	let operand_weights = eq_ind_partial_eval_scalars(&operand_batch_challenges);
+
+	let eval = zero_data.batched_eval(&operation_weights[0], &operand_weights)
```
jimpo: pull the operation weight out of the call. batched_eval should only combine with operand_weights. Then here, frame this as an inner product between operation weights and the batched_eval results.

crates/prover/src/protocols/shift/claims.rs
```diff
@@ -39,34 +42,52 @@ pub struct OperatorClaims<F: Field> {
...
+	/// these orders are protocol, not detail.
 	///
 	/// The arities are erased on the way out, since both proving phases dispatch on a shift key.
 	///
 	/// # Arguments
 	///
-	/// - `sample`: draws the next batching coefficient from the transcript.
+	/// - `sample`: draws the next batching challenge from the transcript.
 	pub fn prepare(self, mut sample: impl FnMut() -> F) -> PreparedOperatorClaims<F> {
+		let operation_batch_challenges = (0..LOG_OPERATION_COUNT)
```
jimpo: nit: sample_many

jimpo (changes requested):

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done in 5858984. `batched_eval` now takes only `operand_weights` and returns the operand-batched value; the four are combined here as an inner product against the operation tensor:

```rust
let eval = inner_product(
    operation_weights,
    [
        zero_data.batched_eval(&operand_weights),
        bitand_data.batched_eval(&operand_weights),
        intmul_data.batched_eval(&operand_weights),
        binmul_data.batched_eval(&operand_weights),
    ],
);
```

One asymmetry I left alone, flagging in case you want it matched: on the prover side `PreparedOperatorData::new` still folds the operation weight into its `batched_eval` field. It takes that weight regardless — it also seeds `weighted_r_x_prime_tensor`, which is where the weight rides for free — so the two-jobs-in-one-argument smell you're pointing at isn't there. Separating it would mean `PreparedOperatorClaims` storing the operation tensor so its own `batched_eval()` could do the inner product, and it would ripple into the m4-prover phase-1 identity test that reads the per-operation fields directly. Happy to do it if you'd rather the two sides read identically — for a value both must compute the same way, there's an argument for that.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done in 5858984 — `chain!` with a single trailing `.cloned()`, which also drops the eleven per-section `.iter().cloned()` calls.

crates/prover/src/protocols/shift/claims.rs
benji-potamus: Done in 5858984, but it needed a signature change rather than a one-liner, so flagging what it touched.

`prepare` took `impl FnMut() -> F`, which has no bulk form — so I changed it to take the channel: `prepare(self, channel: &mut impl IPProverChannel<F>)`, then `channel.sample_many(..)` for both axes. The closure indirection wasn't buying anything; binius-prover already depends on binius-ip-prover. Production call sites just become `claims.prepare(channel)`.

The cost was three test/bench sites that were passing fake samplers. Two are trivial (`&mut ProverTranscript::<StdChallenger>::default()`), and the third — the test pinning that the operation axis is drawn before the operand axis — I think came out better: instead of a stand-in that hands out 1, 2, 3…, it now draws the axes by hand from a second transcript seeded identically and compares, so it pins the order against the real channel. It also gained a check the old one couldn't make, that `prepare` drew *exactly* those five:

```rust
// The two stay in lockstep only if `prepare` drew exactly those five and no more.
assert_eq!(
    IPProverChannel::<B128>::sample(&mut channel),
    IPProverChannel::<B128>::sample(&mut expected)
);
```

Say the word if you'd rather keep `prepare` transcript-agnostic and I'll take the narrower `impl FnMut(usize) -> Vec<F>` route instead.

jimpo (approved):

PR #2382, opened 2026-08-28
BINIUS-4: Bit-index reversal in a single pass over memory

tcoratger:
Closes [BINIUS-4](https://linear.app/jimpo/issue/BINIUS-4/bit-index-reversal-with-chunking).

Runs the tiled bit-reversal in one pass over memory instead of two, and sizes its tiles from what a memory stream actually costs. **1.2x to 2.3x faster** across three packings and three lengths. Pure refactor of an internal path — no API change, no behavior change.

### The problem

The permutation splits the scalar index into three fields, with equal width at both ends:

```text
    i = (h, m, l)        |h| = |l| = log_tile
```

and ran two passes over the buffer:

```text
    phase 1:  (h, m, l) -> (rev l, m, rev h)     transpose the tiles at each middle index
    phase 2:  (h, m, l) -> (h, rev m, l)         permute the tiles across middle indices
```

Composing them reverses every bit. But each pass reads and writes the whole buffer, so the permutation costs **four traversals** of it. Bit reversal does no arithmetic, so those traversals are the entire cost.

### The idea

Read the destination first rather than the source:

```text
    new(H, M, L) = old(rev L, rev M, rev H)
```

The middle field alone decides where a scalar goes. Everything under middle index `rev M` lands under `M`, and nothing else does.

The scalars sharing a middle index form a square of `2^log_tile` tiles. So the squares **pair up**, and a pair is the whole unit of work:

```text
    square(M)      <-  transpose of square(rev M)
    square(rev M)  <-  transpose of square(M)
```

Both are read into scratch, transposed there, and written back crossed. A middle index equal to its own reversal is a fixed point and needs one square.

One pass. Every scalar read once, written once — half the traffic.

### The second finding

The first measurement showed one packing *regressing*, so I swept both algorithms across every tile width rather than tune around it. That exposed a separate problem, independent of the pass count: **the tile heuristic was mis-sized**.

| packing | old heuristic picked | measured optimum |
|---|---|---|
| 1xGhash | k=2 | **k=3** |
| 2xGhash | k=0 | **k=3** |
| 4xGhash | k=0 | **k=2** |

Every optimum lands on the same physical quantity — a **256-byte tile row**, four cache lines. The old rule targeted one cache line.

A row is what each gather and scatter moves, and rows sit far apart in the buffer, so each row costs a memory stream of its own. One cache line is too little to amortize a stream. Sizing the row directly reproduces all three optima from one constant.

```diff
- const CACHE_LINE_BYTES: usize = 64;
+ const TILE_ROW_BYTES: usize = 256;
```

### Benchmark

`cargo bench -p binius-math --bench bit_reverse`, criterion `--save-baseline` on unmodified `main`, then `--baseline before`. AMD Ryzen 9 9950X3D, 16 cores, 128 MiB L3. Every cell improved, all `p < 0.01`.

| packing | log_d=16 | log_d=20 | log_d=24 |
|---|---|---|---|
| 1xGhash | 87.0 -> 41.3 us (**2.15x**) | 886 -> 382 us (**2.32x**) | 30.1 -> 16.3 ms (**1.85x**) |
| 2xGhash | 140 -> 70.8 us (**1.89x**) | 1329 -> 586 us (**2.27x**) | 33.0 -> 15.1 ms (**2.18x**) |
| 4xGhash | 101 -> 82.0 us (1.21x) | 733 -> 594 us (1.20x) | 29.7 -> 15.0 ms (**2.00x**) |

At `log_d=24` the buffer is 256 MiB against 128 MiB of L3, so the run is bandwidth-bound and the traffic halving shows up directly. The two smaller lengths gain from the wider tile as much as from the fused pass.

### One step further, measured and rejected

The new heuristic wants a 256-byte row for every packing, but `MAX_LOG_TILE_PACKED = 3` caps the widest instance. That cap binds only for 1xGhash, whose packed element is 16 bytes, leaving it a 128-byte row.

Raising the cap to 4 was measured rather than assumed:

| 1xGhash | k=3 (shipped) | k=4 |
|---|---|---|
| log_d=16 | **41.3 us** | 44.7 us |
| log_d=20 | **382 us** | 432 us |
| log_d=24 | 16.3 ms | **14.2 ms** |

A wash: 15% better past the last cache level, 8-14% worse below it, for double the compiled instances. Run-to-run noise on this bench is 8-13%, checked by comparing configurations whose tile width did not change between runs, so the `log_d=20` gap is inside the noise and the other two are not.

Left at 3, and the constant now says why.

### Scope

- **changed:** `crates/math/src/bit_reverse.rs` only
- the two-pass implementation is deleted; the single-pass one replaces it
- **untouched:** the naive reference, the group permutation, every public signature, and every caller

Documentation changes are confined to what this PR changes: the new constant, the two stale bullets in the tile cap's rationale, the comments around the re-sized heuristic, and the replaced function. Items this PR does not touch keep their existing docs verbatim.

### Correctness

The index algebra was derived on paper before a line was written, then checked against the existing suite, which is unchanged:

- Nine length cases straddling every branch the tile choice can take.
- Three packings — one, two and four scalars per element — driving three different tile widths.
- A proptest pinning the tiled path to the naive definition.
- A proptest pinning the permutation as an involution.

All 15 pass, on the native and the portable arch legs.

### Verification

| check | result |
|---|---|
| `cargo build --workspace --lib --bins --tests --examples` | pass |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | pass |
| `cargo test --workspace` | pass, 0 failed |
| `cargo test --doc --workspace` | pass |
| `RUSTFLAGS="" cargo test -p binius-math` | pass |
| `cargo +nightly-2026-07-01 fmt --all --check` | pass |
| `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items --no-deps` | pass |
| `python3 tools/check_copyright_notice.py` | 0 wrong, 0 missing |

### DIFF
--- a/crates/math/src/bit_reverse.rs
+++ b/crates/math/src/bit_reverse.rs
@@ -224,6 +224,8 @@ fn bit_reverse_paired<P: PackedField, const LOG_TILE_PACKED: usize>(
 					}
 
 					// Otherwise the two squares trade places, each transposed on the way.
+					// Otherwise the two squares trade places, each transposed on the way.
+					// The mirror is read before the scatter that overwrites where it sat.
 					gather(mirror, m_rev);
 					transpose_tile::<P, LOG_TILE_PACKED>(mirror, block);
 					scatter(square, m_rev);
--- a/crates/math/src/bit_reverse.rs
+++ b/crates/math/src/bit_reverse.rs
@@ -1,4 +1,5 @@
 // Copyright 2025 Irreducible Inc.
+// Copyright 2026 The Binius Developers
 
 use std::ptr;
 
@@ -25,8 +26,13 @@ pub const fn reverse_bits(x: usize, bits: u32) -> usize {
 	x.reverse_bits().unbounded_shr(usize::BITS - bits)
 }
 
-/// Bytes a cache miss fetches, on every target this runs on.
-const CACHE_LINE_BYTES: usize = 64;
+/// Bytes one row of a tile should span.
+///
+/// Rows sit far apart in the buffer, so each one costs a memory stream of its own.
+/// Four cache lines is where a stream is wide enough to reach peak bandwidth.
+///
+/// Going wider costs more than it returns: two squares have to fit in the first cache level.
+const TILE_ROW_BYTES: usize = 256;
 
 /// Base-2 log of the widest tile a permutation instance moves, counted in packed elements.
 ///
@@ -34,9 +40,9 @@ const CACHE_LINE_BYTES: usize = 64;
 ///
 /// - Every tile-sized loop takes its bound from this parameter.
 /// - Each step it may take therefore doubles the number of instances compiled.
-/// - Eight elements already reach a cache line for any scalar of eight bytes or wider.
-/// - That covers every field permuted here.
-/// - A narrower scalar keeps the capped tile rather than paying for further instances.
+/// - Eight elements already reach the row width above once an element is 32 bytes or wider.
+/// - A narrower element would prefer one step more, and one packing here is narrower.
+/// - Measured, that step gains past the last cache level and loses below it.
 const MAX_LOG_TILE_PACKED: usize = 3;
 
 /// Permutes the elements of a buffer by reversing the bits of every index, in place.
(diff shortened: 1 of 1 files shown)


crates/math/src/bit_reverse.rs
```diff
@@ -128,69 +146,91 @@ fn bit_reverse_tiled<P: PackedField, const LOG_TILE_PACKED: usize>(
...
+					transpose_tile::<P, LOG_TILE_PACKED>(square, block);
+
+					// A middle index that reverses to itself keeps its own square.
+					if m_rev == m {
+						scatter(square, m);
+						return;
 					}
+
+					// Otherwise the two squares trade places, each transposed on the way.
+					gather(mirror, m_rev);
```
benji-potamus, line 229: The ordering here is load-bearing and worth pinning down in the comment: `gather(mirror, m_rev)` must run before `scatter(square, m_rev)`, because the scatter overwrites the exact square the gather reads.

```suggestion
					// Otherwise the two squares trade places, each transposed on the way.
					// The mirror is read before the scatter that overwrites where it sat.
					gather(mirror, m_rev);
```

benji-potamus (approved):
Approve.

Reading the permutation destination-first is the right way to see it, and pairing the squares is a clean way to fuse the two passes. The index algebra holds up: gathering rows at `rev j`, transposing in scratch, and scattering back at `rev j` composes to `new(H, L) = old(rev L, rev H)` within a square, and claiming each pair at its lower middle index keeps the parallel writes disjoint. Resizing the tile row is a real second finding, and sweeping every width rather than tuning around the one regression is what turned it up.

Non-blocking, and outside the diff so it has no inline anchor: `crates/math/src/bit_reverse.rs:416-418` — the packing table in the tests is now stale. Under the new heuristic 1xGhash and 2xGhash both reach `LOG_TILE_PACKED = 3` and 4xGhash reaches 2, so none of the three "keeps its own width" and none is "one sub-block" any more.

One further non-blocking note on `min_len` is inline on the diff.

jimpo (approved):

crates/math/src/bit_reverse.rs
```diff
@@ -128,69 +146,93 @@ fn bit_reverse_tiled<P: PackedField, const LOG_TILE_PACKED: usize>(
...
 	let data_ptr = data.as_mut_ptr() as usize;
 
-	// Phase 1: transpose the square of tiles sitting at each middle index.
-	//
-	// One iteration reads and writes `2^log_tile` rows of `2^LOG_TILE_PACKED` words each.
-	// The byte budget counts single words, so divide it by that factor.
-	let min_len = (min_len_for_bytes::<P>() >> (log_tile + LOG_TILE_PACKED + 1)).max(1);
+	// One iteration moves two squares, and half of them return at once.
+	// The byte budget counts single words, so divide it by the words a square holds.
+	let min_len = (min_len_for_bytes::<P>() >> log_square).max(1);
```
benji-potamus, line 158: Non-blocking: `min_len` looks 2x high. An active iteration moves two squares in and back out, so four squares of traffic, and half the indices return at once, giving two squares per index on average — `>> (log_square + 1)`. The old phase-1 line carried that read-plus-write factor. Chunk size also matters a little more here than it did, since the active indices are not spread evenly across the range.

PR #2383, opened 2026-08-28
BINIUS-568: Split a wiring claim into one evaluation claim per operation

tcoratger:
Closes [BINIUS-568](https://linear.app/jimpo/issue/BINIUS-568).

Makes a deferred wiring claim foldable by [#2371](https://github.com/binius-zk/binius64/pull/2371)'s reduction. Rebased onto current `main` — past [#2386](https://github.com/binius-zk/binius64/pull/2386), [#2399](https://github.com/binius-zk/binius64/pull/2399) and [#2401](https://github.com/binius-zk/binius64/pull/2401).

**Rewritten to be purely additive.** The previous version changed how the constraint tables carry the batching weight; this one leaves that alone. `call` and `call_native` keep the bodies `main` gave them, down to the line that sums the four operations.

### The problem

[#2357](https://github.com/binius-zk/binius64/pull/2357) lets a recursive circuit export its wiring claim rather than discharge it, and #2371 landed a reduction that folds many evaluation claims on one multilinear into one. The wiring claim cannot use it, for a structural reason.

A wiring claim is **one value covering four operations at once** — Zero, BitAnd, IntMul, BinMul — each summing over its own constraint list, at its own `r_x_prime`. Those are four different multilinears, and claims fold only when they name the same one.

So the claims accumulate instead. Measured over two levels: statement 5 → 107 → 305 words, one claim per level, each larger than the last.

### The idea

Take the claim apart.

Split by operation and each share **is** an ordinary evaluation claim on a single operation's tensor. Shares about one operation fold together; shares about different operations never do — which is exactly the discipline an accumulator needs.

A split is only useful if a holder can check it without redoing it, so what the shares carry is the recombination:

```
    sum over operations of  weight * eval  ==  the claim's own value
```

A share is a `MultilinearEvalClaim` plus that weight, so it goes to `batch_eval::verify` with no adapter — the type says what the split is for.

### The one change to existing code

`wiring_weights` takes the seed for each operation's constraint table rather than deriving it:

```
- let operation_weights = eq_ind_partial_eval_scalars(operation_batch_v);
- scaled_expand(r_x_prime_v[operation], operation_weights[operation].clone())
+ scaled_expand(r_x_prime_v[operation], operation_scales[operation].clone())
```

A batched evaluation passes the operation's weight, which is what `main` already computed in the same place, so the weight still rides in the seed for free. The split passes one, so the same builder serves both and the split does not re-expand four tensors over the padded constraint count just to shed a weight.

This is what the review asked for, and it is why nothing else moved: `WiringInputs` is untouched, both discharges keep their summation line, and no behaviour changes.

It also fixes a doc line that `main` contradicts — `wiring_weights` currently says "the constraint-index tables ask for a scale of one", which stopped being true when #2386 seeded them.

### Soundness

No protocol behaviour changes and nothing here verifies anything. What matters is that the shares really are the claim taken apart — a wrong axis order or a dropped batching weight would give shares that fold cleanly and mean nothing.

Two independent anchors, because one equation cannot pin both halves.

**The recombination.** For a system with constraints in two of the four operations, the shares' weighted sum must equal what the existing wiring evaluation returns. This pins the unscaled share evaluations and the batching weights. It does **not** read a point.

**The reconstruction.** Each share's evaluation must be recoverable *from its point alone*: cut the point back into runs per the documented layout, expand each run into the table its axis is read with, and contract with the flat entry walk. That is a different code path from the one that produced the value, and it is what pins the axis order. A length check would not — swapping two runs of equal width leaves every length intact.

The reconstruction builds the value axis as **one flat indicator over `r_y ++ r_segment`**, rather than as the two scaled half-tensors `value_tensor` uses. So it also holds the "value axis is the padded address space" reading against the cut the evaluation actually walks — which is the reading a folder has to take.

It also asserts the claim is non-zero and that exactly two of the four shares are, so neither anchor can pass vacuously; that the two empty operations report zero while still being present positionally; and that the generic and native discharges agree, which is what covers the seeded-builder refactor on the path `check_native` runs.

One subtlety: the empty-operation case cannot be keyed on a zero evaluation, because a zero-length run and an absent operation are indistinguishable — a single constraint also spans zero variables. It is keyed on the constraint count, as #2386's own table builder is.

### Scope

- **new:** the per-operation split, its result type, and the helper that reads the operation weights off the flat input
- **changed:** `wiring_weights` takes the seed; its two call sites pass what it used to derive
- **no new module, no new error type, nothing else newly public**
- **not yet connected:** nothing folds these shares in production. The accumulator that collects them per operation, and the node that carries it, follow

### Verification

Run on a machine with a C compiler, so `--workspace` is covered this time.

- `cargo clippy -p binius-verifier --all-targets --all-features -- -D warnings` — clean, CI's exact flags
- `cargo +nightly-2026-07-01 fmt --all --check`, `typos`, `tools/check_copyright_notice.py` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-verifier --document-private-items --no-deps` — clean (it caught a public doc linking a private item; the link is gone)
- `cargo test -p binius-verifier --release` — 17 pass, including the two anchors
- `cargo test -p binius-prover --release --test prove_verify` — 17 pass, which is what covers the seeded builder end to end
- `cargo test -p binius-recursion --release` — 71 pass, 1 ignored, which is what consumes the deferred claim
- **the anchors were mutation-checked**, since a green test proves nothing on its own — every one of these fails the test:

| mutation | result |
| --- | --- |
| operand run moved above the shift runs | killed |
| constraint run omitted from the point | killed |
| batching weight dropped from the share | killed |
| split's constraint tables scaled again | killed |
| `wiring_weights` ignores its seed | killed |
| reconstruction swaps the inner and outer shift runs | killed |
| reconstruction reads the value axis at the wrong segment base | killed |

### DIFF
--- a/crates/verifier/src/protocols/shift/mod.rs
+++ b/crates/verifier/src/protocols/shift/mod.rs
@@ -72,6 +72,6 @@ mod verify;
 pub use error::Error;
 pub use shift_ind::evaluate_shift_inds;
 pub use verify::{
-	DeferredWiringClaim, OperationClaim, VerifyOutput, WiringEvalClaim, WiringEvalFn,
-	WiringEvalShape, check_eval, evaluate_words_mle, verify,
+	DeferredWiringClaim, OperationClaim, OperationShare, VerifyOutput, WiringEvalClaim,
+	WiringEvalFn, WiringEvalShape, check_eval, evaluate_words_mle, verify,
 };
--- a/crates/verifier/src/protocols/shift/verify.rs
+++ b/crates/verifier/src/protocols/shift/verify.rs
@@ -9,6 +9,7 @@ use binius_core::{
 };
 use binius_field::{BinaryField, field::FieldOps, util::FieldFn};
 use binius_ip::{
+	MultilinearEvalClaim,
 	channel::IPVerifierChannel,
 	sumcheck::{SumcheckOutput, verify as verify_sumcheck},
 };
@@ -597,6 +598,30 @@ pub struct WiringEvalShape {
 	r_y_len: usize,
 }
 
+/// One operation's share of a wiring claim.
+///
+/// A whole claim covers four operations, each over its own constraint list.
+/// Those are four different multilinears.
+/// Claims combine only when they name the same one, so a share is what folds.
+#[derive(Debug, Clone)]
+pub struct OperationShare<E> {
+	/// This operation's evaluation claim, on the tensor its own constraints span.
+	///
+	/// The point is the tensor's axis runs concatenated, lowest axis first.
+	///
+	/// ```text
+	///     constraint | operand | inner amount, variant | outer amount, variant | word | segment
+	/// ```
+	///
+	/// The word and segment axes address the padded value space, not the runs a term reads.
+	///
+	/// Every run's width comes from the constraint system, never from a prover message.
+	/// So two proofs of one system give shares that fold side by side.
+	pub claim: MultilinearEvalClaim<E>,
+	/// The weight this evaluation carries into the claim it was batched into.
+	pub weight: E,
+}
+
 impl<'a> WiringEvalFn<'a> {
 	/// Rebuilds the evaluation over a constraint system, reading its input with the given shape.
 	///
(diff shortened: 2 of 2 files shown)


benji-potamus (changes requested):
Request changes — one test gap. The split itself is right, and the design is the right one.

Taking the claim apart by operation is the correct move, and the reasoning that no special reduction is needed holds: every axis really is an indicator now, so a share is a plain evaluation claim. The layout refactor is faithful — `sections` cuts in exactly the order `verify`'s `chain!` concatenates, and pulling it out of `wiring_weights` is a genuine improvement over the inline offset arithmetic. I checked the point's axis order by hand against the weight tables (`constraint`, then `operand | inner_shift` with the operand innermost, then `outer`, then the padded value address `r_y ++ r_segment`) and it matches what `OperationShare::point` documents.

**Blocking — `share.point` is untested.** `verify.rs:1199` checks only the point's *length*; the recombination equality never reads `point` at all. So the axis order — the one output a future folder has to trust — is unpinned. I confirmed this: swapping the inner and outer shift runs in `common` (`verify.rs:1003`) leaves the test green. The PR description's claim that the equality "pins the point assembly" isn't true of the test as written. Please add the check: cut `share.point` back into runs per the documented layout, expand each into its weight table, and contract with `OperationEvalFn::contract`, asserting it equals `share.eval`.

**Non-blocking:**

- `verify.rs:989` — `FieldOps` already requires `Clone` (`crates/field/src/field.rs:91`), so `+ Clone` constrains nothing. Drop it.
- `verify.rs:992` — `wiring_weights` builds all four *scaled* constraint tables and the split discards every one, then re-expands each unscaled. That is four redundant tensor expansions over the padded constraint count, which is the split's dominant cost. Threading the per-operation seed in as an argument (`eq_ind_partial_eval_scalars(operation_batch)` from `call`, all ones from the split) removes it.
- `verify.rs:987` — the count is a compile-time constant and the doc promises the shares line up positionally, so `[OperationShare<E>; OPERATION_COUNT]` via `array::from_fn` states that in the type and drops the `shares.len()` assertion. `Sections::r_x_prime` already takes that shape.
- `verify.rs:1030` — `match op { 0 => …, _ => bmul }` is the only place in the file that reaches an operation by number rather than by name, and `_` maps everything from 3 up to BinMul. Writing the four out, as `call` and `call_native` do, costs a few lines and keeps that consistent.

jimpo (approved):

benji-potamus (approved):
Approve — this supersedes my earlier request for changes.

jimpo has signed off on the rewrite as well, so this is clear to land; no need to hold it for the remaining wiring-matrix simplifications.

PR #2384, opened 2026-08-28
BINIUS-569: Delete the x86_64 simd module in favour of Divisible's broadcast

tcoratger:
Closes [BINIUS-569](https://linear.app/jimpo/issue/BINIUS-569).

Deletes a five-file module whose only method duplicates an abstraction the crate already has.
Pure refactor — no behaviour change, and the machine code is byte-identical.

### The problem

`crates/field/src/arch/x86_64/simd/` is five files and a trait that exist to supply one method.

```rust
pub trait TowerSimdType: Sized + Copy + Underlier {
    fn set_epi_64(val: i64) -> Self;
}
```

`set_epi_64` broadcasts a 64-bit value across every 64-bit lane.
`Divisible<u64>::broadcast` already does exactly that, and is already implemented for all three widths.
They compile to the same instruction:

| width | `TowerSimdType::set_epi_64` | `Divisible<u64>::broadcast` |
|---|---|---|
| `M128` | `_mm_set1_epi64x` | `_mm_set1_epi64x` |
| `M256` | `_mm256_set1_epi64x` | `_mm256_set1_epi64x` |
| `M512` | `_mm512_set1_epi64` | `_mm512_set1_epi64` |

The trait carries `#[allow(dead_code)]`, because without the `gfni` target feature nothing calls it at all.

There is exactly one call site: `Gfni::invert_or_zero`, broadcasting the identity matrix for `vgf2p8affineinvqb`.

The sibling `binius-arith-bench` crate already writes the form this PR converges on.
`crates/arith-bench/src/rijndael/gfni.rs` keeps an `IDENTITY_MAP: u64` and hands it to `U::broadcast`.

### The idea

Delete the trait and put the bound where the need actually is.

Only the inverse broadcasts a qword; `gf2p8mul_epi8` never did.
So `Divisible<u64>` goes on the `InvertOrZero` impl, not on `GfniType`.
`GfniType` is then exactly what its name claims — an underlier with the two GFNI byte instructions.

### The change

```diff
-pub(super) trait GfniType: Copy + TowerSimdType {
+pub(super) trait GfniType: Underlier {

-impl<U: GfniType + Underlier> InvertOrZero for Gfni<PackedPrimitiveType<U, Rijndael8b>> {
+impl<U: GfniType + Divisible<u64>> InvertOrZero for Gfni<PackedPrimitiveType<U, Rijndael8b>> {

-        let transform_after = U::set_epi_64(IDENTITY_MAP);
-        let inv_gfni = U::gf2p8affineinv_epi64_epi8(val_gfni, transform_after);
+        let identity_map = <U as Divisible<u64>>::broadcast(IDENTITY_MAP);
+        let inv_gfni = U::gf2p8affineinv_epi64_epi8(val_gfni, identity_map);
```

`Copy` disappears from the bound because `Underlier` already implies it.

`IDENTITY_MAP` becomes `u64` rather than `i64`, dropping an `as i64` round trip that existed only to feed `set_epi_64`.
It also becomes private — it never had a reader outside its own file.

Its doc now derives the byte layout from the instruction's semantics, which is the part a reader cannot guess:

```text
out.bit[i] = parity(matrix.byte[7 - i] AND inv(b)) XOR imm8.bit[i]
```

So `matrix.byte[k]` drives output bit `7 - k`, and the identity is a descending single-bit pattern
(`0x80, 0x40, ..., 0x01`) rather than an ascending one.

### Soundness

No arithmetic changes — same constant, same instruction, different broadcast call.

The two claims the new doc makes are checked against the hardware rather than asserted:

- `matrix.byte[k]` drives output bit `7 - k` — confirmed for every `k` in `0..8`.
- `IDENTITY_MAP` is the identity on all 256 input bytes.

Existing coverage is already the right shape and was left alone.
The `packed_aes.rs` proptests pin `Gfni::invert_or_zero` lane-by-lane against the scalar
table-based inverse, which is an independent oracle.
No tests were added, because a no-op refactor with identical codegen has nothing new to pin.

### Scope

- **deleted:** `crates/field/src/arch/x86_64/simd/` — 5 files, 62 lines
- **changed:** `gfni_arithmetics.rs` (bounds, constant type, one call) and one `mod` line
- **untouched:** the unused `Mul` impl on `Gfni` (a sibling PR removes it), `arith-bench`, and
  aarch64's unrelated `simd_arithmetic` module of NEON kernels

### Codegen — byte-identical

The claim is that nothing changes, so it is measured rather than asserted.

A probe binary calls `PackedAESBinaryField{16,32,64}x8b::invert_or_zero` behind `#[inline(never)]`,
built from the parent commit and from this commit with `-Ctarget-cpu=native`.
The full `.text` section hashes the same both ways:

| build | `sha256(.text)` |
|---|---|
| parent | `3d955b5cfd49ff071b412dde15e5ec6119c1e8605964c5084087d631f8d44bc4` |
| this PR | `3d955b5cfd49ff071b412dde15e5ec6119c1e8605964c5084087d631f8d44bc4` |

Not one instruction moved anywhere in the linked binary.

Per width the inverse is three instructions before and after, with the broadcast folded into an
embedded `{1toN}` memory operand:

```text
probe_inv_16x:  vmovdqa            (%rsi),%xmm0
                vgf2p8affineinvqb  $0x0,RIP_REL{1to2},%xmm0,%xmm0
                vmovdqa            %xmm0,(%rdi)
```

A wall-clock benchmark is deliberately **not** reported.
Four other agents were building concurrently on this machine (load average ~120), and the same
binary varied between 2.4 and 10.3 ns/element across runs.
With identical machine code there is nothing for a benchmark to detect, so the disassembly is the
evidence.

### Verification

Both CI arch legs, on the final tree:

| check | `-Ctarget-cpu=native` | portable (no `RUSTFLAGS`) |
|---|---|---|
| `cargo test -p binius-field --release` | 273 passed, 0 failed | 256 passed, 0 failed |
| doctests | 2 passed, 0 failed | 2 passed, 0 failed |
| `cargo clippy -p binius-field --all-targets --all-features -- -D warnings` | clean | clean |

- `cargo check --workspace --all-targets` (native) — clean, so nothing downstream depended on the deleted module
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-field --document-private-items` — clean
- `grep` across all crates: no remaining reference to `TowerSimdType`, `x86_64::simd`, or `set_epi_64`

The native leg is the one that exercises this code: `mod gfni` is `cfg`-gated on the `gfni` target
feature, so on the portable leg `gfni_arithmetics.rs` is not compiled at all.
The portable leg's job here is to prove that removing the unconditional `mod simd;` does not break
the non-GFNI build, which it does.


crates/field/src/arch/x86_64/gfni/gfni_arithmetics.rs
```diff
@@ -66,15 +91,16 @@ impl<U: GfniType + Underlier> WideMul for GfniWideMul<PackedPrimitiveType<U, Rij
 	}
 }
 
-impl<U: GfniType + Underlier> InvertOrZero for Gfni<PackedPrimitiveType<U, Rijndael8b>> {
+// `Divisible<u64>` is bounded here rather than on `GfniType`: the affine matrix is read once per
+// 64-bit lane, so only the inverse needs to broadcast a qword.
```
tcoratger:
```suggestion
```

crates/field/src/arch/x86_64/gfni/gfni_arithmetics.rs
```diff
@@ -21,9 +45,10 @@ pub const IDENTITY_MAP: i64 = u64::from_le_bytes([
 	0b00000100,
 	0b00000010,
 	0b00000001,
-]) as i64;
+]);
 
-pub(super) trait GfniType: Copy + TowerSimdType {
+/// SIMD underlier exposing the two GFNI byte instructions the AES packings need.
+pub(super) trait GfniType: Underlier {
```
benji-potamus, line 51: Non-blocking: now that `GfniType: Underlier` is the supertrait, the `+ Underlier` on the `Mul` (line 61) and `WideMul` (line 78) impls is implied and can go. Neither line is inside a diff hunk, so I'm anchoring here.

Same reasoning that made you drop `Copy` — a bound another bound already implies hides which constraints are real. `impl<U: GfniType>` on both compiles clean with `-Ctarget-cpu=native`.

benji-potamus (approved):
Approve.

Good deletion. Five files and a trait collapse into a bound at the one place that needs it, and moving `Divisible<u64>` onto the `InvertOrZero` impl rather than `GfniType` is the right call — `GfniType` now means exactly what its name says. Confirmed no dangling references to `TowerSimdType`, `x86_64::simd`, or `set_epi_64` remain, and `Divisible<u64>` is implemented for all three widths.

The new `IDENTITY_MAP` doc earns its length: the `7 - i` byte reversal in `vgf2p8affineinvqb` is the one thing a reader can't guess, and it matches the instruction's semantics. Worth having next to the constant.

One non-blocking bound nit inline.

jimpo (approved):

PR #2391, opened 2026-08-28
BINIUS-574: Fold the bit axis with GFNI matrix products instead of subset-sum tables

tcoratger:
Closes [BINIUS-574](https://linear.app/jimpo/issue/BINIUS-574).

Computes the bit-axis fold the same way, but as GF(2) matrix products instead of table gathers.
Output is bit-identical.

### The problem

Folding the bit axis maps one 64-bit word to one field element:

```text
    out = sum_{i < 64} bit_i(word) * weight[i]
```

Today that runs as the Method of Four Russians: one 256-entry subset-sum table per byte of the word.

Per word this costs **eight data-dependent 16-byte gathers plus seven XORs**.

The tables are `8 * 256 * 16 B = 32 KiB`, which is the entire first-level data cache — so the fold evicts the very stream it is reading.

### The idea

The map is GF(2)-linear from 64 bits to 128 bits, so it *is* a 128-by-64 bit matrix.

Cut that matrix into 8x8 blocks: sixteen output byte planes by eight input byte planes.

Those 128 blocks describe the whole fold in **1 KiB** instead of 32 KiB.

An 8x8 GF(2) affine product with a zero constant term applies one such block to 64 input bytes in a single instruction.

So the fold becomes matrix products with **no data-dependent loads at all**.

### Per 64 words

- **before:** 512 gathers + 448 XORs, against 32 KiB of tables
- **after:** 128 matrix products + ~64 XOR-class operations + two transpose networks, against 1 KiB of matrices

→ 15 cycles per word down to 5.

### The bit order

The hardware reads the matrix row that produces output bit `i` from byte `7 - i` of the matrix qword:

```text
    parity( A.qword[q].byte[7 - i] & x.qword[q].byte[n] )
```

Matching that against what the fold needs pins the block layout exactly:

```text
    bit_i(byte_k(out)) = XOR_{t < 8} bit_t(byte_j(word)) & bit_{8k + i}(weight[8j + t])
```

So a block is byte `k` of that plane's eight weights, bit-transposed, laid down in reverse byte order.

A test walks all 64x128 single-bit weights one at a time, which pins that layout column by column.

### Soundness

Both paths sum the same 64 terms — the weights of the word's set bits.

Field addition is XOR, which is associative and commutative, so only the summation order differs.

The affine product is per byte lane with a zero constant term, so an output word depends on its own input word alone.

A zero word therefore folds to zero, matching what the tables give for the zero-padded tail.

### Scope

- **new:** the kernel module, with an architecture-gated block path and a portable table path
- **changed:** the bit-axis folder now batches 64 words
- **untouched:** the subset-sum tables, which remain the reference and the fallback

Of the ~800 added lines, **139 are tests** and the rest is the kernel plus its dispatch.

No bench is included — the numbers below were taken with a throwaway harness, kept out of the diff.

The row-axis fold has the same table shape but a different streaming contract, so it is left for a separate change.

### Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --document-private-items` — clean
- `cargo test --workspace` — clean; `cargo test -p binius-prover` 88 + 17 passed
- proptests pin the batched fold bit-identical to the table fold, including the short-batch tail and the all-zero / all-ones / single-bit words
- two tests pin the dispatch, so the properties above are not comparing the tables against themselves

### Benchmark

Single-threaded, idle 16-core Zen 5. Baseline reproduced to 0.3% across two independent invocations.

| words | tables | GFNI | speedup | ns/word |
|---|---|---|---|---|
| 256 | 0.378 us | 0.159 us | 2.37x | 1.48 -> 0.62 |
| 1,024 | 1.383 us | 0.523 us | **2.65x** | 1.35 -> 0.51 |
| 4,096 | 5.383 us | 2.050 us | 2.63x | 1.31 -> 0.50 |
| 16,384 | 22.89 us | 9.63 us | 2.38x | 1.40 -> 0.59 |
| 65,536 | 89.60 us | 37.24 us | 2.41x | 1.37 -> 0.57 |
| 262,144 | 351.8 us | 150.6 us | 2.34x | 1.34 -> 0.57 |
| 1,048,576 | 1.588 ms | 0.773 ms | 2.05x | 1.51 -> 0.74 |
| 4,194,304 | 9.12 ms | 7.59 ms | 1.20x | 2.17 -> 1.81 |

The taper at the largest size is expected: the kernel is now cheap enough that a 64 MiB output stream becomes store-bandwidth bound.

### DIFF
--- a/crates/prover/src/fold_word/kernel/gfni.rs
+++ b/crates/prover/src/fold_word/kernel/gfni.rs
@@ -1,0 +1,333 @@
+// Copyright 2026 The Binius Developers
+
+//! A table-free bit-axis fold, built on the GFNI 8x8 GF(2) matrix instruction.
+//!
+//! The fold sums the weights of a word's set bits:
+//!
+//! ```text
+//!     out = sum_{i < 64} bit_i(word) * weight[i]
+//! ```
+//!
+//! Field addition is XOR, so this is GF(2)-linear: a 128-by-64 bit matrix with the weights as its
+//! columns.
+//!
+//! Cut into 8x8 blocks that is sixteen output byte planes by eight input byte planes.
+//!
+//! Those 128 blocks describe the fold in 1 KiB, against 32 KiB of subset-sum tables.
+//!
+//! One affine product applies a block to 64 input bytes, so a call folds 64 words with no
+//! data-dependent load.
+//!
+//! # Block layout
+//!
+//! The hardware reads the row producing output bit `i` from byte `7 - i` of the matrix qword:
+//!
+//! ```text
+//!     parity( A.qword[q].byte[7 - i] & x.qword[q].byte[n] )
+//! ```
+//!
+//! Matching that against the fold pins bit `t` of byte `7 - i` of block `(j, k)` to bit `8k + i`
+//! of weight `8j + t`.
+//!
+//! Derived from the flock prover's table-free row fold (MIT/Apache-2.0), written independently.
+
+use std::{
+	arch::x86_64::{
+		__m512i, _mm512_gf2p8affine_epi64_epi8, _mm512_loadu_si512, _mm512_permutex2var_epi64,
+		_mm512_permutexvar_epi8, _mm512_set1_epi64, _mm512_setr_epi64, _mm512_storeu_si512,
+		_mm512_ternarylogic_epi64, _mm512_unpackhi_epi64, _mm512_unpacklo_epi64, _mm512_xor_si512,
+	},
+	array,
+	mem::MaybeUninit,
+};
+
+use binius_core::word::Word;
+use binius_field::{BinaryField, Ghash128b};
+
+use super::WORDS_PER_BATCH;
+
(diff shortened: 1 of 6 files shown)


crates/prover/src/fold_word/kernel/gfni.rs
```diff
@@ -0,0 +1,333 @@
...
+//! The hardware reads the row producing output bit `i` from byte `7 - i` of the matrix qword:
+//!
+//! ```text
+//!     parity( A.qword[q].byte[7 - i] & x.qword[q].byte[n] )
+//! ```
+//!
+//! Matching that against the fold pins bit `t` of byte `7 - i` of block `(j, k)` to bit `8k + i`
+//! of weight `8j + t`.
+//!
+//! Derived from the flock prover's table-free row fold (MIT/Apache-2.0), written independently.
```
jimpo: Link to the flock codebase for attribution

PR #2394, opened 2026-08-28
BINIUS-578: Scale every GHASH lane by X with one vector sequence

tcoratger:
Closes [BINIUS-578](https://linear.app/jimpo/issue/BINIUS-578).

Scales a GHASH element by $X$ with one vector sequence instead of one scalar step per 128-bit lane. Same values out — the identity is pinned by proptest against the field multiply.

### The problem

Multiplying a GHASH element by $X$ is not a multiplication.

GHASH is $\mathbb{F}_2[X]$ modulo $X^{128} + X^7 + X^2 + X + 1$.
An element is a polynomial whose 128 binary coefficients sit in 128 bits.
Scaling by $X$ raises every coefficient one degree, which is a one-bit left shift.
The one complication is the coefficient that lands on $X^{128}$, off the top of the register.
The modulus rewrites it: $X^{128} = X^7 + X^2 + X + 1$, the constant `0x87`.

```
x * X = (x << 1)              when bit 127 of x is clear
      = (x << 1) + 0x87       when bit 127 of x is set
```

A shift and one masked exclusive or. No carry-less multiply, no reduction loop.

The crate had this right at the scalar level and wrong everywhere above it.
`Ghash128b::mul_x` did the shift over a `u128`, and on x86 the GHASH underlier is a vector register — so that is a round trip out to general-purpose registers and back.
`packed_ghash_sq` then called it once per 128-bit lane through `Divisible`, so scaling the four lanes of an `M512` coordinate register paid four such round trips.

### The idea

Do the shift and the fold with vector instructions, on the whole underlier at once.

Eight instructions serve any width, plus two loop-invariant constants:

```
shifted  = shl_1_epi64(x) ^ move_64_to_hi(shr_63_epi64(x))   // the 128-bit shift, bit 127 falls off
overflow = broadcast_bit_127(x) & POLY                       // the X^128 term, folded back
result   = shifted ^ overflow
```

No instruction shifts a whole 128-bit lane by one bit, which is why the shift is built from the two 64-bit halves.
The bit leaving the low half belongs at position 64, which is where moving it up a half-lane puts it.
The bit leaving the high half is the $X^{128}$ term, and it is the one the mask folds back in.

### The change

- `MulX` — a new trait beside `MulXWide` in `arithmetic_traits`, carrying the identity `mul_x(x) == x * X`.
- `GhashMulX` — the per-architecture dispatch point underneath it; one blanket impl covers every GHASH packing width.
- `ClMulUnderlier` splits in two. Scaling by $X$ needs no carry-less multiply, so `move_64_to_hi`, `shl_1_epi64` and `shr_63_epi64` move into a new `GhashLanes` supertrait along with the bit-127 mask, and `ClMulUnderlier: GhashLanes` keeps only `clmulepi64`.

```
- fn ghash_mul_x<U: Divisible<M128>>(u: U) -> U {           // one scalar step per lane
-     U::from_iter(Divisible::<M128>::value_iter(u)
-         .map(|lane| Ghash128b::from_underlier(lane).mul_x().to_underlier()))
- }
+ impl<U: GhashMulX> MulX for PackedPrimitiveType<U, Ghash128b> {   // one sequence, all lanes
+     fn mul_x(self) -> Self { Self::from_underlier(self.to_underlier().ghash_mul_x()) }
+ }
```

Two things came along with it:

- `POLY` had three definitions, one per architecture backend. It is a cryptographic constant, so it now has one, in the portable module the other two import from.
- `SlicedPackedField::square_transpose` allocated a `Vec` inside its per-lane loop — `PSub::WIDTH` allocations per call, column length up to 256. The buffer is allocated once and reused.

### Soundness

No protocol surface changes; this is an arithmetic implementation swap.
`mul_x(x) == x * X` is pinned against the field multiply, an oracle that runs the CLMUL product and the modular reduction and touches none of the new code.
Every specialized path is additionally pinned against the `u128` scalar reference at every vector width.

### Scope

- **new:** `MulX`, `GhashMulX`, `GhashLanes`, the vector `mul_x` for x86 (SSE2 / AVX2 / AVX-512) and NEON, the `u128` reference.
- **changed:** `SlicedGhashSq*::{reduce, square, invert_or_zero}`, `GhashSqSquare`, `GhashSqInvert`, `GhashSqHybridWideMul::reduce`, plus the two items above.
- **untouched:** `Ghash128b::mul_inv_x` — it has no caller inside the field crate, so it keeps its scalar form rather than growing an unused vector path.
- **untouched:** both GHASH² packing layouts. Neither has a consumer outside `crates/field` yet and both call `mul_x`, so this helps whichever one survives; picking between them is separate work.

### Benchmark

Zen 5 (9950X3D), `-Ctarget-cpu=native`, minimum of 60 rounds per measurement, otherwise-idle machine.
Old and new run in the same binary back to back on the same inputs, after asserting they agree. Nanoseconds per operation.

The primitive itself:

| | old | new | speedup |
|---|---|---|---|
| `mul_x` on `M256` | 1.08 | 0.59 | **1.8×** |
| `mul_x` on `M512` | 3.38 | 0.52 | **6.5×** |

The composite GHASH² operations move much less, and I would rather say so than quote the primitive's ratio and let it stand for the whole:

| | old | new | speedup |
|---|---|---|---|
| `SlicedGhashSq4x256b` square | 0.98 | 0.75 | **1.30×** |
| `SlicedGhashSq4x256b` mul | 1.76 | 1.68 | 1.05× |
| `SlicedGhashSq4x256b` invert | 40.9 | 39.8 | 1.03× |
| `SlicedGhashSq2x256b` square | 1.50 | 1.46 | 1.03× |
| `SlicedGhashSq2x256b` mul | 3.27 | 3.27 | 1.00× |
| `SlicedGhashSq2x256b` invert | 61.0 | 59.8 | 1.02× |

`mul_x` is one step among several CLMUL products in `mul`, and among a whole Itoh-Tsujii inversion in `invert_or_zero`.
On an idle core the out-of-order engine hides most of the old path's extra work behind that latency, so only `square` — where `mul_x` is a large share of a short dependency chain — shows a clear gain.

Under CPU contention the gap is much wider: measuring the same harness on a saturated machine, the composites showed 1.3×-2.3× and `mul_x` on `M512` up to 7×, because the old path's extra instructions and its register-file round trips compete for issue slots that are no longer free. A prover saturating every core looks more like that than like the table above, but the idle numbers are the honest baseline.

Plain GHASH is untouched and does not regress: mul 2.30 / 1.22 / 0.60, square 1.55 / 0.82 / 0.39, inner product 1.56 / 0.82 / 0.41 ns per scalar at widths 1 / 2 / 4.

### Verification

Field-crate scoped, both feature legs, since `MulX` has a CLMUL path and a portable one:

- `RUSTFLAGS="-Ctarget-cpu=native" cargo test -p binius-field --release` — 277 passed, 0 failed
- `RUSTFLAGS="-Ctarget-cpu=native" cargo clippy -p binius-field --all-targets --all-features -- -D warnings` — clean
- `cargo test -p binius-field --release` (portable leg) — 260 passed, 0 failed

Extra configurations checked because this PR changes which target features gate what:

- `-Ctarget-feature=+avx2,+avx512f` with no CLMUL feature at all — 276 passed, including all six `mul_x` tests. This is the configuration the `GhashLanes` split unlocks.
- `cargo check -p binius-field --lib --target aarch64-unknown-linux-gnu` with `+neon,+aes` — clean.
- `cargo check -p binius-field --lib --target wasm32-unknown-unknown`, with and without `+simd128` — clean.

The aarch64 NEON `mul_x` is compile-checked but not executed — no emulator on this host — so its correctness rests on the argument in the body comment plus CI's aarch64 leg.

New tests:

- `mul_x_matches_the_scalar_reference` pins each vector width against the `u128` reference. Boundary values (`0`, `1`, `X`, `0x87`, `1 << 127`, `(1 << 127) | 1`, `(1 << 127) | (1 << 63)`, `u128::MAX`) are laid out so neighbouring lanes differ, which is what would catch a sequence letting one lane's carry leak into the next; random input follows.
- `mul_x_is_multiplication_by_x_{1x,2x,4x}` pin every packing against the field multiply.
- The existing GHASH² proptests already check `square`, `invert_or_zero`, `Mul` and deferred `wide_mul` against the scalar reference, so the rewiring is covered end to end without new tests.

### Context

Worth recording: `arith-bench/src/ghash/clmul.rs` has had this exact branch-free sequence all along.
The benchmark crate and the field crate carry two independent abstraction layers over the same instructions — `OpsClmul` with `[U; 3]` limbs against `ClMulUnderlier` with `WideGhashProduct`, each with its own `mul_x_wide`.
This closes one gap between them; unifying them is a separate conversation.

### DIFF
--- a/crates/field/src/arithmetic_traits.rs
+++ b/crates/field/src/arithmetic_traits.rs
@@ -6,10 +6,31 @@ use std::{
 	ops::{Add, AddAssign, Sub, SubAssign},
 };
 
+use crate::underlier::Underlier;
+
 /// Value that can be multiplied by itself
 pub trait Square {
 	/// Returns the value multiplied by itself
 	fn square(self) -> Self;
+}
+
+/// Scales a field element by `X`, the generator of its polynomial basis.
+///
+/// A one-bit shift plus a masked exclusive or, not a field multiply.
+/// The scaling is `GF(2)`-linear, so it applies equally to an unreduced product.
+pub trait MulX {
+	/// Returns the element scaled by `X`.
+	#[must_use]
+	fn mul_x(self) -> Self;
+}
+
+/// An underlier whose 128-bit lanes each hold a GHASH element that can be scaled by `X`.
+///
+/// The per-architecture dispatch point behind [`MulX`] for GHASH.
+pub trait GhashMulX: Underlier {
+	/// Returns the value with every 128-bit lane scaled by `X`.
+	#[must_use]
+	fn ghash_mul_x(self) -> Self;
 }
 
 /// A field type that supports widening (unreduced) multiplication.
--- a/crates/field/src/arithmetic_traits.rs
+++ b/crates/field/src/arithmetic_traits.rs
@@ -6,31 +6,29 @@ use std::{
 	ops::{Add, AddAssign, Sub, SubAssign},
 };
 
-use crate::underlier::Underlier;
-
 /// Value that can be multiplied by itself
 pub trait Square {
 	/// Returns the value multiplied by itself
 	fn square(self) -> Self;
 }
 
-/// Scales a field element by `X`, the generator of its polynomial basis.
+/// Scales a value by `X`, the generator of the field's polynomial basis.
 ///
 /// A one-bit shift plus a masked exclusive or, not a field multiply.
-/// The scaling is `GF(2)`-linear, so it applies equally to an unreduced product.
+///
+/// The scaling is `GF(2)`-linear, so it commutes with the modular reduction.
+/// That is what lets one trait serve both a reduced element and an unreduced product:
+///
+/// ```text
+///     reduce(mul_x(wide)) == mul_x(reduce(wide))
+/// ```
+///
+/// Scaling an unreduced product folds the `X` of an irreducible polynomial into a reduction the
+/// caller is going to pay for anyway.
 pub trait MulX {
-	/// Returns the element scaled by `X`.
+	/// Returns the value scaled by `X`.
 	#[must_use]
 	fn mul_x(self) -> Self;
-}
-
-/// An underlier whose 128-bit lanes each hold a GHASH element that can be scaled by `X`.
-///
-/// The per-architecture dispatch point behind [`MulX`] for GHASH.
-pub trait GhashMulX: Underlier {
-	/// Returns the value with every 128-bit lane scaled by `X`.
-	#[must_use]
-	fn ghash_mul_x(self) -> Self;
 }
 
 /// A field type that supports widening (unreduced) multiplication.
@@ -58,18 +56,6 @@ pub trait WideMul: Sized {
 
 	fn wide_mul(a: Self, b: Self) -> Self::Output;
 	fn reduce(wide: Self::Output) -> Self;
-}
-
-/// An unreduced widening product (a [`WideMul::Output`]) that can be scaled by the field element
-/// `X` while still unreduced.
-///
-/// Scaling by `X` and the modular reduction are both `GF(2)`-linear, and they commute:
-/// `reduce(wide.mul_x_wide()) == reduce(wide).mul_x()`. Doing the scaling on the unreduced product
-/// lets an extension-field multiply fold the `X` of its irreducible polynomial into a product it is
-/// going to reduce anyway, saving a reduction over scaling the reduced coordinate.
-pub trait MulXWide {
-	/// Returns the unreduced product scaled by `X`.
-	fn mul_x_wide(self) -> Self;
 }
 
 /// Value that can be inverted
@@ -108,6 +94,23 @@ macro_rules! impl_square_with {
 }
 
 pub(crate) use impl_square_with;
+
+macro_rules! impl_mul_x_with {
+	($name:ident @ $($strategy:tt)*) => {
+		impl $crate::arithmetic_traits::MulX for $name {
+			#[inline]
+			fn mul_x(self) -> Self {
+				<$($strategy)* <$name> as ::bytemuck::TransparentWrapper<$name>>::peel(
+					$crate::arithmetic_traits::MulX::mul_x(
+						<$($strategy)* <$name> as ::bytemuck::TransparentWrapper<$name>>::wrap(self),
+					),
+				)
+			}
+		}
+	};
+}
+
+pub(crate) use impl_mul_x_with;
 
 macro_rules! impl_invert_with {
 	($name:ident @ $($strategy:tt)*) => {
--- a/crates/field/src/arch/aarch64/arithmetic/ghash.rs
+++ b/crates/field/src/arch/aarch64/arithmetic/ghash.rs
@@ -20,7 +20,7 @@ use super::super::m128::M128;
 use crate::{
 	Ghash128b as GhashB128, WideMul,
 	arch::portable::arithmetic::ghash::POLY,
-	arithmetic_traits::{GhashMulX, MulXWide, Square},
+	arithmetic_traits::{MulX, Square},
 	packed_fields::primitive::PackedPrimitiveType,
 };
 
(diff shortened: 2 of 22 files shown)


crates/field/src/arithmetic_traits.rs
```diff
@@ -6,12 +6,33 @@ use std::{
...
+pub trait MulX {
+	/// Returns the element scaled by `X`.
+	#[must_use]
+	fn mul_x(self) -> Self;
+}
+
+/// An underlier whose 128-bit lanes each hold a GHASH element that can be scaled by `X`.
+///
+/// The per-architecture dispatch point behind [`MulX`] for GHASH.
+pub trait GhashMulX: Underlier {
```
jimpo: I don't understand the need for this. All the arithmetic traits are implemented on PackedField, not on Underliers. Can't MulX be used (on PackedFields) instead?

crates/field/src/arithmetic_traits.rs
```diff
@@ -6,12 +6,33 @@ use std::{
...
 pub trait Square {
 	/// Returns the value multiplied by itself
 	fn square(self) -> Self;
 }
 
+/// Scales a field element by `X`, the generator of its polynomial basis.
+///
+/// A one-bit shift plus a masked exclusive or, not a field multiply.
+/// The scaling is `GF(2)`-linear, so it applies equally to an unreduced product.
+pub trait MulX {
```
jimpo, line 28: Can this completely replace MulXWide? Like switch MulXWide uses to this instead?

tcoratger: Yes — done in `9c049be`.

They were the same operation on two types. Scaling by `X` is `GF(2)`-linear, so it commutes with the reduction:

```
reduce(mul_x(wide)) == mul_x(reduce(wide))
```

which is exactly the property each backend's `mul_x_wide` test was already asserting. So `MulXWide` is now `MulX` implemented for the unreduced product, and the trait is gone. Three impls and one call site renamed; nothing else moved.

`MulXWide` was never exported from the crate root, so this is not a public API change.

crates/field/src/arithmetic_traits.rs
tcoratger: Fair — it is gone in `9c049be`, and you are right that it did not belong on `Underlier`.

To say why it was there: `PackedField for PackedPrimitiveType<U, F>` is a single blanket impl covering every binary packing, because `define_packed_binary_field!` emits a type *alias*. So `impl MulX for PackedPrimitiveType<U, Ghash128b>` written once cannot vary by width — it needs a per-`U` hook, and that hook was `GhashMulX`.

But the crate already has the right mechanism for that, and I should have used it: the strategy wrapper, exactly as `Square`, `InvertOrZero` and `WideMul` do. So now:

```rust
impl_mul_x_with!(PackedBinaryGhash1x128b @ GhashMulX1x);
impl_mul_x_with!(PackedBinaryGhash2x128b @ GhashMulX2x);
impl_mul_x_with!(PackedBinaryGhash4x128b @ GhashMulX4x);
```

with `GhashMulX{1,2,4}x` resolved per architecture beside the other GHASH strategies — the vector sequence where the width has a shift, the portable lane walk otherwise. `MulX` is now implemented only on packings, the scalar field, and the unreduced product.

I invoked the macro directly rather than adding a fifth slot to `define_packed_binary_field!`, since AES has no scaling and would have had to pass a filler.

Two things fell out for free:

- the scalar impl no longer needs its own route — it delegates to the width-one packing, like the other scalar ops.
- the `ScaledUnderlier` impl is deleted. It was hand-rolling per-lane composition that the portable wrapper's `Divisible<u128>` walk already does at any width.

Net −21 lines and one fewer trait.

### Verification

Every arm the dispatch now goes through, compiled and (on x86) run:

| configuration | result |
|---|---|
| baseline x86-64 (portable lane walk) | 262 tests pass |
| `+sse2,+pclmulqdq` (1x vector) | 263 pass |
| `+avx2`, no vpclmulqdq (2x vector) | 270 pass |
| `+avx2,+vpclmulqdq` | 271 pass |
| `+avx512f,+vpclmulqdq` (4x vector) | 279 pass |
| `target-cpu=native` | 279 pass |
| aarch64 (NEON sequence) | lib checks clean |
| wasm32, with and without simd128 | lib checks clean |

Plus the whole workspace: 71 test binaries green, clippy `-D warnings` clean, nightly fmt clean.

One thing I checked and want to record, since it looked wrong at first: moving the shift helpers out of `ClMulUnderlier` into `GhashLanes` leaves `impl GhashLanes for M256`/`M512` without a `target_feature` gate, which would emit AVX2/AVX-512 instructions in a baseline build. It is fine — `arch/x86_64/mod.rs` gates the whole `packed_ghash_256`/`packed_ghash_512` modules on `avx2`/`avx512f`, so each impl only compiles where its instructions exist. The baseline run above is the check: it selects the portable modules instead.

Rebased onto main; the conflict was the import block in `packed_ghash_sq.rs` against #2393 dropping `MulFromWideMul`.

jimpo (approved):

PR #2395, opened 2026-08-28
BINIUS-575: Preprocess loop-constant GHASH multipliers to spend one fewer carry-less multiply

tcoratger:
Closes [BINIUS-575](https://linear.app/jimpo/issue/BINIUS-575).

Spends 5 carry-less multiplies instead of 6 wherever a whole buffer is multiplied by one constant.
Same field element, so proof bytes are unchanged.

### The problem

Every hot loop in the prover multiplies a whole buffer by one **loop-constant** field element.

An NTT butterfly multiplies by a broadcast twiddle, held fixed across a block.

A sumcheck fold, an eq expansion, and a basefold mask all multiply by a broadcast challenge.

Each of those costs 4 carry-less multiplies for the product and 2 for the reduction, plus 2 lane shifts.

The multiplier is the same for millions of values, so work spent once on it is free.

### The idea

Reduction is a ring homomorphism, so the *variable* operand can be split into its two 64-bit limbs:

```text
    v = v_lo + X^64 * v_hi
    t * v == t * v_lo + (t * X^64 mod p) * v_hi   (mod p)
```

Precompute the companion `t * X^64 mod p` once, when the multiplier is prepared.

Each term is then a 128-by-64 product of degree at most 190.

Their sum spans **three** 64-bit limbs instead of the schoolbook product's four, so one reduction stage finishes it instead of two.

### Per multiply

- **before:** 4 product + 2 reduction carry-less multiplies, 2 lane shifts
- **after:** 4 product + 1 reduction carry-less multiply, 1 lane shift

→ 6 down to 5. Preparing costs one carry-less multiply, so break-even is at a single use and preparing is never a loss.

### Soundness

The single reduction stage is exact.

The high limb has degree at most 62, and folding it in multiplies by `0x87`, reaching degree at most 69 — well below 128, so nothing spills.

The result is the same field element the schoolbook path gives.

### Scope

- **new:** one defaulted method on the packed field trait, plus the GHASH implementation and a trivial strategy wrapper
- **changed:** the NTT butterflies, the eq and tensor expansions, the sumcheck fold, the shared-store fold, the basefold mask fold
- **untouched:** single-shot scalar claim folds, where preparing for one use is a loss

The method returns the multiplier as a closure, so no impl declares an associated type and no call site needs a new bound.

Its default is ordinary multiplication, which is what every field but `GF(2^128)` uses.

The per-type hook behind it stays private to the field crate, as the squaring and inversion hooks do.

Every field takes the trivial form — the prepared multiplier is the multiplier — except `GF(2^128)` and its carry-less-multiply packings.

AArch64 keeps the trivial form for now: the identity ports to its carry-less multiply unchanged, but that wants measuring on real hardware first.

No bench is included — the numbers below came from a throwaway harness, kept out of the diff.

### Verification

- `cargo test -p binius-field -p binius-math -p binius-ip-prover -p binius-iop-prover` — 662 passed
- a proptest pins the prepared multiply equal to ordinary multiplication for every packing width, plus boundary patterns rotated so each lane differs
- the NTT reference-comparison tests still pass, so the transform output is unchanged
- cross-compiled for AArch64, x86-64 without AVX-512, both wasm targets, and the empty-`RUSTFLAGS` portable leg

### End-to-end proving

**At full thread count this does not move end-to-end proving.**

The same prove benches built from both trees with identical flags, before/after rounds interleaved, median of 3, on an idle 32-core AVX-512 + VPCLMULQDQ box:

| case | before | after | ratio |
|---|---|---|---|
| keccak 1 MiB, 32 threads | 435.8 ms | 436.3 ms | 1.00x |
| sha256 1 MiB, 32 threads | 390.5 ms | 394.3 ms | 0.99x |
| keccak 64 KiB, 32 threads | 37.19 ms | 37.55 ms | 0.99x |
| keccak 64 KiB, 1 thread | 99.05 ms | 97.49 ms | **1.016x** |

Only the serial figure moves, by 1.6%, with near-disjoint ranges.

At 32 threads these loops are memory-bandwidth bound, so removing one carry-less multiply of six buys nothing -- the same effect this PR already notes for the 24-variable expansion.

Inverting Amdahl on the serial figure puts the loop-constant multiplies at roughly 12% of serial prove time, and effectively none of the parallel critical path.

So the microbenchmark wins below are real but do not reach the prover as shipped.

### Benchmark

Instruction counts come from disassembling the same bench source built both ways, so they do not depend on machine load.

| NTT routine | before | after |
|---|---|---|
| fused pair, 4 butterflies | 24 CLMUL | 20 CLMUL |
| fused pair, 3 butterflies | 18 CLMUL | 15 CLMUL |
| breadth-first layer, 256/512-bit | 36 CLMUL | 32 CLMUL |

Wall clock, single-threaded, idle 16-core Zen 5:

| bench | speedup |
|---|---|
| NTT, median of 36 cases | **1.167x** |
| NTT at `log_d` 16 / 20 / 24 | 1.168x / 1.174x / 1.130x |
| eq expansion, 16 variables | **1.190x** |
| eq expansion, 20 variables | 1.167x |
| eq expansion, 24 variables | neutral |
| prepared vs plain multiply, in-binary | 1.19x median |

The ceiling is 6/5 = 1.20x, and the compute-bound sizes sit on it.

The 24-variable expansion is memory-bandwidth bound at 256 MiB, so removing a multiply buys nothing there.

### DIFF
--- a/crates/field/src/arch/aarch64/packed_ghash_128.rs
+++ b/crates/field/src/arch/aarch64/packed_ghash_128.rs
@@ -11,6 +11,10 @@
 /// `GhashClMulWideMul`.
 pub type GhashWideMul1x<T> = super::arithmetic::ghash::GhashClMulWideMul<T>;
 
+/// Prepared-multiplier wrapper for the `PackedBinaryGhash1x128b` packing: the trivial one for now.
+/// The x86_64 split-multiplier form ports to PMULL unchanged, and this is where it would land.
+pub type GhashPreparedMul1x<T> = crate::arch::PreparedMulFromMul<T>;
+
 /// Square wrapper for the `PackedBinaryGhash1x128b` packing: the PMULL carryless-multiply square.
 pub type GhashSquare1x<T> = super::arithmetic::ghash::GhashClMul<T>;
 
--- a/crates/field/src/arch/mod.rs
+++ b/crates/field/src/arch/mod.rs
@@ -11,9 +11,9 @@ cfg_if! {
 	if #[cfg(all(target_arch = "x86_64"))] {
 		mod x86_64;
 		pub use x86_64::{packed_aes_128, packed_aes_256, packed_aes_512, packed_ghash_128, packed_ghash_256, packed_ghash_512, M128, M256, M512, m256_from_u128s};
-		pub use x86_64::packed_ghash_128::{GhashWideMul1x, GhashSquare1x, GhashInvert1x};
-		pub use x86_64::packed_ghash_256::{GhashWideMul2x, GhashSquare2x, GhashInvert2x};
-		pub use x86_64::packed_ghash_512::{GhashWideMul4x, GhashSquare4x, GhashInvert4x};
+		pub use x86_64::packed_ghash_128::{GhashWideMul1x, GhashSquare1x, GhashInvert1x, GhashPreparedMul1x};
+		pub use x86_64::packed_ghash_256::{GhashWideMul2x, GhashSquare2x, GhashInvert2x, GhashPreparedMul2x};
+		pub use x86_64::packed_ghash_512::{GhashWideMul4x, GhashSquare4x, GhashInvert4x, GhashPreparedMul4x};
 		pub use x86_64::packed_ghash_sq_256::GhashSqWideMul1x;
 		pub use x86_64::packed_aes_128::{AesWideMul16x, AesSquare16x, AesInvert16x};
 		pub use x86_64::packed_aes_256::{AesWideMul32x, AesSquare32x, AesInvert32x};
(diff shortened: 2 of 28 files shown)


jimpo (commented):
Overall, I don't think this 20% for some multiplications is worth the complexity and I'd rather close it, but if you do think it's an important optimization then I'd prefer a different trait interface.

How much does this affect E2E proving performance?

As far as trait interface, I'd recommend just extending the PackedField trait with something like this:

```
trait PackedField {
    type PreprocessedMul: Fn(Self) -> Self;

    fn preprocess_mul(val: Self::Scalar) -> Self::PreprocessedMul;
}
```

This would have to be implemented for every PackedField, whether the default PreprocessedMul impl of course just does a regular multiplication by a stored constant.

PR #2400, opened 2026-08-29
BINIUS-582: Fill the idle 32-bit lanes in SHA-256, so every block costs 364 AND constraints

tcoratger:
Closes [BINIUS-582](https://linear.app/jimpo/issue/BINIUS-582).

Fills two 32-bit lanes the SHA-256 hasher was leaving idle. No behavior change — same digests, same gadget contracts, fewer gates.

### The problem

An AND constraint covers both 32-bit halves of a 64-bit word at once.
So a gate that drives only one half costs the same as one that drives two, and wastes half of what it bought.

The hasher is built around that fact — the schedule and the rounds both pack two lanes.
Two places still leave a lane idle.

### The idea

Count what one compression cannot avoid, then check the circuit against it.

XOR, rotation and shift are affine, so they cost nothing.
Only the AND of two affine forms is witnessed, and each one eliminates exactly one summand:

```
n-summand modular sum      n - 1 AND constraints
Ch, Maj                    1 AND constraint each
```

That fixes a compression's nonlinear work:

```
schedule      48 words x 3 sums                144
rounds        64 x (Ch + Maj + 7 sums)         576
feed-forward  8 sums                             8
                                              ----
one compression, one lane                      728
both 32-bit lanes carry it                     364
```

So 364 is the floor. Two paths sat above it.

### Where the lanes idle

**1. The feed-forward in `sha256_compress`.**

That gadget splits one compression across the two lanes: rounds 0..32 above, 32..64 below.
By the time the feed-forward runs the split is spent and both lanes are free again — but the eight final additions were issued one word per gate.

→ 368 constraints instead of 364.

**2. The trailing odd block in `sha256_fixed`.**

Blocks are compressed in pairs through the two-lane core.
An odd block count leaves one block over, and it went through that same paired core with the second lane dead.

→ 728 constraints for one compression instead of 364.

The comment there claimed the two cores emit the same circuit without a chip. They do not: 728 against 364.

### The change

Pair the feed-forward words, so four gates do the work of eight:

```
- for i in 0..8:  iadd_32(state_in[i], state[i])           // 8 gates, high lane idle
+ for j in 0..4:  iadd_32(pair(in[2j], in[2j+1]),          // 4 gates, both lanes busy
+                         pair(st[2j], st[2j+1]))
```

Send the trailing odd block to the single-lane core, which fills the second lane from the block itself:

```
- sha256_compress_2x(sub, state, block)    // 728, second lane dead
+ sha256_compress(sub, state, block)       // 364, splits its own 64 rounds across both lanes
```

The variable-length hasher already routed its trailing block this way, so it picks up the feed-forward saving only.

### What this is not

This does not change the decomposition of SHA-256, and it is not the kind of win circuit-golf scoreboards chase.

Those exploit a ripple-carry adder that costs one row per bit, where carry-save trees, vanishing low-bit carries and the trailing zeros of `K[t]` all buy rows.
Here a 32-bit addition is one constraint, so an $n$-summand sum costs $n-1$ however it is grouped, and none of those tricks apply.

Lane utilization is the only lever this cost model leaves, and this PR takes it.

### Scope

- **changed:** the feed-forward in `sha256_compress`, and the trailing-block branch in `sha256_fixed`.
- **new:** `a_compression_spends_only_its_nonlinear_work` and `every_block_costs_the_same`, which pin the bound with its derivation.
- **re-blessed:** the four example snapshots that move — `sha256`, `bitcoin_headers`, `bitcoin_p2pkh`, `zklogin`. The other eleven are unchanged.
- **re-derived:** the two recursion cost models this moves, in `merkle.rs` and `challenger.rs`.
- **untouched:** the round function, the message schedule, `sha256_varlen`'s compression chain, and the chip path.

### Soundness

Neither change touches what is constrained, only how many gates carry it.

- The paired feed-forward clears each summand before it enters the low lane, and the shift into the high lane clears its own operand. Carries never cross bit 32, so the two sums stay independent.
- The single-lane core accepts an input state whose high halves are dirty, which is exactly what a pair leaves behind, and it returns clean ones. So the digest clearing now runs on even block counts only.
- The chip path is unaffected: paired compressions still reach a registered `Sha256Compress2x`, and one fewer call is cheaper there too, since instances pad to a power of two.

### Verification

- `cargo clippy -p binius-circuits -p binius-examples --all-targets --all-features -- -D warnings` — clean.
- `cargo +nightly-2026-07-01 fmt --all --check` — clean.
- `cargo test -p binius-circuits --release` — 409 + 3 + 2 pass, including the RFC KATs, the chip tests and the two new bounds.
- `cargo test -p binius-recursion --release` — 34 + 9 + 2 + 1 + 12 + 13 pass.
- `check-snapshot` on all fifteen examples — matches.

### Benchmark

Circuit cost, from `cargo run --release --example sha256 -- stat --message-len N`:

| message | blocks | AND before | AND after | AND per block | committed trace |
|---|---|---|---|---|---|
| 128 B | 3 | 1,407 | **1,092** | 469 → **364** | 1,786 → 1,407 (both $2^{11}$) |
| 192 B | 4 | 1,456 | 1,456 | 364 → 364 | unchanged |
| 1024 B | 17 | 6,503 | **6,188** | 382.5 → **364** | 8,254 of $2^{14}$ → 7,875 of $2^{13}$ |

Prover, from `HASH_MAX_BYTES=N cargo bench -p binius-examples --bench sha256` (Basefold, `log_inv_rate` 1, same machine and session):

| message | witness gen | proof gen | verify | proof size |
|---|---|---|---|---|
| 128 B | 8.59 → **7.26 µs** (−15.5%) | 7.79 → **6.02 ms** (−22.7%) | 1.393 → **1.216 ms** (−12.7%) | 45.42 → 45.45 KiB |
| 1024 B | 39.78 → **38.55 µs** (−3.1%) | 16.35 → **12.00 ms** (−26.6%) | 5.389 → **4.970 ms** (−7.8%) | 98.78 → **75.45 KiB** |

Even block counts are the control and do not move.
The outsized 1024-byte win is the committed trace crossing a power-of-two boundary — a windfall of where that boundary happens to sit, not a general 26%.
The durable result is the "AND per block" column: **every fixed-length SHA-256 now costs exactly 364 AND constraints per block, for any message length**, where an odd block count previously paid up to 469.

### Effect on the other SHA-256 examples

The two changes land differently across the tree, so here is each one measured on its own.

| example | trailing block | feed-forward | net AND | net committed |
|---|---|---|---|---|
| `sha256` (1024 B) | −311 AND, −369 committed | −4 AND, −10 committed | **−315** | **−379** |
| `bitcoin_headers` | none | −40 AND, −60 committed | **−40** | **−60** |
| `bitcoin_p2pkh` | none | −4 AND, −6 committed | **−4** | **−6** |
| `zklogin` | none | −12 AND, +12 committed | **−12** | **+12** |

Only `sha256` has an odd trailing block to reclaim, so it takes the whole of that win — including the trace crossing $2^{14}$ down to $2^{13}$.

`zklogin` is the one place the feed-forward pairing is not a free win.
Packing two words into one gate needs the low operand masked, which is two shifts, and a lowered term carries at most two.
Where the input state is already shift-deep, that pushes a linear definition over the budget and it gets committed instead of inlined — here 12 AND constraints traded for 12 committed values and 24 ZERO constraints.

That is 0.005% of a 253,647-word trace and moves no table across a power of two, against a clean win in the other three, so the pairing stays.

### The recursion cost models

`binius-recursion` pins SHA-256 costs in two module-level tables and asserts them in tests, so this
change moves them. One consequence is worth stating rather than renumbering past.

A lone compression used to cost 368 against 364 per block on a shared core, so pairing bought four
constraints. Both are 364 now, so **a shared core and two lone cores cost the same per compression**.

| what is measured | before | after |
|---|---|---|
| a one-element leaf digest | 376 | **372** |
| a sixteen-element leaf digest | 2,263 | **1,948** |
| one inner node on its own core | 384 | **380** |
| one inner node sharing a core | 380 per node | 380 per node |
| one climbed level of a lone opening | 384 | **380** |
| one climbed level of a paired opening | 377 per query | 377 per query |
| one observed 64-byte block, amortized | 396 | **395** |

The sixteen-element leaf is the one large move: 256 bytes is five blocks, so it had an odd trailing
block to reclaim.

Every paired figure is unchanged, so the layer-depth analysis stands as written — its break-even
`2^L = 0.99 * n_q` comes from `377 / 380`, and neither number moved.

What did need rewriting is the claim that sharing a core saves the split's overhead. It no longer
saves anything on the compression; a leaf and a climbed level still gain, but from the work around
the compression being paid once instead of twice. `a_shared_core_barely_beats_two_lone_nodes` is
renamed to `a_shared_core_costs_what_two_lone_nodes_cost` and its invariant restated to match.

### DIFF
--- a/crates/circuits/src/sha256/compress.rs
+++ b/crates/circuits/src/sha256/compress.rs
@@ -211,13 +211,37 @@ pub fn sha256_compress(builder: &CircuitBuilder, state_in: State, m: [Wire; 16])
 		builder.assert_eq("sha256_compress.round_state", hinted, expected);
 	}
 
-	// Feed-forward, then drop the high lane.
+	// Feed-forward, two output words per gate.
 	//
-	// It carries the round-31 state the split already consumed.
-	// A caller reads one compression's output, in the low 32 bits.
+	// The rounds have consumed the split, so both lanes are free again.
+	// Pairing the words refills them, and 8 additions become 4:
+	//
+	//     low lane  <- H[2j]
+	//     high lane <- H[2j + 1]
+	//
+	// Each summand is cleared before it enters the low lane.
+	//
+	// - A state word's high half holds the round-31 state the split already spent.
+	// - A caller's high half is not required to be empty.
+	//
+	// The shift into the high lane clears its own operand, so only the low one needs it.
+	let pair =
+		|lo: Wire, hi: Wire| builder.bxor(clear_high_bits(builder, lo, 32), builder.shl(hi, 32));
+	let sums: [Wire; 4] = array::from_fn(|j| {
+		builder.iadd_32(
+			pair(state_in.0[2 * j], state_in.0[2 * j + 1]),
+			pair(state.0[2 * j], state.0[2 * j + 1]),
+		)
+	});
+
+	// A caller reads one compression's output, in the low 32 bits with the high 32 zero.
 	State(array::from_fn(|i| {
-		let sum = builder.iadd_32(state_in.0[i], state.0[i]);
-		clear_high_bits(builder, sum, 32)
+		let sum = sums[i / 2];
+		if i % 2 == 0 {
+			clear_high_bits(builder, sum, 32)
+		} else {
+			builder.shr(sum, 32)
+		}
 	}))
 }
 
@@ -676,13 +700,82 @@ fn small_sigma_1(b: &CircuitBuilder, x: Wire) -> Wire {
 #[cfg(test)]
 mod tests {
 	use binius_core::word::Word;
-	use binius_frontend::{CircuitBuilder, Hint, Wire};
+	use binius_frontend::{CircuitBuilder, CircuitStat, Hint, Wire};
 	use proptest::prelude::*;
 
 	use super::{
 		IV, Sha256Compress2x, State, compress_2x_gates, populate_message_block, ref_compress,
 		sha256_compress, sha256_compress_2x, sha256_compress_2x_seq,
 	};
+
+	/// AND constraints one 32-bit SHA-256 compression cannot go below.
+	///
+	/// XOR, rotation and shift are affine, so they cost nothing.
+	/// Only the AND of two affine forms is witnessed, and each one eliminates exactly one summand:
+	///
+	/// ```text
+	///     n-summand modular sum          n - 1
+	///     Ch, Maj                            1  each
+	/// ```
+	///
+	/// A compression's nonlinear work is therefore fixed:
+	///
+	/// ```text
+	///     schedule    48 words x 3 sums                      144
+	///     rounds      64 x (Ch + Maj + 7 sums)               576
+	///     feedforward 8 sums                                   8
+	///                                                       ----
+	///                                                        728
+	/// ```
+	///
+	/// A gate constrains both 32-bit halves of its word at once, so two lanes share that total.
+	const AND_PER_COMPRESSION: usize = 364;
+
+	/// AND constraints a circuit spends on the state `f` returns.
+	///
+	/// The output is bound to public wires, since dead code elimination would drop it otherwise.
+	/// Binding is affine, so it lands on ZERO constraints and leaves the AND count alone.
+	fn and_cost(f: impl FnOnce(&CircuitBuilder) -> State) -> usize {
+		let builder = CircuitBuilder::new();
+		let out = f(&builder);
+		for wire in out.0 {
+			let public = builder.add_inout();
+			builder.assert_eq("out", wire, public);
+		}
+		CircuitStat::collect(&builder.build()).n_and_constraints
+	}
+
+	#[test]
+	fn a_compression_spends_only_its_nonlinear_work() {
+		let inputs = |b: &CircuitBuilder| {
+			let state = State::public(b);
+			let m: [Wire; 16] = std::array::from_fn(|_| b.add_inout());
+			(state, m)
+		};
+
+		// One compression across two lanes: 32 rounds each, and paired feed-forward words.
+		let single = and_cost(|b| {
+			let (state, m) = inputs(b);
+			sha256_compress(b, state, m)
+		});
+		assert_eq!(single, AND_PER_COMPRESSION);
+
+		// Two compressions across two lanes: one lane each, all 64 rounds.
+		let paired = and_cost(|b| {
+			let (state, m) = inputs(b);
+			sha256_compress_2x(b, state, m)
+		});
+		assert_eq!(paired, 2 * AND_PER_COMPRESSION);
+
+		// Chaining the pair costs no more: the hint that breaks the dependency is affine to bind.
+		let chained = and_cost(|b| {
+			let (state, _) = inputs(b);
+			let blocks: [[Wire; 16]; 2] =
+				std::array::from_fn(|_| std::array::from_fn(|_| b.add_inout()));
+			sha256_compress_2x_seq(b, state, blocks)
+		});
+		assert_eq!(chained, 2 * AND_PER_COMPRESSION);
+	}
 
 	/// A test circuit that proves a knowledge of preimage for a given state vector S in
 	///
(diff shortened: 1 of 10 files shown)


crates/circuits/src/sha256/mod.rs
```diff
@@ -144,29 +144,21 @@ pub fn sha256_fixed(builder: &CircuitBuilder, message: &[Wire], len_bytes: usize
 		block_idx += 2;
 	}
 	if block_idx < n_blocks {
+		// The trailing odd block has no partner, so the paired core would idle a whole lane on
+		// it. The single-lane core fills that lane from the block itself, splitting its own 64
+		// rounds across the two halves, and costs half as much.
```
benji-potamus, line 149: Non-blocking: the fact that makes this routing safe is that `state` here still carries the previous pair's first compression in its high half, and the single-lane core tolerates that. The comment above the loop only grants it to "the paired core", so a reader arriving at this branch has to go and check `sha256_compress`'s preconditions. Worth one line here:

```suggestion
		// rounds across the two halves, and costs half as much.
		//
		// It takes a pair's leftover high halves, since it reads an input state's low half only.
```

crates/circuits/src/sha256/compress.rs
```diff
@@ -211,13 +211,37 @@ pub fn sha256_compress(builder: &CircuitBuilder, state_in: State, m: [Wire; 16])
...
+	//     high lane <- H[2j + 1]
+	//
+	// Each summand is cleared before it enters the low lane.
+	//
+	// - A state word's high half holds the round-31 state the split already spent.
+	// - A caller's high half is not required to be empty.
+	//
+	// The shift into the high lane clears its own operand, so only the low one needs it.
+	let pair =
+		|lo: Wire, hi: Wire| builder.bxor(clear_high_bits(builder, lo, 32), builder.shl(hi, 32));
```
benji-potamus: Non-blocking: this is the third copy of the same lane packer in this file — `State::pack_4x64b`'s `pack_pair`, `sha256_compress_2x_seq`'s `pack`/`clear`, and now `pair`. They differ only in whether the low operand is cleared, and this one's variant (clear the low, let `shl` clear the high) is the safe default for all three. A single free function beside `clear_high_bits` would let all three callers drop their local lambda.

benji-potamus (approved):
Approve.

Nice change. The cost model in the PR description is the right frame — a gate constrains both halves, so lane utilization is the only lever — and the two idle lanes it finds are real. Both changes hold up: `iadd_32` keeps the halves independent by construction, the paired feed-forward masks the low operand and lets `shl` clear the high one, and the single-lane core's documented tolerance of dirty high halves is exactly what the paired core leaves behind, so moving the digest clearing to even block counts is sound.

The tests earn their place. `every_block_costs_the_same` pins the actual claim (per-block cost invariant across parities), not just a number, and the derivation in `a_compression_spends_only_its_nonlinear_work` gives the constant a reason to be 364. Snapshot deltas and the re-derived recursion tables are all consistent with the claimed deltas, including the 374 in the challenger docs.

Two non-blocking notes inline.

jimpo (approved):

crates/circuits/src/sha256/compress.rs
tcoratger: Done in d8861b91 — `pack_u32_words` now sits beside `clear_high_bits` and the two lambdas in `sha256_compress` and `sha256_compress_2x_seq` are gone.

One correction on making it the default for all three: at `pack_4x64b` it is not free. Every caller already hands it cleared words — the varlen chain masks `state_second` explicitly for exactly this reason — so masking again commits one redundant intermediate per word:

```
zklogin, pack_4x64b using the masking packer
  AND constraints  186,820 -> 186,820   (unchanged)
  ZERO constraints  33,513 ->  33,601   (+88)
  gates            388,535 -> 388,735   (+200)
  committed trace  253,647 -> 253,735   (+88)
```

So the state packer keeps the unmasked form, and I wrote down the precondition it was relying on silently, which is the part that actually made it a footgun:

```rust
/// # Preconditions
///
/// - Every state word holds its value in the low 32 bits, with the high 32 empty.
///   - The word lifted into the high half has its own high bits discarded by the shift.
///   - The word entering the low half is passed through unmasked.
```

Net effect is a pure refactor: identical gate sequence at all three sites, `every_block_costs_the_same` still pins `n_blocks * 364`, and the sha256 and zklogin snapshots match unchanged.

PR #2404, opened 2026-08-31
BINIUS-581: Mask the evaluation every oracle opening publishes

tcoratger:
Closes [BINIUS-581](https://linear.app/jimpo/issue/BINIUS-581).

Gives the precommit segment the blinding that masks its revealed evaluation, so both committed segments carry the same padding.

### The problem

Each oracle opening hands the verifier **one evaluation of that oracle in the clear**.

The ZK wrapper decrypts the claim and records it as a public wire, so the value itself is not hidden.

That evaluation is a linear function of the committed segment, so it needs masking like anything else.

### The dummy wires cannot mask it

This is the concern the issue raises parenthetically, and it holds — it is structural, not hypothetical.

An evaluation weights each wire by its coefficient in the wiring relation.

That relation only ever sums over wires that appear in a multiplication constraint:

```
for (r_x_tensor_i, MulConstraint { a, b, c }) in zip(r_x_tensor, mul_constraints) {
    for (dst, operand) in zip(&mut acc, [a, b, c]) {
        // only wires named by some operand contribute
        *dst += r_x_tensor_i * sum(r_y_tensor[w.index] for w in operand.wires() if w.segment == segment);
    }
}
```

The plain dummy wires appear in no constraint, so their coefficient is **exactly zero**.

Adding more of them is therefore inert against an evaluation, which is why this PR does not touch the wire count.

### Where the randomness has to live

A dummy constraint's three wires *do* sit in a constraint, so they reach the relation with a nonzero weight.

That is the only place a revealed evaluation can be masked from.

| segment | dummy wires (weight 0) | dummy constraints (in the relation) |
|---|---|---|
| precommit | `q + 1` | **0** |
| private | `q + 1` | 2 |

The private segment was already covered. The precommit segment had **nothing** masking its evaluation.

That matters: in the plain Spartan path the precommit segment holds real witness data, allocated by the circuit author.

### The change

The precommit segment now gets dummy constraints too, in the padded constraint system and the witness layout.

That lets the two ad-hoc `n_dummy_constraints: 0` overrides in the prover go, rather than being mirrored a third time.

### Why the count is two

The old `TODO: Document why these are necessary` on `N_DUMMY_CONSTRAINTS` is now answered, and the answer is a budget match:

- Verification publishes **four** linear functionals of a committed segment: the three operand evaluations `a_eval`, `b_eval`, `c_eval`, which the mulcheck verifier reads straight off the transcript, plus the segment's own batched claim.
- A dummy constraint spends three wires but supplies only **two** free field elements, because the prover sets its `c` wire to `a * b`.

Two constraints therefore carry `2 * 2 = 4` free elements against four revealed values.

### Soundness

No protocol, transcript or challenge-distribution change — this only widens the random padding a committed segment carries.

A new test pins the mechanism the whole issue turns on: the dummy wires are asserted **absent** from the wiring relation, and every dummy-constraint wire asserted **present** in it.

### Scope

- **changed:** both committed segments get dummy constraints, in the padded constraint system, the witness layout, and the two prover packing sites that mirrored them
- **new:** a test pinning which blinding reaches the wiring relation
- **untouched:** `n_dummy_wires` and `n_dummy_constraints` keep their values, so no existing blinding is reduced, and no call site changes

### Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --document-private-items --no-deps` — clean
- `RUSTFLAGS="-D warnings" cargo build --target wasm32-unknown-unknown` — clean
- `cargo test` on the four affected crates — green, including the ZK prove/verify round trips and the proof-size fixture

### Cost

| | before | after |
|---|---|---|
| proof size, one-constraint fixture | 73,936 B | 73,968 B |

The fixture has a single real constraint, so two more pad it from four to eight and add one mulcheck round.

Segment sizes do not change (`log_precommit` and `log_private` stay at 4), and a circuit with real constraints absorbs the two extra without moving the power-of-two rounding at all.

### DIFF
--- a/crates/spartan-frontend/src/constraint_system.rs
+++ b/crates/spartan-frontend/src/constraint_system.rs
@@ -371,19 +371,18 @@ pub struct BlindingInfo {
 ///
 /// # Why this value
 ///
-/// Verification publishes four linear functionals of a committed segment:
+/// Everything verification publishes touches a committed segment only through that segment's
+/// three operand contributions `(A_S, B_S, C_S)`. The operand evaluations are those plus the
+/// other segments' shares, and the segment's own batched claim is `A_S + lambda * B_S +
+/// lambda^2 * C_S`, already in their span. So three values need masking, not four.
 ///
-/// ```text
-///     3  operand evaluations, read from the transcript in the clear
-///     1  the segment's own batched claim
-///     -----
-///     4  values needing a free random field element each
-/// ```
+/// One dummy constraint cannot mask them. Its wires contribute `(alpha * a, alpha * b,
+/// alpha * a * b)`, because the prover pins the third wire to the product of the other two.
+/// Its share of `C_S` is therefore fixed by its shares of `A_S` and `B_S`, and a verifier
+/// holding all three recovers a relation among the real wires.
 ///
-/// A dummy constraint spends three wires but supplies only two free field elements.
-/// Its third wire is pinned to the product of the other two, so the constraint holds.
-///
-/// Two constraints therefore carry `2 * 2 = 4` free elements, matching the four revealed values.
+/// Two constraints break that: `a` and `b` of each are free, and the distribution they induce
+/// on `(A_S, B_S, C_S)` is statistically close to uniform.
 const N_DUMMY_CONSTRAINTS: usize = 2;
 
 impl BlindingInfo {
--- a/crates/spartan-frontend/src/constraint_system.rs
+++ b/crates/spartan-frontend/src/constraint_system.rs
@@ -362,44 +362,41 @@ pub struct BlindingInfo {
 	pub n_dummy_constraints: usize,
 }
 
-/// Dummy multiplication constraints added to every committed segment.
+/// Dummy multiplication constraints appended to every committed segment.
 ///
 /// A wire in no constraint has coefficient zero in the wiring relation.
-/// So the plain dummy wires cannot mask a revealed evaluation.
+/// So no count of plain dummy wires masks a value the relation reveals.
 ///
-/// The three wires of a dummy constraint do sit in a constraint.
-/// That is what carries free randomness into the relation.
+/// The three wires of a dummy constraint do sit in a constraint, so they reach the relation.
 ///
-/// Two of them cover the one evaluation revealed per oracle, with room to spare.
+/// # Why this value
+///
+/// Verification publishes four linear functionals of a committed segment:
+///
+/// ```text
+///     3  operand evaluations, read from the transcript in the clear
+///     1  the segment's own batched claim
+///     -----
+///     4  values needing a free random field element each
+/// ```
+///
+/// A dummy constraint spends three wires but supplies only two free field elements.
+/// Its third wire is pinned to the product of the other two, so the constraint holds.
+///
+/// Two constraints therefore carry `2 * 2 = 4` free elements, matching the four revealed values.
 const N_DUMMY_CONSTRAINTS: usize = 2;
 
 impl BlindingInfo {
-	/// The blinding a committed segment needs, given how much of it the verifier sees.
+	/// The blinding a committed segment needs when FRI opens it at `n_test_queries` positions.
 	///
-	/// Two different things are revealed, and each needs its own randomness.
-	///
-	/// Each of the `n_test_queries` FRI queries opens one Merkle leaf.
-	/// That reveals one codeword symbol, a linear function of the whole padded segment.
+	/// Each query opens one Merkle leaf, revealing one codeword symbol of the segment.
+	/// A symbol is a fixed linear function of the segment.
 	/// So `n_test_queries` random wires make every opened symbol uniform and independent.
 	///
 	/// One further wire covers the leaves that are never opened.
 	/// Their hashes still travel in the authentication paths, and the leaves carry no salt.
 	/// The spare degree of randomness keeps an unopened leaf unguessable, as a salt would.
-	///
-	/// Separately the verifier learns `n_oracle_openings` evaluations of the segment in the clear.
-	/// An evaluation weights each wire by its coefficient in the wiring relation.
-	/// A wire in no constraint is weighted by zero, so no count of plain dummy wires masks one.
-	///
-	/// The dummy constraints mask it instead, because their wires do sit in a constraint.
-	///
-	/// # Panics
-	///
-	/// Panics unless the dummy constraints cover every revealed evaluation.
-	pub const fn for_fri_queries(n_test_queries: usize, n_oracle_openings: usize) -> Self {
-		assert!(
-			N_DUMMY_CONSTRAINTS >= n_oracle_openings,
-			"a revealed evaluation with no dummy constraint behind it is unmasked"
-		);
+	pub const fn for_fri_queries(n_test_queries: usize) -> Self {
 		Self {
 			n_dummy_wires: n_test_queries + 1,
 			n_dummy_constraints: N_DUMMY_CONSTRAINTS,
@@ -473,9 +470,12 @@ impl<F: Field> WitnessLayout<F> {
 	}
 
 	pub fn with_blinding(self, info: BlindingInfo) -> Self {
-		// Both committed segments carry the same blinding: dummy wires to mask the query
-		// openings, and three wires per dummy constraint to mask the revealed evaluation. Keep
-		// this in sync with the padded constraint system in the verifier crate.
+		// Both committed segments carry the same blinding:
+		//
+		//     dummy wires                   -> mask the codeword symbols FRI opens
+		//     3 wires per dummy constraint  -> mask the evaluations sent in the clear
+		//
+		// Keep this in sync with the padded constraint system in the verifier crate.
 		let blinding_size = info.n_dummy_wires + 3 * info.n_dummy_constraints;
 
 		let total_precommit = self.n_precommit as usize + blinding_size;
--- a/crates/spartan-frontend/src/constraint_system.rs
+++ b/crates/spartan-frontend/src/constraint_system.rs
@@ -362,21 +362,47 @@ pub struct BlindingInfo {
 	pub n_dummy_constraints: usize,
 }
 
+/// Dummy multiplication constraints added to every committed segment.
+///
+/// A wire in no constraint has coefficient zero in the wiring relation.
+/// So the plain dummy wires cannot mask a revealed evaluation.
+///
+/// The three wires of a dummy constraint do sit in a constraint.
+/// That is what carries free randomness into the relation.
+///
+/// Two of them cover the one evaluation revealed per oracle, with room to spare.
+const N_DUMMY_CONSTRAINTS: usize = 2;
+
 impl BlindingInfo {
-	/// The blinding a committed segment needs when FRI opens it at `n_test_queries` positions.
+	/// The blinding a committed segment needs, given how much of it the verifier sees.
 	///
-	/// Each query opens one Merkle leaf, revealing one codeword symbol of the segment.
-	/// A symbol is a fixed linear function of the segment.
+	/// Two different things are revealed, and each needs its own randomness.
+	///
+	/// Each of the `n_test_queries` FRI queries opens one Merkle leaf.
+	/// That reveals one codeword symbol, a linear function of the whole padded segment.
 	/// So `n_test_queries` random wires make every opened symbol uniform and independent.
 	///
 	/// One further wire covers the leaves that are never opened.
 	/// Their hashes still travel in the authentication paths, and the leaves carry no salt.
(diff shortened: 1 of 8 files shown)


crates/spartan-frontend/src/constraint_system.rs
```diff
@@ -362,21 +362,47 @@ pub struct BlindingInfo {
...
+	/// A wire in no constraint is weighted by zero, so no count of plain dummy wires masks one.
+	///
+	/// The dummy constraints mask it instead, because their wires do sit in a constraint.
+	///
+	/// # Panics
+	///
+	/// Panics unless the dummy constraints cover every revealed evaluation.
+	pub const fn for_fri_queries(n_test_queries: usize, n_oracle_openings: usize) -> Self {
+		assert!(
+			N_DUMMY_CONSTRAINTS >= n_oracle_openings,
```
benji-potamus: Blocking: please drop `n_oracle_openings` and this assert.

`N_DUMMY_CONSTRAINTS` is a constant, so the parameter never influences the returned `BlindingInfo` — every accepted argument yields the same value, and the only work the parameter does is compare `2 >= 1` at every call site. It also gives assurance it cannot back up: `BlindingInfo`'s fields are `pub`, and `crates/spartan-prover/tests/wrapper_integration_test.rs:61` already builds one by struct literal, bypassing this path entirely.

The reasoning is worth keeping — put it in the doc comment on `N_DUMMY_CONSTRAINTS` and revert the signature to `for_fri_queries(n_test_queries)`. That deletes this parameter, the `OPENINGS_PER_ORACLE` constant, and the four call-site churn, and leaves the actual fix (precommit gets dummy constraints) as the whole diff.

crates/spartan-frontend/src/constraint_system.rs
```diff
@@ -362,21 +362,47 @@ pub struct BlindingInfo {
...
 
+/// Dummy multiplication constraints added to every committed segment.
+///
+/// A wire in no constraint has coefficient zero in the wiring relation.
+/// So the plain dummy wires cannot mask a revealed evaluation.
+///
+/// The three wires of a dummy constraint do sit in a constraint.
+/// That is what carries free randomness into the relation.
+///
+/// Two of them cover the one evaluation revealed per oracle, with room to spare.
```
benji-potamus: Blocking: this is the line that replaces the old `TODO: Document why these are necessary`, and as written it documents the wrong reason, which is worse than the TODO.

Two things the count actually turns on, neither of which is one-per-opening:

1. A dummy constraint has three wires but supplies only **two** free field elements — `pack_and_blind_witness` sets `c = a * b`.
2. `precommit_claim` is not the only thing revealed. `verify_mulcheck` reads `a_eval`, `b_eval`, `c_eval` from the transcript in the clear, and each is a linear functional of the whole witness, committed segments included. So the randomness a segment carries has to cover those three operand evaluations as well, not just its own batched claim.

Four free elements against three operand evaluations is why two is the number. Suggested:

```suggestion
/// Two is what the revealed values need. The verifier reads `a_eval`, `b_eval` and `c_eval`
/// in the clear, on top of each segment's own batched claim, and one dummy constraint carries
/// only two free field elements, because the prover sets its `c` wire to `a * b`.
```

crates/spartan-verifier/src/lib.rs
```diff
@@ -58,6 +58,15 @@ use crate::constraint_system::{BlindingInfo, ConstraintSystemPadded};
...
 pub const SECURITY_BITS: usize = 96;
 
+/// Evaluations of each committed oracle that verification publishes in the clear.
+///
+/// The precommit, private and mask oracles are each opened once.
+/// Every opening hands the verifier one evaluation of that oracle.
+///
+/// A revealed evaluation needs its own randomness inside the wiring relation.
+/// So this count bounds how much blinding a committed segment must carry.
+pub const OPENINGS_PER_ORACLE: usize = 1;
```
benji-potamus: Goes away with the parameter above. Two things to note if you keep any of it: the rule this constant feeds is enforced in `binius-spartan-frontend` against a frontend-private constant, so the two halves of one invariant sit in different crates; and the mask oracle named in the doc is not a committed segment and carries no `BlindingInfo`, so it is outside what this count bounds.

crates/spartan-verifier/src/constraint_system.rs
```diff
@@ -209,13 +213,76 @@ mod tests {
...
+			let dummy_constraint_base = n_circuit + info.n_dummy_wires;
+			for offset in 0..3 * info.n_dummy_constraints {
+				assert!(
+					segment.contains(&(dummy_constraint_base + offset)),
+					"a dummy constraint wire never reached the wiring relation"
+				);
+			}
+
+			// One free random value per revealed evaluation is the whole point of the count.
+			assert!(info.n_dummy_constraints >= crate::OPENINGS_PER_ORACLE);
```
benji-potamus: Non-blocking: this restates the constructor's own assert, and it is loop-invariant, so it re-checks the same constant once per segment. Drop it — the two support assertions above are what this test is for.

benji-potamus (changes requested):
Request changes — on the new constant and its assert only. The behaviour change itself is right and wanted.

The leak is real and the diagnosis is correct: the wiring transparent polynomial in `spartan-verifier/src/wiring.rs` sums only over wires named by some operand, so a wire in no constraint has coefficient zero, and `precommit_claim` was therefore a linear functional of the precommit segment with no randomness anywhere behind it. Giving both committed segments the same blinding fixes that, and it lets the two ad-hoc `n_dummy_constraints: 0` overrides in the prover go — nice to see the special case deleted rather than mirrored a third time.

What I'd like changed is the surface added around it. `OPENINGS_PER_ORACLE` and the `n_oracle_openings` parameter never affect the result, and the doc line justifying `N_DUMMY_CONSTRAINTS = 2` names a reason that isn't the operative one. Details inline. Landing this as just the segment-symmetry change plus a corrected comment on the constant would be the strongest version.

crates/spartan-frontend/src/constraint_system.rs
tcoratger: Done — parameter, assert, `OPENINGS_PER_ORACLE` and all four call sites are gone in 05678d6. You are right on both counts: the parameter could not influence the result, and the public fields mean the check was never load-bearing anyway. The reasoning now lives on `N_DUMMY_CONSTRAINTS`, so the diff is the segment-symmetry change plus a corrected comment.

crates/spartan-frontend/src/constraint_system.rs
tcoratger: You are right, and I checked both halves:

- `pack_and_blind_witness` sets `c = a * b`, so a dummy constraint spends three wires and supplies two free elements.
- `verify_mulcheck` does `channel.recv_array()` for `[a_eval, b_eval, c_eval]`, so those three are in the clear on top of each segment's batched claim.

Four revealed functionals against `2 * 2 = 4` free elements is the reason, and the doc now says that. I split your suggestion across lines and put the count in a small block, to stay inside the one-sentence-per-line house style:

```rust
/// # Why this value
///
/// Verification publishes four linear functionals of a committed segment:
///
/// ```text
///     3  operand evaluations, read from the transcript in the clear
///     1  the segment's own batched claim
///     -----
///     4  values needing a free random field element each
/// ```
///
/// A dummy constraint spends three wires but supplies only two free field elements.
/// Its third wire is pinned to the product of the other two, so the constraint holds.
///
/// Two constraints therefore carry `2 * 2 = 4` free elements, matching the four revealed values.
```

I also pluralised the two inline comments in the padded constraint system and the test, which said "the revealed evaluation" as if the batched claim were the only one.

crates/spartan-verifier/src/lib.rs
tcoratger: Deleted. Your two notes are also why it was the wrong shape even kept: the invariant would have straddled two crates against a frontend-private constant, and the mask oracle carries no `BlindingInfo` at all, so it was never in scope for this count.

crates/spartan-verifier/src/constraint_system.rs
tcoratger: Dropped. The two support assertions are what the test is for — the dummy wires absent from the wiring relation, every dummy-constraint wire present.

jimpo (approved):

crates/spartan-frontend/src/constraint_system.rs
```diff
@@ -362,6 +362,30 @@ pub struct BlindingInfo {
...
+///     3  operand evaluations, read from the transcript in the clear
+///     1  the segment's own batched claim
+///     -----
+///     4  values needing a free random field element each
+/// ```
+///
+/// A dummy constraint spends three wires but supplies only two free field elements.
+/// Its third wire is pinned to the product of the other two, so the constraint holds.
+///
+/// Two constraints therefore carry `2 * 2 = 4` free elements, matching the four revealed values.
```
benji-potamus: Blocking: the count is still off, and the "budget match" framing is not how the masking works.

The batched claim is not a fourth independent functional of the segment. `evaluate_segment_wiring_mle` computes it as `A_S + λ·B_S + λ²·C_S`, so it lies in the span of the same three operand contributions the operand evaluations touch. Everything the transcript publishes reaches segment `S` only through `(A_S, B_S, C_S)` — three values, not four. (`private_claim` is then derived by subtraction, publishing nothing new.)

And the free elements are not a fungible budget of one per revealed value. One dummy constraint contributes `(α·a, α·b, α·a·b)` to `(A_S, B_S, C_S)`, so its share of `C_S` is *determined* by its shares of `A_S` and `B_S`. That is the actual reason one is not enough: a verifier holding all three recovers `C_S - C_real = (A_S - A_real)(B_S - B_real)/α`, a relation on the real wires. Two constraints leave `(a1, b1, a2, b2)` free and make the triple statistically close to uniform.

`2` is the right value, but the stated rationale gives the wrong scaling rule — under it, publishing another λ-combination of the same three would demand a third constraint, when it would demand nothing.

```suggestion
/// # Why this value
///
/// Everything verification publishes touches a committed segment only through that segment's
/// three operand contributions `(A_S, B_S, C_S)`. The operand evaluations are those plus the
/// other segments' shares, and the segment's own batched claim is `A_S + lambda * B_S +
/// lambda^2 * C_S`, already in their span. So three values need masking, not four.
///
/// One dummy constraint cannot mask them. Its wires contribute `(alpha * a, alpha * b,
/// alpha * a * b)`, because the prover pins the third wire to the product of the other two.
/// Its share of `C_S` is therefore fixed by its shares of `A_S` and `B_S`, and a verifier
/// holding all three recovers a relation among the real wires.
///
/// Two constraints break that: `a` and `b` of each are free, and the distribution they induce
/// on `(A_S, B_S, C_S)` is statistically close to uniform.
```

crates/spartan-verifier/src/constraint_system.rs
```diff
@@ -209,13 +213,73 @@ mod tests {
...
+			for operand in [&constraint.a, &constraint.b, &constraint.c] {
+				for wire in operand.wires() {
+					let slot = match wire.segment {
+						WitnessSegment::Precommit => Some(0),
+						WitnessSegment::Private => Some(1),
+						WitnessSegment::Public => None,
+					};
+					if let Some(slot) = slot {
+						in_support[slot].insert(wire.index as usize);
+					}
```
benji-potamus: Non-blocking: `continue` in the match arm drops the `Option` and the `if let`.

```suggestion
					let slot = match wire.segment {
						WitnessSegment::Precommit => 0,
						WitnessSegment::Private => 1,
						WitnessSegment::Public => continue,
					};
					in_support[slot].insert(wire.index as usize);
```

benji-potamus (changes requested):
The core change is right and I'm happy with it: both committed segments now get dummy constraints, the frontend layout and the padded constraint system compute the same `n_dummy_wires + 3 * n_dummy_constraints` per segment, and the two prover packing sites just take `*cs.blinding_info()` instead of mirroring a third override. The new test pins the one property the whole issue turns on — dummy wires absent from the wiring relation, dummy-constraint wires present — and it is the right test to have. The parameter, the `OPENINGS_PER_ORACLE` constant, the assert and the call-site churn are all gone; the workspace has no trace of them left.

The remaining blocker is the `N_DUMMY_CONSTRAINTS` rationale, which still miscounts. Details inline.

jimpo (approved):

benji-potamus (approved):
Approve — this supersedes my earlier request for changes.

The `N_DUMMY_CONSTRAINTS` rationale now counts correctly: three values need masking, since the operand evaluations and the segment's batched claim all reach the segment through `(A_S, B_S, C_S)`, and the argument for why one constraint is not enough — its `C` share being pinned to the product — is the operative one.

One thing left over: the PR description still carries the old four-revealed-values budget match (`2 * 2 = 4`), which the code comment no longer says. Worth updating before merge so the two agree.

PR #2407, opened 2026-08-31
BINIUS-584: Build a large equality expansion as an outer product

tcoratger:
Closes [BINIUS-584](https://linear.app/jimpo/issue/BINIUS-584).

Computes the same equality-indicator expansion, writing every coefficient once instead of rewriting it once per remaining coordinate.
Output is bit-identical; the multiplication count does not change.

### The problem

The expansion is built one coordinate at a time.
Each round doubles the length: it reads everything expanded so far, rewrites all of it, and appends as much again.

Summed over the rounds, a result written once costs three passes' worth of traffic:

```
    reads   2^n - 1
    writes  2^(n+1) - 2
```

Past the last cache level, that traffic — not the multiplications — is what the routine is bound by.

### The idea

A coefficient is a product over the coordinates, so cutting the index cuts the product:

```
    index = high * 2^k + low
    coeff = high_expansion[high] * low_expansion[low]
```

The low expansion is therefore *one block* of the result, and every other block is that block scaled by a single coefficient of the high expansion.

```
    block 0   [ lo_0  lo_1  ...  lo_m ]        <- built in place, scaled last
    block 1   [ lo_0  lo_1  ...  lo_m ] * hi_1
    block 2   [ lo_0  lo_1  ...  lo_m ] * hi_2
    ...
```

Every coefficient is written once, and the block read once per block is small enough to stay in cache.

### Per coefficient

- **before:** written once per remaining coordinate, read once per remaining coordinate
- **after:** written once, against a cache-resident block

→ three passes of traffic collapse to one, for the same $2^n$ multiplications.

### The change

```
- tensor_prod_eq_ind_reserved::<Cube, P, Data>(FieldBuffer::new(0, buffer), point)
+ match split_low_len::<P>(point.len()) {
+     Some(low_len) => split_expand::<Cube, P, Data>(seed, point, low_len),
+     None => tensor_prod_eq_ind_reserved::<Cube, P, Data>(seed, point),
+ }
```

The low expansion is built straight into the store's first block, which is where the result wants it anyway, so nothing beyond the result's own length is ever allocated at full size.
The high expansion is one scalar per block.

### Where the cut goes

At the midpoint, floored at the packing width:

```rust
let low_len = (n_vars / 2).max(P::LOG_WIDTH);
(n_vars > low_len).then_some(low_len)
```

The floor is what makes a block a whole number of packed words, which is what lets the outer product address a block as a run of the store.
The `None` is the degenerate case: a point that fits in one block has no high half to multiply against, so it takes the doubling rounds.

There is deliberately no tuned constant here.
An earlier revision placed the cut where one block stops fitting the second cache level, but that value is a property of the machine that measured it rather than of the problem, so it was wrong by construction anywhere else.
The midpoint needs no calibration, and it lands on the same cut the constant computed at the largest size this is called at — see the benchmark.

Appending coordinates to an *existing* expansion keeps the rounds — there is nothing to cut there.

### Soundness

No protocol change: this is a different way to compute the same coefficients.

- A property test runs the split against the doubling rounds over every cut shape, for both bases, with a random scale.
- An end-to-end test runs the public entry point against the scalar engine, which shares no code with either packed path, at the shortest point that splits at all, at the sizes where the packing-width floor still overrides the midpoint, and at sizes that split many ways.
- The pointer pattern that reads the first block while writing the spare blocks was checked under Miri in isolation, under both Stacked Borrows and Tree Borrows.

### Scope

- **changed:** the entry point every full expansion funnels through, so both bases and all four public forms get it
- **unchanged:** appending to an existing expansion
- **bench:** two more sizes on the existing group

### Verification

- `cargo test -p binius-math` (debug, so the `debug_assert`s run) — 178 passed, 0 failed
- `cargo test --workspace` (debug) — clean
- `cargo +nightly-2026-07-01 fmt --all --check`, `typos`, `check_copyright_notice.py` — clean
- `examples/sha256 prove --message-len 1000000 --pcs whir` — proof verifies

### Benchmark

A/B on one tree, swapping only how the cut is chosen so the two paths compile identically otherwise.
128-bit scalars, 4-lane packing, 32 cores, `-C target-cpu=native`.

**Against the doubling rounds**, at the sizes that used to take them:

| variables | rounds | split | delta |
| ---: | ---: | ---: | ---: |
| 3 | 26.6 ns | 32.6 ns | +6 ns |
| 4 | 33.9 ns | 38.5 ns | +5 ns |
| 5 | 44.8 ns | 58.9 ns | +14 ns |
| 6 | 61.7 ns | 78.9 ns | +17 ns |
| 8 | 202 ns | 243 ns | +41 ns |
| 10 | 631 ns | 720 ns | +89 ns |
| 12 | 2.31 us | 2.47 us | +0.16 us |

The split does strictly more work below the point where the traffic saving repays it — a second sub-expansion and a rescale pass — and that costs at most 0.16 us, on a result that comfortably fits L2.

Between the two *split* rules, the midpoint is the faster of the two where they differ and it is measurable: 9.42 us against 11.28 us at 14 variables, 37.1 us against 38.6 us at 16.

**Against the doubling rounds**, at the larger sizes, measured with the cut held at 12:

| coefficients | rounds | split | speedup |
|---|---|---|---|
| 2^15 | 0.0180 ms | 0.0203 ms | 0.89x |
| 2^16 | 0.0360 ms | 0.0383 ms | 0.94x |
| 2^17 | 0.0725 ms | 0.0741 ms | 0.98x |
| 2^18 | 0.1450 ms | 0.0885 ms | **1.64x** |
| 2^20 | 0.4102 ms | 0.2243 ms | **1.83x** |
| 2^22 | 7.2661 ms | 4.3999 ms | **1.65x** |
| 2^24 | 31.072 ms | 22.346 ms | **1.39x** |

The three losing rows are reached now, where an earlier revision skipped them.
They are also the rows where the midpoint picks a smaller cut than 12 and measures slightly faster than this table's split column, so the real losses there are a little smaller than shown — 0.89x is the floor, not the value.

At 24 variables the midpoint picks `24/2 = 12`, which is exactly the cut the deleted cache constant computed.
So for every size this primitive is actually called at, a block is at most 64 KB either way — the cache property the constant was chasing holds without anyone naming it.

That row also serves as a noise control: it is the same code under both rules, and it moved 20.2 / 21.2 / 23.8 ms across three runs on this box.
An 18% spread on an identical path is why no delta is claimed between the two *split* rules above 2^18.

End to end on `sha256 --message-len 1000000 --pcs whir`, three runs each:

| span | rounds | split |
|---|---|---|
| Expand tensor queries | 12.1 / 12.0 / 11.8 ms | 8.27 / 7.90 / 8.08 ms |
| Prove (total) | 450 / 444 / 444 ms | 445 / 462 / 446 ms |

The span it dominates improves 1.48x and repeatably.
**Total prove time does not move measurably on this workload** — 4 ms is inside the run-to-run noise of a 445 ms prove.
The win is worth taking because the primitive has references from every prover crate and the saving grows with the expansion, not because this particular proof gets faster.

### DIFF
--- a/crates/math/src/multilinear/hypercube.rs
+++ b/crates/math/src/multilinear/hypercube.rs
@@ -361,24 +361,14 @@ pub fn scaled_eq_ind_partial_eval_into<Cube: Hypercube, P: PackedField, Data: Ve
 	}
 }
 
-/// Bytes one block of a split expansion spans.
-///
-/// # Why this value
-///
-/// A block is read once per block of the result and written once in total.
-/// Sizing it for the second cache level keeps those repeated reads off memory, while leaving room
-/// beside them for the writes they feed.
-///
-/// Halving it measurably loses at the largest sizes, and quadrupling it gains nothing.
-const BLOCK_BYTES: usize = 1 << 16;
-
 /// Coordinates the low half of a split takes, or nothing when the point is too short to split.
 ///
-/// A block spans whole packed words, so it never falls below one.
+/// The cut sits at the midpoint, so neither expansion is much larger than the square root of the
+/// result and the two cost the same to build.
+///
+/// A block spans whole packed words, so the cut never falls below the packing width.
 fn split_low_len<P: PackedField>(n_vars: usize) -> Option<usize> {
-	let log_scalar_bytes = size_of::<P::Scalar>().next_power_of_two().ilog2();
-	let low_len = BLOCK_BYTES.ilog2().saturating_sub(log_scalar_bytes) as usize;
-	let low_len = low_len.max(P::LOG_WIDTH);
+	let low_len = (n_vars / 2).max(P::LOG_WIDTH);
 
 	// A point no longer than one block leaves no high half to take the outer product against.
 	(n_vars > low_len).then_some(low_len)
@@ -884,8 +874,8 @@ mod tests {
 			// Property: cutting the point and multiplying the two expansions together is the same
 			// map as appending its coordinates one at a time.
 			//
-			// The production cut is far above these sizes, so it is passed in directly here.
-			// That covers every cut shape, including the ones a tuning change could start picking.
+			// The cut is passed in directly rather than taken from the midpoint rule, so this
+			// covers every cut shape, not only the one that rule picks.
 			//
 			//     n_vars = 5, low_len = 2  ->  4 blocks of 4 coefficients
 			prop_assume!(low_len >= P::LOG_WIDTH);
@@ -920,9 +910,11 @@ mod tests {
 		// The public entry point picks the cut itself, so this runs at sizes that reach it,
 		// against the scalar engine, which shares no code with either packed path.
 		//
-		// 13 is one coordinate past the cut, where the result spans two blocks and the parallel
-		// loop has a single item. 19 is far enough past it to split many ways.
-		for n_vars in [13, 14, 18, 19] {
+		// 3 is the shortest point that splits at all, where the result spans two blocks and the
+		// parallel loop has a single item.
+		// 4 and 5 are where the packing-width floor still overrides the midpoint.
+		// 6 is where the midpoint takes over, and 13, 14 and 19 split many ways, odd and even.
+		for n_vars in [3, 4, 5, 6, 13, 14, 19] {
 			assert!(split_low_len::<P>(n_vars).is_some(), "expected the split at {n_vars} vars");
 
 			let point = random_scalars::<F>(&mut rng, n_vars);
@@ -939,7 +931,7 @@ mod tests {
 	}
 
 	#[test]
-	fn the_crossover_leaves_a_whole_block_below_the_cut() {
+	fn the_cut_leaves_a_whole_block_below_it() {
 		// Invariant: a block spans whole packed words, and the result spans whole blocks.
 		//
 		// Both are what let the outer product address a block as a run of the store.
--- a/crates/math/src/multilinear/hypercube.rs
+++ b/crates/math/src/multilinear/hypercube.rs
@@ -423,6 +423,12 @@ fn split_expand<Cube: Hypercube, P: PackedField, Data: VecLike<P>>(
 	point: &[P::Scalar],
 	low_len: usize,
 ) -> FieldBuffer<P, Data> {
+	// precondition
+	debug_assert!(P::LOG_WIDTH <= low_len && low_len < point.len());
+
+	// A block is `2^low_len` scalars, which is a whole number of packed words exactly because
+	// the cut sits at or above the packing width. That is what makes a block's position in the
+	// store equal its index into the high expansion.
 	let (low_coords, high_coords) = point.split_at(low_len);
 
 	// The low expansion is built straight into the store's first block, which is where the result
--- a/crates/math/src/multilinear/hypercube.rs
+++ b/crates/math/src/multilinear/hypercube.rs
@@ -372,23 +372,6 @@ pub fn scaled_eq_ind_partial_eval_into<Cube: Hypercube, P: PackedField, Data: Ve
 /// Halving it measurably loses at the largest sizes, and quadrupling it gains nothing.
 const BLOCK_BYTES: usize = 1 << 16;
 
-/// Blocks the result must span before the split earns its keep, as a base-2 logarithm.
-///
-/// # Why this value
-///
-/// The split trades the rounds' repeated rewrites for two sub-expansions and one extra pass.
-/// That trade only pays once the result stops fitting in the last cache level, since below it
-/// the traffic the split saves was never memory traffic.
-///
-/// Measured over 128-bit scalars, the two paths cross where the result reaches a few megabytes,
-/// which this puts at 64 blocks:
-///
-/// ```text
-///     coefficients   2^15   2^16   2^17   2^18   2^20   2^22   2^24
-///     split / rounds 0.89x  0.94x  0.98x  1.64x  1.83x  1.65x  1.39x
-/// ```
-const LOG_MIN_BLOCKS: usize = 6;
-
 /// Coordinates the low half of a split takes, or nothing when the point is too short to split.
 ///
 /// A block spans whole packed words, so it never falls below one.
@@ -397,7 +380,8 @@ fn split_low_len<P: PackedField>(n_vars: usize) -> Option<usize> {
 	let low_len = BLOCK_BYTES.ilog2().saturating_sub(log_scalar_bytes) as usize;
 	let low_len = low_len.max(P::LOG_WIDTH);
 
-	(n_vars >= low_len + LOG_MIN_BLOCKS).then_some(low_len)
+	// A point no longer than one block leaves no high half to take the outer product against.
+	(n_vars > low_len).then_some(low_len)
 }
 
 /// Expands a point by cutting it in two and taking the outer product of the two expansions.
@@ -426,9 +410,9 @@ fn split_expand<Cube: Hypercube, P: PackedField, Data: VecLike<P>>(
 	// precondition
 	debug_assert!(P::LOG_WIDTH <= low_len && low_len < point.len());
 
-	// A block is `2^low_len` scalars, which is a whole number of packed words exactly because
-	// the cut sits at or above the packing width. That is what makes a block's position in the
-	// store equal its index into the high expansion.
+	// A block is `2^low_len` scalars, a whole number of packed words exactly because the cut sits
+	// at or above the packing width.
+	// That is what makes a block's position in the store equal its index into the high expansion.
 	let (low_coords, high_coords) = point.split_at(low_len);
 
 	// The low expansion is built straight into the store's first block, which is where the result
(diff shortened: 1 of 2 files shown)


crates/math/src/multilinear/hypercube.rs
```diff
@@ -353,8 +353,121 @@ pub fn scaled_eq_ind_partial_eval_into<Cube: Hypercube, P: PackedField, Data: Ve
...
+/// # Preconditions
+///
+/// * the seed holds one coefficient, and its store has room for the whole expansion
+/// * the cut leaves at least one packed word below it, and at least one block above it
+fn split_expand<Cube: Hypercube, P: PackedField, Data: VecLike<P>>(
+	seed: FieldBuffer<P, Data>,
+	point: &[P::Scalar],
+	low_len: usize,
+) -> FieldBuffer<P, Data> {
+	let (low_coords, high_coords) = point.split_at(low_len);
```
benji-potamus, line 371: Non-blocking: the two documented preconditions aren't enforced, and CONTRIBUTING asks for a `debug_assert!` next to the doc contract. The lower one is also the invariant that makes the whole block layout work, and it's the one a future tuning change could break silently — worth stating where the cut is made rather than leaving it to the reader:

```suggestion
	// precondition
	debug_assert!(P::LOG_WIDTH <= low_len && low_len < point.len());

	// A block is `2^low_len` scalars, which is a whole number of packed words exactly because
	// the cut sits at or above the packing width. That is what makes a block's position in the
	// store equal its index into the high expansion.
	let (low_coords, high_coords) = point.split_at(low_len);
```

crates/math/src/multilinear/hypercube.rs
```diff
@@ -353,8 +353,121 @@ pub fn scaled_eq_ind_partial_eval_into<Cube: Hypercube, P: PackedField, Data: Ve
...
+///     split / rounds 0.89x  0.94x  0.98x  1.64x  1.83x  1.65x  1.39x
+/// ```
+const LOG_MIN_BLOCKS: usize = 6;
+
+/// Coordinates the low half of a split takes, or nothing when the point is too short to split.
+///
+/// A block spans whole packed words, so it never falls below one.
+fn split_low_len<P: PackedField>(n_vars: usize) -> Option<usize> {
+	let log_scalar_bytes = size_of::<P::Scalar>().next_power_of_two().ilog2() as usize;
+	let low_len = BLOCK_BYTES.ilog2().saturating_sub(log_scalar_bytes as u32) as usize;
```
benji-potamus: Non-blocking: `ilog2` already returns `u32`, so the round trip through `usize` and back cancels out.

```suggestion
	let log_scalar_bytes = size_of::<P::Scalar>().next_power_of_two().ilog2();
	let low_len = BLOCK_BYTES.ilog2().saturating_sub(log_scalar_bytes) as usize;
```

benji-potamus (approved):
Approve.

Good decomposition. Writing each coefficient once against a cache-resident block is the right shape for this primitive, and keeping the doubling rounds below the crossover means nothing regresses at the sizes that lose. The aliasing in `split_expand` is sound and mirrors the pattern already in `tensor_prod_eq_ind_reserved` — the first block's pointer is taken before the spare borrow, the two ranges are disjoint, and the parallel fill has completed before block 0 takes its own coefficient. Hardcoding the tuning constants is fine; it's what `task_size` already does, and both carry their reasoning.

Two non-blocking suggestions inline.

crates/math/src/multilinear/hypercube.rs
```diff
@@ -353,8 +353,121 @@ pub fn scaled_eq_ind_partial_eval_into<Cube: Hypercube, P: PackedField, Data: Ve
...
+/// # Why this value
+///
+/// A block is read once per block of the result and written once in total.
+/// Sizing it for the second cache level keeps those repeated reads off memory, while leaving room
+/// beside them for the writes they feed.
+///
+/// Halving it measurably loses at the largest sizes, and quadrupling it gains nothing.
+const BLOCK_BYTES: usize = 1 << 16;
+
+/// Blocks the result must span before the split earns its keep, as a base-2 logarithm.
```
jimpo: Do we need this conditional? At the sizes below this cutoff point, it's just a small vector and the operation is very fast already. Just always do the split algorithm

tcoratger: Agreed — dropped in 1015921, and the threshold constant with it.

I measured the range it would newly cover before removing it, since I had only benchmarked from `2^15` up. 128-bit scalars, 4-lane packing, `-C target-cpu=native`, A/B on one tree with only the guard flipped:

| coefficients | rounds | split | delta |
| ---: | ---: | ---: | ---: |
| 2^13 | 6.5 us | 7.0 us | +0.5 us |
| 2^14 | 10.4 us | 11.7 us | +1.3 us |
| 2^15 | 19.5 us | 21.1 us | +1.6 us |
| 2^16 | 38.7 us | 40.0 us | +1.3 us |
| 2^17 | 78.4 us | 77.4 us | -1.0 us |
| 2^18 | 169 us | 88 us | -81 us |

Worst case is 1.6 microseconds. You are right that it does not buy anything, and the constant was worse than useless: I had set it from where the result stops fitting *this* machine's last cache level, so it was wrong by construction on any other one.

What remains is a structural guard rather than a tuning one — a point no longer than one block has no high half to take an outer product against:

```rust
// A point no longer than one block leaves no high half to take the outer product against.
(n_vars > low_len).then_some(low_len)
```

The end-to-end test now covers `2^13` and `2^14` too, so the smallest split — two blocks, one item in the parallel loop — is pinned against the scalar engine rather than merely reachable.

One note on the measurements: three other agents were compiling on this box, so the absolute numbers run 10-40% high against my earlier ones. The ratios and the sign are unaffected, and both arms were measured back to back under the same load.

crates/math/src/multilinear/hypercube.rs
```diff
@@ -353,8 +353,111 @@ pub fn scaled_eq_ind_partial_eval_into<Cube: Hypercube, P: PackedField, Data: Ve
...
+/// Coordinates the low half of a split takes, or nothing when the point is too short to split.
+///
+/// A block spans whole packed words, so it never falls below one.
+fn split_low_len<P: PackedField>(n_vars: usize) -> Option<usize> {
+	let log_scalar_bytes = size_of::<P::Scalar>().next_power_of_two().ilog2();
+	let low_len = BLOCK_BYTES.ilog2().saturating_sub(log_scalar_bytes) as usize;
+	let low_len = low_len.max(P::LOG_WIDTH);
+
+	// A point no longer than one block leaves no high half to take the outer product against.
+	(n_vars > low_len).then_some(low_len)
```
benji-potamus: The last commit dropped `LOG_MIN_BLOCKS`, so with 128-bit scalars this splits from 13 variables up. The benchmark in the PR body measures that region as a loss — 0.89x at 2^15, 0.94x at 2^16, 0.98x at 2^17 — and 2^13/2^14 sit below everything measured, where the split does strictly more work (a second sub-expansion, plus a full rescale pass) over a result that comfortably fits L2.

Please restore the crossover. The reason given for deleting it — a constant whose right value depends on the machine — applies just as much to `BLOCK_BYTES` above, which is a cache-size guess that stays either way. As it stands the change buys no constant back and pays a measured regression for it.

If you would rather keep the always-split shape, it needs numbers at 2^13 and 2^14 showing the loss there is small too.

benji-potamus (changes requested):
Request changes.

Good piece of work. The outer-product decomposition is the right idea, building the low expansion straight into the store's first block avoids any extra full-size allocation, and the unsafe follows the pattern `tensor_prod_eq_ind_reserved` already establishes a few lines up. The tests are well chosen: the property test covers every cut shape, and the end-to-end one checks the public entry point against the scalar engine, which shares no code with either packed path.

To answer your question on the thread: worth keeping, not worth closing. 1.65x at 2^22 on a primitive every prover crate reaches is a real win, even where one workload's total prove time does not move.

One item has no line to sit on: the description is now out of date with the code. It still describes a crossover at 2^18, says the doubling rounds stay below it, and says the three losing rows are never reached — the last commit made all three false. Please bring it back in line with whatever the code ends up doing, along with the change requested inline.

crates/math/src/multilinear/hypercube.rs
benji-potamus: Retracting this — I missed that @jimpo asked for exactly this removal in the thread above, and that you already posted the 2^13/2^14 measurements I was asking for. Worst case is 1.6 us; that settles it. Sorry for the noise. Approving separately.

benji-potamus (approved):
Approve — this supersedes my previous review, whose one finding I have retracted inline. It asked for the crossover back without noticing that @jimpo had already asked for its removal and that you had already measured 2^13 and 2^14. Worst case 1.6 us settles it, and the guard that remains is structural rather than tuned.

On the change itself: the outer-product decomposition is the right shape for this primitive, building the low expansion straight into the store's first block avoids any extra full-size allocation, and the aliasing in `split_expand` mirrors the pattern `tensor_prod_eq_ind_reserved` already establishes — first block's pointer taken before the spare borrow, disjoint ranges, parallel fill complete before block 0 takes its own coefficient. The tests are well chosen: the property test covers every cut shape, and the end-to-end one pins the smallest split against the scalar engine, which shares no code with either packed path.

One thing with no line to sit on: the description still describes the 2^18 crossover, says the doubling rounds stay below it, and says the three losing rows are never reached. The last commit made all three false, and the corrected table now lives only in a review thread. Worth folding into the body before merge.

jimpo (changes requested):
Get rid of the BLOCK_BYTES threshold. Just split the point in half, do the iterative tensor expansion on the two halfs, then take the outer product into a new buffer. You can set the halfway point to at least P::LOG_WIDTH and early return if the tail is empty.

jimpo (approved):

PR #2408, opened 2026-08-31
BINIUS-585: Give the packed NTT engine all three of its transforms

tcoratger:
Closes [BINIUS-585](https://linear.app/jimpo/issue/BINIUS-585).

The additive NTT has three linear maps and only one of them was fast.
This makes the existing packed engine serve all three, which is a **17x** speedup on the two that were not, and fills in an inverse that used to panic.

### The problem

| | reference | breadth-first | single thread | multi thread |
|---|---|---|---|---|
| forward | slow | fast | fast | fast |
| inverse | slow | `todo!()` | `unimplemented!()` | `unimplemented!()` |
| transpose | slow | slow | slow | slow |

Three of four implementations panic on the inverse.
All four inherit a trait default for the adjoint that walks the network one element at a time through indexed reads and writes.

The trait's own doc said so:

> The default body mirrors the reference forward transform and is correspondingly slow.
> An implementation with a fast forward transform should override it.

None did.

And the adjoint is not a corner. The WHIR prover runs it once per level, to turn a level's opened rows into the weight they induce on the message. It is essentially the whole of that step.

### The idea

All three are the **same** network over the **same** twiddles, with the same butterfly count.
They differ in exactly two things:

```
    forward     layers ascending     u += v * t;  v += u
    inverse     layers descending    v += u;      u += v * t
    transpose   layers descending    u += v;      v += u * t
```

So the engine does not need three copies. It needs one parameter.

```rust
trait Butterfly {
    const ASCENDING: bool;
    fn apply<P: PackedField>(u: &mut P, v: &mut P, twiddle: P);
    fn apply_zero<P: PackedField>(u: &mut P, v: &mut P);
}
```

`ASCENDING` is a `const`, so each map compiles to a single arm — no branch survives into the inner loop.

### Where the order enters

Four places, and nowhere else:

- the depth-first recursion applies a level's own layer **before** recursing, or **after**
- the breadth-first base case runs whole-word layers then in-word layers, or the reverse
- the dispatcher that fuses layer pairs walks its window from one end or the other
- the multithreaded driver runs its shared phase before the independent one, or after

Everything else — the twiddle indexing, the block geometry, the zero-block peeling, the plane fusion, the task splitting — is shared unchanged.

### The change

```
- fn forward_depth_first<P: PackedField>(..)         + fn depth_first<B: Butterfly, P: PackedField>(..)
- fn forward_breadth_first<P: PackedField>(..)       + fn breadth_first<B: Butterfly, P: PackedField>(..)
- fn forward_shared_layer<P: PackedField>(..)        + fn shared_layer<B: Butterfly, P: PackedField>(..)
- fn forward_shared_layer_pair<P: PackedField>(..)   + fn shared_layer_pair<B: Butterfly, P: PackedField>(..)
- fn forward_shared_layers<P: PackedField>(..)       + fn shared_layers<B: Butterfly, P: PackedField>(..)
```

Each of the three engines then gets one private `run::<B, P>` and three one-line trait methods.

The breadth-first base case is also split into its two layer shapes, so the two orders become a four-line `if` rather than a duplicated body.

### Soundness

No protocol change, and the forward transform runs the code that was already there.

Two properties that name no implementation now cover the two new maps:

- **round trip** — forward then inverse is the identity, at every split of the layer range
- **adjoint identity** — `<transpose(a), m> = <a, forward(m)>`, at every split

Both run on the multithreaded engine. On top of that, all three maps are checked against the scalar reference on all three packed engines, over a sweep and a proptest of `(log_d, skip_early, skip_late, log_num_shares)`.

I checked the tests actually bite, by breaking the code two ways:

| mutation | tests that failed |
|---|---|
| transpose butterfly written as the inverse's | adjoint identity, reference match, proptest |
| descending base case run ascending | adjoint identity, round trip, reference match, proptest |

### Scope

- **removed:** the trait's default adjoint, so an implementation must now make a choice rather than inherit a slow one
- **moved:** that body into the scalar reference, which is what the fast engines are tested against
- no API change beyond the two methods that used to panic now working
- the benchmark used for the numbers below is **not** in the diff — it is reproduced at the end of this description for anyone who wants to rerun it

### Verification

- `cargo test -p binius-math --release` — 177 passed, 0 failed
- `cargo +nightly-2026-07-01 fmt --all --check`, `typos` — clean
- `examples/sha256 prove --message-len 1000000 --pcs whir` — proof verifies

### Benchmark

128-bit scalars, 4-lane packing, 32 cores, `-C target-cpu=native`.
The benchmark source is at the end of this description; it is deliberately not part of the diff.

**19 layers, one skipped early** — the shape WHIR's induced weight runs:

| map | reference | multithread | speedup |
|---|---|---|---|
| forward | 10.76 ms | 0.587 ms | 18.3x |
| inverse | 10.69 ms | 0.605 ms | **17.7x** |
| transpose | 10.79 ms | 0.631 ms | **17.1x** |

**23 layers, rate 1, 16 lanes** — the commit-path encode shape:

| map | reference | multithread | speedup |
|---|---|---|---|
| forward | 205.6 ms | 12.91 ms | 15.9x |
| inverse | 195.9 ms | 11.03 ms | **17.8x** |
| transpose | 193.8 ms | 11.23 ms | **17.3x** |

All three now cost the same, which is the point: they are the same network.

End to end on `sha256 --message-len 1000000 --pcs whir`, three runs each, A/B on one tree with only the multithreaded adjoint flipped back to the reference:

| span | reference adjoint | packed adjoint |
|---|---|---|
| Glue induced weight, level 0 | 12.8 / 15.3 / 12.8 ms | 3.16 / 3.12 / 3.58 ms |
| WHIR level 0 | 41.1 / 47.7 / 38.9 ms | 31.2 / 30.3 / 30.3 ms |
| Prove (total) | 452 / 629 / 462 ms | 449 / 434 / 440 ms |

Two honest notes on that table.

The level-0 span drops 4x rather than 17x because it is not all transform. Measured separately, everything the encoder's adjoint does around the transform -- summing the repeats, the bit reversal, its own zeroed buffer -- comes to about 0.6 ms, and the fold in there is a contiguous streaming add already running at memory bandwidth (I tried splitting it across workers; it got slower). The rest of the remaining span sits above that call, in building the sparse codeword weight: an 8 MiB zeroed buffer taken fresh from the global allocator once per level. That is where the next bite is, not in the transform.

Level 1's span does not move, because at its shape the caller picks the row-by-row route instead. That choice is calibrated against a constant measured from the *old* adjoint — a "layer entry at 1.6 ns". That number is now roughly 17x smaller, so the crossover wants re-deriving, which is a follow-up on this.

<details>
<summary>The benchmark used above, for anyone who wants to rerun it</summary>

Drop this at `crates/math/benches/ntt_transforms.rs`, add the matching `[[bench]]` entry to
`crates/math/Cargo.toml`, and run `cargo bench -p binius-math --bench ntt_transforms`.

```rust

use binius_field::{Ghash128b, PackedBinaryGhash4x128b};
use binius_math::{
	ntt::{
		AdditiveNTT, NeighborsLastMultiThread, NeighborsLastReference,
		domain_context::GaoMateerPreExpanded,
	},
	test_utils::random_field_buffer,
};

type P = PackedBinaryGhash4x128b;
type F = Ghash128b;

/// The three maps the butterfly network computes, over the same network and the same twiddles.
///
/// The scalar reference is the baseline: it walks the network one element at a time, which is what
/// every map but the forward one used to cost.
fn bench_transforms(c: &mut Criterion) {
	let mut group = c.benchmark_group("ntt_transforms");
	group.sample_size(10);

	// Two shapes the prover actually runs:
	//
	//     19 layers, one skipped early    the weight a WHIR level's opened rows induce
	//     23 layers, rate 1, 16 lanes     the commit-path encode
	for (log_d, skip_early, skip_late) in [(19usize, 1usize, 0usize), (23, 1, 4)] {
		let domain_context = GaoMateerPreExpanded::<F>::generate(log_d);
		let reference = NeighborsLastReference {
			domain_context: &domain_context,
		};
		let multi_thread = NeighborsLastMultiThread::new(&domain_context, 5);

		// Throughput counts butterflies, so every row of the group is directly comparable.
		let layers = (log_d - skip_early - skip_late) as u64;
		group.throughput(Throughput::Elements(layers << (log_d - 1)));

		let parameter = format!("log_d={log_d}/skip_early={skip_early}/skip_late={skip_late}");
		let mut data = random_field_buffer::<P>(&mut rng, log_d);

		macro_rules! bench_map {
			($engine:expr, $name:literal, $map:ident) => {
				group.bench_function(BenchmarkId::new($name, &parameter), |b| {
					b.iter(|| $engine.$map(data.as_mut_view(), skip_early, skip_late));
				});
			};
		}

		bench_map!(reference, "reference/forward", forward_transform);
		bench_map!(reference, "reference/inverse", inverse_transform);
		bench_map!(reference, "reference/transpose", transpose_transform);
		bench_map!(multi_thread, "multithread/forward", forward_transform);
		bench_map!(multi_thread, "multithread/inverse", inverse_transform);
		bench_map!(multi_thread, "multithread/transpose", transpose_transform);
	}

	group.finish();
}

criterion_group!(benches, bench_transforms);
criterion_main!(benches);
```

</details>

### DIFF
--- a/crates/math/src/ntt/neighbors_last.rs
+++ b/crates/math/src/ntt/neighbors_last.rs
@@ -25,6 +25,151 @@ use crate::field_buffer::FieldSliceMut;
 // Empirically it performs well and is small enough for the buffer to fit comfortably in L1 cache.
 const DEFAULT_LOG_BASE_LEN: usize = 10;
 
+/// One of the three linear maps the butterfly network computes.
+///
+/// All three run the same network over the same twiddles, and cost the same.
+/// They differ in the order the layers run and in what one butterfly does:
+///
+/// ```text
+///     forward     layers ascending     u += v * t;  v += u
+///     inverse     layers descending    v += u;      u += v * t
+///     transpose   layers descending    u += v;      v += u * t
+/// ```
+///
+/// Everything below is written once against this, so all three get the same engine.
+trait Butterfly {
+	/// Whether the layers run from the earliest to the latest.
+	const ASCENDING: bool;
+
+	/// Applies one butterfly under a twiddle.
+	fn apply<P: PackedField>(u: &mut P, v: &mut P, twiddle: P);
+
+	/// Applies one butterfly where the twiddle is zero.
+	///
+	/// One of the two values then keeps what it held, so its cache lines stay clean.
+	fn apply_zero<P: PackedField>(u: &mut P, v: &mut P);
+
+	/// Runs the same map through the scalar reference implementation.
+	fn reference<P: PackedField>(
+		domain_context: &impl DomainContext<Field = P::Scalar>,
+		data: FieldSliceMut<'_, P>,
+		skip_early: usize,
+		skip_late: usize,
+	);
+}
+
+/// The transformation the additive NTT is named for.
+struct Forward;
+
+impl Butterfly for Forward {
+	const ASCENDING: bool = true;
+
+	#[inline(always)]
+	fn apply<P: PackedField>(u: &mut P, v: &mut P, twiddle: P) {
+		*u += *v * twiddle;
+		*v += *u;
+	}
+
+	#[inline(always)]
+	fn apply_zero<P: PackedField>(u: &mut P, v: &mut P) {
+		*v += *u;
+	}
+
+	fn reference<P: PackedField>(
+		domain_context: &impl DomainContext<Field = P::Scalar>,
+		data: FieldSliceMut<'_, P>,
+		skip_early: usize,
+		skip_late: usize,
+	) {
+		NeighborsLastReference { domain_context }.forward_transform(data, skip_early, skip_late);
+	}
+}
+
+/// The inverse of the forward transformation.
+struct Inverse;
+
+impl Butterfly for Inverse {
+	const ASCENDING: bool = false;
+
+	#[inline(always)]
+	fn apply<P: PackedField>(u: &mut P, v: &mut P, twiddle: P) {
+		*v += *u;
+		*u += *v * twiddle;
+	}
+
+	#[inline(always)]
+	fn apply_zero<P: PackedField>(u: &mut P, v: &mut P) {
+		*v += *u;
+	}
+
+	fn reference<P: PackedField>(
+		domain_context: &impl DomainContext<Field = P::Scalar>,
+		data: FieldSliceMut<'_, P>,
+		skip_early: usize,
+		skip_late: usize,
+	) {
+		NeighborsLastReference { domain_context }.inverse_transform(data, skip_early, skip_late);
+	}
+}
+
+/// The adjoint of the forward transformation.
+///
+/// The forward butterfly has determinant one, so its inverse and its transpose agree only where
+/// the twiddle vanishes.
+struct Transpose;
+
+impl Butterfly for Transpose {
+	const ASCENDING: bool = false;
+
+	#[inline(always)]
+	fn apply<P: PackedField>(u: &mut P, v: &mut P, twiddle: P) {
+		*u += *v;
+		*v += *u * twiddle;
+	}
+
+	#[inline(always)]
+	fn apply_zero<P: PackedField>(u: &mut P, v: &mut P) {
+		*u += *v;
+	}
+
+	fn reference<P: PackedField>(
+		domain_context: &impl DomainContext<Field = P::Scalar>,
+		data: FieldSliceMut<'_, P>,
+		skip_early: usize,
+		skip_late: usize,
+	) {
+		NeighborsLastReference { domain_context }.transpose_transform(data, skip_early, skip_late);
+	}
+}
+
+/// Applies one layer to the block a recursion level owns.
+///
+/// ## Preconditions
+///
+/// - `2 * block_size_half == data.len()`
+#[inline]
+fn one_block<B: Butterfly, P: PackedField>(
+	domain_context: &impl DomainContext<Field = P::Scalar>,
+	data: &mut [P],
+	block_size_half: usize,
+	layer: usize,
+	block: usize,
+) {
+	let (block0, block1) = data.split_at_mut(block_size_half);
+	if block == 0 {
+		// `domain_context.twiddle(layer, 0)` is always zero (see `DomainContext::twiddle`).
+		// So the butterfly collapses to a single addition.
+		for (u, v) in iter::zip(block0, block1) {
+			B::apply_zero(u, v);
+		}
+	} else {
+		let packed_twiddle = P::broadcast(domain_context.twiddle(layer, block));
+		for (u, v) in iter::zip(block0, block1) {
+			B::apply(u, v, packed_twiddle);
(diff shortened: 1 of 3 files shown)


crates/math/src/ntt/neighbors_last.rs
```diff
@@ -330,26 +511,40 @@ fn fused_pair<P: PackedField>(
...
 
-/// Same as [`fused_pair`], for the super-block whose index is zero.
+/// Same as the fused pair above, for the super-block whose index is zero.
 ///
 /// There `twiddle_0` and `twiddle_1_even` are both the layer's block-0 twiddle, which is zero.
-/// Each of their butterflies collapses to `v += u`, leaving `u` untouched.
-/// So plane 0 is never written, and its cache lines stay clean.
-fn fused_pair_zero_block<P: PackedField>(planes: [&mut [P]; 4], twiddle_1_odd: P) {
+/// Each of their butterflies collapses to a single addition, leaving one plane untouched.
+/// So that plane is never written, and its cache lines stay clean.
```
benji-potamus, line 532: This no longer holds for all three maps. `Transpose::apply_zero` is `*u += *v`, so under the descending order below every one of the four planes is written: `x_0` by the first and third calls, `x_1` by the fourth, `x_2` and `x_3` by the `apply`. The clean-plane saving is real for `Forward` and `Inverse` only, where `apply_zero` leaves `u` and so spares plane 0.

Worth stating, since the next reader optimising this will otherwise assume the saving is uniform.

```suggestion
/// Each of their butterflies collapses to a single addition.
/// For the forward and inverse maps that addition leaves `u`, so plane 0 is never written and its
/// cache lines stay clean. The transpose adds the other way and writes all four.
```

benji-potamus (changes requested):
Request changes.

Good change, and the hard part is right. Each of the three butterflies matches the forward matrix `[[1, t], [1, 1 + t]]` — the transpose is the genuine adjoint, not a mis-transcribed inverse — and the layer direction is handled everywhere it enters, including inside the fused pair itself, which the description's list of four does not mention. The `const ASCENDING` parameterisation earns its complexity: the alternative is three copies of a 500-line network. The tests are the strongest part, since the round trip and the adjoint identity pin the two new maps without naming an implementation, and all three maps are then checked bit for bit against the scalar reference on all three engines.

Three things:

- One inline comment below.
- Needs a rebase onto `main`: it conflicts with the `PackedBinaryGhash*` → `PackedGhash*` rename, in five test call sites and nothing else.
- The `1.6 ns` layer-entry constant in `adjoint_is_cheaper` (`crates/iop-prover/src/whir/induced_weight.rs`, no inline anchor from this diff) — agreed that re-deriving the crossover is a follow-up, not this PR's job. But that doc comment states the number as a current measurement, and this PR falsifies it. Please either add a line there saying it predates the fast adjoint, or link the follow-up issue from this PR, so the next reader does not trust a stale constant.

PR #2410, opened 2026-08-31
BINIUS-587: Build the barycentric weight from the subspace recurrence

tcoratger:
Closes [BINIUS-587](https://linear.app/jimpo/issue/BINIUS-587).

Same weight, computed from the subspace polynomial's recurrence instead of by multiplying out the domain — and computed in the domain's own field rather than the arithmetic one.

This started as a question about the `E::from` in that fold. Measured, the conversion itself costs nothing (see **On the conversion** below); the loop around it is the real problem.

### The problem

Interpolating over a binary subspace is cheap because every point of the domain shares **one** barycentric weight, where the textbook formula needs one per point. That weight is the inverse of the product of every nonzero point:

```
    w = (prod_{d != 0} d)^{-1}
```

It was computed exactly that way — one multiplication per point of the domain. So "one weight, whatever the domain size" was paid for by a loop that is itself linear in the domain size, landing on the same order as the interpolation it exists to make cheap. It measures at 30% of a `lagrange_evals` call and 37% of an `extrapolate`.

Two more things were wrong with it.

**It ran in the wrong field.** Every point was lifted into the arithmetic field `E` before being multiplied:

```rust
subspace.iter().skip(1).map(E::from).fold(E::one(), |acc, d| acc * d)
```

A recursion channel therefore pays a symbolic multiplication per point, plus a symbolic inversion, to produce a value that depends on nothing the channel knows.

**Its safety argument was wrong.** The unchecked inversion was justified by:

> the product runs over the subspace's non-zero elements — `skip(1)` drops index 0, which is the zero element — so it is non-zero by construction, whatever the caller passes

`skip(1)` drops the zero *index*. Under a linearly dependent basis some other index also maps to the zero point, the product vanishes, and `unsafe { invert() }` is called on zero. `BinarySubspace::new_unchecked` says in its own doc that it does not check independence, so "whatever the caller passes" was not true.

### The idea

The subspace polynomial of the span of the first `i` basis elements satisfies

```
    W_0(X)     = X
    W_{i+1}(X) = W_i(X) * (W_i(X) + W_i(b_i))
```

$W_i$ is $\mathbb{F}_2$-linear, so it is a sum of terms $X^{2^j}$. Squaring doubles every exponent, so $W_i(X)^2$ has **no linear term**. Expanding the recurrence, each step therefore multiplies the linear coefficient by a single number:

```
    prod_{d != 0} d = prod_i W_i(b_i)
```

and those numbers come from one triangular sweep over the basis.

### Per domain

| dimension | before | after |
|---|---:|---:|
| 6 | 63 multiplications | ~21 |
| 12 | 4095 multiplications | ~78 |

→ one multiplication per *point* becomes a square of the *dimension*.

### The change

```rust
- let product = subspace.iter().skip(1).map(E::from).fold(E::one(), |acc, d| acc * d);
- unsafe { product.invert() }

+ let mut evals = subspace.basis().to_vec();
+ let mut product = F::ONE;
+ for i in 0..evals.len() {
+     let normalizer = evals[i];
+     product *= normalizer;
+     for eval in &mut evals[i + 1..] {
+         *eval *= *eval + normalizer;
+     }
+ }
+ assert_ne!(product, F::ZERO, "precondition: the subspace basis must be independent");
+ E::from(product.invert_or_zero())
```

Everything runs in `F`; `E` sees exactly one embedding, of a finished constant.

### On the conversion

The original prompt for this was: is lifting into `E` before multiplying wasteful, since `E * F` is usually cheaper than `E * E`?

Measured, no — it costs exactly nothing, because every native caller has `E = F`, so `E::from` is the identity and `E * E` *is* `F * F`:

| dim | product built in `E` | product built in `F`, one lift at the end |
|---|---|---|
| 3 | 116 ns | 116 ns |
| 8 | 1.064 us | 1.063 us |
| 12 | 15.441 us | 15.421 us |

The instinct is right in general; there is just no subfield multiplication to reach for here. What the field choice *does* buy is the recursion channel, where `E * E` is constraint rows rather than nanoseconds.

### Soundness

The value is unchanged, pinned three ways:

- against the product it is defined as, at every dimension from 0 to 8
- against the textbook per-point Lagrange weights, which an existing test already compares to
- by checking the two field parameters land on the same weight

Breaking the recurrence (dropping the addition — a plausible slip) fails 5 tests including all three of these, so they bite.

The unchecked inversion becomes checked, with the precondition stated. Because the inversion moved into `F`, that check costs a recursion channel nothing — which is exactly what the old comment was trying to buy by skipping it.

### Scope

One function, one file. No API change, no signature change.

### Verification

- `cargo test -p binius-math` (debug) — 178 passed; and `--release` — 178 passed
- `cargo test -p binius-verifier -p binius-prover -p binius-m4-prover` — all pass
- `cargo +nightly-2026-07-01 fmt --all --check`, `typos` — clean

### Benchmark

A/B on one tree with only this function's body swapped. 128-bit scalars, `-C target-cpu=native`.

| dimension | operation | before | after | speedup |
|---|---|---:|---:|---|
| 6 | `lagrange_evals` | 0.907 us | 0.739 us | 1.23x |
| 6 | `extrapolate` | 0.727 us | 0.563 us | 1.29x |
| 7 | `lagrange_evals` | 1.722 us | 1.341 us | 1.28x |
| 7 | `extrapolate` | 1.379 us | 0.993 us | 1.39x |
| 10 | `lagrange_evals` | 13.058 us | 9.695 us | 1.35x |
| 12 | `lagrange_evals` | 51.959 us | 37.135 us | 1.40x |
| 12 | `extrapolate` | 41.990 us | 27.089 us | 1.55x |

**Read this honestly.** Every production domain is dimension 6 or 7, and the weight is built a handful of times per proof, so this is worth **microseconds** on a 440 ms prove — it will not show up in a proof timing at all. The reasons to take it are that the algorithm is right, the safety argument is now true, and a recursion channel stops spending constraints on a constant.

<details>
<summary>The benchmark used above</summary>

```rust
use std::time::Instant;

use binius_field::{Ghash128b, Random};
use binius_math::{BinarySubspace, univariate::EvaluationDomain};

type F = Ghash128b;

fn bench<T>(name: &str, iters: usize, mut f: impl FnMut() -> T) {
    for _ in 0..3 {
        std::hint::black_box(f());
    }
    let t = Instant::now();
    for _ in 0..iters {
        std::hint::black_box(f());
    }
    let e = t.elapsed().as_secs_f64() / iters as f64;
    println!("{name:<44} {:>10.3} us", e * 1e6);
}

fn main() {
    for dim in [6usize, 7, 10, 12] {
        let s = BinarySubspace::<F>::with_dim(dim);
        let z = F::random(&mut rng);
        let values: Vec<F> = (0..1 << dim).map(|_| F::random(&mut rng)).collect();
        bench(&format!("lagrange_evals dim={dim}"), 20000, || s.lagrange_evals(&z));
        bench(&format!("extrapolate    dim={dim}"), 20000, || s.extrapolate(&values, &z));
    }
}
```

</details>

### DIFF
--- a/crates/math/src/univariate.rs
+++ b/crates/math/src/univariate.rs
@@ -154,30 +154,71 @@ impl<F: BinaryField, Data: Deref<Target = [F]>> EvaluationDomain<F> for BinarySu
 /// One weight therefore serves every point:
 ///
 /// ```text
-///     w = (prod_{j >= 1} d_j)^{-1}
+///     w = (prod_{d != 0} d)^{-1}
 /// ```
+///
+/// # Algorithm
+///
+/// That product is the linear coefficient of the subspace polynomial, and the subspace polynomial
+/// has a recurrence:
+///
+/// ```text
+///     W_0(X)     = X
+///     W_{i+1}(X) = W_i(X) * (W_i(X) + W_i(b_i))
+/// ```
+///
+/// Squaring a linearized polynomial doubles every exponent, so it contributes no linear term.
+/// Each step therefore multiplies the linear coefficient by one number, leaving
+///
+/// ```text
+///     prod_{d != 0} d = prod_i W_i(b_i)
+/// ```
+///
+/// which costs a square of the dimension rather than one multiplication per point of the domain.
+///
+/// The weight depends on the subspace alone, so all of it runs in the domain's own field and
+/// crosses into the arithmetic field once.
+/// That is what allows the checked inversion below: one wrapper channel's element type offers the
+/// unchecked inverse alone.
+///
+/// # Panics
+///
+/// Panics if the basis is linearly dependent, which makes the product vanish.
 fn barycentric_weight<F, E, Data>(subspace: &BinarySubspace<F, Data>) -> E
 where
 	F: BinaryField,
 	E: FieldOps + From<F>,
 	Data: Deref<Target = [F]>,
 {
-	let product = subspace
-		.iter()
-		.skip(1)
-		.map(E::from)
-		.fold(E::one(), |acc, d| acc * d);
+	// Seed the recurrence at the polynomial `X`, whose values on the basis are the basis itself.
+	let mut evals = subspace.basis().to_vec();
 
-	// SAFETY: the product runs over the subspace's non-zero elements — `skip(1)` drops index 0,
-	// which is the zero element — so it is non-zero by construction, whatever the caller passes.
-	// Inverting without the zero case spares the wrapper channels a constraint that could never
-	// fire.
-	unsafe { product.invert() }
+	let mut product = F::ONE;
+	for i in 0..evals.len() {
+		// Entry `i` has reached the polynomial vanishing on everything below it, so it is the
+		// factor this step contributes.
+		let normalizer = evals[i];
+		product *= normalizer;
+
+		// Advance the entries still to come one polynomial along.
+		for eval in &mut evals[i + 1..] {
+			*eval *= *eval + normalizer;
+		}
+	}
+
+	// Invariant: each factor is a subspace polynomial evaluated off the subspace it vanishes on,
+	// which is nonzero exactly when the basis is independent. The subspace type leaves that to its
+	// caller, so it is checked here rather than assumed.
+	assert_ne!(product, F::ZERO, "precondition: the subspace basis must be independent");
+
+	E::from(product.invert_or_zero())
 }
 
 #[cfg(test)]
 mod tests {
-	use binius_field::{Field, Ghash128b, Random, util::powers};
+	use binius_field::{
+		Field, Ghash128b, Random, Rijndael8b, arithmetic_traits::InvertOrZero, util::powers,
+	};
 	use proptest::prelude::*;
 	use rand::prelude::*;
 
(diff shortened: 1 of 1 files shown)


crates/math/src/univariate.rs
```diff
@@ -154,30 +154,69 @@ impl<F: BinaryField, Data: Deref<Target = [F]>> EvaluationDomain<F> for BinarySu
...
+/// Each step therefore multiplies the linear coefficient by one number, leaving
+///
+/// ```text
+///     prod_{d != 0} d = prod_i W_i(b_i)
+/// ```
+///
+/// which costs a square of the dimension rather than one multiplication per point of the domain.
+///
+/// The weight depends on the subspace alone, so all of it runs in the domain's own field and
+/// crosses into the arithmetic field once. A wrapper channel then spends nothing on it.
```
benji-potamus: Non-blocking, but this claim is not true, and unlike the old SAFETY comment that made the same claim it now sits in the docs.

`CircuitElem::combine` takes the concrete path whenever none of its inputs is a `Wire`, and `From<F> for CircuitElem` yields `Constant`. Every domain point entered the old fold as a constant, so the per-point multiplications *and* the `unsafe { invert() }` were already constant-folded at build time — a wrapper channel spent zero constraints on the weight before this change too.

What running in `F` does buy is the checked inversion: `CircuitElem::invert_or_zero` is `unimplemented!`, so `assert_ne!` + `invert_or_zero` is only reachable because the arithmetic never leaves the domain field. That is the argument worth stating here.

```suggestion
/// The weight depends on the subspace alone, so all of it runs in the domain's own field and
/// crosses into the arithmetic field once. That is what allows the checked inversion below: a
/// wrapper channel's element type implements only the unchecked `invert`.
```

benji-potamus (approved):
Approve.

The recurrence is right and the sweep implements it faithfully: `L_{i+1} = W_i(b_i) * L_i` because `W_i^2` carries no linear term, `L_n` is the product over the nonzero points, and `evals[j]` holds `W_i(b_j)` at the top of each step. Dim 0 lands on the empty product correctly. The first test pins the value against the definition rather than re-deriving it, which is what makes the change safe to take.

The safety finding is real — `new_unchecked` documents that it does not check independence, so a dependent basis does put a zero past index 0 and the old `unsafe { invert() }` was reachable on zero. Trading that for a release-time `assert_ne!` is the right call; it is off any hot path and every in-tree caller builds from a fixed basis.

The perf framing in the description is honest and I read it the same way — this does not earn its keep on wall clock. It earns it by deleting an `unsafe` whose stated justification was false.

crates/math/src/univariate.rs
jimpo: Huh, even that doesn't make sense. It should call invert -- invert_or_zero is unimplemented, no?

tcoratger: You're right, and I've taken the correction — pushed in 92a4758.

I checked the constant-folding claim rather than take it on trust, and it holds: `combine` returns `Constant(fold(a, b))` when both inputs are constants (`circuit_elem.rs:85`), and `From<F>` yields `Constant` (`:271`). Every domain point entered the old fold as a constant, so the whole product *and* the inversion folded at build time. My line implied this change saves a wrapper channel something. It doesn't, and saying so was wrong.

Your replacement says the right thing, so I've used its substance:

```
/// The weight depends on the subspace alone, so all of it runs in the domain's own field and
/// crosses into the arithmetic field once.
/// That is what allows the checked inversion below: one wrapper channel's element type offers the
/// unchecked inverse alone.
```

One correction to the correction, which matters for anyone reading this thread later: there are two wrapper element types and they differ exactly here.

| type | `invert_or_zero` |
| --- | --- |
| `spartan-verifier`'s `CircuitElem` | `unimplemented!()` — `circuit_elem.rs:433` |
| `recursion`'s `SymbolicElem` | implemented — constant folds, wire goes through a hint plus `constrain_inverse` |

So "a wrapper channel's element type implements only the unchecked `invert`" is true of the one you named and not of the other. Hence "one wrapper channel's" in the wording above — it is the `CircuitElem` case that constrains the design, and one such type is enough to rule the checked form out in `E`.

tcoratger: Good question, and the answer is that the unimplemented one is never reached here.

`invert_or_zero` is called on `product`, whose type is `F` — the domain's own field, always a concrete binary field. `E` is only ever constructed from the finished weight, by `E::from`. So `CircuitElem::invert_or_zero` is not on this path at all; the call resolves to the concrete field's, which is implemented everywhere.

It is the move into `F` that makes the checked form available. In the old code the product had type `E`, and a checked inversion there would have needed `E::invert_or_zero` — which is exactly the `unimplemented!()` you are pointing at. So the old code had no choice but the unchecked one.

On calling `invert` instead: I'd rather not, for two reasons.

- Deleting that `unsafe` is most of what this PR is for. Its stated justification was false — it claimed the product is non-zero "whatever the caller passes", but `new_unchecked` documents that it does not check independence, and a dependent basis puts a zero past index 0. Re-introducing `unsafe` to skip one comparison on a value computed once per proof trades the fix away for nothing.
- It matches the house pattern. `NormalizedSubspacePolys::new` already does `assert_ne!(step, F::ZERO, ...)` then `step.invert_or_zero()`, on the same kind of value, twenty files away.

I've also fixed the doc line above this that prompted the thread — it claimed a constraint saving that does not exist, since constants already fold at build time.

jimpo (approved):

PR #2417, opened 2026-08-31
BINIUS-592: Assert that a ZK mask covers the round polynomial it masks

tcoratger:
Closes [BINIUS-592](https://linear.app/jimpo/issue/BINIUS-592).

Refuses a Libra mask whose degree is below the round polynomial it is supposed to hide.
Prover-side precondition only — the transcript is byte-for-byte unchanged on every input that was already valid.

### The problem

The ZK MLE-check hides each round polynomial by adding a random mask polynomial to it, then sends the sum.

That addition is the coefficient-wise sum of two univariates, and it zero-pads the shorter side.

So a mask of lower degree than the main prover's round polynomial contributes a zero to the top coefficients.
Those coefficients ship exactly as the witness produced them.

Nothing compared the two degrees, and no precondition said they had to be compared.

### The evidence

Same quadratic main prover, run twice with mask buffers drawn from two different seeds.
Reading the top coefficient of round 1 straight off the transcript:

| mask degree | seed 11 | seed 22 |
|---|---|---|
| 2 | `0x8a0a0156a6c9c5259c768960fc6d95ec` | `0x8e302ec31945a5a66a87b270f786557a` |
| 1 | `0x166d45842a6f3545de2674d5410142c3` | `0x166d45842a6f3545de2674d5410142c3` |

A degree-2 mask moves the coefficient and moves it differently per seed — the mask is doing its job.
A degree-1 mask leaves it bit-identical across seeds, because it never reached that coefficient.

One witness-dependent $\mathbb{F}_{2^{128}}$ element in the clear, every round.

(The MLE-check round proof truncates the *constant* coefficient, not the top one, so the leaking coefficient is genuinely on the wire.)

### The change

One comparison per round, before the batching step.

```
main   : [a_0, a_1, a_2]
mask   : [b_0, b_1]      -> zero-padded to [b_0, b_1, 0]
batched: [a_0 + c*b_0, a_1 + c*b_1, a_2]
                                    ^^^ unmasked
```

The invariant enforced is: **every coefficient of the main round polynomial has a partner in the mask's**.
That is "at least as many coefficients", not "exactly as many" — a longer mask still hides, and the verifier is told the batched degree independently.
A test pins that overshoot case so nobody later tightens the relation to equality.

### Soundness

- The check runs before anything is written, so an accepted run produces the identical transcript.
- The masking algebra itself was already correct — the audit of the Libra prime polynomial found no defect. Only the contract between the two degrees was missing.
- The one live caller already satisfies it: the Spartan prover pairs a degree-2 mask with a quadratic composition. The hazard was that this pairing lived as a comment in another crate, so a cubic composition added later would have broken hiding with no signal.

### Scope

- **changed:** the batching step and the proving routine's docs in `crates/ip-prover/src/sumcheck/zk_mlecheck.rs`
- **tests:** the existing end-to-end test is parameterized by the two degrees, then driven at equal, over-, and under-degree
- **untouched:** the mask type, the mask prover, the verifier, and every other crate

### Verification

The new test was watched failing before it was trusted:

- assert removed → `test_prove_mask_below_main_degree ... FAILED` (`test did not panic as expected`)
- assert restored → 9 passed in the module, 108 passed in the crate

Gates, all clean:

- `cargo test -p binius-ip-prover` — 108 passed
- `cargo test -p binius-ip-prover --release` — 107 passed, 1 pre-existing failure (`logup_star::prove::tests::test_out_of_range_index_panics`, unrelated to this change and failing on `main` too)
- `cargo clippy -p binius-ip-prover --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-ip-prover --document-private-items --no-deps`
- `cargo +nightly fmt --all --check`, `typos crates/ip-prover/`, `cargo check --workspace --all-targets`
- `cargo test -p binius-spartan-prover` — the live caller, all green

### DIFF
--- a/crates/ip-prover/src/sumcheck/zk_mlecheck.rs
+++ b/crates/ip-prover/src/sumcheck/zk_mlecheck.rs
@@ -347,9 +347,17 @@ impl<F: Field, P: PackedField<Scalar = F>, Data: Deref<Target = [P]>> MleCheckPr
 /// Returns [`ProveZKOutput`] containing the main polynomial's multilinear evaluations
 /// and the round challenges.
 ///
+/// # Pre-conditions
+///
+/// * The mask's univariate degree must be at least the degree of the main prover's round
+///   polynomials.
+/// * The two round polynomials are added coefficient by coefficient, so a shorter mask leaves the
+///   high-degree coefficients uncovered and the protocol is no longer hiding.
+///
 /// # Panics
 ///
 /// Panics if the main prover emits more than one round polynomial.
+/// Panics if the mask's round polynomial is shorter than the main prover's.
 pub fn prove<F: Field, P: PackedField<Scalar = F>, Data: Deref<Target = [P]>>(
 	mut main_prover: impl MleCheckProver<F>,
 	mask: Mask<P, Data>,
@@ -384,6 +392,21 @@ pub fn prove<F: Field, P: PackedField<Scalar = F>, Data: Deref<Target = [P]>>(
 		let mask_round_coeffs = mask_round_coeffs_vec
 			.pop()
 			.expect("mask prover has 1 claim");
+
+		// The batching below pads the shorter polynomial with zeros, so any coefficient beyond
+		// the mask's degree would reach the channel exactly as the witness produced it.
+		//
+		//     main   : [a_0, a_1, a_2]
+		//     mask   : [b_0, b_1]      -> padded to [b_0, b_1, 0]
+		//     batched: [a_0 + c*b_0, a_1 + c*b_1, a_2]
+		//                                         ^^^ unmasked
+		assert!(
+			mask_round_coeffs.0.len() >= main_round_coeffs.0.len(),
+			"the mask round polynomial has {} coefficients against the main round polynomial's \
+			 {}, so the excess would be sent unmasked",
+			mask_round_coeffs.0.len(),
+			main_round_coeffs.0.len()
+		);
 
 		// Batch the round coefficients: batched = main + batch_challenge * mask
 		let batched_round_coeffs = main_round_coeffs + &(mask_round_coeffs * batch_challenge);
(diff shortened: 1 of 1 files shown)


crates/ip-prover/src/sumcheck/zk_mlecheck.rs
```diff
@@ -385,6 +393,21 @@ pub fn prove<F: Field, P: PackedField<Scalar = F>, Data: Deref<Target = [P]>>(
 			.pop()
 			.expect("mask prover has 1 claim");
 
+		// The batching below pads the shorter polynomial with zeros, so any coefficient beyond
+		// the mask's degree would reach the channel exactly as the witness produced it.
+		//
+		//     main   : [a_0, a_1, a_2]
+		//     mask   : [b_0, b_1]      -> padded to [b_0, b_1, 0]
+		//     batched: [a_0 + c*b_0, a_1 + c*b_1, a_2]
+		//                                         ^^^ unmasked
```
benji-potamus, line 402: The first question a reader has here is "why compare lengths and not degrees?" — a coefficient vector whose top entry is zero has the length of degree `d` but the effective degree of `d-1`, which is exactly the leak this guards against. The answer is that `MleCheckMaskProver::execute` builds the round polynomial from `coeffs_for_var`, which iterates `0..=degree` unconditionally, so the length is structural and no coefficient *value* can shorten it. That isn't visible from this file's `prove`, so it's worth a line next to the diagram:

```suggestion
		//                                         ^^^ unmasked
		//
		// The mask emits one coefficient per power up to its degree, so the length is
		// structural: a zero coefficient never shortens the vector.
```

Suggestion, not a blocker.

benji-potamus (approved):
Approve.

Real gap, fixed in the right place. `prove` is the only site in the tree that batches a mask into a round polynomial, and the assert lands before that round's `send_many` with nothing witness-dependent written earlier in the function (`mask_eval` is witness-independent). A plain `assert!` is right for a hiding premise, and it matches the one-claim `assert_eq!` immediately above it.

The severity argument holds: `mlecheck::RoundProof::truncate` removes the *constant* coefficient, so the top one is exactly what goes on the wire. And the Spartan caller pairs a degree-2 mask with the quadratic composition (3 coefficients each), so this is a no-op there and the transcript is byte-identical.

On whether this belongs in the type system instead of a prover-side assert: `MleCheckProver` exposes no degree, and `execute` is free to return a different coefficient count each round, so the count is only knowable at the point you put the check. The assert is the right resting place, and hoisting it would mean widening the trait for nothing.

Parameterizing the existing end-to-end test over the two degrees to cover equal/over/under is a tidy way to get the coverage without a new test harness.

One suggestion inline.

jimpo (approved):

PR #2425, opened 2026-08-31
BINIUS-608: Give the batched interpreter row slices instead of per-element access

tcoratger:
Closes [BINIUS-608](https://linear.app/jimpo/issue/BINIUS-608).

Makes the batched interpreter work a row at a time instead of a word at a time, so the optimiser can vectorise it.
Same bytecode, same results — the equivalence test against the single-instance interpreter is the proof.

### The problem

The batched interpreter runs one bytecode instruction across every instance of the batch.
It does that one word at a time.

```rust
for i in 0..ctx.n_instances() {
    let val = ctx.load(src1, i) ^ ctx.load(src2, i);
    ctx.store(dst, i, val);
}
```

Two things stop that from going fast.

- `load` and `store` resolve through the 2D view's `Index` impl, which recomputes `row * data_width + col` on every access.
- `store` takes `&mut self`, so the compiler must assume a store may alias the next load, and cannot widen the loop.

Here is the whole fused-AND-XOR loop body, disassembled out of `Executor::run` monomorphised for the batched context, from a release build with `-Ctarget-cpu=native`:

```
imul   %r10,%r15            ; recompute row * data_width
imul   %r8,%r13
imul   %r14,%rbp
imul   %rax,%r14
mov    (%rbx,%r15,8),%r15
and    (%rbx,%r13,8),%r15   ; one word
xor    (%rbx,%r14,8),%r15   ; one word
mov    %r15,(%r11,%rsi,8)   ; one word
inc    %rsi
jne    <top>
```

Four multiplies and three memory operations to move one 64-bit word.
Across the whole 2,480-instruction function there is not one vector arithmetic instruction.

A row of a stripe is contiguous in memory. That fact never reaches the optimiser.

### The idea

Give the execution-context trait row-wise operations, with today's scalar loop as the default body.

```rust
fn map<const D: usize, const S: usize, F>(&mut self, dsts: [u32; D], srcs: [u32; S], op: F)
where F: Fn([Word; S]) -> [Word; D];
```

The single-instance context inherits the default and does not change at all.
The batched context overrides it with contiguous row slices and one loop.

Two more methods cover what `map` cannot express:

- `update` folds one row into another, which is what the multi-way exclusive-or needs.
- `check` scans rows for an assertion violation, and only then walks instances to record any.

The row split lives in `StridedArray2DViewMut`, the type that owns the layout.
That keeps the pointer arithmetic where the layout is defined and leaves the frontend with a safe surface and no `unsafe` of its own.

### The change

```diff
 fn exec_fax<C: EvalContext>(&mut self, ctx: &mut C) {
     let (dst, src1, src2, src3) = (...);
-    for i in 0..ctx.n_instances() {
-        let val = (ctx.load(src1, i) & ctx.load(src2, i)) ^ ctx.load(src3, i);
-        ctx.store(dst, i, val);
-    }
+    ctx.map([dst], [src1, src2, src3], |[a, b, c]| [(a & b) ^ c]);
 }
```

Same instruction after the change, on the same machine:

```
vmovdqu64 -0xc0(%r15,%r13,8),%zmm0
vmovdqu64 -0x80(%r15,%r13,8),%zmm1
vmovdqu64 -0x40(%r15,%r13,8),%zmm2
vmovdqu64 (%r15,%r13,8),%zmm3
...
vpternlogq $0x78,-0xc0(%r12,%r13,8),%zmm0,%zmm4   ; (a & b) ^ c, 8 words
vpternlogq $0x78,-0x80(%r12,%r13,8),%zmm1,%zmm5
vpternlogq $0x78,-0x40(%r12,%r13,8),%zmm2,%zmm6
vpternlogq $0x78,(%r12,%r13,8),%zmm3,%zmm7
```

Thirty-two words per iteration, and the AND and the XOR fold into one instruction.

### Vector instructions in the batched run loop

Counted over `Executor::run` monomorphised for the batched context plus the `map` / `update` / `check` bodies the compiler outlines from it.

| | before | after |
|---|---|---|
| functions | 1 | 21 |
| total instructions | 2,480 | 6,196 |
| vector logic ops (`vpternlogq` / `vpxorq` / `vpandq` / `vporq`) | 3 | 34 |
| `%ymm` operands | 16 | 150 |
| `%zmm` operands | 4 | 353 |

The three "before" logic ops are the GHASH reduction on `%xmm`, which is per-word scalar helper code and unchanged.
Every wide operand in the before column is a `vmovdqu` move: no arithmetic on instances was vectorised at all.

### Scope

- **new:** `StridedArray2DViewMut::row` and `StridedArray2DViewMut::rows_mut`.
- **new:** `EvalContext::map`, `EvalContext::update`, `EvalContext::check`, each with the current scalar loop as its default body.
- **changed:** 20 of the 21 opcodes routed through them, and the batched overrides.
- **untouched:** the single-instance context, which keeps the default bodies verbatim.
- **untouched:** the hint opcode's per-instance loop, the constant broadcast, and the tile gather.

### Soundness

`rows_mut` carves mutable and shared slices out of one allocation, so it must never be handed a destination row that is also a source row.

That invariant already holds in the bytecode.
A gate's outputs are freshly allocated wires.
`pool_slots` reclaims a scratch slot only once the whole gate that last read it has run, so no gate is ever handed a slot still holding one of its own inputs.

Rather than assume it, `rows_mut` asserts it — destinations pairwise distinct, and disjoint from the sources — at the point the slices are formed.
The check is a handful of register compares per instruction against a loop over hundreds of words, and it is what the `unsafe` block below it rests on.
A future scratch-allocator change that broke the invariant would panic, not miscompile.

`batched_matches_scalar_per_instance` is the equivalence proof: it runs a mixed-opcode circuit over 1024 instances and asserts every instance's committed witness equals what the single-instance interpreter produces for the same inputs.
Its fixture now exercises all 20 converted opcodes — multiply, GHASH multiply, select, fused-AND-XOR, subtract with borrow, both 32-bit lane adds, a five-way exclusive-or, all eight shift variants, and all six assertions.
A bad row split is exactly what that test would catch.

Miri, on the tests that exercise the row path:

| | Stacked Borrows | Tree Borrows |
|---|---|---|
| `binius-frontend` `eval_form::batch` (4 tests) | pass | pass |
| `binius-utils` row / row-split tests (4 tests) | pass | pass |

Two frontend tests are not covered by those runs, and neither is skipped silently:

- `const_eval::differential_against_interpreter` enumerates roughly 13,000 gate builds and evaluations, which does not finish under Miri in a workable time.
- `parallel_population_matches_serial_for_varied_stripe_widths` is rayon-based, and additionally trips a pre-existing Stacked Borrows violation in `StridedArray2DViewMut::into_strides`: each stride re-derives a mutable slice over the whole buffer, invalidating the tag of every stride handed out before it. That is untouched by this PR — the existing `test_strides` unit test reproduces it on its own, on `main`.

### Verification

All clean:

- `cargo test -p binius-frontend -p binius-utils` (debug, so the row-split assert is live)
- `RUSTFLAGS="-Ctarget-cpu=native" cargo test -p binius-frontend -p binius-utils --release`
- `RUSTFLAGS="-Ctarget-cpu=native" cargo clippy -p binius-frontend -p binius-utils --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-frontend -p binius-utils --document-private-items --no-deps`
- `cargo +nightly-2026-07-01 fmt --all --check`, `python3 tools/check_copyright_notice.py`, `typos`

`cargo test -p binius-circuits` also passes end to end (407 tests), which is the broadest evidence that no gate in the circuit library ever emits an instruction whose destination is one of its own sources.

### Benchmark

AMD Ryzen 9 9950X3D, 16 cores / 32 threads, `RUSTFLAGS="-Ctarget-cpu=native"`.
Circuit: one `keccak_f1600`, 2,760 gates and 2,760 eval instructions.

One 256-instance tile through `populate_wire_witness_batched`, single-threaded and core-pinned, 300 reps, best of 5 runs:

| | before | after | speedup |
|---|---|---|---|
| ms per tile | 0.613 | 0.083 | 7.4x |
| M word-ops/s | 1,152 | 8,520 | 7.4x |

`populate_batch_parallel` at `log_instances = 13`, with the `BufferPool` held outside the timed loop, best of 3 runs:

| threads | before | after | speedup |
|---|---|---|---|
| 1 | 24.02 ms | 6.75 ms | 3.56x |
| 4 | 7.63 ms | 2.81 ms | 2.71x |
| 16 | 4.02 ms | 2.36 ms | 1.70x |
| 32 (all) | 4.06 ms | 3.48 ms | 1.17x |

The interpretation phase itself gets 7.4x faster, but `populate_batch_parallel` also zeroes a per-tile buffer, runs the caller's fill, broadcasts constants, and gathers the tile into the batch layout — none of which this PR touches.
So the end-to-end gain shrinks as thread count rises and the path becomes memory-bandwidth bound.
The single-thread and low-core columns are where this change shows up.

### DIFF
--- a/crates/frontend/src/eval_form/exec.rs
+++ b/crates/frontend/src/eval_form/exec.rs
@@ -44,6 +44,56 @@ pub trait EvalContext {
 	/// Writes `value` to the register at index `reg` within instance `instance`.
 	fn store(&mut self, reg: u32, instance: usize, value: Word);
 
+	/// Applies `op` to every instance, reading the `srcs` registers and writing the `dsts`.
+	///
+	/// Every destination register must differ from every source register.
+	fn map<const D: usize, const S: usize, F>(&mut self, dsts: [u32; D], srcs: [u32; S], op: F)
+	where
+		F: Fn([Word; S]) -> [Word; D],
+	{
+		for i in 0..self.n_instances() {
+			let out = op(srcs.map(|reg| self.load(reg, i)));
+			for (dst, value) in dsts.into_iter().zip(out) {
+				self.store(dst, i, value);
+			}
+		}
+	}
+
+	/// Folds `src` into `dst` across every instance.
+	///
+	/// `dst` must differ from `src`.
+	fn update<F>(&mut self, dst: u32, src: u32, op: F)
+	where
+		F: Fn(Word, Word) -> Word,
+	{
+		for i in 0..self.n_instances() {
+			let value = op(self.load(dst, i), self.load(src, i));
+			self.store(dst, i, value);
+		}
+	}
+
+	/// Records a failure against `path_spec` for every instance whose `srcs` satisfy `fails`.
+	///
+	/// `message` renders the same words into the detail line of that failure.
+	/// It runs only for a failing instance, so it may allocate freely.
+	fn check<const S: usize, P, M>(
+		&mut self,
+		srcs: [u32; S],
+		path_spec: PathSpec,
+		fails: P,
+		message: M,
+	) where
+		P: Fn([Word; S]) -> bool,
+		M: Fn([Word; S]) -> String,
+	{
+		for i in 0..self.n_instances() {
+			let words = srcs.map(|reg| self.load(reg, i));
+			if fails(words) {
+				self.note_assertion_failure(i, path_spec, message(words));
+			}
+		}
+	}
+
 	/// Records an assertion violation for one instance.
 	///
 	/// The index is local to this context.
@@ -130,30 +180,21 @@ impl<'a> Executor<'a> {
 		let dst = self.read_reg();
 		let src1 = self.read_reg();
 		let src2 = self.read_reg();
-		for i in 0..ctx.n_instances() {
-			let val = ctx.load(src1, i) & ctx.load(src2, i);
-			ctx.store(dst, i, val);
-		}
+		ctx.map([dst], [src1, src2], |[a, b]| [a & b]);
 	}
 
 	fn exec_bor<C: EvalContext>(&mut self, ctx: &mut C) {
 		let dst = self.read_reg();
 		let src1 = self.read_reg();
 		let src2 = self.read_reg();
-		for i in 0..ctx.n_instances() {
-			let val = ctx.load(src1, i) | ctx.load(src2, i);
-			ctx.store(dst, i, val);
-		}
+		ctx.map([dst], [src1, src2], |[a, b]| [a | b]);
 	}
 
 	fn exec_bxor<C: EvalContext>(&mut self, ctx: &mut C) {
 		let dst = self.read_reg();
 		let src1 = self.read_reg();
 		let src2 = self.read_reg();
-		for i in 0..ctx.n_instances() {
-			let val = ctx.load(src1, i) ^ ctx.load(src2, i);
-			ctx.store(dst, i, val);
-		}
+		ctx.map([dst], [src1, src2], |[a, b]| [a ^ b]);
 	}
 
 	fn exec_select<C: EvalContext>(&mut self, ctx: &mut C) {
@@ -161,15 +202,8 @@ impl<'a> Executor<'a> {
 		let cond = self.read_reg();
 		let t = self.read_reg();
 		let f = self.read_reg();
-		for i in 0..ctx.n_instances() {
-			// Select t if MSB(cond) is 1, otherwise select f.
-			let val = if ctx.load(cond, i).is_msb_true() {
-				ctx.load(t, i)
-			} else {
-				ctx.load(f, i)
-			};
-			ctx.store(dst, i, val);
-		}
+		// Select t if MSB(cond) is 1, otherwise select f.
+		ctx.map([dst], [cond, t, f], |[cond, t, f]| [if cond.is_msb_true() { t } else { f }]);
 	}
 
 	fn exec_bxor_multi<C: EvalContext>(&mut self, ctx: &mut C) {
@@ -180,12 +214,15 @@ impl<'a> Executor<'a> {
 		let srcs = (0..n)
 			.map(|_| self.read_reg())
 			.collect::<SmallVec<[u32; 8]>>();
-		for i in 0..ctx.n_instances() {
-			let mut val = Word::ZERO;
-			for &src in &srcs {
-				val = val ^ ctx.load(src, i);
+		// The destination doubles as the accumulator, so it is seeded before the fold runs.
+		match srcs.split_first() {
+			None => ctx.map([dst], [], |[]| [Word::ZERO]),
+			Some((&first, rest)) => {
+				ctx.map([dst], [first], |[a]| [a]);
+				for &src in rest {
+					ctx.update(dst, src, |acc, w| acc ^ w);
+				}
 			}
-			ctx.store(dst, i, val);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


crates/frontend/src/eval_form/exec.rs
```diff
@@ -44,6 +44,56 @@ pub trait EvalContext {
 	/// Writes `value` to the register at index `reg` within instance `instance`.
 	fn store(&mut self, reg: u32, instance: usize, value: Word);
 
+	/// Applies `op` to every instance, reading the `srcs` registers and writing the `dsts`.
+	///
+	/// Every destination register must differ from every source register.
```
benji-potamus: Suggestion: why emitted bytecode always satisfies this precondition is worked out in the PR description only, and it is what the `unsafe` in `rows_mut` ultimately rests on. Worth keeping next to the contract:

```suggestion
	/// Every destination register must differ from every source register. Emitted bytecode
	/// satisfies this: a gate's outputs are freshly allocated wires, and `pool_slots` reclaims
	/// a scratch slot only after the whole gate that last read it has run.
```

crates/frontend/src/eval_form/exec.rs
```diff
@@ -229,9 +263,7 @@ impl<'a> Executor<'a> {
 	/// So the compiler monomorphizes this into a branch-free tight loop with the shift inlined.
 	#[inline]
 	fn shift_each<C: EvalContext>(ctx: &mut C, dst: u32, src: u32, op: impl Fn(Word) -> Word) {
-		for i in 0..ctx.n_instances() {
-			ctx.store(dst, i, op(ctx.load(src, i)));
-		}
+		ctx.map([dst], [src], |[w]| [op(w)]);
```
benji-potamus: Suggestion: `shift_each` is now a one-line forwarder, and the doc comment above it describes what `map` does rather than what this function does. Folding it into the eight match arms (`ShiftVariant::Sll => ctx.map([dst], [src], |[w]| [w << amount]),`) deletes the helper and the stale comment at about the same width per arm. Your call — keeping it does confine the `[dst], [src]` boilerplate to one place.

crates/utils/src/strided_array.rs
```diff
@@ -63,6 +63,53 @@ impl<'a, T> StridedArray2DViewMut<'a, T> {
...
+		&self.data[start..start + self.width()]
+	}
+
+	/// Returns the `dsts` rows mutably alongside the `srcs` rows.
+	///
+	/// # Panics
+	///
+	/// Panics if any index is out of bounds.
+	/// Panics if a destination row repeats, or appears among the sources.
+	pub fn rows_mut<const D: usize, const S: usize>(
```
benji-potamus: Suggestion: `dsts`/`srcs` names the interpreter's use of this method inside an otherwise domain-neutral 2D view. Something like `mut_rows`/`shared_rows` would keep the utility's vocabulary its own; the disjointness contract is already carried by the doc and the asserts, which is the part that matters.

benji-potamus (approved):
Approve.

Good change. The row split lives in the type that owns the layout, so the frontend keeps a safe surface; the single-instance context inherits the default bodies verbatim and is genuinely untouched; and the aliasing precondition is enforced by an always-on assert rather than assumed, so a future scratch-allocator change panics instead of miscompiling. Widening `batched_matches_scalar_per_instance` to cover all 20 converted opcodes is what makes a rewrite this broad checkable.

Rebased onto current `main` cleanly. Three suggestions inline — all preferences, take them or leave them.

benji-potamus (approved):

benji-potamus (approved):

jimpo (approved):

PR #2426, opened 2026-08-31
BINIUS-611: Build the logUp* table denominator in one packed pass

tcoratger:
Closes [BINIUS-611](https://linear.app/jimpo/issue/BINIUS-611).

Builds the same table denominator with one pass and no temporary. Output is bit-identical, dead lanes included.

### The problem

The table-side logUp denominator is `iota(j) - c` for every position `j` of a table.

It was built by filling a fresh vector of $2^m$ scalars and then packing that vector into the arena buffer.

That is one extra allocation, one extra full write and one extra full read, on top of the pack.

The pack is not cheap either — it inserts scalars into a word one lane at a time.

### The idea

Positions that share a packed word differ only in the low bits the word index leaves clear.

The embedding is $\mathbb{F}_2$-linear, so a whole word is one fixed lane pattern shifted by a single scalar:

```
word w, lane l  ->  iota(w * WIDTH) + (iota(l) - c)
```

The bracketed pattern is the same for every word, so it is built once and reused.

### Per packed word

- **before:** `WIDTH` scalar subtractions, `WIDTH` lane inserts, plus a scalar written to a temporary vector and read back out of it
- **after:** one broadcast, one addition, written straight into the arena

→ the per-lane inserts and the temporary both disappear.

### The dead lanes

A table shorter than one packed word shares that word with lanes that are not its entries.

Keeping the challenge inside the lane pattern rather than in the broadcast is what leaves those lanes zero: only the first word can have such lanes, and its shift is `iota(0) = 0`.

That matters because the old build packed through a helper that zero-fills a short final word, so anything else would not be bit-identical.

It is also the wart the sibling looker-side builder has — it subtracts a broadcast challenge from a short word and leaves `c` sitting in the spare lanes. This builder does not inherit it, and a test pins that.

### On parallelism — measured, and mostly a loss

The obvious follow-up is to fan the fill out over the thread pool. It is a loss at every table size the code runs at.

Both arms below are the same packed fill, differing only in `iter_mut` versus `par_iter_mut`, measured as two paired sweeps:

| table size | output | one thread | thread pool |
|---|---|---|---|
| $2^{16}$ | 1 MiB | 5.58 µs | 66.5 µs |
| $2^{18}$ | 4 MiB | 30.60 µs | 109.8 µs |
| $2^{20}$ | 16 MiB | 131.3 µs | 234.6 µs |
| $2^{21}$ | 32 MiB | 537.8 µs | 446.6 µs |
| $2^{22}$ | 64 MiB | 2.604 ms | 779.7 µs |

The crossover is where the output stops fitting in cache. So the fill stays on the calling thread below 32 MiB of output and goes to the pool above it.

This matters for the workload: the only live caller is the integer-multiplication reduction, whose shared power table is $2^{16}$ rows, four orders of magnitude below the crossover. A version that simply parallelised the loop would have been a **4.8x regression** there.

### Scope

- **changed:** the table denominator builder in `crates/ip-prover/src/logup_star/witness.rs`, and its tests
- **untouched:** the looker-side builder, the scatter, the pushforward, every call site
- **new tests:** a scalar reference for the table side, a bit-for-bit comparison against packing that reference, and a dead-lane assertion, all run at a four-scalar packing width so that a table shorter than one word actually exists

The default test profile builds at baseline x86-64, where the optimal packing is one scalar per word and no short word can occur. The new tests therefore pin a four-scalar type explicitly; the existing looker proptest's claim to span the packing width is vacuous for that reason.

### Verification

- `cargo test -p binius-ip-prover` — 108 passed (106 before).
- `cargo test -p binius-ip-prover --release` — 107 passed, 1 failed: `logup_star::prove::tests::test_out_of_range_index_panics`, which fails on `main` today and is fixed separately.
- `RUSTFLAGS="-Ctarget-cpu=native" cargo clippy -p binius-ip-prover --all-targets --all-features -- -D warnings` — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-ip-prover --document-private-items --no-deps` — clean.
- `cargo +nightly-2026-07-01 fmt --all --check`, `typos crates/ip-prover/`, `cargo check --workspace --all-targets` — clean.
- Mutation check: moving the challenge from the lane pattern into the broadcast leaves `c` in the dead lanes, and the new tests fail on exactly that.

### Benchmark

A/B on one tree, `RUSTFLAGS="-Ctarget-cpu=native"`, only this function's body flipped between the two arms. AMD Ryzen 9 9950X3D, 16 cores, 128 MiB L3, AVX-512 so four scalars to a packed word.

| table size | before | after | speedup |
|---|---|---|---|
| $2^8$ | 113.54 ns | 45.85 ns | 2.5x |
| $2^{12}$ | 1.5705 µs | 309.50 ns | 5.1x |
| $2^{16}$ | 27.025 µs | 5.6351 µs | 4.8x |
| $2^{20}$ | 922.46 µs | 122.83 µs | 7.5x |
| $2^{22}$ | 21.237 ms | 521.41 µs | 40.7x |

The line that matters for today's workload is $2^{16}$, the size of the shared power table the integer-multiplication reduction looks up: **4.8x**, all of it from the packing rather than from threads.

The bench itself lands under [BINIUS-597](https://linear.app/jimpo/issue/BINIUS-597), on its own branch — this PR does not depend on it and does not include it.

This shape is **not** on the `sha256` example's path: that workload reports no integer multiplications, so the whole logUp* reduction is skipped. The numbers above are against the crate's own bench, not a fraction of prove time.

### DIFF
--- a/crates/ip-prover/src/logup_star/witness.rs
+++ b/crates/ip-prover/src/logup_star/witness.rs
@@ -322,12 +322,6 @@ where
 	FieldBuffer::new(log_len, packed)
 }
 
-/// Output bytes below which the table denominator is built on the calling thread.
-///
-/// One core keeps up while the output fits in cache.
-/// Below that size, handing the work out costs more than it saves.
-const MIN_PARALLEL_DENOMINATOR_BYTES: usize = 32 << 20;
-
 /// Build the negated table denominator `J - c` over the `m`-variable table cube.
 ///
 /// Entry `j` is `iota(j) - c`. The logUp denominator for table position `j` is `c - iota(j)`; the
@@ -358,12 +352,8 @@ where
 	let mut packed = alloc.alloc::<P>(packed_len);
 	// The allocator rounds its blocks up, so the fill is bounded to the words that are entries.
 	let words = &mut packed.spare_capacity_mut()[..packed_len];
-	if packed_len * size_of::<P>() < MIN_PARALLEL_DENOMINATOR_BYTES {
-		words.iter_mut().enumerate().for_each(fill);
-	} else {
-		words.par_iter_mut().enumerate().for_each(fill);
-	}
-	// Safety: either branch writes each of the first `packed_len` slots exactly once.
+	words.iter_mut().enumerate().for_each(fill);
+	// Safety: the loop writes each of the first `packed_len` slots exactly once.
 	unsafe { packed.set_len(packed_len) };
 
 	FieldBuffer::new(table_n_vars, packed)
--- a/crates/ip-prover/src/logup_star/witness.rs
+++ b/crates/ip-prover/src/logup_star/witness.rs
@@ -9,7 +9,7 @@
 //! - the negated table denominator `J - c`, with `J` the embedded table positions,
 //! - the pushforward `Y = I_* eq_r`, the looker numerator scattered onto table positions.
 
-use std::iter;
+use std::{iter, mem::MaybeUninit};
 
 use binius_compute::{Allocator, VecLike};
 use binius_field::{BinaryField, Divisible, Field, PackedField, util::powers};
@@ -322,6 +322,12 @@ where
 	FieldBuffer::new(log_len, packed)
 }
 
+/// Output bytes below which the table denominator is built on the calling thread.
+///
+/// One core keeps up while the output fits in cache.
+/// Below that size, handing the work out costs more than it saves.
+const MIN_PARALLEL_DENOMINATOR_BYTES: usize = 32 << 20;
+
 /// Build the negated table denominator `J - c` over the `m`-variable table cube.
 ///
 /// Entry `j` is `iota(j) - c`. The logUp denominator for table position `j` is `c - iota(j)`; the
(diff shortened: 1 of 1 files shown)


crates/ip-prover/src/logup_star/witness.rs
```diff
@@ -333,11 +339,34 @@ where
...
+	// So a short table's dead lanes stay zero: only the first word has them, and its shift is zero.
+	let lanes = P::from_scalars((0..live_lanes).map(|l| embed_position::<F>(l) - c));
+	let fill = |(word, slot): (usize, &mut MaybeUninit<P>)| {
+		slot.write(P::broadcast(embed_position::<F>(word << P::LOG_WIDTH)) + lanes);
+	};
+
+	let mut packed = alloc.alloc::<P>(packed_len);
+	// The allocator rounds its blocks up, so the fill is bounded to the words that are entries.
+	let words = &mut packed.spare_capacity_mut()[..packed_len];
+	if packed_len * size_of::<P>() < MIN_PARALLEL_DENOMINATOR_BYTES {
```
benji-potamus: Non-blocking: consider dropping the threshold and the parallel arm, and just always taking `iter_mut()`.

By your own measurement the pool only wins above 32 MiB of output, and the only live caller is four orders of magnitude below that — so `MIN_PARALLEL_DENOMINATOR_BYTES` and this branch are unreachable today. Even at 2^22 the single-threaded fill is still ~40x faster than what's on `main`, so deleting the arm costs nothing real and removes a tuned constant that nothing exercises (and which the sibling `looker_denominator` right above doesn't have — it parallelises unconditionally, so the two builders now disagree on policy for no visible reason).

Happy to merge as-is if you'd rather keep it; it's just a constant and a branch that no test or caller reaches.

benji-potamus (approved):
Approve.

The linearity argument holds: within a word the position is `(w << LOG_WIDTH) | l` with the low bits clear, so the split into a broadcast shift plus a fixed lane pattern is exact rather than approximate. Keeping the challenge inside the lane pattern is what reproduces the old helper's zero-fill in a short word's dead lanes, and the new dead-lane assertion pins it. Building at an explicit four-scalar width so a short word exists at all is the right call — the existing proptest really was vacuous on that axis.

One non-blocking note inline about the parallel arm.

jimpo (changes requested):
This two-code-path thing looks unnecessary. This seems like a proper use of `with_min_task` (missing) would solve the parallelism regression. If that's the only code path (rayon using with_min_task), how does it perform?

crates/ip-prover/src/logup_star/witness.rs
```diff
@@ -333,11 +333,30 @@ where
...
+
+	// Positions sharing a word differ only in the low bits the word index leaves clear.
+	// The embedding is `GF(2)`-linear, so a word is one lane pattern shifted by a single scalar.
+	//
+	//     word w, lane l  ->  iota(w * WIDTH) + (iota(l) - c)
+	//
+	// Invariant: the challenge rides in the lane pattern, not in the shift.
+	// So a short table's dead lanes stay zero: only the first word has them, and its shift is zero.
+	let lanes = P::from_scalars((0..live_lanes).map(|l| embed_position::<F>(l) - c));
+	let fill = |(word, slot): (usize, &mut MaybeUninit<P>)| {
```
jimpo: nit: inline this closure now that it's only called in one place

jimpo (changes requested):

crates/ip-prover/src/logup_star/witness.rs
tcoratger: done

jimpo (approved):

PR #2440, opened 2026-08-31
BINIUS-624: Give every field-buffer copy the parallel path

tcoratger:
Closes [BINIUS-624](https://linear.app/jimpo/issue/BINIUS-624).

Makes the plain copy constructor call the one that reserves growth room, so both split a long copy across workers.
No signature changes and no call site moves — the copy is bit-identical, only the thread it runs on differs.

### The problem

Two constructors copy a borrowed field buffer into memory drawn from an allocator.
One also reserves room to grow; the other does not.

Given the same capacity they compute exactly the same thing: the same number of words requested, the same words copied, the same length recorded.
Only the growth-reserving one splits the copy across workers.
The plain one is a single-threaded `extend_from_slice`, and nothing in its name or its doc says so.

So a caller with no need for growth room reaches for the plain one and silently gets one core of thirty-two.
The WHIR opening does that twice per proof at 64 MiB.

### The idea

```
from_view_in(alloc, src)  ==  from_view_with_capacity_in(alloc, src, src.log_len())
```

That is an identity, not an approximation.
A field slice's word count is fixed by its length, so both ask the allocator for `1 << log_len.saturating_sub(P::LOG_WIDTH)` words, and the capacity assertion in the second is trivially satisfied.

So the first becomes a call to the second, and the duplicated body goes away.

### The one-run shortcut

A run is the words one worker takes, floored by the byte budget for one task, which is 1 MiB.
A source at or below that is a single chunk: the parallel loop cannot split it, and entering the worker pool for a one-item loop costs about 10 ns.

That is small, but it would be paid by every caller of the plain constructor at every size, so the shared copy now takes a direct memcpy when the source fits one run.
The growth-reserving constructor stops paying it too — the Reed-Solomon encoder is the caller that benefits there.

### The change

```
 pub fn from_view_in<A>(alloc: &A, src: FieldSlice<'_, P>) -> Self {
-    let mut words = alloc.alloc::<P>(src.as_ref().len());
-    words.extend_from_slice(src.as_ref());
-    FieldBuffer::new(src.log_len(), words)
+    Self::from_view_with_capacity_in(alloc, src, src.log_len())
 }

 // inside from_view_with_capacity_in
+if source.len() <= run {
+    head.write_copy_of_slice(source);
+} else {
     (head.par_chunks_mut(run), source.par_chunks(run)) ...
+}
```

### Why here and not at the call sites

The narrower fix is to change the three WHIR-opening call sites to name the growth-reserving constructor with the capacity they already span.
That recovers the same milliseconds and touches one crate.

It is the wrong fix.
The two constructors compute the same thing, so having one of them be silently serial is a trap rather than a choice, and the next caller falls into it exactly as this one did.
Every caller of the plain constructor in the workspace — eight of them, none inside a parallel loop — wants the split copy at large sizes and is unharmed at small ones.

The cost of choosing this is the wider blast radius: `binius-math` is depended on by every prover crate, so the whole workspace test suite is the gate rather than one crate's.
That ran clean.

### Scope

- **changed:** one constructor becomes a call to its sibling; the shared copy gains a one-run branch
- **new:** a test copying a source of exactly one run and one of two, so both branches are pinned against the source
- **untouched:** every call site, in this crate and out of it

### Verification

- `cargo test --workspace` — 60 test binaries, all pass
- `RUSTFLAGS="-Ctarget-cpu=native" cargo clippy -p binius-math -p binius-iop-prover --all-targets --all-features -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc -p binius-math -p binius-iop-prover --document-private-items --no-deps` — clean
- `cargo +nightly-2026-07-01 fmt --all --check`, `typos crates/math/` — clean

### Benchmark

Both constructors alternating inside one process, pooled allocator, median of 200 rounds, arm order rotated each round so none inherits another's cache state.
Three repetitions of the whole harness; the figures below are medians of those three.
32 cores, `-Ctarget-cpu=native`, machine shared with other builds, load average 1.0 to 3.6.

| source | before (serial) | after | change |
|---|---:|---:|---|
| 16 B | 0.0347 us | 0.0353 us | +0.6 ns |
| 4 KiB | 0.0456 us | 0.0473 us | +1.7 ns |
| 16 KiB | 0.0807 us | 0.0822 us | +1.5 ns |
| 64 KiB | 0.561 us | 0.556 us | -5 ns |
| 256 KiB | 2.85 us | 2.78 us | -74 ns |
| 1 MiB | 13.34 us | 13.61 us | +2%, inside the spread |
| 2 MiB | 30.0 us | 21.3 us | **-29%** |
| 4 MiB | 60.3 us | 68.4 us | a wash; every arm lands in 55-69 us here |
| 16 MiB | 364 us | 264 us | **-28%** |
| 64 MiB | 4273 us | 2292 us | **-46%** |

Below the 1 MiB run boundary the change is neutral, which is what the one-run shortcut buys.
Above it the split wins.

End to end, `./target/release/examples/sha256 prove --message-len 1000000 --pcs whir`, the two binaries alternating back to back, ten pairs — five with the unpatched arm first and five with the patched arm first.
Load average 1.7 to 5.3.

| span | median before | median after | paired wins |
|---|---:|---:|---|
| `WHIR level 0` | 42.4 ms | 34.2 ms | 9 of 10 |
| `Fold rounds` (summed over levels) | 8.6 ms | 7.7 ms | 9 of 10 |
| `Commit witness` — untouched by this change | 30.6 ms | 31.5 ms | 5 of 10 |
| `Prove` | 464 ms | 443 ms | 7 of 10 |

The level-0 span is where the two 64 MiB copies live and is the number to read.
It drops by more than the isolated 3.8 ms the two copies predict, because in situ the source is not sitting in cache the way it is inside a tight benchmark loop, so the serial copy there runs well below the 15 GB/s the microbenchmark gives it.

The untouched span is listed as the control: it moves by a similar magnitude in no consistent direction, which is the size of the run-to-run noise on this box and the reason `Prove` itself is quoted last.

### DIFF
--- a/crates/math/src/field_buffer/mod.rs
+++ b/crates/math/src/field_buffer/mod.rs
@@ -181,13 +181,12 @@ impl<P: PackedField, Data: VecLike<P>> FieldBuffer<P, Data> {
 	/// Copies a borrowed buffer into memory drawn from `alloc`.
 	///
 	/// Whole packed words are copied, dead lanes and all, so the copy is bit-identical.
+	/// The copy splits into runs, so more than one worker carries a long source.
 	pub fn from_view_in<A>(alloc: &A, src: FieldSlice<'_, P>) -> Self
 	where
 		A: Allocator<Vec<P> = Data>,
 	{
-		let mut words = alloc.alloc::<P>(src.as_ref().len());
-		words.extend_from_slice(src.as_ref());
-		FieldBuffer::new(src.log_len(), words)
+		Self::from_view_with_capacity_in(alloc, src, src.log_len())
 	}
 
 	/// Copies a borrowed buffer into memory drawn from `alloc`, with room reserved to grow.
@@ -224,13 +223,18 @@ impl<P: PackedField, Data: VecLike<P>> FieldBuffer<P, Data> {
 		let run = source.len().min(task_chunk_len::<P>().next_power_of_two());
 
 		let head = &mut words.spare_capacity_mut()[..source.len()];
-		(head.par_chunks_mut(run), source.par_chunks(run))
-			.into_par_iter()
-			.for_each(|(dst, src)| {
-				dst.write_copy_of_slice(src);
-			});
+		if source.len() <= run {
+			// One run holds the whole source, so a split would only buy a worker handoff.
+			head.write_copy_of_slice(source);
+		} else {
+			(head.par_chunks_mut(run), source.par_chunks(run))
+				.into_par_iter()
+				.for_each(|(dst, src)| {
+					dst.write_copy_of_slice(src);
+				});
+		}
 
-		// SAFETY: the loop above wrote every word of the source.
+		// SAFETY: both branches above wrote every word of the source.
 		unsafe { words.set_len(source.len()) };
 
 		FieldBuffer::new(src.log_len(), words)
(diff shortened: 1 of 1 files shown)


jimpo (changes requested):
Why not `with_min_task_bytes` here? Does the chunk size just need tuning? (it should use that instead of par_chunks, basically)

jimpo (approved):

PR #2444, opened 2026-08-31
BINIUS-629: Make the algebraic-folding option control all of the algebraic folding

tcoratger:
Closes [BINIUS-629](https://linear.app/jimpo/issue/BINIUS-629).

Makes `enable_algebraic_folding` mean one thing: off, every operation emits its gate.

Two halves. Gating the constant identities changes nothing about the default build. Adding the missing zero-shift fast paths does — it removes gates from four of the thirteen snapshot circuits, and every movement is a reduction.

### The problem

The option is documented as "apply algebraic identities that let a gate return one of its operands".

It gates four rules and leaves fourteen running whatever it says.

`band` is the worked example — the first rule is behind the option, the five below it are not:

```rust
if shared.opts.enable_algebraic_folding && x == y {
    return x;
}
// Identities that hold bit for bit, so they need no AND constraint:
//   c & d  -> fold        0 & y -> 0        all-1 & y -> y
match (const_of(&shared.graph, x), const_of(&shared.graph, y)) {
    (Some(a), Some(b)) => return shared.graph.add_constant(Word(a.0 & b.0)),
    (Some(a), _) if a == Word::ZERO => return x,
    (Some(a), _) if a == Word::ALL_ONE => return y,
    (_, Some(b)) if b == Word::ZERO => return y,
    (_, Some(b)) if b == Word::ALL_ONE => return x,
    _ => {}
}
```

The same split is in `bxor` (self-inverse gated, three constant arms not), `bor` (idempotent gated, five not), and `select` (identical arms gated, constant condition not).

So `false` does not give an unfolded circuit — it gives one with four rules off and fourteen on.

That matters because the option exists so a caller can build the same circuit twice and compare; a half-working option makes the comparison quietly wrong.

There is a matching gap in the shift builders: `rotl`, `rotr`, `rotl32` and `rotr32` fast-path a zero amount, and `shl`, `shr`, `sar`, `sll32`, `srl32` and `sra32` do not. The four that do, do it unconditionally, so they were ungated too.

### The idea

One option, one meaning.

`false` emits a gate for every operation. `true` folds every identity, including the six shifts that were not folding before.

### The tally

| | gated before | gated now |
|---|---|---|
| `band` | 1 of 6 | 6 |
| `bxor` | 1 of 4 | 4 |
| `bor` | 1 of 6 | 6 |
| `select` | 1 of 2 | 2 |
| shift / rotate by 0 | 0 of 10 | 10 |
| **total** | **4 of 28** | **28** |

### The change

The constant arms move inside the option check, and the identity table above them absorbs the rule that was already gated:

```diff
-		// Idempotent: x & x = x, bit for bit, so return x and emit no gate.
-		if shared.opts.enable_algebraic_folding && x == y {
-			return x;
-		}
 		// Identities that hold bit for bit, so they need no AND constraint:
-		//   c & d  -> fold        0 & y -> 0        all-1 & y -> y
-		match (const_of(&shared.graph, x), const_of(&shared.graph, y)) {
+		//   x & x  -> x           c & d  -> fold
+		//   0 & y  -> 0           all-1 & y -> y
+		if shared.opts.enable_algebraic_folding {
+			if x == y {
+				return x;
+			}
+			match (const_of(&shared.graph, x), const_of(&shared.graph, y)) {
```

The shift half lands in the shared helper every shift builder already routes through, so all ten entry points get it from one place rather than four:

```diff
 	fn emit_shift(&self, variant: ShiftVariant, x: Wire, n: u32) -> Wire {
 		let mut shared = self.shared.borrow_mut();
+		// Identity in every variant:
+		//   shift(x, 0) -> x
+		if shared.opts.enable_algebraic_folding && n == 0 {
+			return x;
+		}
```

### Why `rotl` grew a modulo

`rotl` and `rotl32` express a left rotate as a right rotate by the complement, and the gate's amount must be strictly less than the width:

```diff
-		if n == 0 {
-			return x;
-		}
-		self.emit_shift(ShiftVariant::Rotr, x, 64 - n)
+		self.emit_shift(ShiftVariant::Rotr, x, (64 - n) % 64)
```

The old code was correct: the unconditional early return meant `64 - n` was never evaluated at `n == 0`, so `Rotr 64` was unreachable.

Making the guard conditional removes that protection. With the option off, `rotl(x, 0)` now has to reach the gate — and it must arrive as `Rotr 0`, not `Rotr 64`.

So this is not a latent bug being fixed; it is the correction that had to come with hoisting the guard, and getting it wrong would have produced an out-of-range immediate on exactly the path the new tests exercise.

A zero-amount shift gate is otherwise well-formed already: `shifted_term` maps every variant at zero to the single canonical identity spelling, which the existing lowering test covers.

### Direction chosen, and why

Two answers were defensible: gate everything, or split the option because the constant arms are really eager constant propagation.

Gating everything, for three reasons.

**The name still fits.** `0 & y -> 0` is an annihilator, `all-1 & y -> y` a neutral element, `x & x -> x` idempotence — all algebraic laws, all about returning a wire the caller already has. Only the both-constant arm (`c & d -> fold`) is constant folding in the strict sense.

**One knob beats two for a debugging switch.** The option's job is "build it twice and diff". Splitting it means a caller has to know two names to get an unfolded circuit, and it forces an arbitrary call on where `select`'s constant condition belongs.

**It makes the option a usable bisection tool.** Off now genuinely means "no builder-level rewriting", which is the property a differential comparison needs.

### How this relates to the constant-propagation pass

The both-constant arms do overlap `enable_constant_propagation`, which is a whole pass in `pass/const_prop.rs` that folds gates whose inputs are all constant — and which is **off** by default while this eager folding was unconditional.

That overlap is real and predates this PR. The eager version is strictly cheaper (no gate is ever created) and the pass is strictly more general (it reaches gates the builder cannot, such as a comparison of two constants).

This PR does not resolve the overlap; it only makes the eager half controllable. Deciding whether the two should be one knob is a follow-up.

### What this is not

Constant folding in the shift builders is deliberately out of scope — `shl(const, n)` still emits a gate. That is a separate queued item whose value is unproven.

### Scope

- **changed:** `crates/frontend/src/builder/mod.rs` — four bit-op builders and the shift helper
- **changed:** `crates/frontend/src/gates/shift.rs` — the zero-amount lowering test now builds with folding off, so its eight gates actually reach the lowering instead of being folded away
- **changed:** `crates/frontend/src/builder/tests.rs` — the option test, kept localised
- **changed:** four `.snap` files, all reductions (below)
- **untouched:** the constant-propagation pass, and every option default

### Testing

`algebraic_folding_flag_is_honoured` used to assert that one AND count moved. It now pins the full semantics against a battery of all 28 gated identities:

- with folding on, the circuit has **0 gates**
- with folding off, it has **exactly 28** — one per identity
- both builds are then filled with a witness and run through `ConstraintSystem::verify`, so a folded and an unfolded circuit are shown to compute the same 28 words

### What the default build does

The gating half moves nothing — off is a new behaviour, on is exactly what it was.

The zero-shift half does move the default, and the thirteen `circuits-snapshot` examples are the measurement. Four moved:

| example | gates | eval instructions | other |
|---|---|---|---|
| `zklogin` | 388,463 → 385,516 (−2,947) | 456,550 → 453,603 (−2,947) | scratch peak 1,547 → 1,484 (−63) |
| `ethsign` | 214,965 → 214,829 (−136) | 250,294 → 250,158 (−136) | — |
| `ec_msm` | 208,914 → 208,778 (−136) | 244,030 → 243,894 (−136) | — |
| `bitcoin_p2pkh` | 150,119 → 150,051 (−68) | 175,937 → 175,869 (−68) | — |

Nine are byte-identical: `sha256`, `sha512`, `keccak`, `blake2s`, `blake2b`, `blake3_compress`, `blake3`, `hashsign`, `bitcoin_header_chain`.

Two things are worth reading off that table.

**Every movement is a reduction, and no constraint count moves at all** — not ZERO, not AND, not IMUL, not BMUL, and the committed trace is unchanged everywhere. Fusion, CSE and dead-code elimination were already collapsing those identity shifts, so they cost nothing in the proof. What they cost was a wire, a gate and a bytecode instruction each, paid at build time and again on every witness fill. That is what this recovers, plus 63 scratch slots on `zklogin`.

**`zklogin` dominates because of `extract_byte`.** `extract_byte(word, 0)` lowers to a shift by zero, and base64 decoding calls it once every eight bytes, which is where the bulk of the 2,947 comes from.

The two fixtures originally used to check for drift, `sha256_fixed` over 8192 bytes (47,271 AND / 12,735 ZERO) and `keccak256` over 8 blocks, were verified identical in all 19 `CircuitStat` fields; both also appear unchanged in the snapshot set above.

### Verification

All from the worktree:

- all thirteen `cargo run --example <name> -- check-snapshot` — pass
- `cargo test -p binius-frontend` (debug, so `debug_assert!` fires) — 265 pass
- `RUSTFLAGS="-Ctarget-cpu=native" cargo test -p binius-frontend --release` — 265 pass
- `RUSTFLAGS="-Ctarget-cpu=native" cargo test -p binius-circuits --release` — 412 pass
- `RUSTFLAGS="-Ctarget-cpu=native" cargo test -p binius-prover --release --test prove_verify` — 17 pass (this file builds two circuits with the option off, so it exercises the new unfolded paths end to end)
- `clippy -p binius-frontend --all-targets --all-features -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items` — clean
- `cargo +nightly-2026-07-01 fmt --all --check`, `tools/check_copyright_notice.py`, `typos` — clean

### DIFF
--- a/crates/examples/snapshots/bitcoin_p2pkh.snap
+++ b/crates/examples/snapshots/bitcoin_p2pkh.snap
@@ -1,8 +1,8 @@
 bitcoin_p2pkh circuit
 --
 Gates & Instructions
-├─ Number of gates: 149,311
-└─ Number of evaluation instructions: 161,206
+├─ Number of gates: 149,307
+└─ Number of evaluation instructions: 161,202
 
 Constraints
 ├─ ZERO constraints: 15,896 used (97.0% of 2^14)
--- a/crates/frontend/src/builder/mod.rs
+++ b/crates/frontend/src/builder/mod.rs
@@ -68,7 +68,10 @@ pub struct Options {
 	pub enable_common_subexpression_elimination: bool,
 	/// Drop gates that cannot affect the constraint system.
 	pub enable_dead_code_elimination: bool,
-	/// Apply algebraic identities that let a gate return one of its operands.
+	/// Apply the identities that make an operation return a wire it already has.
+	///
+	/// - Covers a repeated operand, an absorbing or neutral constant operand, and a zero shift.
+	/// - Off means every operation emits its gate.
 	pub enable_algebraic_folding: bool,
 	/// Share scratch slots between values whose lifetimes do not overlap.
 	///
@@ -940,22 +943,24 @@ impl CircuitBuilder {
 	///
 	/// # Cost
 	///
-	/// 1 AND constraint, or none when both operands are the same wire.
+	/// 1 AND constraint, or none when an algebraic identity resolves it.
 	pub fn band(&self, x: Wire, y: Wire) -> Wire {
 		let mut shared = self.shared.borrow_mut();
-		// Idempotent: x & x = x, bit for bit, so return x and emit no gate.
-		if shared.opts.enable_algebraic_folding && x == y {
-			return x;
-		}
 		// Identities that hold bit for bit, so they need no AND constraint:
-		//   c & d  -> fold        0 & y -> 0        all-1 & y -> y
-		match (const_of(&shared.graph, x), const_of(&shared.graph, y)) {
-			(Some(a), Some(b)) => return shared.graph.add_constant(Word(a.0 & b.0)),
-			(Some(a), _) if a == Word::ZERO => return x,
-			(Some(a), _) if a == Word::ALL_ONE => return y,
-			(_, Some(b)) if b == Word::ZERO => return y,
-			(_, Some(b)) if b == Word::ALL_ONE => return x,
-			_ => {}
+		//   x & x  -> x           c & d  -> fold
+		//   0 & y  -> 0           all-1 & y -> y
+		if shared.opts.enable_algebraic_folding {
+			if x == y {
+				return x;
+			}
+			match (const_of(&shared.graph, x), const_of(&shared.graph, y)) {
+				(Some(a), Some(b)) => return shared.graph.add_constant(Word(a.0 & b.0)),
+				(Some(a), _) if a == Word::ZERO => return x,
+				(Some(a), _) if a == Word::ALL_ONE => return y,
+				(_, Some(b)) if b == Word::ZERO => return y,
+				(_, Some(b)) if b == Word::ALL_ONE => return x,
+				_ => {}
+			}
 		}
 		let z = shared.graph.add_internal();
 		shared
@@ -970,20 +975,22 @@ impl CircuitBuilder {
 	///
 	/// # Cost
 	///
-	/// 1 linear constraint, or none when both operands are the same wire.
+	/// 1 linear constraint, or none when an algebraic identity resolves it.
 	pub fn bxor(&self, a: Wire, b: Wire) -> Wire {
 		let mut shared = self.shared.borrow_mut();
-		// Self-inverse: x ^ x = 0, so return the zero constant and emit no gate.
-		if shared.opts.enable_algebraic_folding && a == b {
-			return shared.graph.add_constant(Word::ZERO);
-		}
 		// Identities that hold bit for bit, so they need no linear constraint:
-		//   c ^ d  -> fold        0 ^ b -> b        a ^ 0 -> a
-		match (const_of(&shared.graph, a), const_of(&shared.graph, b)) {
-			(Some(x), Some(y)) => return shared.graph.add_constant(Word(x.0 ^ y.0)),
-			(Some(x), _) if x == Word::ZERO => return b,
-			(_, Some(y)) if y == Word::ZERO => return a,
-			_ => {}
+		//   x ^ x  -> 0           c ^ d  -> fold
+		//   0 ^ b  -> b           a ^ 0  -> a
+		if shared.opts.enable_algebraic_folding {
+			if a == b {
+				return shared.graph.add_constant(Word::ZERO);
+			}
+			match (const_of(&shared.graph, a), const_of(&shared.graph, b)) {
+				(Some(x), Some(y)) => return shared.graph.add_constant(Word(x.0 ^ y.0)),
+				(Some(x), _) if x == Word::ZERO => return b,
+				(_, Some(y)) if y == Word::ZERO => return a,
+				_ => {}
+			}
 		}
 		let z = shared.graph.add_internal();
 		shared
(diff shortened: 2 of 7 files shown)


crates/frontend/src/builder/mod.rs
```diff
@@ -1207,10 +1221,7 @@ impl CircuitBuilder {
 	/// 1 AND constraint (0 if n = 0).
 	pub fn rotl32(&self, x: Wire, n: u32) -> Wire {
 		assert!(n < 32, "rotate amount n={n} out of range");
-		if n == 0 {
-			return x;
-		}
-		self.emit_shift(ShiftVariant::Rotr32, x, 32 - n)
+		self.emit_shift(ShiftVariant::Rotr32, x, (32 - n) % 32)
```
benji-potamus, line 1223: The `% 32` is load-bearing and invisible: it is the only thing keeping `Rotr32 32` — an out-of-range immediate — off the `n == 0` path once the fast-path became conditional. Your PR description explains it well; please put a line of it next to the code, so the next reader does not "simplify" it away.

```suggestion
		assert!(n < 32, "rotate amount n={n} out of range");
		// The gate's amount must be < 32, so a rotate by 0 must arrive as Rotr32 0, not Rotr32 32.
		self.emit_shift(ShiftVariant::Rotr32, x, (32 - n) % 32)
```

Same note wanted on `rotl`'s `(64 - n) % 64`.

crates/examples/snapshots/bitcoin_p2pkh.snap
```diff
@@ -1,8 +1,8 @@
 bitcoin_p2pkh circuit
 --
 Gates & Instructions
-├─ Number of gates: 150,119
-└─ Number of evaluation instructions: 175,937
+├─ Number of gates: 150,051
+└─ Number of evaluation instructions: 175,869
```
benji-potamus: These numbers are stale. On current `main` this circuit is 149,311 gates / 161,206 evaluation instructions, not the 150,119 / 175,937 this diff subtracts from — so the four blessed snapshots conflict on rebase, and the deltas in the PR description no longer describe what the change buys.

Rebase onto `main` and re-run `check-snapshot` across all thirteen examples, then update the table. Worth re-checking which nine are still byte-identical: the gate set moved under you, so the split may not be the same four.

crates/frontend/src/builder/mod.rs
```diff
@@ -1292,7 +1294,7 @@ impl CircuitBuilder {
 	///
 	/// # Cost
 	///
-	/// 1 AND constraint.
+	/// 1 AND constraint (0 if n = 0).
```
benji-potamus, line 1296: Suggestion, take it or leave it: "0 if n = 0" is now only true when the option is on, whereas the bit-op docs you touched went to the option-neutral "or none when an algebraic identity resolves it". Same wording here — on all six shift builders plus the four rotates — would keep the six `# Cost` blocks honest.

benji-potamus (changes requested):
Request changes.

The direction is right and the execution is careful. Gating everything under one knob is the correct call for a switch whose job is "build it twice and diff" — a half-working option is worse than no option. Every eager fold in the builder now sits behind it, the zero-shift half closes a real gap in the shared `emit_shift` helper rather than at ten call sites, and the rewritten `algebraic_folding_flag_is_honoured` is a genuinely strong test: pinning all 28 identities in both directions and then filling and verifying both circuits is much better than the single AND count it replaced.

What needs doing before this lands is the snapshot half — `main` has moved and the four blessed files are now stale (details inline). Two smaller inline notes as well, one of which I'd like in the code.

benji-potamus (approved):
Approve.

Makes the option's semantics match its documentation instead of half-matching it, with the tally table making the before/after precise. The `rotl`/`rotl32` modulo catch on hoisting the guard is exactly the kind of thing that would otherwise slip through quietly, and it's called out and tested. Every measured movement is a gate-count reduction with constraint counts and traces unchanged, so this is a clean win with no behavior change on the default path.

Rebased onto current `main` (past the BINIUS-660 example-binary collapse) and pushed. The four `.snap` files were regenerated with `bless-snapshot` against the merged code rather than hand-merged, and `check-snapshot` passes for all four; `cargo test -p binius-frontend` (272 pass) and `cargo clippy -p binius-frontend --all-targets -- -D warnings` are clean on the rebased head.

jimpo (approved):

PR #2447, opened 2026-09-01
BINIUS-632: Narrow binius-frontend's public surface and turn on missing_docs

tcoratger:
Closes [BINIUS-632](https://linear.app/jimpo/issue/BINIUS-632).

Removes two accidental entries from `binius-frontend`'s public API, keeps three that earn their place, and turns on `missing_docs`. Pure API and documentation work — no behavior change.

### The problem

`binius-frontend` is the crate circuit authors type against, so every `pub` item in it is a promise.

A census over the workspace, run from outside the crate, found three public items with no external caller:

| item | location | callers outside the crate |
|---|---|---|
| `Circuit::witness_row` | `artifact/circuit.rs` | 0 |
| `WitnessFiller::value_vec_mut` | `artifact/witness.rs` | 0 |
| `hints::hint_id_of` | `ir/hints.rs` | 0 |

Two more look like the same case and are not. `Circuit::scratch_peak_live` and `Circuit::n_eval_insn` have no direct external callers either, but both are read by `CircuitStat::collect`, which many crates use.

Separately, the crate warns on a missing crate-level doc but not on missing item docs, so a public item can ship with no explanation at all.

### The idea

"No caller today" is a different question from "should this be public".

The question that decides it: does a supported public API force a caller down this path?

- A public function that hands back a raw buffer and offers no other way to address it makes the addressing function part of its contract.
- A method whose only purpose is a soundness test that bypasses the evaluator is not part of any contract.
- A registry's own key derivation is not part of any contract.

### The change

**Deleted — `WitnessFiller::value_vec_mut`**

Marking it `pub(crate)` immediately surfaced a `dead_code` warning, which settled it: the one and only caller in the entire workspace is a `#[cfg(test)]` soundness test. There is no non-test caller at all.

A method documented as the way to *plant a witness the evaluator would never produce* does not belong on a public surface. The test now builds the forged vector directly, which is a more honest spelling of "bypass the prover" and needs nothing from the frontend:

```diff
-let mut w = circuit.new_witness_filler();
 let cs = circuit.constraint_system();
+let mut value_vec = ValueVec::new(circuit.value_vec_layout());
 for (i, c) in cs.constants.iter().enumerate() {
-    w.value_vec_mut()[ValueIndex::constant(i as u32)] = *c;
+    value_vec[ValueIndex::constant(i as u32)] = *c;
 }
-let result = cs.verify(w.value_vec());
+let result = cs.verify(&value_vec);
```

**Narrowed to `pub(crate)` — `hints::hint_id_of`**

Callers reach hints through `CircuitBuilder::call_hint`, which takes the handler type, registers it, and returns the id itself. Hashing a name into a key is how the registry works internally, not a step any caller performs.

This is not "delete everything with no caller". Three items were examined and kept.

**Kept — `Circuit::witness_row`**

The public `populate_wire_witness_batched` takes a raw `StridedArray2DViewMut` the caller must fill, and its rows are numbered by exactly this function. There is no other way to know which row a wire occupies.

The only alternative spelling composes two public calls, `value_vec_layout().word_offset(witness_index(w))` — precisely the kind of thing that should be hard to express, not the thing a caller is pushed toward. It is also the natural companion to the already-public `witness_index`.

**Kept — `Circuit::scratch_peak_live` and `Circuit::n_eval_insn`**

Both feed public fields of `CircuitStat`, which is used across `binius-recursion` and elsewhere. They are already reachable through a supported path, so narrowing them would only force the same read through a longer one. Confirming the pre-existing read here: these should stay.

### missing_docs

Adding `#![warn(missing_docs)]` and building produced **three** warnings, all in the hint registry:

- `HintId`, the type alias
- `HintRegistry::new`
- `HintRegistry::execute`

Three is small enough to fix in the same change rather than split it out, so the attribute lands on and the crate is clean under `-D warnings`.

### Scope

- **deleted:** one method
- **narrowed:** one function to `pub(crate)`
- **kept, with reasoning recorded:** three items
- **left untouched:** every dependent crate — no call-site edits were needed anywhere

### Verification

Every dependent crate builds with **zero** call-site changes, which is the census confirming itself:

- `cargo build -p binius-circuits -p binius-recursion -p binius-prover -p binius-m4-prover -p binius-m4-verifier -p binius-examples --tests` — clean
- `cargo test -p binius-frontend` — 266 pass (debug, so `debug_assert`s ran)
- `RUSTFLAGS="-Ctarget-cpu=native" cargo test -p binius-frontend --release` — 266 pass
- `cargo test -p binius-circuits` — 414 pass
- `tools/check_copyright_notice.py` — 0 wrong format, 0 missing
- `typos` — clean

No behavior change, so no snapshot can move.

### DIFF
--- a/crates/frontend/src/ir/hints.rs
+++ b/crates/frontend/src/ir/hints.rs
@@ -14,13 +14,16 @@ use std::{
 
 use binius_core::Word;
 
+/// Registry key for one prover-side computation.
+///
+/// Derived from the declared name rather than assigned in order, so it is stable across runs.
 pub type HintId = u32;
 
 /// Hint handler trait for extensible operations.
 ///
 /// Each implementor declares a globally unique `NAME`. The registry identifies hints by the
-/// hash of this name (see [`hint_id_of`]), so registering the same hint twice is a no-op
-/// and every gate using the same hint type shares a single handler entry.
+/// hash of this name, so registering the same hint twice is a no-op and every gate using the
+/// same hint type shares a single handler entry.
 ///
 /// # The `dimensions` parameter
 ///
@@ -66,7 +69,7 @@ pub trait Hint: Send + Sync + 'static {
 ///
 /// Hashes the name with `std::hash::DefaultHasher` (fixed seed, deterministic across runs)
 /// and folds the resulting 64-bit value down to 32 bits by XORing its two halves.
-pub fn hint_id_of(name: &str) -> HintId {
+pub(crate) fn hint_id_of(name: &str) -> HintId {
 	let mut hasher = DefaultHasher::new();
 	name.hash(&mut hasher);
 	let h = hasher.finish();
@@ -101,6 +104,7 @@ pub struct HintRegistry {
 }
 
 impl HintRegistry {
+	/// An empty registry.
 	pub fn new() -> Self {
 		Self {
 			handlers: HashMap::new(),
@@ -120,6 +124,11 @@ impl HintRegistry {
 		self.handlers[&hint_id].shape(dimensions)
 	}
 
+	/// Run the handler under `hint_id`, writing its results into `outputs`.
+	///
+	/// # Panics
+	///
+	/// Panics if nothing is registered under `hint_id`.
 	pub fn execute(
 		&self,
 		hint_id: HintId,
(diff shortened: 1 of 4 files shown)


crates/frontend/src/ir/hints.rs
```diff
@@ -66,7 +69,7 @@ pub trait Hint: Send + Sync + 'static {
 ///
 /// Hashes the name with `std::hash::DefaultHasher` (fixed seed, deterministic across runs)
 /// and folds the resulting 64-bit value down to 32 bits by XORing its two halves.
-pub fn hint_id_of(name: &str) -> HintId {
+pub(crate) fn hint_id_of(name: &str) -> HintId {
```
tcoratger:
```suggestion
pub fn hint_id_of(name: &str) -> HintId {
```

jimpo (approved):

PR #2453, opened 2026-09-01
BINIUS-638: Move the packed AES/GHASH modules under packed_fields and rename their types

tcoratger:
Closes [BINIUS-638](https://linear.app/jimpo/issue/BINIUS-638).

Moves three modules and renames seven public types so a packed type's name says which field it packs. Pure refactor — no behaviour change, no deprecated aliases.

### The problem

The crate already sorts its field code into two drawers.

- Scalar field definitions live in `fields/`: `fields::rijndael`, `fields::ghash`, `fields::ghash_sq`.
- Packed types live in `packed_fields/`: `packed_fields::primitive`, `packed_fields::sliced`.

Three modules never made it into either drawer and sat at the crate top level instead: `packed_aes`, `packed_ghash`, `packed_ghash_sq`.

Their type names had drifted too.

- `PackedAESBinaryField16x8b` packs `Rijndael8b` — the scalar field was renamed away from AES, the packing was not.
- `PackedBinaryGhash1x128b` packs `Ghash128b` — the `BinaryField` infix says nothing the rest of the name doesn't.
- `PackedGhashSq1x256b` packs `GhashSq256b` — this one already reads correctly, which is what the other two should look like.

### The idea

Put the packed modules in the drawer their scalar counterparts are already in, and name each packed type after the field it packs.

```
fields::rijndael  -> Rijndael8b    packed by  PackedRijndael{1,16,32,64}x8b
fields::ghash     -> Ghash128b     packed by  PackedGhash{1,2,4}x128b
fields::ghash_sq  -> GhashSq256b   packed by  PackedGhashSq{1,2}x256b
```

The third line is what the first two now imitate.

### The moves

```
crates/field/src/packed_aes.rs      -> crates/field/src/packed_fields/rijndael.rs
crates/field/src/packed_ghash.rs    -> crates/field/src/packed_fields/ghash.rs
crates/field/src/packed_ghash_sq.rs -> crates/field/src/packed_fields/ghash_sq.rs
```

All three are recorded as git renames, so the diff shows a move rather than a delete plus an add.

### The renames, and how many call sites each touched

| old name | new name | occurrences | files |
|---|---|---|---|
| `PackedAESBinaryField1x8b` | `PackedRijndael1x8b` | 8 | 4 |
| `PackedAESBinaryField16x8b` | `PackedRijndael16x8b` | 32 | 11 |
| `PackedAESBinaryField32x8b` | `PackedRijndael32x8b` | 21 | 9 |
| `PackedAESBinaryField64x8b` | `PackedRijndael64x8b` | 23 | 11 |
| `PackedBinaryGhash1x128b` | `PackedGhash1x128b` | 126 | 40 |
| `PackedBinaryGhash2x128b` | `PackedGhash2x128b` | 91 | 36 |
| `PackedBinaryGhash4x128b` | `PackedGhash4x128b` | 69 | 26 |

370 occurrences across 69 files. 73 files change in total, once the module declarations and the doc references are counted.

By crate: `field` 33, `prover` 12, `iop-prover` 10, `math` 8, `m4-prover` 6, `recursion` 2, `ip-prover` 2.

### Keeping the crate-root paths working

Before the move, each module re-exported itself from `lib.rs`.

```
-pub mod packed_aes;              -pub use packed_aes::*;
-mod packed_ghash;                -pub use packed_ghash::*;
-mod packed_ghash_sq;             -pub use packed_ghash_sq::*;
```

A glob over a parent module does not reach into its submodules, so `packed_fields/mod.rs` now carries the three submodules' items itself.

```
+pub mod ghash;                   +pub use ghash::*;
+pub mod ghash_sq;                +pub use ghash_sq::*;
+pub mod rijndael;                +pub use rijndael::*;
```

`lib.rs` keeps its single `pub use packed_fields::*`, so every crate-root path a caller writes today still resolves — under the new type name.

The three modules are declared `pub mod`, matching how `primitive` and `sliced` are declared beside them. That is a small widening for `ghash` and `ghash_sq`, which used to be private modules whose types were only reachable at the crate root.

### One consequence worth calling out

`packed_ghash_sq` was a private module, so its module doc was private documentation, and rustdoc allows a private doc to link a private item.

As `packed_fields::ghash_sq` that doc becomes public, and its link to a private helper is then an error under `RUSTDOCFLAGS="-D warnings"`.

The link is dropped and the sentence is otherwise unchanged. This is the only content edit in the whole PR; nothing else gains or loses documentation.

### Scope

- **moved:** the three module files, as git renames.
- **renamed:** seven public types, in place, everywhere — including the `arch/` backends that alias them behind `#[cfg(target_feature = ...)]` for aarch64 NEON+AES, wasm32 simd128 and AVX-512, none of which a baseline x86-64 build compiles.
- **untouched:** `PackedGhashSq{1,2}x256b` and the `SlicedGhashSq*` aliases, which already follow the target pattern; the `arch/` module names `packed_aes_128`, `packed_ghash_256`, `packed_ghash_sq_256` and friends, which name per-width backends rather than the moved modules; the informal test-function label `..._packed_aes16x8b_over_1b`, which is a test name, not one of the renamed types.
- **not added:** no deprecated aliases, no re-export shims.

### Verification

Every old name is gone: a whole-tree grep for `PackedAESBinaryField` and `PackedBinaryGhash`, and for the three old module paths, returns zero — including untracked files.

Clean, native (`RUSTFLAGS="-Ctarget-cpu=native"`):

- `cargo clippy -p binius-field --all-targets --all-features -- -D warnings`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test -p binius-field --release` — 275 passed
- `cargo test --workspace` (debug, so the `debug_assert`s actually run) — all green

Clean, portable (`RUSTFLAGS` unset, so the fallback paths compile):

- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test -p binius-field --release` — 258 passed
- `cargo test --workspace` (debug) — all green

Clean, other gates:

- `cargo +nightly-2026-07-01 fmt --all --check`
- `typos` (v1.32.0)
- `python3 tools/check_copyright_notice.py` — 690 files, 0 wrong, 0 missing
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --document-private-items --no-deps`

The `#[cfg]`-gated backends no local build compiles were checked by cross-compiling them:

- `--target aarch64-unknown-linux-gnu` with `-Ctarget-feature=+neon,+aes`
- `--target wasm32-unknown-unknown` with `-Ctarget-feature=+simd128`
- host with `-Ctarget-feature=+avx512f,+vpclmulqdq,...`

`cargo test --workspace --release` reports one failure, on `main` as much as here: `logup_star::prove::tests::test_out_of_range_index_panics`. It is a `#[should_panic]` test over a `debug_assert` precondition, so in a release build the assertion is compiled out and the panic arrives from the scatter's slice bound with a different message. It cannot pass in any release build, is untouched by this PR, and both release legs are otherwise green — 2027 tests passing native, 1925 portable, with `--no-fail-fast`.


crates/field/benches/ghash_invert.rs
```diff
@@ -9,8 +9,7 @@
 use std::hint::black_box;
 
 use binius_field::{
-	Ghash128b as GhashB128, PackedBinaryGhash1x128b, PackedBinaryGhash2x128b,
-	PackedBinaryGhash4x128b, PackedField,
+	Ghash128b as GhashB128, PackedField, PackedGhash1x128b, PackedGhash2x128b, PackedGhash4x128b,
```
tcoratger, line 12: I think that we can do a followup to avoid this type of aliases `Ghash128b as GhashB128` but let's keep for a followup to avoid mixing things.

jimpo (approved):

PR #2457, opened 2026-09-02
BINIUS-635: Add `collect_into_alloc_vec` and adopt it where it fits

benji-potamus:
Closes [BINIUS-635](https://linear.app/jimpo/issue/BINIUS-635/collect-into-alloc-vec-for-rayon).

### The problem

Rayon's `collect_into_vec` targets a `&mut Vec`, which an `Allocator`'s buffer is not. So the prover builds those buffers by hand:

```rust
let mut buffer = alloc.alloc(len);
(inputs, buffer.spare_capacity_mut()[..len].par_iter_mut())
    .into_par_iter()
    .for_each(|(input, out)| out.write(f(input)));
// SAFETY: the zip is over equal-length sides, so ...
unsafe { buffer.set_len(len) };
```

Every site repeats the same soundness argument — a rayon zip yields as many items as its shorter side holds, so a length mismatch leaves trailing slots uninitialized — and several sites also have to remember that the allocator may hand back more capacity than it was asked for, and bound the spare slice themselves.

### The change

**Commit 1** adds `CollectIntoAllocVec` to `binius-compute`: an extension trait on `IndexedParallelIterator`, blanket-implemented, mirroring `binius-utils`' existing `IndexedParallelIteratorExt`. It sizes a buffer from the iterator's length, fills the spare capacity in parallel, and hands the buffer back, so the argument above is made once.

It lives in `binius-compute` because it is about `Allocator`. That gives the crate its first dependency, on `binius-utils`, for the `rayon` re-export the rest of the workspace already builds its parallel loops from — `binius-utils` has no `binius-*` dependencies, so the edge is new but not circular.

**Commit 2** adopts it at the seven sites where the loop produces exactly one element per output slot:

| crate | site |
|---|---|
| `binius-math` | `multilinear/fold.rs` — `fold_highest_var` |
| `binius-iop-prover` | `fri/encode.rs` — the random mask |
| `binius-ip-prover` | `logup_star/witness.rs` — `looker_denominator` |
| `binius-spartan-prover` | `wiring.rs` — `fold_constraints` |
| `binius-m4-prover` | `witness/instance_fold.rs` |
| `binius-m4-prover` | `witness/operand_columns.rs` (×2) |

Seven `unsafe { set_len }` blocks go away, along with the per-site length reasoning. The loops now say what one item is rather than where it lands:

```rust
let data = (lo.as_ref(), hi.as_ref())
    .into_par_iter()
    .with_min_task(WorkPerItem::FieldMuls)
    .map(|(&lo_i, &hi_i)| extrapolate_line(lo_i, hi_i, broadcast_scalar))
    .collect_into_alloc_vec(alloc);
```

The net is 88 insertions / 88 deletions — the consumer costs about what it removes at seven sites, and pays off at the eighth.

### What is deliberately left alone

I went through every `alloc` + `spare_capacity_mut` + `set_len` site in the workspace. The rest do not fit this consumer:

- **Several outputs per pass.** `fracaddcheck/circuit.rs` writes a numerator and a denominator; `spartan-prover/wiring.rs`'s mul witness writes `a`, `b` and `c`. Collecting each separately would repeat the shared work.
- **Chunk-shaped writes, not one slot per item.** `ring_switch.rs`, `FieldBuffer`'s copy, `fri/fold.rs`, `operand_columns.rs`'s striped writer, `binary_merkle_tree.rs`, and `intmul/witness.rs`'s strided column view.
- **The buffer is pre-sized past the iterator.** `prove.rs`'s witness packing and `shift/key_collection/collection.rs` allocate the whole padded power-of-two length, then push a tail element and zero-pad into it. Collecting would size the buffer to the aligned prefix and reallocate the largest prover buffer. `prove.rs` carried a comment naming `collect_into_vec`'s `&mut Vec` as the obstacle, which this PR would have made stale; it now states the reason that actually holds.
- **Not parallel.** `logup_star`'s `table_denominator` and `shift/monster.rs` are sequential loops.

### Testing

`cargo test --workspace` passes (I excluded `binius-examples` on the final run only because the box ran out of disk linking its example binaries; it type-checks under the clippy run below). `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo doc --no-deps --document-private-items`, `cargo +nightly fmt --check` and `typos` are all clean. The new trait has a doc example and a unit test over `BufferPool`, whose block rounding is what makes the bounded spare slice load-bearing.

https://claude.ai/code/session_012HhgduRiSH5dYWtrDXNkiL

### DIFF
--- a/crates/compute/Cargo.toml
+++ b/crates/compute/Cargo.toml
@@ -8,3 +8,4 @@ authors.workspace = true
 workspace = true
 
 [dependencies]
+rayon.workspace = true
--- a/crates/compute/src/lib.rs
+++ b/crates/compute/src/lib.rs
@@ -10,8 +10,13 @@
 //! [`VecLike`] buffers, letting the prover's allocation code be written against `&impl Allocator`
 //! rather than a concrete pool. `&BufferPool` is the primary [`Allocator`], producing [`PoolVec`]
 //! buffers.
+//!
+//! [`CollectIntoAllocVec`] is the rayon seam over the same machinery: it collects a parallel
+//! iterator straight into one of those buffers.
 
 use std::{mem, mem::MaybeUninit, ops::DerefMut};
+
+use rayon::prelude::*;
 
 pub mod buffer_pool;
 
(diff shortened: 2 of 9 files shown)


crates/compute/Cargo.toml
```diff
@@ -8,3 +8,4 @@ authors.workspace = true
 workspace = true
 
 [dependencies]
+binius-utils = { path = "../utils", default-features = false }
```
jimpo: Change dependency to the rayon crate itself, not binius-utils

jimpo (changes requested):

crates/compute/Cargo.toml
benji-potamus: Done — `binius-compute` now depends on `rayon` directly (`rayon.workspace = true`), and `crates/compute/src/lib.rs` imports `rayon::prelude::*` instead of going through `binius_utils::rayon`. Squashed into the first commit and pushed.

jimpo (approved):

PR #2473, opened 2026-09-09
Scope pull-request CI by architecture and pin target features

benji-potamus:
Closes [BINIUS-665](https://linear.app/jimpo/issue/BINIUS-665/scope-pull-request-ci-by-architecture-arch-specific-crates).

Two commits, each standalone.

### 1. Pin CI target features instead of `-Ctarget-cpu=native`

`native` made two things a property of whichever runner SKU a job landed on: the machine code it built, and — through `cfg(target_feature)` — which source paths it compiled and ran at all.

The first forced the CPU feature hash into the cache key, splitting the cache into one bucket per SKU. The repo is at its 10 GB quota, so the rare buckets get evicted and a job that draws one rebuilds the whole dependency tree; on run 34381460029 the native leg restored no cache and rebuilt 289 third-party crates. The second made coverage nondeterministic: on that same run the native leg's SKU reported `avx512f, gfni` and the portable leg's reported them absent.

Each leg now pins an explicit feature set, so artifacts are portable across runners, the CPU feature hash leaves the cache key, and the per-SKU buckets collapse into one per architecture:

* x86_64: `-Ctarget-cpu=x86-64-v3 -Ctarget-feature=+pclmulqdq,+aes`
* arm64: `-Ctarget-feature=+neon,+aes,+sha2`

**The pinned sets are narrower than the issue proposed, deliberately.** RUSTFLAGS apply to everything the job *executes*, so a feature that some SKU lacks is not a lost optimization but a SIGILL. AES-NI and PCLMULQDQ are Westmere-era, AVX2 is Haswell-era, and Armv8 AES/SHA2 are on every arm64 runner; VAES, VPCLMULQDQ, SHA-NI, GFNI and AVX-512 are not on the whole fleet (Cascade Lake, for one, has AVX-512 but none of VAES/VPCLMULQDQ/GFNI/SHA-NI). Raising the executed baseline is a follow-up worth doing once the fleet's guaranteed set is confirmed from the `Run platform diagnostics` output, which prints it on every run.

So the wider features are type-checked instead of executed, by a check-only clippy pass over the three crates that hold architecture-specific code. Nothing else in CI compiles those paths.

`--target x86_64-unknown-linux-gnu` on that step is load-bearing, not decoration: without it cargo applies RUSTFLAGS to build scripts and proc macros as well, and *runs* them. Locally, the same command without `--target` dies before reaching any Binius crate:

```
error: failed to run custom build command for `proc-macro2 v1.0.107`
  process didn't exit successfully: [...] (signal: 4, SIGILL: illegal instruction)
```

Also drops `-Ctarget-cpu=native` from the documentation build, where it bought nothing and cost a cache bucket of its own.

### 2. Scope pull-request test legs to the arch-specific crates

Only `binius-field`, `binius-hash-prover` and `binius-arith-bench` contain architecture-specific code, but every pull request built and tested the whole workspace on all three legs. On `pull_request` the portable and arm64 legs now build, test and doctest those three crates alone. The x86_64 leg stays full on every event, `merge_group` stays full on every leg, and the `push` warmers stay full so the merge queue restores a full cache.

Workspace clippy stays on every leg and every event — it is the cheap way to type-check the arch-gated code under each leg's cfg.

**Risk accepted, as the issue states it:** the `field` SIMD paths are also exercised transitively by the `prover` and `math` tests, which no longer run on the two secondary PR legs. A SIMD bug that `field`'s own tests miss now surfaces in the merge queue, bouncing a merge rather than reaching main.

### Verification

Run locally on this tree, since nothing here can be checked without executing CI:

* Wide-feature clippy, with `--target`: clean in 28s.
* `cargo build`/`nextest run`/`test --doc` over the three-crate set under the pinned x86_64 flags: 399 tests pass.
* `cargo clippy --workspace --all-targets -- -D warnings` under the pinned x86_64 flags: clean.

This PR's own CI run is the first real measurement — it runs on `pull_request`, so the reduced legs are live here.

### Noted, not changed

`DEVELOPMENT.md`'s cross-compilation section names `binius-field` and `binius-arith-bench` as the arch-specific crates and omits `binius-hash-prover`, whose SHA-NI/AVX-512/NEON paths are equally cross-arch. Pre-existing, and a separate change.

https://claude.ai/code/session_013ZGaaPNGLAjY4BXQU1HPVy

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,12 +61,6 @@ jobs:
       fail-fast: false
       matrix:
         arch:
-          # binius-circuits tests are architecture-independent (the crate depends only on
-          # binius-frontend + binius-core, which contain no arch-specific code), so they run only on
-          # the x86_64 leg. The other legs skip them with nextest's run-layer filter (`-E`) rather
-          # than `cargo --exclude`: keeping the build set at `--workspace` matches the Compile step's
-          # feature unification, so the dependency tree is not recompiled. See BINIUS-151.
-          #
           # The feature-enabled legs pin an explicit target-feature set rather than asking for
           # `-Ctarget-cpu=native`. GitHub's runner fleet is heterogeneous, so `native` makes the
           # generated code — and therefore which `cfg(target_feature)` paths CI compiles and runs
@@ -76,18 +70,20 @@ jobs:
           # AES/SHA2 are present on every arm64 runner. Wider x86_64 features — AVX-512, GFNI,
           # VAES, VPCLMULQDQ, SHA-NI — are only on some SKUs, so they are type-checked by the
           # check-only clippy pass below instead of executed.
+          #
+          # `arch_packages_on_pr` marks a leg that narrows to the architecture-specific packages on
+          # a pull request. See `PACKAGES` below.
           - name: x86_64-portable
             runner: ubuntu-26.04
             rustflags: # portable
-            test_args: "-E 'not package(binius-circuits)'"
+            arch_packages_on_pr: true
           - name: x86_64
             runner: ubuntu-26.04
             rustflags: "-Ctarget-cpu=x86-64-v3 -Ctarget-feature=+pclmulqdq,+aes"
-            test_args: ""
           - name: arm64
             runner: ubuntu-26.04-arm
             rustflags: "-Ctarget-feature=+neon,+aes,+sha2"
-            test_args: "-E 'not package(binius-circuits)'"
+            arch_packages_on_pr: true
     name: rust-checks-${{ matrix.arch.name }}
     # This job also runs on pushes to main, where it stops after Compile and exists only to write a
     # cache entry that pull-request and merge-queue runs can restore. The verification steps below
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -64,20 +64,42 @@ jobs:
...
+          # testing it on all three legs asks the same question three times, and the two secondary
+          # legs are where a pull request pays for the redundant two. The `field` SIMD paths are
+          # also exercised transitively by the `prover` and `math` tests, which no longer run here
+          # on a pull request; the merge queue runs every leg in full, so a SIMD bug that `field`'s
+          # own tests miss bounces a merge rather than reaching main.
           - name: x86_64-portable
             runner: ubuntu-24.04
             rustflags: # portable
             test_args: "-E 'not package(binius-circuits)'"
+            pr_packages: "-p binius-field -p binius-hash-prover -p binius-arith-bench"
```
jimpo: This option is weird. This generalizes the wrong thing. These are the arch-dependent packages, which is a fixed set. The logic is that on some triggers (PR), we want to build & test only arch-dependent packages, and on some triggers we want to build & test the whole workspaces.

.github/workflows/ci.yml
```diff
@@ -64,20 +64,42 @@ jobs:
...
+          # on every event. The rest of the workspace is architecture-independent, so building and
+          # testing it on all three legs asks the same question three times, and the two secondary
+          # legs are where a pull request pays for the redundant two. The `field` SIMD paths are
+          # also exercised transitively by the `prover` and `math` tests, which no longer run here
+          # on a pull request; the merge queue runs every leg in full, so a SIMD bug that `field`'s
+          # own tests miss bounces a merge rather than reaching main.
           - name: x86_64-portable
             runner: ubuntu-24.04
             rustflags: # portable
             test_args: "-E 'not package(binius-circuits)'"
```
jimpo: The rationale of this was that circuits are arch-independent. Now that we restrict non-arch-independent stuff even more, just remove this extra `not package(binius-circuits)` stuff. We either run the select arch-dependent packages or everything (inc circuits).

jimpo (changes requested):

.github/workflows/ci.yml
benji-potamus: Agreed — the package list is a constant, not a per-leg parameter. It now lives once in the `PACKAGES` expression, and the matrix carries only the boolean that says which legs narrow on a pull request:

```yaml
arch_packages_on_pr: true
...
PACKAGES: ${{ github.event_name == 'pull_request' && matrix.arch.arch_packages_on_pr && '-p binius-field -p binius-hash-prover -p binius-arith-bench' || '--workspace' }}
```

The leg dimension has to stay, since the x86_64 leg runs the whole workspace on a pull request too; the boolean is what remains once the list moves out.

.github/workflows/ci.yml
benji-potamus: Done — `test_args` and the `not package(binius-circuits)` filter are gone from the matrix and from the nextest command, along with the BINIUS-151 comment that justified them. A leg now either takes the three architecture-specific packages or the whole workspace, circuits included.

Effect on the merge queue: the portable and arm64 legs run the circuits tests there now, where before only x86_64 did.

jimpo (approved):

PR #2478, opened 2026-09-10
[shift] Take one constraint point and build one tensor expansion

benji-potamus:
Closes [BINIUS-649](https://linear.app/jimpo/issue/BINIUS-649/shift-take-one-constraint-point-and-build-one-tensor-expansion).

After #2477, every operation's claim sits at a prefix of the unified constraint point `r_x`, of length `ℓ_cons = max(ℓ_zero, ℓ_and, ℓ_imul, ℓ_bmul)`. This change makes the shift reduction carry that one point, not four.

## Verifier (`crates/verifier/src/protocols/shift/`)

- `OperationClaim` is gone. `verify` takes `operation_claims: &[E]`, the four operations' operand evaluations concatenated in `[zero, bitand, intmul, binmul]` order, each run as long as its operation's arity (the new `OPERATION_ARITIES`). `check_eval` takes the point `r_x: &[E]`, which is the only place that reads it.
- New helper `log_constraints(cs)` returns each operation's prefix length. Both sides read the prefix lengths from it.
- The wiring claim's flat input carries one `r_x` section: `operation_batch.. | operand_batch.. | r_x.. | r_s_inner.. | r_v_inner.. | r_s_outer.. | r_v_outer.. | r_y.. | r_segment`. `WiringEvalShape::r_x_len` replaces `r_x_prime_lens`. The input shrinks by `Σ log_z − ℓ_cons`, and the recursion circuit's deferred-claim wires shrink with it.
- `wiring_weights` expands `r_x` once. `constraint_tables` then truncates that expansion to each operation's prefix: widest first, summing halves, so additions only. Each table equals the stage that a lowest-up build passes through.
- The tables stay unscaled. `call` and `call_native` multiply each operation's evaluation by its batch weight: 4 multiplications, where scaling the tables would cost `Σ 2^ℓ_z`. As a result, `split_by_operation` no longer needs its seed of one.

## Prover (`crates/prover/src/protocols/shift/`)

- `OperatorClaims` holds `r_x`, `log_constraints`, `r_zhat_prime` and the per-arity eval arrays (`[F; 1]`, `[F; 3]`, `[F; 4]`, `[F; 6]`). `OperatorData` and `PreparedOperatorData` are removed. `shift::prove` does not take the constraint system, so the prefix lengths travel in the claims.
- `prepare` expands `r_x` once, cuts it with the same `constraint_tables`, and scales each table by its operation weight. `PreparedOperatorClaims[operation]` returns that table. `Key::accumulate` / `accumulate_wide` and `build_g` / `build_monster_segments` read it, and are otherwise unchanged.

## Transcript

The transcript does not change. `r_x = zero::reduction_point(r_x_star, max(r_x_star.len(), ℓ_zero), sample)` samples the same extra challenges that the Zero point sampled before. `r_x_star` already spans the widest AND, IMUL and BMUL set, because `rerand::verify` runs over the longest summand.

## Follow-up

Evaluating every operation matrix at the *same* full `r_x` — padding the matrices with empty rows so one tensor expansion serves both sides, with no per-operation table — is [BINIUS-667](https://linear.app/jimpo/issue/BINIUS-667/shift-evaluate-every-operation-matrix-at-the-full-constraint-point). It waits on @jimpo's protocol design input, as he asked.

## Tests

All re-run after the review fixes.

- New `constraint_tables_are_the_prefix_expansions_of_the_point` checks each table against `eq_ind_partial_eval_scalars(&r_x[..ℓ])`. The widths in the test are unsorted, one repeats, and one is zero.
- `cargo nextest run --workspace`: 1931 passed. This covers `prove_verify.rs`, `prover/tests/shift.rs`, the M4 shift round trip, and `recursion_end_to_end.rs`. The deferred-claim round trip (`check_deferred`) passes, and the exported claim is now 44 elements.
- The 9 `#[ignore]`d M4 prove tests (`prove_hash_primitives`, `prove_hash_based_sig`) also pass (`--run-ignored only`).
- The `#[ignore]`d `deferring_shrinks_a_level_by_the_cost_of_the_level_below` did not finish on my machine (15 GB of RAM, no swap). The kernel killed it with `SIGKILL` after 32 s. It is killed the same way on `main` (after 34 s), so the kill does not come from this change. The test builds two multi-million-row circuits and needs a machine with more memory.
- `cargo clippy --all-targets --all-features -D warnings` on the four touched crates is clean, and so is `cargo doc` with `RUSTDOCFLAGS=-D warnings`.

https://claude.ai/code/session_01DCnTP9k12rd8kBk3kzUuJK


crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -196,7 +218,7 @@ impl<F> VerifyOutput<F> {
 pub fn verify<F, C>(
 	constraint_system: &ConstraintSystem,
 	inout: InoutSegment,
-	claims: [&OperationClaim<C::Elem>; OPERATION_COUNT],
+	claims: &OperationClaims<C::Elem>,
```
jimpo: don't take the struct, take the `r_x` point and `operation_claims: &[C::Elem]` separately

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -210,12 +232,18 @@ where
...
+	// operations' evals are held. An operation of lower arity reads a prefix of the operand
+	// weights; the slots above its arity name no claim.
 	let operation_weights = eq_ind_partial_eval_scalars(&operation_batch_challenges);
 	let operand_weights = eq_ind_partial_eval_scalars(&operand_batch_challenges);
 
-	let eval =
-		inner_product(operation_weights, claims.map(|claim| claim.batched_eval(&operand_weights)));
+	let operation_evals = claims.evals.iter().map(|evals| {
+		assert!(evals.len() <= operand_weights.len()); // precondition
+		iter::zip(evals, &operand_weights)
```
jimpo: prob use inner_product helper function

jimpo (changes requested):
Two minor inline comments, and one bigger one to be handled as a follow-up.

The fact that the sparse matrices are evaluated at different `r_x` points, which are prefixes of the same point, is still not simple enough. We should have them all be evaluated at *the same* full r_x point. Basically, we'll think of these padded operation matrices as padded with fully empty rows, up to the maximum r_x length. Then in both the prover side of the shift argument and the final wiring matrix evaluation, they can use a single r_x tensor expansion. Create a linear ticket for the desired outcome (I'll follow-up on that ticket with protocol design input).

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done. `OperationClaims` is gone.

`verify` now takes `operation_claims: &[C::Elem]`: the four operations' operand evaluations concatenated in `[zero, bitand, intmul, binmul]` order, each run as long as its operation's arity. A new `OPERATION_ARITIES` constant in the module gives those lengths, and `verify` splits the slice with it.

`r_x` stays separate, as the argument of `check_eval`, which is the only place that reads it — `verify` batches the evaluations and never touches the point. So I did not also add an `r_x` parameter to `verify`: it would be unused. If you meant it to take the point too, say so and I will move it.

The four callers build the flat slice with `.concat()`.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done. Each operation's run is now `inner_product(evals.iter().cloned(), operand_weights[..arity].iter().cloned())`, and the four results pair with the operation weights through `inner_product` as before.

jimpo (approved):

PR #2479, opened 2026-09-13
[shift] Evaluate every operation matrix at the full constraint point

benji-potamus:
Linear: [BINIUS-667](https://linear.app/jimpo/issue/BINIUS-667/shift-evaluate-every-operation-matrix-at-the-full-constraint-point). Follow-up to #2478.

## What

Every operation matrix is padded with empty rows up to the length of `r_x`. As in jimpo's design note on the ticket, the constraint reductions still claim each operation at its own prefix `r_x[..n]`. Each claim is then multiplied by `eq(0, r_x[n..])`, which gives the evaluation of the padded matrix at the full `r_x`.

One tensor expansion of `r_x` now serves both sides.

**Verifier**
- New `shift::padding_scales(cs, r_x)` returns the four factors. `reduction.rs` scales the operand claims with them before `shift::verify`.
- New `shift::log_constraint_point(cs)` returns the length of `r_x`.
- `WiringEvalFn` builds one constraint table. `split_by_operation` gives every share the same full point.
- Removed: `constraint_tables`, `log_constraints`.

**Prover**
- `OperatorClaims::from_rerand` applies the same factors. `OperatorClaims::log_constraints` is removed.
- `PreparedOperatorClaims` holds one `r_x_tensor`. The operation batching weight moves out of the per-operation tables and into the operand weights, one 8-entry run per operation. `prepared[op]` now returns that run.
- `build_monster_segments` builds one scalar table per operation, indexed by the operation's packed code.

## Testing

- `cargo nextest run --release -p binius-verifier -p binius-prover -p binius-m4-prover`: 156 passed. This covers `prove_verify.rs`, the M4 tests and `prover/tests/shift.rs`.
- Mutation check: removing the verifier's scaling fails `test_prove_verify_zero_constraints` and `test_prove_verify_public_wider_than_hidden`.
- I did not run `recursion_end_to_end.rs` locally, because it is slow. CI covers it.
- clippy (`--workspace --all-targets --all-features`), fmt, typos and rustdoc are clean.

https://claude.ai/code/session_01KQU9dM9KNtBYBsj2rAJ1K8


crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -859,12 +837,12 @@ impl WiringEvalFn<'_> {
 		let outer_shift_scalars = eq_ind_partial_eval_scalars(&[r_s_outer_v, r_v_outer_v].concat());
 		debug_assert_eq!(outer_shift_scalars.len(), SHIFT_COUNT);
 
-		// The constraint-index table is the one left that differs between operations. All four
-		// are prefixes of one point, so one expansion of it serves them all.
-		let constraint_tables = constraint_tables(scaled_expand(r_x_v, E::one()), log_constraints);
+		// Every operation matrix is padded to the same rows, so one expansion of `r_x` serves all
+		// four as well.
+		let constraint_table = scaled_expand(r_x_v, E::one());
```
jimpo: `scaled_expand` is confusing. This is just `eq_ind_partial_eval_scalars`.

crates/verifier/src/reduction.rs
```diff
@@ -205,15 +208,20 @@ where
...
+	let operation_claims = iter::zip(
+		[
+			vec![Channel::Elem::zero()],
+			bitand_evals.to_vec(),
+			absent_or_claimed::<_, INTMUL_ARITY>(cs.log_imul_constraints(), &mut operand_evals),
+			absent_or_claimed::<_, BINMUL_ARITY>(cs.log_bmul_constraints(), &mut operand_evals),
+		],
+		shift::padding_scales(cs, &r_x),
+	)
+	.flat_map(|(claims, scale)| claims.into_iter().map(move |claim| claim * scale.clone()))
```
jimpo: why flat_map?

crates/verifier/src/protocols/shift/verify.rs
```diff
@@ -859,12 +837,12 @@ impl WiringEvalFn<'_> {
...
 
 		WiringInputs {
-			constraint_tables,
+			constraint_table,
```
jimpo: This field should just be called `r_x_tensor`, I believe.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Done: the table is now `eq_ind_partial_eval_scalars(r_x_v)`. `scaled_expand` stays only for the two word-index tensors, which do carry a scale. One trade-off: `call_native` no longer uses the parallel packed expansion for `r_x`, so that table is now built on one thread.

crates/verifier/src/reduction.rs
benji-potamus: There was no good reason: it only paired each run with its scale before flattening. It is gone now. The claims are built as before with `.concat()`: the BitAnd evals are scaled in place, and `absent_or_claimed` takes the operation's scale. This matches `OperatorClaims::from_rerand` on the prover side.

crates/verifier/src/protocols/shift/verify.rs
benji-potamus: Renamed to `r_x_tensor`, which matches `PreparedOperatorClaims::r_x_tensor` on the prover side.

jimpo (approved):

PR #2482, opened 2026-09-15
fix(math): make BinarySubspace::get work when the dimension reaches usize::BITS

erkancamli:
### Problem

`BinarySubspace::get` bounds its index with

```rust
assert!(index < 1 << self.dim(), "precondition: index must be less than 2^dim");
```

`dim()` is `basis.len()`, which is not bounded below `usize::BITS` anywhere. `BinarySubspace::default()` spans the whole field, so `dim()` is 128 for `Ghash128b`, and `with_dim` accepts anything up to `F::DEGREE`. For those subspaces `1 << dim` overflows: in debug it panics with `attempt to shift left with overflow`, and in release the shift amount is masked to `128 & 63 == 0`, so the bound becomes `index < 1` and every non-zero index is rejected as out of range.

`element_at` has the same problem one level down. It does `index >> i` for every basis position `i`, and `i` runs to `dim() - 1`, so a basis longer than the index width shifts past it.

The sibling code already knows about this: `BinarySubspaceIterator::new` asserts `basis.len() < usize::BITS as usize` and `iter` documents the panic. `get` has no such guard, and its own doc promises only that it panics when `index` is at least `2^dim`, which is not what happens.

```rust
let subspace = BinarySubspace::<Ghash128b>::default();
let _ = subspace.get(0);   // panics
```

### Fix

Two lines, both narrow.

`get` skips the comparison once the dimension reaches `usize::BITS`, because at that point every `usize` index is below `2^dim` and the bound cannot be computed:

```rust
self.dim() >= usize::BITS as usize || index < 1 << self.dim()
```

`element_at` uses `checked_shr`, treating a basis position above the top of the index as unselected, which is what the index semantics already imply:

```rust
let selected = index.checked_shr(i as u32).unwrap_or(0) & 1 == 1;
```

Behaviour for every subspace narrower than `usize::BITS` is unchanged, including the existing `test_binary_subspace_range_error`, which still panics on an out of range index.

### Test

`test_get_on_a_subspace_wider_than_a_usize` pins the 128-dimensional case: `get(0)` is zero, `get(1)` and `get(5)` select the expected basis elements, and `get(usize::MAX)` is the sum of the basis elements the index can actually reach.

Before the fix it fails on the first call:

```
panicked at core/src/panicking.rs: attempt to shift left with overflow
   3: binius_math::binary_subspace::BinarySubspace<F,Data>::get
             at ./src/binary_subspace.rs:66:19
```

After, `cargo test -p binius-math --lib` is 179 passed, and `cargo test --workspace --lib` is green across every crate. `cargo clippy -p binius-math --all-targets -- -D warnings` is clean and `cargo fmt --check` reports nothing.

### Environment note

My sandbox cannot reach `static.rust-lang.org`, so rustup could not install the pinned 1.98.1. It has 1.95.0, and I ran everything with `--ignore-rust-version` after moving `rust-toolchain.toml` aside locally; that file is untouched in this PR. For the same reason `cargo fmt` ran on stable rather than the `nightly-2026-07-01` that `prek.toml` pins, so the nightly-only comment wrapping options were skipped. Worth a second look from CI on that one point.


erkancamli:
Could someone approve the workflow runs on this one? No checks have run yet, so there is nothing green for a reviewer to look at.

For what it is worth, locally: `cargo test -p binius-math --lib` is 179 passed, `cargo test --workspace --lib` is green across every crate, `cargo clippy -p binius-math --all-targets -- -D warnings` is clean and `cargo fmt --check` reports nothing. The one caveat I noted in the description still stands: my sandbox cannot reach `static.rust-lang.org`, so I ran on 1.95.0 with `--ignore-rust-version` rather than the pinned 1.98.1, and `cargo fmt` ran on stable so the nightly-only comment options were skipped. CI is the real check on both of those.

erkancamli:
Pushed a simplification and a doc fix, and stating the reachability plainly since the description did not.

**`element_at` now bounds the iterator instead of branching per element.**

```rust
.take(usize::BITS as usize)
.enumerate()
.map(|(i, &basis_i)| basis_i * BinaryField1b::from((index >> i) & 1 == 1))
```

`checked_shr` put a branch on every basis element, and `element_at` is on the NTT twiddle path (`domain_context.rs:95,260` and `subspace_polys.rs:179`). It also carried an unguarded `i as u32`. `take` says the same thing directly, a basis element past the top of the index has no bit to select it, and the `map` goes back to the one-liner `main` already had, so the diff is smaller. Verified equivalent to the `checked_shr` version over basis lengths 0 to 200 and indices around every power-of-two boundary up to `usize::MAX`: zero mismatches.

**`get`'s `# Panics` section was stale.** It still said an index at or above `2^dim` panics, which stops being true at `dim >= usize::BITS` precisely because of this change.

**Reachability, stated honestly.** I went through every call site rather than leaving this implied:

- Every `.get()` call in the workspace outside this module's own tests is in `ntt/domain_context.rs` and `ntt/subspace_polys.rs`, on subspaces built from an NTT basis of length `log_domain_size`, which is tens.
- The only non-test `BinarySubspace::default()` in the workspace is `crates/verifier/src/reduction.rs:126`, over `B8`, so `dim = 8`.
- The largest `with_dim` anywhere is `with_dim(32)` in a test; production ones are `Word::LOG_BITS` (6) or 7.

So **no current caller constructs a subspace with `dim >= 64`, and this fixes nothing that is broken today.** What it does fix is a public API that contradicts itself at a boundary the type explicitly permits: `with_dim` asserts `dim <= F::DEGREE`, which is 128 for `Ghash128b`, and `get` cannot serve that range. `BinarySubspaceIterator::new` already handles the same boundary with an explicit `assert!(basis.len() < usize::BITS as usize)` and documents it, so the codebase already treats it as real; `get` is the one place that did not. That is the whole claim, and I would rather scope it that way than call it critical.

Verification on this head, with `rust-toolchain.toml` moved aside so I could build on stable 1.95.0:

- `cargo test -p binius-math --lib`: 179 passed, 0 failed, including `test_get_on_a_subspace_wider_than_a_usize` and the pre-existing `test_binary_subspace_range_error` (so the loosened assert still rejects out-of-range indices at small dims).
- `cargo test --workspace --lib`: 1867 passed, 0 failed, across 24 suites.
- `cargo clippy -p binius-math --all-targets`: diagnostic multiset byte-identical to `origin/main`.
- Mutation check, both halves: reverting only the assert makes the test fail with `attempt to shift left with overflow`; restoring it and removing only the `.take` makes it fail with `attempt to shift right with overflow`. Neither half is unexercised.

Caveats I cannot close from here: I am on 1.95.0 rather than the pinned 1.98.1 because my sandbox cannot reach `static.rust-lang.org`, and stable rustfmt skipped four nightly-only options from `.rustfmt.toml` (`wrap_comments`, `comment_width`, `imports_granularity`, `group_imports`), so comment wrapping and import grouping in the new code are unverified. CI is the real check on both, which is why the workflow approval would help.

